PR #4, opened 2022-07-22
WIP implementation of dynamic forms, used for Search and Intake

gigxz:
WIP dynamic form component, which is being used for  "Advanced Search" client search pane, and Client Intake form.

I left comments on the parts that are worth looking at to get an idea of how it works. This is still a work in progress! Requesting review mostly for visibility, but feel free to look closely if you want to.


src/modules/form/components/DynamicForm.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { FormDefinition, Item } from '../types';
+
+import DynamicField from './DynamicField';
+
+interface Props {
+  definition: FormDefinition;
+  onSubmit: (values: Record<string, any>) => void;
+}
+
+const DynamicForm: React.FC<Props> = ({ definition, onSubmit }) => {
```
gigxz, line 13: This is the component for dynamically rendering a form based on a form definition, which looks like this: https://github.com/greenriver/hmis-frontend/blob/46726c11749f49b3d8bb8767f027b500a89e00af/src/modules/form/data/intake.json

The linked definition looks like this rendered with DynamicForm:
hmis dev test_5173_intake

src/modules/form/components/DynamicForm.tsx
```diff
@@ -0,0 +1,78 @@
...
+    <Box component='form' onSubmit={submitHandler} sx={{ mt: 4, pb: 2 }}>
+      <Grid
+        container
+        // direction='row'
+        direction='column'
+        rowSpacing={2}
+        columnSpacing={2}
+        sx={{ mb: 2 }}
+      >
+        {definition?.item.map(renderItem)}
```
gigxz, line 54: Recursively renders each form "item" as a `DynamicField`

src/modules/form/components/DynamicField.tsx
```diff
@@ -0,0 +1,130 @@
...
+    }}
+  >
+    {title && <Typography sx={{ mb: 1 }}>{title}</Typography>}
+    <Stack direction='column' spacing={1}>
+      {children}
+    </Stack>
+  </Paper>
+);
+
+const DynamicField: React.FC<Props> = ({
```
gigxz, line 44: `DynamicField` is where we map a given Item to a Component, based on it's type ("string", "choice", etc)

src/modules/form/data/search.json
```diff
@@ -0,0 +1,61 @@
...
+      }
+    },
+    {
+      "type": "string",
+      "required": false,
+      "linkId": "4",
+      "text": "Last 4 Social",
+      "mapping": {
+        "clientSearchInput": "ssnSerial"
+      }
```
gigxz, line 38: Experimenting with a "mapping" field here to map fields into a ClientSearchInput object, for the "Advanced Search" section of the client search page. Will definitely need to come back to this when figuring out how to submit the client creation form and assessments

eanders (approved):

PR #5, opened 2022-07-22
Add codegen

ZachCMP:
https://www.pivotaltracker.com/story/show/182709885

You'll need to have `SCHEMA_PATH` in your environment to make codegen work. I just pointed it to the schema file I dumped in the backend repo on my local filesystem, but it'll take [lots of different options for paths](https://www.graphql-code-generator.com/docs/config-reference/schema-field)


graphql.schema.json
```diff
@@ -0,0 +1,2082 @@
+{
```
gigxz, line 1: Do we want to check this schema in? From the PR description it sounded like we would want to point `SCHEMA_PATH` either to schema file generated in the warehouse repo, or (later, maybe?) to a URL.

gqlTypes.tsx
```diff
@@ -0,0 +1,175 @@
...
+export type Scalars = {
+  ID: string;
+  String: string;
+  Boolean: boolean;
+  Int: number;
+  Float: number;
+  /** A Date object, transporated as a string */
+  Date: any;
+  /** A DateTime object, transporated as a string */
+  DateTime: any;
```
gigxz, line 22: When I regenerate locally, the comments on line 19 and 21 get dropped. Not sure why that is, because I see them in the schema file. I generated with `SCHEMA_PATH=graphql.schema.json yarn graphql:codegen`.

graphql.schema.json
ZachCMP: @gigxz Frankly I'm not sure that we need this JSON schema. We generate it in HNMI, but I'm not sure what consumes it. Also, I'm NOT generating the types from this schema, this schema is actually generated FROM the `schema.graphql` file in the warehouse. My `SCHEMA_PATH` is:

```
SCHEMA_PATH=../hmis-warehouse/drivers/hmis/app/graphql/schema.graphql
```

gqlTypes.tsx
ZachCMP: See above ^^. The JSON schema here is actually generated FROM the real schema in the warehouse repo. You should use the `schema.graphql` file in the warehouse repo to generate things from, not this JSON representation of the schema.

graphql.schema.json
gigxz: Oh okay, gotcha. I think we can remove (or comment out) the `./graphql.schema.json` step in `codegen.yml` if we don't have a use for it yet

gigxz (approved):

graphql.schema.json
ZachCMP: @gigxz Agreed

PR #17, opened 2022-08-18
Paginate enrollments table, add Household member table, use readFragment to get enrollment name

gigxz:
Depends on https://github.com/greenriver/hmis-warehouse/pull/2388


src/components/dashboard/enrollments/ViewEnrollment.tsx
```diff
@@ -1,38 +1,66 @@
...
+
+  const { data, loading, error } = useGetEnrollmentQuery({
+    variables: { id: enrollmentId },
+    skip: !!enrollmentFragment,
+  });
+  if (error) throw error;
+  if (loading) return <Loading />;
+
+  const enrollment = enrollmentFragment || data?.enrollment;
+  if (!enrollment) throw Error('Enrollment not found');
```
gigxz: see if this can be done using `fragment-matcher` codegen and `typePolicies` instead

gigxz: @ZachCMP here I'm using `readFragment` to grab the enrollment if we have it in the cache, otherwise doing a query. On the previous page we queried for a list of enrollments (so we have `Enrollment:<id>` fragments in the cache). On this page, we should be able to use that cached enrollment to show the name of the enrollment at the top of the page. This seems like it will be a common pattern, but I'm wondering if we can configure to the cache to do this automatically? What do you think?

PR #57, opened 2022-10-20
Adjusts proxy timeouts

yakloinsteak:
We've ruled out the instance sizing fixing the instability problem. We'll run with these longer proxy timeouts for a bit to see if the problem comes back. We're back down to two replicas as well. 

@gigxz Note that we think redeploying temporarily fixes this problem, so if you're in dire need, that's a stopgap.

I have commented out some other things we can try in the proxy config if the timeouts don't fix it.


gigxz (approved):

gigxz:
sounds good. thanks!

PR #95, opened 2022-12-12
Add validation logic

ZachCMP:
https://www.pivotaltracker.com/story/show/183735687

### DIFF
--- a/src/modules/search/components/SearchForm.tsx
+++ b/src/modules/search/components/SearchForm.tsx
@@ -37,39 +37,6 @@ export interface SearchFormProps {
 
 const defaultSearchKeys = ['textSearch', 'projects'];
 
-export const validateSearchFormInput = (values: Record<string, any>) => {
-  const requiredValueSet = omitBy(
-    pick(values, [
-      'personalId',
-      'firstName',
-      'lastName',
-      'ssnSerial',
-      'dob',
-      'textSearch',
-    ]),
-    (e) => isNil(e) || isEmpty(e)
-  );
-
-  const errors: Record<keyof typeof requiredValueSet, string> = {};
-
-  if (isEmpty(requiredValueSet)) {
-    errors.textSearch =
-      'Please enter a search term or specific search criteria';
-  } else {
-    if (
-      values.textSearch &&
-      typeof values.textSearch === 'string' &&
-      isEmpty(omit(requiredValueSet, ['textSearch'])) &&
-      values.textSearch.length < 4
-    ) {
-      errors.textSearch =
-        'Please enter more than three characters or add more specific search criteria';
-    }
-  }
-
-  return errors;
-};
-
 const SearchForm: React.FC<SearchFormProps> = ({
   definition,
   onSubmit,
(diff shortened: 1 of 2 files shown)


src/modules/search/components/SearchForm.tsx
```diff
@@ -37,6 +37,39 @@ export interface SearchFormProps {
...
 const defaultSearchKeys = ['textSearch', 'projects'];
 
+export const validateSearchFormInput = (values: Record<string, any>) => {
+  const requiredValueSet = omitBy(
+    pick(values, [
+      'personalId',
+      'firstName',
+      'lastName',
+      'ssnSerial',
+      'dob',
```
gigxz: I don't think we want these values in the JS at all. Eventually the advanced search FormDefinition will be coming from the backend, so the frontend shouldn't make any assumptions about what's in it. The backend could choose to return a different form based on which installation it is, for example, if some particular client has additional search fields they want to use.

I'm speaking theoretically because the search form form def is actually being served in the frontend right now ([here](https://github.com/greenriver/hmis-frontend/blob/stable/src/modules/form/data/search.json)) but we do need to update it so it's fetched from the backend, like all the rest of them. 

The idea is that the forms are "config," and we don't want to couple config and code, because then we would need to redeploy to make changes (not to mention needing to make a change in two places). This here couples code to config, because if we add another field to the search form, we would need to add the field name here as well.

src/modules/search/components/SearchForm.tsx
```diff
@@ -37,6 +37,39 @@ export interface SearchFormProps {
...
+      'Please enter a search term or specific search criteria';
+  } else {
+    if (
+      values.textSearch &&
+      typeof values.textSearch === 'string' &&
+      isEmpty(omit(requiredValueSet, ['textSearch'])) &&
+      values.textSearch.length < 4
+    ) {
+      errors.textSearch =
+        'Please enter more than three characters or add more specific search criteria';
```
gigxz: The button below says "Advanced Search" so lets keep the language consistent

```suggestion
        'Please enter more than three characters or add advanced search criteria';
```

src/modules/search/components/SearchForm.tsx
```diff
@@ -37,6 +37,39 @@ export interface SearchFormProps {
...
+      'textSearch',
+    ]),
+    (e) => isNil(e) || isEmpty(e)
+  );
+
+  const errors: Record<keyof typeof requiredValueSet, string> = {};
+
+  if (isEmpty(requiredValueSet)) {
+    errors.textSearch =
+      'Please enter a search term or specific search criteria';
```
gigxz: Can you put these strings in a translation? We're not using it consistently yet, but we are using it for other parts of the search form! (see `useTranslation` usage in this component)

gigxz (commented):
Looks good!! Suggested a few minor changes

src/modules/search/components/SearchForm.tsx
ZachCMP: @gigxz Can you describe how you would do the validation in that case? I'm not sure how I can do the validation we discussed unless that validation is also in the form definition. Is that the case, or there some other source of truth I could use here?

src/modules/search/components/SearchForm.tsx
ZachCMP: Fixed

src/modules/search/components/SearchForm.tsx
ZachCMP: Fixed

src/modules/search/components/SearchForm.tsx
gigxz: Yeah, the source of truth is the FormDefinition. We already use it to filter down the query params to known params. For example if you navigate to `https://qa-hmis.openpath.host/?firstName=first&foo=bar` it will only save `firstName` into the initial state, because "foo" is not in the form definition.

We also use it in `transformSubmitValues`, which is what generates the map that you're trying to validate here. As long as we're adequately testing `transformSubmitValues`, we should be confident that it's returning a map that only contains keys that are in the form definition.

In other words, at this point we have already confirmed that the maps keys are correct.

What we _haven't_ done is validate that all the fields in the form definition (`firstName`, `lastName`, etc) exist on `ClientSearchInput`. That's up to the form author (us) to ensure. That part is a bit unsafe, but we could do additional validation on the backend –– when loading the form–– to ensure it. However that will become irrelevant with [this ticket](https://www.pivotaltracker.com/story/show/184014535) so I think we should hold off on worrying about that until we get to implementing that change.

gigxz: if that's not clear, let me know and we can chat!

PR #97, opened 2022-12-14
Add image to client card

ZachCMP:
https://www.pivotaltracker.com/story/show/183190250

Depends on https://github.com/greenriver/hmis-warehouse/pull/2680


src/api/operations/clients.queries.graphql
```diff
@@ -15,6 +15,14 @@ query GetClient($id: ID!) {
   }
 }
 
+query GetClientImage($id: ID!) {
+  client(id: $id) {
+    image {
```
gigxz, line 40: Should add ID for caching
```suggestion
  client(id: $id) {
    id
    image {
```


I get the following warning in the console
> invariant.ts:42 Cache data may be lost when replacing the client field of a Query object.

> To address this problem (which is not a bug in Apollo Client), either ensure all objects of type Client have an ID or a custom merge function, or define a custom merge function for the Query.client field, so InMemoryCache can safely merge these objects:

>  existing: {"__typename":"Client","image":{"__ref":"ClientImage:8042"}}
>  incoming: {"__ref":"Client:8042"}

gigxz (approved):
One small comment, otherwise LGTM! Waiting to merge until https://github.com/greenriver/hmis-warehouse/pull/2680 is merged.

PR #101, opened 2022-12-15
Add client sort option input

ZachCMP:
https://www.pivotaltracker.com/story/show/183189700

### DIFF
--- a/src/modules/search/components/ClientSearch.tsx
+++ b/src/modules/search/components/ClientSearch.tsx
@@ -119,6 +119,11 @@ const ClientSearch: React.FC<Props> = ({
   );
   const [offset, setOffset] = useState(0);
 
+  const handleSetSortOrder: typeof setSortOrder = useCallback((value) => {
+    setOffset(0);
+    return setSortOrder(value);
+  }, []);
+
   const [searchClients, { data, loading, error }] = useSearchClientsLazyQuery({
     variables: {
       input: {},
@@ -212,7 +217,7 @@ const ClientSearch: React.FC<Props> = ({
           cardsEnabled={!!cards}
           onChangeCards={handleChangeDisplayType}
           sortOrder={sortOrder}
-          onChangeSortOrder={setSortOrder}
+          onChangeSortOrder={handleSetSortOrder}
         />
       )}
       {loading && <Loading />}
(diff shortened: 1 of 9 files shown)


src/modules/search/components/SearchResultsHeader.tsx
```diff
@@ -65,6 +73,20 @@ const SearchResultsHeader = ({
...
+          }
+          select
+          label='Sort Order'
+        >
+          {Object.values(ClientSortOption).map((sortOrder) => (
+            <MenuItem key={sortOrder} value={sortOrder}>
+              {HmisEnums.ClientSortOption[sortOrder]}
+            </MenuItem>
+          ))}
+        </TextField>
```
gigxz: Let's use the `GenericSelect` input so that the styling is consistent with other dropdowns.


Here's a working example. The generated `HmisEnums` object isn't typed correcty yet (but it could be!) so some casting is needed for the option list:



```
        <GenericSelect<ClientSortOption, false, false>
          sx={{ width: 250 }}
          getOptionLabel={(option) => HmisEnums.ClientSortOption[option]}
          label='Sort Order'
          onChange={(_e, value) => value && onChangeSortOrder(value)}
          value={sortOrder}
          options={
            Object.keys(HmisEnums.ClientSortOption) as ClientSortOption[]
          }
        />
```

This will lift the label up above, so we can you add a label for the display option as well so they are aligned? Like this:

Screen Shot 2022-12-16 at 10 46 38 AM

src/modules/search/components/ClientSearch.tsx
```diff
@@ -205,6 +211,8 @@ const ClientSearch: React.FC<Props> = ({
           disabled={!hasResults}
           cardsEnabled={!!cards}
           onChangeCards={handleChangeDisplayType}
+          sortOrder={sortOrder}
+          onChangeSortOrder={setSortOrder}
```
gigxz: Can we make it so changing the sort order also brings the offset back to 0?

ZachCMP: Fixed

src/modules/search/components/SearchResultsHeader.tsx
ZachCMP: Fixed

gigxz (approved):
looks good!

PR #104, opened 2022-12-16
183561015 client omnisearch

ZachCMP:
https://www.pivotaltracker.com/story/show/183561015


src/modules/search/components/OmniSearch.tsx
```diff
@@ -0,0 +1,136 @@
...
+                        <Typography variant='overline'>{label}</Typography>
+                        {isEmpty(options) ? (
+                          <Typography color='text.disabled' variant='body2'>
+                            No {label} found
+                          </Typography>
+                        ) : (
+                          options.map((option) => {
+                            return (
+                              <div key={option.id}>
+                                <MenuItem
```
gigxz, line 131: I'm not able to tab through the menu items. Can we update so it is keyboard navigable?

src/modules/search/components/OmniSearch.tsx
```diff
@@ -0,0 +1,136 @@
...
+                                  component={RouterLink}
+                                  to={`/client/${option.id}`}
```
gigxz: I have been using `generatePath` across the board, so we can easily change route paths if needed without changing all the components:

`generatePath(Routes.CLIENT_DASHBOARD, { clientId: id })`

src/modules/search/components/OmniSearch.tsx
```diff
@@ -0,0 +1,136 @@
...
+                                  onClick={() => values.anchorEl?.blur()}
+                                >
+                                  {[option.firstName, option.lastName].join(
+                                    ' '
+                                  )}
```
gigxz: Lets use `clientName(client)` so the name matches the profile.

src/modules/search/components/OmniSearch.tsx
```diff
@@ -0,0 +1,136 @@
...
+import React, { useState } from 'react';
+import { Link as RouterLink } from 'react-router-dom';
+
+import TextInput from '@/components/elements/input/TextInput';
+import { useOmniSearchClientsQuery } from '@/types/gqlTypes';
+
+const OmniSearch: React.FC = () => {
+  const [value, setValue] = useState<string | null>(null);
+
+  const { data, loading } = useOmniSearchClientsQuery({
```
gigxz, line 28: Ideally the results would be sorted by relevancy, not last name. Probably a separate ticket, but let's look at what that would take to add that to the `ClientSearch` concern

src/modules/search/components/OmniSearch.tsx
ZachCMP: Sure, I'll take a look and see if we can do that. There might be some stuff in the `useAutocomplete` hook I can use

src/modules/search/components/OmniSearch.tsx
ZachCMP: @gigxz Taking a cursory look at this, it seems like we could more easily do this if we were actually using the postgres full-text search mechanisms, which would be able to give us a clear ranking. Because it looks like we're currently doing a multi-condition search, we would probably have to roll our own ranking logic based on conditions matched, which honestly sounds like a nightmare.

One option to solve this could be to pre-generate the text search vectors for a given client, persist that information into the clients table (or some other place), and then search on it whenever we need to do text search ([source](https://www.compose.com/articles/mastering-postgresql-tools-full-text-search-and-phrase-search/)). That would mean adding some mechanism that would assemble the correct search vectors and handle maintaining that persisted information, and then refactoring the client search to use the text vectors instead of the multi-condition search it uses now. Alternatively we could also generate the vectors on every search (we do this in HNMI), which would trade the denormalization maintenance cost for a performance cost per query.

I'm not deeply experienced in postgres text searching–the above option came up from some googling and some experience in HNMI–so I don't know what the lift would be, or if it's even feasible. Definitely worth a new issue to handle that work, I think it's outside the scope here.

gigxz: Interesting, thanks for the research. Added this info to https://www.pivotaltracker.com/story/show/184065163

src/modules/search/components/OmniSearch.tsx
ZachCMP: @gigxz This is fixed. it doesn't actually do tab navigation, as that would make the autocomplete behave poorly, but the arrow key navigation now works as expected. I also cleaned up some of the routing behavior and added an indication if we're currently looking at a given client in the list. Let me know if you need me to do more here

src/modules/search/components/OmniSearch.tsx
ZachCMP: fixed

src/modules/search/components/OmniSearch.tsx
ZachCMP: fixed

src/modules/search/components/OmniSearch.tsx
gigxz: Nice. This looks good, I'll merge and deploy! Then we will hopefully get some of Ash's time this week to give feedback on the interactions and UI.

PR #108, opened 2022-12-20
Add project search

ZachCMP:
https://www.pivotaltracker.com/story/show/184046922

Relies on https://github.com/greenriver/hmis-warehouse/pull/2700


ZachCMP:
@gigxz I'd like to wait until we have the backend PR and any other GraphQL schema changes merged before I fix the merge conflicts. I don't want to introduce more problems by merging schemas that aren't based on the schema in the current release.

src/modules/search/components/OmniSearch.tsx
```diff
@@ -11,37 +11,82 @@ import {
...
+  const getOptionTargetPath = useCallback(
+    (option: NonNullable<typeof options>[number]) => {
+      let targetPath: string | null = null;
+      if (option.__typename === 'Client') {
+        targetPath = generatePath(Routes.CLIENT_DASHBOARD, {
+          clientId: option.id,
+        });
+      }
+      if (option.__typename === 'Project') {
+        targetPath = generatePath(Routes.EDIT_PROJECT, {
```
gigxz: I think it should go to VIEW_PROJECT

src/modules/search/components/OmniSearch.tsx
```diff
@@ -11,37 +11,82 @@ import {
...
+          projectId: option.id,
+        });
+      }
+      return targetPath;
+    },
+    []
```
gigxz, line 65: As a rule I try to move functions that don't have dependencies outside of the component, so they're not redeclared every render. (like this callback and the one below). The performance impact is probably not noticable, though, so it's more of a style thing. Fine with leaving as-is

gigxz (approved):
LGTM after fixing project path. I had a comment about the search behavior, left that on the backend PR

src/modules/search/components/OmniSearch.tsx
ZachCMP: @gigxz I remember why I did this now. I wanted to use `typeof options` in the type definitions so I could take advantage of the TS type detection on the `options` variable. Since that's defined inside the component body, I had to put these here to make that work. If I move them outside the component I'd have to statically type them, which is fine, but I like that a lot less than being able to use the inferred type from `options`.

ZachCMP: Also there should be pretty much zero performance impact to have them in the component with no deps. That should actually prevent them from being redeclared every render, since `useCallback` will only redeclare them if the deps change. No deps means it'll declare them on component mount and then keep that reference through the entire lifetime of the component.

src/modules/search/components/OmniSearch.tsx
ZachCMP: Fixed

src/modules/search/components/OmniSearch.tsx
gigxz: Ok sounds good! yeah it will keep the same reference to the function, but it will still redeclare the variable--  agreed near zero impact, more of a style thing like I said. keeping it here for the types makes sense. lgtm to merge once the backend PR goes in!

PR #113, opened 2022-12-23
184052669 ssn input

ZachCMP:
https://www.pivotaltracker.com/story/show/184052669

Relies on https://github.com/greenriver/hmis-warehouse/pull/2714 for the added enum value and updated form definition


package.json
```diff
@@ -78,8 +79,10 @@
     "react-error-boundary": "^3.1.4",
     "react-i18next": "^12.0.0",
     "react-router-dom": "^6.3.0",
+    "storybook-addon-state": "^1.0.3",
     "typescript": "^4.4.2",
-    "vite": "^3.0.0"
+    "vite": "^3.0.0",
+    "yarn": "^1.22.19"
```
gigxz: Why is yarn added? The binary in `.yarn/releases` should be used, so that we're all using the same version

src/components/elements/input/SsnInput.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+import LabelWithContent from '../LabelWithContent';
+import MultiFieldInput from '../MultiFieldInput';
+
+type SsnInputProps = {
+  onChange?: (value: string) => any;
+  onlylast4?: boolean;
+} & Omit<TextFieldProps, 'onChange'>;
+
+const SsnInput = ({
```
gigxz, line 15: Can we set a placeholder by default that has the correct number of zeros in each field? The placeholder should only appear if there are NO input values in any field.

The convention we are following is to use placeholders for dropdowns ("Select <x>...") and for fields with masked inputs (like this, and `mm/dd/yyyy`).


I tried entering the placeholder prop in storybook but it looks wonky:
Screen Shot 2022-12-23 at 12 59 01 PM

src/components/elements/input/SsnInput.tsx
```diff
@@ -0,0 +1,132 @@
...
+  inputProps,
+  InputProps,
+  value,
+  label,
+  onChange,
+  helperText,
+  onlylast4 = false,
```
gigxz, line 22: I don't think we have a use case for this, but seems fine to keep. It looks like the first placeholder is cut off, but we can wait to worry about that until/if we use this variant

Screen Shot 2022-12-23 at 1 02 13 PM

gigxz (changes requested):
looks excellent!! requested changes: remove yarn package, and add zeros placeholder

package.json
ZachCMP: No idea, I'll remove it. Maybe I had a typo in my `yarn add` command for `storybook-addon-state`

src/components/elements/input/SsnInput.tsx
ZachCMP: Yeah, I noticed that, sounds good.

src/components/elements/input/SsnInput.tsx
ZachCMP: Fixed

ZachCMP:
@gigxz All things fixed. Let me know if you need more on this.

PR #115, opened 2023-01-04
183190260 client photo upload

ZachCMP:
https://www.pivotaltracker.com/story/show/183190260


ZachCMP:
@gigxz It looks like some cypress tests are failing, but I'm not sure it's related to my work. Can you take a look and see if it's related?

gigxz:
Yeah I'll take a look.

Branch needs https://github.com/greenriver/hmis-warehouse/pull/2679  - linking for future reference

gigxz:
If you merge in the latest `stable` it should fix it. Needed to add a mock for Base64: https://github.com/greenriver/hmis-frontend/pull/116/files

src/components/elements/input/ClientImageUploadDialog.tsx
```diff
@@ -0,0 +1,111 @@
...
+    <Dialog {...omit(props, 'children')} onClose={onClose}>
+      <DialogTitle>Upload Client Photo</DialogTitle>
+      <DialogContent>
+        {loading ? (
+          <CircularProgress />
+        ) : (
+          client && (
+            <Grid container spacing={2} justifyContent='space-around'>
+              <Grid item>
+                <Typography>Current Photo</Typography>
```
gigxz: Can we hide the "Current Photo" and "New Photo" labels if the client doesn't already have a photo?

src/components/elements/input/ClientImageUploadDialog.tsx
```diff
@@ -0,0 +1,111 @@
...
+                {client.image && <ClientCardImageElement client={client} />}
+              </Grid>
+              <Grid item>
+                <Typography>New Photo</Typography>
+                {client.image && <ClientCardImageElement url={newPhotoSrc} />}
+              </Grid>
+              <Grid item xs={12}>
+                <Uploader
```
gigxz, line 103: Not everyone is familiar with the drag and drop. There should also be an "Choose File" button that pulls up the standard file selection tool, both here and in the form. Thanks!

ZachCMP: Fixed

src/components/elements/input/ClientImageUploadDialog.tsx
ZachCMP: done

ZachCMP:
@gigxz I implemented these fixes:

- Clicking on uploader will now open the normal file dialog
- Current and New image sections in the dialog are not shown if the client has an image already
- You can now clear the uploaded file in the uploader

In terms of removing the photo from the client, I think we need to have some more conversation about that. Since the image that's shown is just the most recent client headshot, it's unclear what the action should be. Here's some options:

1. Just remove the most recent one. Currently that will mean that the photo will change to the most recent one before the deleted one.
2. Remove all photos for the client. This will definitely clear the image, but doesn't seem right to me.
3. Add some mechanism to point to a specific photo explicitly. We can then remove the link and the photo to clear the image.

eanders:
> @gigxz I implemented these fixes:
> 
> * Clicking on uploader will now open the normal file dialog
> * Current and New image sections in the dialog are not shown if the client has an image already
> * You can now clear the uploaded file in the uploader
> 
> In terms of removing the photo from the client, I think we need to have some more conversation about that. Since the image that's shown is just the most recent client headshot, it's unclear what the action should be. Here's some options:
> 
> 1. Just remove the most recent one. Currently that will mean that the photo will change to the most recent one before the deleted one.
> 2. Remove all photos for the client. This will definitely clear the image, but doesn't seem right to me.
> 3. Add some mechanism to point to a specific photo explicitly. We can then remove the link and the photo to clear the image.

Hmm, good points on the removal will revert to the previous image, hadn't thought about that.  I think that should be the behavior though.  They can always remove that one too.  I see the removal as an "oops, I uploaded a picture of Joe, but this is James" situation.  Once we have general file management we may just direct people there, and maybe we can just give them a button to "make this the client photo" which touches the record to up the modification date.

I was confused about the current and new labels until I saw it with a current image showing.  Did you mean those are hidden _unless_ you have a current file?

ZachCMP:
@eanders Okay, so I can implement it such that it removes the most recent image for now. I think it makes sense to fix it later once we have some more file management.

And yes, the Current/New section is hidden unless there is a current image (i.e. `client.image` is non-null).

eanders:
Thanks!

gigxz:
I agree with all that!
The build is failing because of some flakiness that I'm looking into, you can ignore it

ZachCMP:
@eanders @gigxz I implemented the image removal logic mentioned above. Same changes were made in this PR, and some were made in the [warehouse PR](https://github.com/greenriver/hmis-warehouse/pull/2679)

gigxz (approved):
Looks good!

PR #127, opened 2023-01-11
Add user dropdown

ZachCMP:
https://www.pivotaltracker.com/story/show/184093713


package.json
```diff
@@ -75,6 +75,7 @@
     "jest": "^28.1.0",
     "jest-environment-jsdom": "^28.1.0",
     "lodash-es": "^4.17.21",
+    "material-ui-popup-state": "^5.0.4",
```
gigxz, line 78: What's your workflow for adding dependencies? I typically would run `yarn add material-ui-popup-state`.  It doesn't look like this library requires anything that we don't already have, so all these tarballs shouldn't be showing up.

src/components/layout/UserMenu.tsx
```diff
@@ -0,0 +1,67 @@
...
+          horizontal: 'right',
+        }}
+        {...bindMenu(popupState)}
+      >
+        <MenuItem component={Link} href={getManageAccountLink()}>
+          <ListItemIcon>
+            <OpenInNewIcon />
+          </ListItemIcon>
+          <ListItemText>Manage Account</ListItemText>
+        </MenuItem>
```
gigxz: we should only show this menu item if `import.meta.env.PUBLIC_WAREHOUSE_URL` is defined

src/components/layout/UserMenu.tsx
```diff
@@ -0,0 +1,67 @@
...
+          vertical: 'bottom',
+          horizontal: 'right',
+        }}
+        transformOrigin={{
+          vertical: 'top',
+          horizontal: 'right',
+        }}
+        {...bindMenu(popupState)}
+      >
+        <MenuItem component={Link} href={getManageAccountLink()}>
```
gigxz:
```suggestion
        <MenuItem component={Link} href={getManageAccountLink()} target='_blank'>
```

src/components/layout/UserMenu.tsx
```diff
@@ -0,0 +1,67 @@
...
+          <ListItemIcon>
+            <OpenInNewIcon />
```
gigxz:
```suggestion
            <OpenInNewIcon fontSize='small' />
```

src/components/layout/UserMenu.tsx
```diff
@@ -0,0 +1,67 @@
...
+          </ListItemIcon>
+          <ListItemText>Manage Account</ListItemText>
+        </MenuItem>
+        <MenuItem onClick={logoutUser}>
+          <ListItemIcon>
+            <LogoutIcon />
```
gigxz:
```suggestion
            <LogoutIcon fontSize='small' />
```

package.json
ZachCMP: Yeah, that's what I did. I don't know why it would include those if we already had the deps

src/components/layout/UserMenu.tsx
ZachCMP: fixed

PR #129, opened 2023-01-11
182883868 obfuscate id urls

ZachCMP:
https://www.pivotaltracker.com/story/show/182883868

I based the encoder class on the ruby class. However, I don't have a depth of knowledge in cryptographic algorithms, so the hashes generated don't match the backend, so there's more work to do to align the algorithm. I'm happy to go back and take another crack at it, but someone more knowledgeable about those algorithms could probably do it much faster than me.


src/modules/hmis/ClientIdEncoder.ts
```diff
@@ -0,0 +1,64 @@
...
+  static encode(id: number | string): string {
+    if (!this.PROTECT_IDS) return String(id);
+
+    const dayStamp = startOfToday().valueOf() / (60 * 60 * 24);
+
+    return this.obfuscate(parseInt(String(id)), dayStamp);
+  }
+
+  static obfuscate(id: number, dayStamp: number): string {
+    // TODO: Make this match the backend
```
eanders: If this obfuscates and deobfuscates, I'm not sure why it needs to match the back-end.  It's not like the id will match, or that the paths would be the same.

eanders (approved):

src/modules/hmis/ClientIdEncoder.ts
gigxz: Agreed, they don't need to match

gigxz (commented):
This looks like a good start! Working as expected for links to the client dashboard from the client card. It just needs to be expanded to apply to all links and routes:

1. Obfuscation should be implemented for all database IDs, not just Clients. [Here are all the different routes we have](https://github.com/greenriver/hmis-frontend/blob/stable/src/routes/routes.ts). Let's go with obfuscating any param that ends with `Id`. That will, for example, obfuscate `:assessmentId` while leaving `assessmentRole`as-is.
2. Instead of calling `ClientIdEncoder.decode(param)` each time we want to access an ID in a component, I think it would be cleaner to replace all usages of `useParams()` with a custom hook like `useSafeParams()` that takes care of decoding all of the IDs.
3. Instead of calling `ClientIdEncoder.encode(param)` each time we are generating a path, it would be cleaner to replace all usages of `generatePath(path, params)` with `generateSafePath(path, params)` and let the `generateSafePath` handle obfuscating any database IDs if necessary.

src/env.d.ts
```diff
@@ -5,6 +5,9 @@ interface ImportMetaEnv {
   readonly PUBLIC_WAREHOUSE_URL: string;
   readonly PUBLIC_CAS_URL: string;
   readonly PUBLIC_SENTRY_DSN: string;
+  readonly PUBLIC_PROTECTED_IDS: string;
```
gigxz, line 8: Is there a reason this can't be a `boolean`?

src/modules/hmis/ClientIdEncoder.ts
```diff
@@ -0,0 +1,64 @@
+import CryptoJS from 'crypto-js';
+import { startOfToday } from 'date-fns';
+
+export default class ClientIdEncoder {
```
gigxz: Can you add some unit tests for this class @ZachCMP ?

src/modules/hmis/ClientIdEncoder.ts
```diff
@@ -0,0 +1,64 @@
...
+    const encrypted = atob(slug.replace(this.DELIMITER_REGEX, ''));
+
+    const keyHex = CryptoJS.enc.Base64.parse(this.KEY);
+
+    const decrypted = CryptoJS.DES.decrypt(encrypted, keyHex, {
+      mode: CryptoJS.mode.ECB,
+    }).toString(CryptoJS.enc.Utf8);
+
+    const composed = parseInt(decrypted, 16);
+    const dayStamp = composed >> 32;
```
eanders: Let's note for the future we could expire these based on the date stamp.

src/env.d.ts
ZachCMP: @gigxz I'm not sure. I think this is the vite environment config, and I'm not sure whether we've parsed the env vars by the time we're getting here. If not, then the value of `PUBLIC_PROTECTED_IDS` would be a string–`'true' or 'false' or ''`. I'm not very familiar with Vite, so I'm not sure.  Do you know if this represents a raw value or a parsed value?

gigxz: Not sure off the top of my head but I just tried it and it's not parsed

src/modules/hmis/ClientIdEncoder.ts
ZachCMP: done

ZachCMP:
@gigxz I addressed the feedback and pushed. I'll re-request a review from you in PT

gigxz (approved):
Looks good!! I ran all the cypress tests locally and everything seems to be working correctly. I'll enable the obfuscation in QA in a separate PR.

PR #145, opened 2023-01-20
183046855 add service

ZachCMP:
https://www.pivotaltracker.com/story/show/183046855

Requires https://github.com/greenriver/hmis-warehouse/pull/2775


src/components/elements/input/GenericSelect.tsx
```diff
@@ -41,18 +47,33 @@ const GenericSelect = <
...
+          }}
           disabled={rest.disabled}
           // Only render placeholder if no values are selected
           placeholder={hasValue(value) ? undefined : placeholder}
           label={label}
         />
       )}
       {...rest}
+      // Allow parent to control disabled state; Otherwise just disable when loading
+      disabled={!isNil(rest.disabled) ? rest.disabled : rest.loading}
```
gigxz: I don't think we need this. You should still be able to click on the element, you just won't be able to select an option yet. I think it's more frustrating for a user to not be able to click at all (and would require updates to cypress commands). These are the 2 states you could get into (given the suggested changes) and I think they're fine:

Screen Shot 2023-01-20 at 3 28 55 PM

Screen Shot 2023-01-20 at 3 39 08 PM

src/components/elements/input/GenericSelect.tsx
```diff
@@ -41,18 +47,33 @@ const GenericSelect = <
         <TextInput
           {...params}
           {...inputProps}
-          InputProps={{ ...params.InputProps, ...inputProps.InputProps }}
+          InputProps={{
+            ...params.InputProps,
+            startAdornment: rest.loading ? (
+              <InputAdornment position='start'>
```
gigxz: We should only show the indicator if the selector has a value. Or else we are causing an unnecessary glitchy flash while the loading indicator shows up and then disappears.

```suggestion
            // Show a loading indicator if we have a value but the picklist is still loading
            startAdornment: rest.loading && params.inputProps.value ? (
              <InputAdornment position='start' sx={{ pl: 1 }}>
```

src/components/pages/Service.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+const Service = ({ create = false }: { create?: boolean }) => {
+  const navigate = useNavigate();
+  const { clientId, enrollmentId, serviceId } = useSafeParams() as {
+    enrollmentId: string;
+    serviceId: string;
+    clientId: string;
+  };
+  const title = create ? `Add Service` : `Edit Service`;
+  // const [crumbs, crumbsLoading, enrollment] = useProjectCrumbs(title);
```
gigxz:
```suggestion
```

gigxz (commented):
Left some comments on the new loading indicator. All the services pieces look good! The validation behavior is off but that should all be addressed in the backend PR.

src/components/elements/input/GenericSelect.tsx
ZachCMP: fixed

src/components/elements/input/GenericSelect.tsx
ZachCMP: fixed

src/components/pages/Service.tsx
ZachCMP: fixed

gigxz (approved):

ZachCMP:
@gigxz Fixed merge conflicts

gigxz (approved):

PR #146, opened 2023-01-23
184093619 new client profile

ZachCMP:
https://www.pivotaltracker.com/story/show/184093619

Couple notes:

- I wasn't sure what you meant by the fetchPolicy task in the PT story. Happy to make that change if you point me in the right direction
- Some of the content for the client accordions is just `"TK"`, since none was provided in the mocks
- The actions in the "Quick Actions" card don't match the mocks, they are just what was available on the old client card. Ash said this was fine and made sense, so that's what I went with.
- I came up with some treatments for the "uploading" and "uploaded" states of the uploader component, which were not in the mocks. I showed Ash and she said they looked good, but just note that there's no reference for them.
- Overall I erred on the side of the mocks, even if the content there seemed placeholder or not quite right. Please feel free to correct my assumptions there as needed.


gigxz (approved):
Thanks @ZachCMP this is great!! It looks so much better.
I updated a few things, notably:
1) Using the `HmisEnum` component to render enums instead of using `HmisEnums` directly. This componet renders the "invalid value" enum in red, as well as rendering "non-value" enum values (like data not collected) in lighter gray.
2) Fixing some validateDomNesting errors that appeared because of nested buttons, and tables nested within paragraphs
3) Some style changes to the accordion to make it match the accordions we have on the organization page. I will update those to use the same SimpleAccordion component. Thanks for creating that!

PR #153, opened 2023-02-02
184349231 delete services

ZachCMP:
https://www.pivotaltracker.com/story/show/184349231

No backend work was needed


gigxz (approved):
thanks!

PR #156, opened 2023-02-06
183046860 bulk add services

ZachCMP:
https://www.pivotaltracker.com/story/show/183046860

Needs https://github.com/greenriver/hmis-warehouse/pull/2801

### DIFF
--- a/src/components/pages/BulkAddServices.tsx
+++ b/src/components/pages/BulkAddServices.tsx
@@ -1,4 +1,4 @@
-import { Stack, Typography } from '@mui/material';
+import { Button, Stack, Typography } from '@mui/material';
 
 import Loading from '../elements/Loading';
 
@@ -34,14 +34,31 @@ const BulkAddServices = () => {
         AddServiceToEnrollmentMutationVariables
       >
         mutationDocument={AddServiceToEnrollmentDocument}
-        renderTable={(additionalColumns) => (
+        renderList={(items, { onSelect }) => (
           <ProjectEnrollmentsTable
+            noLinks
             projectId={projectId}
-            additionalColumns={additionalColumns}
+            additionalColumns={[
+              ...items.map((item) => ({
+                header: item.label,
+                render: () => item.node,
+              })),
+              {
+                header: '',
+                render: (enrollment) => (
+                  <Button
+                    color='secondary'
+                    onClick={() => onSelect(enrollment)}
+                  >
+                    Assign
+                  </Button>
+                ),
+              },
+            ]}
           />
         )}
         definitionIdentifier='service'
-        getInputFromTarget={(formData, enrollment) => ({
+        getInputFromItem={(formData, enrollment) => ({
           input: { ...formData, enrollmentId: enrollment.id },
         })}
         getErrors={(data) => data.createService?.errors}
--- a/src/components/pages/BulkAddServices.tsx
+++ b/src/components/pages/BulkAddServices.tsx
@@ -4,50 +4,56 @@ import Loading from '../elements/Loading';
 
 import { InactiveChip } from './Project';
 
-import EditRecord from '@/modules/form/components/EditRecord';
+import useSafeParams from '@/hooks/useSafeParams';
+import BulkAdd from '@/modules/bulk/components/BulkAdd';
+import ProjectEnrollmentsTable from '@/modules/inventory/components/ProjectEnrollmentsTable';
 import ProjectLayout from '@/modules/inventory/components/ProjectLayout';
 import { useProjectCrumbs } from '@/modules/inventory/components/useProjectCrumbs';
 import {
-  ServiceFieldsFragment,
-  UpdateProjectCocDocument,
-  UpdateProjectCocMutation,
-  UpdateProjectCocMutationVariables,
+  AddServiceToEnrollmentDocument,
+  AddServiceToEnrollmentMutation,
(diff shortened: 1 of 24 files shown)


src/components/pages/BulkAddServices.tsx
```diff
@@ -0,0 +1,125 @@
...
+                          <CircularProgress color='inherit' size={15} />
+                        ) : enrollmentsAdded.includes(enrollment.id) ? (
+                          <CheckIcon />
+                        ) : undefined
+                      }
+                    >
+                      {enrollmentsAdded.includes(enrollment.id)
+                        ? 'Assigned'
+                        : 'Assign'}
+                    </Button>
```
gigxz: I'm seeing a bug that is allowing me to create duplicate records if I assign services to multiple clients quickly without waiting for each one to finish loading. The loading spinner disappears because we only have one "workingEnrollmentId" even though several can be "working" at the same time. I'm able to go click on a previously assigned service again, and it creates a duplicate service.

A quick way to fix this might be to change `workingEnrollmentId` state to a list ("loadingEnrollmentIds") so you can store all of them, and remove them one by one as mutations succeed. Finally, you want to make the button `disabled` if the mutation is currently loading.

The way I handled this for adding household members was to pull the mutation and loading state into the button itself (`AddToHouseholdButton`) which keeps track of it's own loading/success/error status, and calls `onSuccess` hook when it has successfully completed. I think that approach could work well here too, and that could potentially be refactored into a generic component MutationButton. I'll keep that in mind, but for now, if the above suggestion works to fix this bug, I'd go with that.

src/modules/bulk/components/BulkAdd.tsx
```diff
@@ -0,0 +1,250 @@
...
+          ...acc,
+          {
+            ...item,
+            item: traverse(item.item),
+          },
+        ];
+      }
+      return [...acc, item];
+    }, [] as FormItem[]);
+  };
```
gigxz: Let's move this function `extractClientItemsFromDefinition` into `formUtil.ts`.

I also suggest replacing the recursion with a call to `getItemMap` which flattens the form definition. I try to use that wherever possible to avoid repeating the recursion logic all over the place.


```suggestion
const extractClientItemsFromDefinition = (definition: FormDefinitionJson) => {
    const itemMap = getItemMap(definition, false); // flattened map { linkId => item }
    const targetItems = Object.values(itemMap).filter(
      ({ serviceDetailType }) => serviceDetailType === ServiceDetailType.Client
    );
```

src/modules/form/components/DynamicFormFields.tsx
```diff
@@ -0,0 +1,242 @@
...
+
+  // Recursively render an item
+  const renderItem = (
+    item: FormItem,
+    nestingLevel: number,
+    props?: OverrideableDynamicFieldProps,
+    renderFn?: (children: ReactNode) => ReactNode
+  ) => {
+    const isDisabled = !isEnabled(item, disabledLinkIds);
+    if (!isShown(item, disabledLinkIds)) return null;
```
gigxz: This is running `isEnabled` twice which is unnecessary and potentially expenseive.

I think it's clearer written as it was before:

```suggestion
    const isDisabled = !isEnabled(item, disabledLinkIds);
    if (isDisabled && item.disabledDisplay !== DisabledDisplay.Protected) return null;
```

src/modules/form/components/DynamicFormFields.tsx
```diff
@@ -0,0 +1,242 @@
...
+
+export const isShown = (item: FormItem, disabledLinkIds: string[] = []) => {
+  if (
+    !isEnabled(item, disabledLinkIds) &&
+    item.disabledDisplay !== DisabledDisplay.Protected
+  )
+    return false;
+
+  return true;
+};
```
gigxz, line 65: Can we move this into formUtil, maybe as `itemIsEnabled` and `itemIsVisible`

src/modules/form/hooks/useDynamicFormFields.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const getCleanedValues = useCallback(() => {
+    if (!definition) return values;
+    const excluded = addDescendants(disabledLinkIds, definition);
+    return omit(values, excluded);
+  }, [definition, disabledLinkIds, values]);
+
+  const shouldShowItem = useCallback(
+    (item: FormItem) => isShown(item, disabledLinkIds),
+    [disabledLinkIds]
+  );
```
gigxz, line 46: `shouldShowItem` feels somewhat unnecessary to pass through this hook since it's only used in one place

src/modules/form/hooks/useDynamicFormFields.tsx
```diff
@@ -0,0 +1,100 @@
...
+  isShown,
+  Props as DynamicFormFieldsProps,
+} from '../components/DynamicFormFields';
+import { addDescendants, FormValues } from '../util/formUtil';
+
+import useComputedData from './useComputedData';
+
+import { FormDefinitionJson, FormItem } from '@/types/gqlTypes';
+
+const useDynamicFormFields = ({
```
gigxz, line 14: This refactor looks great!

ZachCMP:
@gigxz Fixes the the feedback above is fixed and pushed

gigxz (approved):

PR #157, opened 2023-02-06
184086440 fix omnisearch

ZachCMP:
https://www.pivotaltracker.com/story/show/184086440

### DIFF
--- a/src/modules/search/components/OmniSearch.tsx
+++ b/src/modules/search/components/OmniSearch.tsx
@@ -92,7 +92,6 @@ const OmniSearch: React.FC = () => {
   const getOptionTargetPath = useCallback(
     (option: Option) => {
       let targetPath: string | null = null;
-
       if (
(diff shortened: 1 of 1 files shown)


gigxz (changes requested):
Thanks, this fixes one bug! But there is still another:

1. Enter a search term
2. Click on an item
3. Click on the app title to navigate home (or navigate anywhere)
4. Click on the omnisearch box (which still has the query in it)
5. Click on the same item you clicked on before

Expected behavior: navigate to the item I clicked on
Actual behavior: nothing happens

ZachCMP:
@gigxz Fixed. The autocomplete thought that it had already selected that item and therefore its "value" had not actually changed. Since we're only using the autocomplete for its UX and not as an actual input here, I chose to override its value mechanism by passing `value: null`

gigxz (approved):

PR #163, opened 2023-02-10
Add custom theming

ZachCMP:
https://www.pivotaltracker.com/story/show/182931779

Backend PR: https://github.com/greenriver/hmis-warehouse/pull/2816


src/env.d.ts
```diff
@@ -8,6 +8,8 @@ interface ImportMetaEnv {
   readonly PUBLIC_PROTECTED_IDS: string;
   readonly PUBLIC_INITIAL_DELIMITER: string;
   readonly PUBLIC_PROTECTED_ID_KEY: string;
+  readonly PUBLIC_ALLOW_SET_THEME: string;
+  readonly PUBLIC_DEFAULT_THEME: string;
```
gigxz: I don't think we need this env var here. The default should be decided by the backend.

gigxz (approved):
Looks great, thanks!

src/env.d.ts
ZachCMP: How do you want the frontend in QA to be able to change themes? Would that also use a backend-only mechanism, or do we need some mechanism on the frontend to do that?

gigxz: I think the `window` mechanism you added for QA-ing is good! We don't need `PUBLIC_DEFAULT_THEME` for that approach to work.

src/config/useMergedTheme.ts
```diff
@@ -0,0 +1,45 @@
...
+  useEffect(() => {
+    let newTheme = fullTheme;
+
+    try {
+      const merged = createTheme(
+        deepmerge(
+          deepmerge(baseThemeDef, fullThemeDef),
+          remoteThemeOptions || {}
+        )
+      );
```
gigxz: Ah @ZachCMP sorry I missed this the first time. This is not allowing us to "compose" themes. The reason I split up the "base" and the theme in the way I did is to allow for theme composition: see [Theme composition: using theme options to define other options](https://mui.com/material-ui/customization/theming/#createtheme-options-args-theme).

I was trying to figure out how this was feasible on your branch, and I basically ended up implementing that, so I pushed that up.. sorry for no warning on that! Can you take a look at https://github.com/greenriver/hmis-frontend/pull/163/commits/a02358a554147605398d6b2262668adcb952064f and let me know what you think?

I am looking at the outlined MuiButtons as an example. On the Enrollment page there are several. In that commit, I updated it to set the background for those to `theme.palette.background.paper`. In order for that to work, we need the actual composition. Deepmerging everything won't do the trick. In the "halloween" they successfully have a black background.

ZachCMP: @gigxz Gotcha. I looked at your changes and they look good to me. The most important part is that the base theme and the remote theme are combined via `deepmerge` and not `createTheme`, which you have implemented there. So that all looks good to me.

PR #172, opened 2023-02-24
Add graphql permissions

ZachCMP:
https://www.pivotaltracker.com/story/show/184497361

Requires backend code in https://github.com/greenriver/hmis-warehouse/pull/2840


src/modules/permissions/useHasPermissionsHooks.tsx
```diff
@@ -0,0 +1,46 @@
...
+export type ProjectPermissions = keyof Omit<ProjectAccess, '__typename'>;
+export const useHasProjectPermissions = (
+  id: string,
+  permissions: ProjectPermissions[],
+  mode?: PermissionsMode
+) => {
+  const projectPermissions = useGetProjectPermissionsQuery({
+    variables: { id },
+  })?.data?.project?.access;
+  return useHasPermissions(projectPermissions, permissions, mode);
```
gigxz: We shouldn't ignore the loading state here

src/components/pages/EditProject.tsx
```diff
@@ -44,6 +49,7 @@ const EditProject = () => {
 
   if (loading) return <Loading />;
```
gigxz: This shoudl consider whether the project permission query is loading too

src/components/pages/Organization.tsx
```diff
@@ -72,6 +72,7 @@ const Organization = () => {
           )}
         </Grid>
         <Grid item xs>
+          {organization?.contactInformation}
```
gigxz: Why is this added?

src/components/pages/EditProject.tsx
gigxz: Actually, this is wrapped in `ProjectEditRoute`. So we shouldn't need to make. project permission query in this component at all.

gigxz (changes requested):

src/components/pages/Organization.tsx
ZachCMP: I had made the create project action use the `can_edit_project_details` permission, in which case we should totally hide this paper if there's no contact info and we can't create a project. So I think this was initially a condition to check for that. I thought better of using that permission, so I removed the permission filter, but must have not completely removed the code that would hide the paper, thus this line. I'll remove it

src/components/pages/EditProject.tsx
ZachCMP: fixed

src/modules/permissions/useHasPermissionsHooks.tsx
ZachCMP: I made this return a tuple with the form `[<permission boolean>, <query data>]` so that we can get everything we could need from query data.

ZachCMP:
@gigxz Addressed all your feedback. I'll branch off this for the rest of the react work around the root permissions

gigxz (approved):

PR #173, opened 2023-02-28
Implement view and edit client permissions

ZachCMP:
https://www.pivotaltracker.com/story/show/184497372

Relies on backend PR https://github.com/greenriver/hmis-warehouse/pull/2848, which adds more permissions to the root access type.


gigxz (approved):
Looks great!

PR #175, opened 2023-03-10
184630467 add more permissions

ZachCMP:
https://www.pivotaltracker.com/story/show/184630467


gigxz:
@ZachCMP is there a backend PR to test this with? I need the `age` field (among other things, maybe).

ZachCMP:
@gigxz Yeah, looks like there is a dependency there. I initially thought that this was independent, but looking again it looks like I was incorrect. I'll get it up soon, I'm just adding tests for the permissions now. Apologies for jumping the gun!

src/modules/client/components/ClientProfileCard.tsx
```diff
@@ -18,15 +18,16 @@ import ClientImageUploadDialog from '@/components/elements/input/ClientImageUplo
 import NotSpecified from '@/components/elements/NotSpecified';
 import SimpleAccordion from '@/components/elements/SimpleAccordion';
 import SimpleTable from '@/components/elements/SimpleTable';
-import ClientDobAge from '@/modules/hmis/components/ClientDobAge';
-import ClientSsn from '@/modules/hmis/components/ClientSsn';
+import { ClientSafeDobAge } from '@/modules/hmis/components/ClientDobAge';
```
gigxz: I don't think we want to have two components (`ClientSafeDobAge`/`ClientDobAge`) because we should _always_ use the safe one. Can you make that the default export, and either remove the inner component or stop exporting it from the file.

src/components/pages/EnrollmentRoute.tsx
```diff
@@ -0,0 +1,64 @@
...
+}) => {
+  const { [clientIdParam]: clientId, [enrollmentIdParam]: enrollmentId } =
+    useSafeParams();
+
+  const [allowed, { loading, data }] = useHasClientPermissions(
+    clientId || '',
+    compact([
+      view ? 'canViewEnrollmentDetails' : null,
+      edit ? 'canEditEnrollments' : null,
+    ])
```
gigxz: You can't edit things you can't view, as a rule.
We should be able to remove the `view` prop, from both this and `ClientRoute`. And replace this with:
```suggestion
     [edit ? 'canEditEnrollments' : 'canViewEnrollmentDetails'],
```

gigxz:
Alright, no worries @ZachCMP. I left a few comments, but I'll come back and re-review once it is ready to test locally.

src/modules/client/components/ClientProfileCard.tsx
ZachCMP: `ClientSafeDobAge` will do a permissions query, whereas the other one is just layout logic, so I think it's a good idea to have and export both. I have no issue making the safe one the default export though. Do you want me to do that?

src/modules/hmis/components/ClientDobAge.tsx
```diff
@@ -2,37 +2,52 @@ import { Stack, Typography, TypographyProps } from '@mui/material';
...
     </Stack>
   );
 };
 
+export const ClientSafeDobAge: React.FC<Props> = (props) => {
+  const { client } = props;
+
+  const [canViewDob] = useHasClientPermissions(client.id, ['canViewDob']);
+
+  return <ClientDobAge client={client} onlyAge={!canViewDob} />;
```
gigxz: The rest of the props need to be passed down

src/modules/hmis/components/ClientSsn.tsx
```diff
@@ -26,4 +28,28 @@ const ClientSsn = ({ client, noValue, lastFour, variant = 'body2' }: Props) => {
...
 
+export const ClientSafeSsn: React.FC<Props> = (props) => {
+  const { client } = props;
+
+  const [canViewFullSsn] = useHasClientPermissions(client.id, [
+    'canViewFullSsn',
+  ]);
+  const [canViewPartialSsn] = useHasClientPermissions(client.id, [
+    'canViewPartialSsn',
+  ]);
```
gigxz: Can we skip these queries if `client.ssn` is already null? The backend won't return the SSN at all if it can't be seen, so it's unnecessary to do an extra check.

src/modules/hmis/components/ClientDobAge.tsx
gigxz: Do we need to make this query? I think we could just as well say `onlyAge={client.dob && !isNil(client.age)}`. The backend should control the DOB visibility already.

src/modules/hmis/components/ClientDobAge.tsx
```diff
@@ -2,37 +2,52 @@ import { Stack, Typography, TypographyProps } from '@mui/material';
...
 
 const ClientDobAge = ({
   client,
   noValue,
   reveal,
   variant = 'body2',
+  onlyAge,
 }: Props) => {
-  if (!client.dob) return <>{noValue}</> || null;
+  if (!client.dob && !client.age) return <>{noValue}</> || null;
```
gigxz: This will mess up if the age is zero (for a baby). I have been using `isNil`.

src/modules/client/components/ClientProfileCard.tsx
gigxz: Hmm. See my comment on https://github.com/greenriver/hmis-frontend/pull/175#discussion_r1132806511. I think the extra query component is redundant, and we should just update `ClientDobAge` to appropriately handle the case where `client.age` is present and `client.dob` is not.

ZachCMP:
@gigxz Backend PR: https://github.com/greenriver/hmis-warehouse/pull/2868

src/modules/hmis/components/ClientSsn.tsx
ZachCMP: @gigxz Yes, we can skip the `canViewFullSsn` query if it's null. However, the `canViewPartialSsn` is not as straightforward. If the SSN is `123456789`, the backend will return `XXXXX6789`, but that is also a valid full SSN, so without checking the permission we can't know whether it only gave us the last four because that's all that we have for that client, or whether it's because we only have permission to see the partial SSN.

If you're okay with it we could just add the permissions to the `ClientIdentificationFieldsFragment` so we just fetch it with the rest of the fields and then should have it present within the `ClientSsn` component. That feels like the easiest way to do it. Does that sound good to you?

src/modules/client/components/ClientProfileCard.tsx
ZachCMP: I just removed the `ClientSafeDob` component and switched to checking for nullability in the original one

src/components/pages/EnrollmentRoute.tsx
ZachCMP: Fixed

src/modules/hmis/components/ClientDobAge.tsx
ZachCMP: Fixed

ZachCMP: Fixed

src/modules/hmis/components/ClientDobAge.tsx
ZachCMP: Fixed

src/modules/hmis/components/ClientSsn.tsx
gigxz: Hey @ZachCMP sorry for the delayed response on this one. I went ahead and made that change while reviewing so you don't need to context switch back to this: https://github.com/greenriver/hmis-frontend/pull/175/commits/c017a3408afb4f29952af0bcb1811aaad7c83d3b

Overall it's looking great, there is just a performance issue: https://github.com/greenriver/hmis-warehouse/pull/2868#discussion_r1135459887

gigxz (approved):
👍

PR #190, opened 2023-03-23
184628960 files UI

ZachCMP:
https://www.pivotaltracker.com/story/show/184713004
https://www.pivotaltracker.com/story/show/184628960
https://www.pivotaltracker.com/story/show/184713017

Requires https://github.com/greenriver/hmis-warehouse/pull/2901


package.json
```diff
@@ -23,9 +23,7 @@
   },
   "lint-staged": {
     "*.{js,ts,jsx,tsx}": [
-      "eslint --fix --ignore-path .gitignore"
-    ],
-    "*": [
+      "eslint --fix --ignore-path .gitignore",
```
ZachCMP: @gigxz This change was made because prettier was trying to format the nginx config file. It seemed to me like it should only format JS files, but please correct me if I'm mistaken about that.

gigxz: Prettier was formatting json and *.md files, which I think we should keep. I'll update to add the docker directory to the prettierignore instead.

src/components/dashboard/files/FileModal.tsx
```diff
@@ -0,0 +1,223 @@
...
+  Dialog,
+  DialogProps,
+  DialogTitle,
+  Divider,
+  IconButton,
+  Pagination,
+  Stack,
+  Typography,
+} from '@mui/material';
+import pdfWorker from 'pdfjs-dist/build/pdf.worker.js?url';
```
gigxz, line 15: Lint is complaining that `'pdfjs-dist' should be listed in the project's dependencies`, which I think is true. Can you add this to package.json

gigxz, line 15: Tests are failing because of this `?url` syntax, what's going on there?

src/components/dashboard/AllFiles.tsx
```diff
@@ -0,0 +1,162 @@
...
+        justifyContent={'space-between'}
+        sx={{ mb: 2, pr: 1, alignItems: 'center' }}
+      >
+        <Typography variant='h4'>All Files</Typography>
+        {canEdit && (
+          <ButtonLink
+            to={generateSafePath(DashboardRoutes.NEW_FILE, {
+              clientId,
+            })}
+            id='add-client-file'
```
gigxz: (for consistency)
```suggestion
            data-testid='addClientFileButton'
```

cypress/e2e/files.cy.ts
```diff
@@ -0,0 +1,75 @@
...
+    viewportHeight: 1000, // extra long height so its easier to see what's going on
+    viewportWidth: 1024,
+  },
+  () => {
+    // Set up new client
+    cy.createClient('Cy First', 'Cy Last');
+
+    // Got to files tab and create new file
+    cy.get('#side-nav-files').click();
+    cy.get('#add-client-file').click();
```
gigxz: for consistency, can we use test IDs instead of html IDs
```suggestion
    cy.testId('addClientFileButton').click();
```

src/components/dashboard/files/useFileActions.tsx
```diff
@@ -0,0 +1,151 @@
...
+          onClick={() => {
+            setFileToDelete(file);
+          }}
+          size='small'
+          variant='outlined'
+          endIcon={
+            deletingIds.includes(file.id) && (
+              <CircularProgress size={15} color='inherit' />
+            )
+          }
```
gigxz: We should use `LoadingButton` from `import LoadingButton from '@/components/elements/LoadingButton';` instead

gigxz:
Record editing isn't working yet. When I go to Edit a file, the "tags" dropdown doesn't populate with my saved values.

src/modules/form/components/DynamicField.tsx
```diff
@@ -1,5 +1,5 @@
 import { Stack, Typography } from '@mui/material';
```
gigxz, line 1: Once this issue https://github.com/greenriver/hmis-warehouse/pull/2901/files#r1148026223 is resolved, you'll also need to make  a change to line 120 of this file to handle the case when `value` is an array

src/components/dashboard/AllFiles.tsx
```diff
@@ -0,0 +1,162 @@
...
+    'canManageOwnClientFiles',
+  ]);
+  const { data: pickListData } = useGetPickListQuery({
+    variables: { pickListType: PickListType.AvailableFileTypes },
+  });
+
+  const columns: ColumnDef<FileFieldsFragment>[] = useMemo(() => {
+    return [
+      {
+        header: 'Name',
```
gigxz, line 47: Wrapped titles get centered for some reason, can we have them left-aligned to match
Screen Shot 2023-03-24 at 4 46 17 PM

src/components/dashboard/AllFiles.tsx
```diff
@@ -0,0 +1,162 @@
...
+                const { editButton, deleteButton } = getActionsForFile(file);
+                return (
+                  <Stack
+                    direction='row'
+                    spacing={1}
+                    justifyContent='flex-end'
+                    flexGrow={1}
+                  >
+                    {editButton}
+                    {deleteButton}
```
gigxz, line 102: I'm getting a unique key error from this

ZachCMP:
@gigxz All the mentioned items should be fixed now

src/components/dashboard/enrollments/AllEnrollments.tsx
```diff
@@ -96,7 +96,6 @@ const AllEnrollments = () => {
           rowLinkTo={rowLinkTo}
           columns={columns}
           pagePath='client.enrollments'
-          fetchPolicy='cache-and-network'
```
gigxz: Why is this removed?

gigxz:
The file picker isn't working for me anymore at all. Can you reproduce that? The file picker is just frozen, I can't click anywhere within it, but I can tab through it and use Esc to close it. It happens on both chrome and safari.

src/components/dashboard/files/useFileActions.tsx
```diff
@@ -0,0 +1,151 @@
...
+  const [deleteFile] = useDeleteClientFileMutation();
+
+  const handleDeleteFile = useCallback(
+    (fileId: string) => {
+      setDeletingIds((ids) => uniq([...ids, fileId]));
+      setDeleting(true);
+      return deleteFile({
+        variables: { input: { fileId } },
+      }).finally(() => {
+        // * Using refetch here instead of eviction because it prevents a table load after delete and makes it easier to set deleting ids correctly
```
gigxz: I'm not sure what you're after here. The table _should_ reload after delete. Can we use a cache eviction here instead?

The behavior I'm seeing is: I delete an item, and it remains in the table (but the tags disappear?). I need to refresh the page makes the row disappear.

src/components/dashboard/AllFiles.tsx
```diff
@@ -0,0 +1,167 @@
...
+          GetClientFilesQuery,
+          GetClientFilesQueryVariables,
+          FileFieldsFragment
+        >
+          queryVariables={{ id: clientId }}
+          queryDocument={GetClientFilesDocument}
+          // rowLinkTo={canEdit ? rowLinkTo : undefined}
+          columns={columns}
+          pagePath='client.files'
+          // fetchPolicy='cache-and-network'
```
gigxz: Can you remove the comments

src/components/dashboard/enrollments/AllEnrollments.tsx
ZachCMP: No reason, probably should be reverted. I went ahead and did that

ZachCMP:
@gigxz I'm not able to reproduce either file picker issue or the item deletion issue you mention above. Everything seems to be working fine for me.

https://user-images.githubusercontent.com/6409951/228278377-e49b36a7-9043-44c4-a792-25db95eeebaf.mov

src/components/dashboard/files/useFileActions.tsx
ZachCMP: That seems like the wrong behavior, but I'm not seeing that behavior locally. Regardless, here's what I'm on about here:

These things have to happen once the file is deleted:

1. Have to reload the items in the table
2. Have to remove the deleted ID from `deletingIds`, which is what makes the loading indicator show while deleting
3. Have to set `fileToDelete` to undefined, which hides the confirm deletion modal
4. Have to set `deleting` to false, which clears the loading state on the confirmation modal
5. Have to call `onDeleteFile`, which is a callback the consumer can pass in to fire when the file is done deleting

Either an eviction or a refetch will reload the items in the table, but I think a refetch is better here. It gives me a much better place to do all the things above. With a refetch, the logic becomes:

```
deleteFileMutation -- finally --> refetch -- then --> do items 2-5 above
```

This means that when I do items 2-5, I know for sure that that file mutation has fired, returned, and the items have been refetched, which means that the operation has definitively been completed. Further, the table won't hit a spinner, it will just remove the deleted item once things are refetched.

If I use an eviction, then it will cause some number of queries to refetch, but I have no introspection into when that will happen and whether or not it's been completed. If I do items 2-5 above before things are refetched, then I'll clear all the loading state of the deletion before I've pulled new information, which could cause the indicators to behave strangely. It will also cause the table to get a full-reload with a spinner, which I think is slightly worse UX.

I could deal with this by adding a `useEffect` that checks file IDs against the previous file IDs values and does items 2-5 based on that comparison, but it feels more fragile to me. I could probably mitigate some of that with cache manipulation, but I'd rather not open that can of worms if I don't have to.

Hopefully that helps explain my logic here. I'm happy to adjust this in whatever way you'd like if you don't agree with my reasoning. Let me know if that's the case and I'll go ahead and make those changes.

ZachCMP:
@gigxz Also I got the tests to pass by mocking `react-pdf`. I wasn't able to get it to transpile through jest easily, so I decided that was the easiest path forward, as it doesn't seem very important to me that we check that the PDF rendering is working correctly in the jest tests. Let me know if you disagree and I can take another crack at trying to get it to transpile.

src/components/dashboard/AllFiles.tsx
ZachCMP: Yup, done

src/components/dashboard/files/useFileActions.tsx
gigxz: Try with pagination and see if you can reproduce the bug. You can pass `defaultPageSize={2}` to the `GenericTableWithData` in `AllFiles` to test that. If you want to use refetch that's fine with me, but it should work on a paginated table.

For all the other record tables, we use cache eviction to handle record deletion and creation. I agree it's not optimal to show the spinner. If this is a better pattern that works with pagination, we can use it everywhere. (No need to do that as part of this PR, just saying - I'd prefer if things were consistent).

gigxz:
@ZachCMP sounds good, that's fine for tests. I'm still getting this error, it is related to pagination so you should be able to reproduce it that way.

src/components/dashboard/files/useFileActions.tsx
ZachCMP: Ah, gotcha. Yeah, pagination complicates the refetch enough to make it not worth using. I've switched to eviction. Apologies, I failed to take the pagination facet into account.

gigxz: Alright, this works. It doesn't actually reload the table. In other places we would do `cache.evict({ id: Client:${clientId}, fieldName: 'files' });` instead of `cache.evict({ id: File:${id} });`, so that the whole table reloads and the pagination number is correct, but I think I like this better. The total count is off, but there's no reload. I think its a reasonable tradeoff.

Kapture 2023-03-28 at 12 32 23

gigxz (approved):
I think this is good to go. The file picker issue I was having is something local, I'm seeing it in other apps too.

PR #193, opened 2023-03-23
Various container related adjustments  183573574

yakloinsteak:
This gets rid of some github actions deprecations and lets you test builds locally from the docker directory.


gigxz (approved):
thanks!

PR #194, opened 2023-03-23
Nginx exporter and json logs

yakloinsteak:
https://www.pivotaltracker.com/story/show/183469895

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -53,6 +53,9 @@ jobs:
           export GITHASH="${SHA::9}"
           export IMAGE_TAG="githash-${GITHASH}"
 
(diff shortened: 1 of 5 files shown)


yakloinsteak:
@gigxz  To enable nginx compression, we need to change the load balancer's security policy to only allow tls 1.3.  Let me know if that would be okay and when it would be okay to test that as well as the other changes in this PR.

gigxz:
Great, thanks. Yeah 1.3 should be fine.  You're welcome to deploy this anytime this afternoon or tomorrow am, just let me know when you're doing it. We need it to be stable by 3pm tomorrow, so I'll re-deploy the latest release-70 sometime tomorrow afternoon. How does that sound?

gigxz (approved):

PR #195, opened 2023-03-24
183560765 okta hmis support

ttoomey:
### _Please squash-merge this PR_

Reference
[warehouse PR #2906](https://github.com/greenriver/hmis-warehouse/pull/2906)
[#183560765](https://www.pivotaltracker.com/story/show/183560765)

Adds "Sign in with Okta" button, rolls up password sign-in form when okta is abled. Add app configuration to track if okta is enabled. Config is stored in context and discovered from new endpoint on warehouse backend.

Adjusted the Referrer-Policy for compatibility with  origin-based forgery protection in rails

Note, I'm running the app in docker locally. Happy to share my docker config with others for now I'm keeping them out of source control since I'm probably the only one using it (gitignore change).


ttoomey:
rebased off release-70

vite.config.js
```diff
@@ -53,6 +53,14 @@ export default defineConfig(({ command, mode }) => {
               console.debug('Starting proxy with options:', options);
             },
           },
+          '/users/auth': {
```
gigxz: This works for local development, but the nginx config will need to be updated to support reverse proxying the new routes in production https://github.com/greenriver/hmis-warehouse/pull/2906/files#r1149341126

src/components/pages/OktaLoginForm.tsx
```diff
@@ -0,0 +1,45 @@
...
+}
+const OktaLoginForm: React.FC<Props> = ({ path }) => {
+  // fixme - csrf should probably be in a context or other shared state
+  const [csrf, setCsrf] = useState<string>();
+  useEffect(() => {
+    const token = getCsrfToken();
+    if (token) {
+      setCsrf(token);
+    }
+  }, []);
```
gigxz: I don't think we want to store the cookie in state, because it changes with every request. We only need to grab it when we are making an auth request.

So, this seems good. I don't think the effect hooks are _really_ necessary, because this component shouldn't be re-rendering. I think `const csrf = getCsrfToken();` would be just fine. Am I missing something?

src/components/pages/OktaLoginForm.tsx
```diff
@@ -0,0 +1,45 @@
...
+  if (!csrf) {
+    // FIXME- if no crsrf yet, maybe reloading the window
```
gigxz, line 11: Reloading sounds good. Before this all renders, we call [fetchCurrentUser](https://github.com/greenriver/hmis-frontend/blob/ab1190e0b09615fd46bfa7630350d15520764eb5/src/modules/auth/hooks/useAuth.tsx#L154-L165) on initial page load, which will set the csrf cookie for us. So, we should only hit line 18 if the cookie was cleared somehow.

.gitignore
```diff
@@ -20,10 +20,13 @@
 .env.development.local
 .env.test.local
 .env.production.local
+untracked
+docker-compose.yml
+docker-dev
```
gigxz, line 25: Todd added a docker-compose file with https://github.com/greenriver/hmis-frontend/pull/193/files, but I haven't tested it yet. It would be great if you can test that one, or share your file. I think it is a good idea to have the docker setup checked in to source control.

gigxz (changes requested):
This looks great! A couple small comments inline.

Requested change: adding the okta routes to the nginx config to support reverse-proxying

src/env.d.ts
```diff
@@ -9,6 +9,7 @@ interface ImportMetaEnv {
   readonly PUBLIC_INITIAL_DELIMITER: string;
   readonly PUBLIC_PROTECTED_ID_KEY: string;
   readonly PUBLIC_ALLOW_SET_THEME: string;
+  readonly PUBLIC_OKTA_SSO_PATH: string | undefined;
```
gigxz: Oops sorry, there was one thing I missed when reviewing @ttoomey. We don't want to specify whether or not to use okta at build time, we want to specify it at runtime. The frontend should be built once and able to deploy to multiple environments. These variables listed here are installation-wide. (APP_NAME, WAREHOUSE_URL, and CAS_URL all need to be removed, but I haven't gotten to it yet).

We don't support injecting runtime variables into the frontend. Instead, our approach for "configurations" is to have the frontend fetch them on page load and store them in local storage. This is currently how we are fetching themes (`MergedThemeProvider` fetches the theme from `ThemeController`). 

We should be doing something similar for Okta. There are going to be other settings and feature flags, too. To accomplish this, I think we need:
* A `SettingsController` similar to `ThemeController` that returns a config blob `{ okta_enabled: boolean }` or whatever name/type you think makes sense.
* A `SettingsProvider` that hits that unauthenticated JSON endpoint (unless we already have it in local storage) and provides settings in a context, wrapped here `src/providers/app.tsx`
* Update Login page to look at context instead of `PUBLIC_OKTA_SSO_PATH` var

src/components/pages/OktaLoginForm.tsx
ttoomey: I encountered issues with missing/invalid csrf tokens in development. I suspect there's a timing issue; maybe since the api endpoints often takes some time to resolve, particularly on a cold start.

This useEffect is a bit of a placeholder / timing hack that seemed to make the form more reliable for me. You're right though, it's not what we'd want in production. I'll take it out.

My thought was that if the token is in state, we can update it as API calls resolve, allowing the login to be more robust against missing / invalid tokens. But maybe just reloading the page when the token is missing is sufficient

src/components/pages/OktaLoginForm.tsx
ttoomey: I was encountering this situation but maybe it was just an artifact of my local dev environment.

.gitignore
ttoomey: The docker setup I use is a bit custom, it expects nginx-proxy and uses debian instead of alpine, which I prefer for development. Happy to share but it might not fit everyone's environment / preference.
[Here's what I'm using](https://gist.github.com/ttoomey/c6acecc5c22e6087cf76f95c68e5e985)

I can make another PR for this docker config if it would be helpful

src/env.d.ts
ttoomey: Got it, good to know. I'll replace the env var with a settings endpoint and adjust the login UI. I'll use local store as a cache. In other apps I've used service workers for this kind of cache, maybe something to consider for the future.

gigxz: Ok, cool. If you think service worker is a better mechanism for caching settings, I'm fine with that. I haven't used them much, so it wasn't an intentional choice.

.gitignore
gigxz: Cool, thanks for sharing. I think there's no need for a PR now but it is great to know that you have a working setup!

src/components/pages/OktaLoginForm.tsx
gigxz: I can't repro it (maybe a browser add on blocking the cookie?), but I think this is a fine solution to keep in here.

gigxz (approved):

ttoomey:
Hi @gigxz - is it okay if I merge this PR into 70?

docker/templates/default.conf.template
```diff
@@ -19,7 +19,7 @@ server {
         # use INLINE_RUNTIME_CHECK=false yarn build and nonce / hash to apply a CSP
         #add_header Content-Security-Policy "default-src 'none'; script-src 'self'; connect-src 'self'; img-src 'self'; font-src fonts.gstatic.com; style-src 'self' fonts.googleapis.com;" always;
 
-        add_header Referrer-Policy "no-referrer";
+        add_header Referrer-Policy "strict-origin-when-cross-origin";
```
ttoomey, line 22: Hi @schwartzkylej - I'm loosening this a bit from what you had in 0b9924c6. This is for compatibility with rails origin-based forgery protection. Do you have any concerns with this change? Thanks

schwartzkylej: No that change is fine. Thanks for the heads up.

gigxz:
@ttoomey yep, as long as the front end and backend PRs are merged at the same time

PR #202, opened 2023-03-29
184800714 ssn dob entry

ZachCMP:
https://www.pivotaltracker.com/story/show/184800714

Relies on warehouse PR https://github.com/greenriver/hmis-warehouse/pull/2922


src/components/dashboard/EditClient.tsx
```diff
@@ -10,12 +10,22 @@ import { useDashboardClient } from '../pages/ClientDashboard';
...
 import generateSafePath from '@/utils/generateSafePath';
 
 const Profile = () => {
   const { client } = useDashboardClient();
   const navigate = useNavigate();
+  const { data } = useGetClientPermissionsQuery({
+    variables: { id: client.id },
+  });
+
+  const { canViewDob, canViewFullSsn } = data?.client?.access || {};
```
gigxz, line 27: We should wait until the permissions are loaded until we render the `EditRecord` component

gigxz (commented):
Looks good to me after loading fix

src/components/dashboard/EditClient.tsx
ZachCMP: So this component is actually loaded in a `PermissionsFilter` component which should have a loading state while the permissions are still loading, but I also added a specific check to this component just in case the `PermissionsFilter` is loading other permissions.

gigxz (approved):
Looks good. I pushed a change to make SSN and DOB to always appear on the Client creation form, because they were always hidden. We may want to revisit later, but I think there are reasons to want to allow inputting those fields even if the user can't see them, for example to do lookups with matching services (which may require DOB/SSN).

PR #204, opened 2023-03-29
Prevent refetching of permissions

ZachCMP:
https://www.pivotaltracker.com/story/show/184630508


src/api/operations/clients.queries.graphql
```diff
@@ -34,6 +34,7 @@ query GetClientName($id: ID!) {
 
 query GetClientPermissions($id: ID!) {
   client(id: $id) {
+    id
     ...ClientPermissions
```
gigxz: `ClientPermissions` already has `id` on it, so I don't think this is doing anything

src/modules/permissions/PermissionsFilters.tsx
```diff
@@ -79,6 +79,7 @@ export const ClientPermissionsFilter: React.FC<
 > = ({ id, ...props }) => {
   const { data, loading } = useGetClientPermissionsQuery({
     variables: { id },
+    skip: !id,
```
gigxz, line 82: When is this happening? It looks like a case of us expecting `clientId` to be present in the route, but not finding it?

gigxz:
Hmm, this isn't fixing the repeated requests for me. I'm still seeing repeated requests as I click around the dashboard for `GetClientPermissions` for the same client ID. I thought Apollo would save us from that! I'm seeing this warning in the console.


```
Cache data may be lost when replacing the access field of a Client object.

To address this problem (which is not a bug in Apollo Client), either ensure all objects of type ClientAccess have an ID or a custom merge function, or define a custom merge function for the Client.access field, so InMemoryCache can safely merge these objects:

  existing: {"__typename":"ClientAccess","canViewDob":true,"canViewFullSsn":true,"canViewPartialSsn":false,"canEditEnrollments":true,"canDeleteEnrollments":true,"canViewEnrollmentDetails":true,"canManageAnyClientFiles":true,"canManageOwnClientFiles":true,"canViewAnyConfidentialClientFiles":true,"canViewAnyNonconfidentialClientFiles":true}
  incoming: {"__typename":"ClientAccess","canViewFullSsn":true,"canViewPartialSsn":false}

For more information about these options, please refer to the documentation:

  * Ensuring entity objects have IDs: https://go.apollo.dev/c/generating-unique-identifiers
  * Defining custom merge functions: https://go.apollo.dev/c/merging-non-normalized-objects
```

ZachCMP:
Now relies on backend PR https://github.com/greenriver/hmis-warehouse/pull/2932 for the access type IDs

ZachCMP:
@gigxz I'm not seeing the overfetching behavior now. This branch also needed a lot merged in from `release-70`, which I've done. It's possible something in there fixed the issue.

gigxz:
Cool. I tried merging in the release-70 changes when I tested with your branch yesterday, and it didn't fix the issue for me. So, sounds like your backend change fixed it. It works for me. thanks!

PR #214, opened 2023-04-10
184085879 dynamic form view

ZachCMP:
https://www.pivotaltracker.com/story/show/184085879

### DIFF
--- a/src/components/dashboard/enrollments/ViewAssessmentPage.tsx
+++ b/src/components/dashboard/enrollments/ViewAssessmentPage.tsx
@@ -5,6 +5,8 @@ import { useOutletContext } from 'react-router-dom';
 import { DashboardContext } from '@/components/pages/ClientDashboard';
 import useSafeParams from '@/hooks/useSafeParams';
 import { useAssessment } from '@/modules/assessments/components/useAssessment';
+import DynamicView from '@/modules/form/components/viewable/DynamicView';
+import { createInitialValuesFromSavedValues } from '@/modules/form/util/formUtil';
 import { EnrollmentFieldsFragment } from '@/types/gqlTypes';
 
 // PLACEHOLDER: implement read-only view of assessments
@@ -30,7 +32,20 @@ const ViewAssessmentPage = () => {
     assessmentId,
   });
 
-  return <Typography>{assessmentTitle}</Typography>;
+  if (!definition || !assessment) return <Typography>NOOOOO</Typography>;
+
+  const values = createInitialValuesFromSavedValues(
+    definition.definition,
+    assessment.customForm?.values
+  );
+
+  return (
+    <DynamicView
+      values={values}
+      definition={definition.definition}
+      pickListRelationId={enrollment?.project?.id}
+    />
+  );
 };
 
 export default ViewAssessmentPage;
--- a/src/components/dashboard/enrollments/ViewAssessmentPage.tsx
+++ b/src/components/dashboard/enrollments/ViewAssessmentPage.tsx
@@ -1,16 +1,32 @@
 /* eslint-disable @typescript-eslint/no-unused-vars */
-import { Typography } from '@mui/material';
-import { useOutletContext } from 'react-router-dom';
+import EditIcon from '@mui/icons-material/Edit';
+import { Box, Button, Grid, Paper, Typography } from '@mui/material';
+import { ReactNode } from 'react';
+import { useOutletContext, Link as ReactRouterLink } from 'react-router-dom';
 
+import Loading from '@/components/elements/Loading';
+import {
+  CONTEXT_HEADER_HEIGHT,
+  STICKY_BAR_HEIGHT,
+} from '@/components/layout/layoutConstants';
 import { DashboardContext } from '@/components/pages/ClientDashboard';
 import useSafeParams from '@/hooks/useSafeParams';
+import AssessmentTitle from '@/modules/assessments/components/AssessmentTitle';
 import { useAssessment } from '@/modules/assessments/components/useAssessment';
+import FormStepper from '@/modules/form/components/FormStepper';
 import DynamicView from '@/modules/form/components/viewable/DynamicView';
 import { createInitialValuesFromSavedValues } from '@/modules/form/util/formUtil';
+import IdDisplay from '@/modules/hmis/components/IdDisplay';
+import { DashboardRoutes } from '@/routes/routes';
+import { HmisEnums } from '@/types/gqlEnums';
 import { EnrollmentFieldsFragment } from '@/types/gqlTypes';
+import generateSafePath from '@/utils/generateSafePath';
 
-// PLACEHOLDER: implement read-only view of assessments
-const ViewAssessmentPage = () => {
+const ViewAssessmentPage = ({
+  top = STICKY_BAR_HEIGHT + CONTEXT_HEADER_HEIGHT,
+}: {
+  top?: number;
+}) => {
   const { client, enrollment } = useOutletContext<DashboardContext>();
   const { clientId, enrollmentId, assessmentId } = useSafeParams() as {
     clientId: string;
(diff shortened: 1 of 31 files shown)


src/components/dashboard/enrollments/ViewAssessmentPage.tsx
```diff
@@ -1,21 +1,44 @@
...
+  top?: number;
+}) => {
   const { client, enrollment } = useOutletContext<DashboardContext>();
   const { clientId, enrollmentId, assessmentId } = useSafeParams() as {
     clientId: string;
     enrollmentId: string;
     assessmentId: string;
   };
 
+  const canEdit = useHasClientPermissions(clientId, ['canEditEnrollments']);
```
gigxz: This should be canViewEnrollmentDetails

src/modules/form/components/viewable/DynamicViewField.tsx
```diff
@@ -0,0 +1,189 @@
...
+        adjustValue({ linkId: item.linkId, value: fullOption });
+      } else {
+        console.warn(
+          `Selected value '${JSON.stringify(
+            value
+          )}' is not present in option list '${item.pickListReference}'`
+        );
+      }
+    },
+  });
```
gigxz: Can we move this to a shared function so it's not repeated? Or move the logic into the hook.

src/modules/form/components/viewable/DynamicViewField.tsx
```diff
@@ -0,0 +1,189 @@
...
+    case ItemType.Date:
+      return (
+        <TextContent
+          {...commonProps}
+          renderValue={(val) => {
+            let formatted;
+            if (val instanceof Date) formatted = format(val, 'M/d/yyyy');
+            if (typeof val === 'string') {
+              const parsed = parseHmisDateString(val);
+              if (parsed) formatted = format(parsed, 'M/d/yyyy');
```
gigxz:
```suggestion
              formatted = parseAndFormatDate(val)
```

src/modules/form/components/viewable/DynamicViewField.tsx
```diff
@@ -0,0 +1,189 @@
...
+          renderValue={(val) => `$${val}`}
+        />
+      );
+    case ItemType.Date:
+      return (
+        <TextContent
+          {...commonProps}
+          renderValue={(val) => {
+            let formatted;
+            if (val instanceof Date) formatted = format(val, 'M/d/yyyy');
```
gigxz:
```suggestion
            if (val instanceof Date) formatted = formatDateForDisplay(val);
```

gigxz:
```suggestion
            if (val instanceof Date) formatted = formatDateForDisplay(val);
```

src/modules/form/components/viewable/DynamicViewField.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+            return (
+              <Typography variant='body2'>
+                {ensureArray(value)
+                  .map((val) => {
+                    let label = val.label;
+                    if (options)
+                      label =
+                        options?.find((opt) => opt.code === val.code)?.label ||
+                        label;
```
gigxz: Why do we need this? The `onCompleted` callback of the usePickList query should be able to handle updating the value after the list loads.

src/modules/form/components/viewable/item/TextContent.tsx
```diff
@@ -0,0 +1,41 @@
...
+  horizontal?: boolean;
+  hasValue?: (value: T) => boolean;
+  renderValue?: (value: T) => ReactNode;
+}
+
+const TextContent = <T extends ReactNode>({
+  label,
+  value,
+  horizontal = false,
+  hasValue = (x) => !!x,
```
gigxz: Can we just use `hasMeaningfulValue` for the default? I think it should work for everything, including booleans. If it works for everything, it doesn't need to be a prop.

src/modules/form/components/viewable/DynamicViewFields.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+    const itemComponent = (
+      <DynamicViewField
+        key={item.linkId}
+        item={item}
+        value={isDisabled ? undefined : values[item.linkId]}
+        nestingLevel={nestingLevel}
+        horizontal={horizontal}
+        pickListRelationId={pickListRelationId}
+        adjustValue={adjustValue}
```
gigxz, line 116: Passing a change hook into a read-only component seems off to me. The only "change" that's happening is visually replacing enum codes with enum labels, which does not affect disabled link ID state.

I think it would be cleaner to handle the code->label display change locally in the `DynamicViewField` components. Something simple like:
```
  const [displayValue, setDisplayValue] = useState(value);

  // in usePicklist onCompleted:
  // setDisplayValue(fullOption)
```

src/modules/form/components/viewable/item/NotCollectedText.tsx
```diff
@@ -0,0 +1,27 @@
...
+}: NotCollectedTextProps): JSX.Element => {
+  return (
+    <Typography
+      {...props}
+      sx={({ palette }) =>
+        Object.assign({
+          color: palette.text.disabled,
+          ...(props.sx ? props.sx : {}),
+        })
+      }
```
gigxz: Can be simplified, I think..
```suggestion
       color='text.disabled'
      {...props}
```

src/modules/form/hooks/useDynamicViewFields.tsx
```diff
@@ -0,0 +1,107 @@
...
+      <DynamicViewFields
+        {...props}
+        {...{
+          definition,
+          itemMap,
+          autofillDependencyMap,
+          enabledDependencyMap,
+          disabledLinkIds,
+          setDisabledLinkIds,
+          setValues,
```
gigxz, line 77: As mentioned in my other comment, it would be great if we can avoid passing change hooks to the read-only component. Are these only needed because of the picklist option callback currently?

src/modules/form/components/viewable/DynamicViewField.tsx
```diff
@@ -0,0 +1,189 @@
...
+          }}
+        />
+      );
+    case ItemType.Choice:
+      return (
+        <TextContent
+          {...commonProps}
+          hasValue={(val) => hasMeaningfulValue(val)}
+          renderValue={(value) => {
+            if (pickListLoading) return <Skeleton />;
```
gigxz: The skeletons span the whole width of the card, which is kind of awkward. Can we give them a max width around 200 px or so, and some height?

src/modules/form/components/viewable/DynamicViewField.tsx
```diff
@@ -0,0 +1,189 @@
...
+        <TextContent
+          {...commonProps}
+          hasValue={(val) => hasMeaningfulValue(val)}
+          renderValue={(val) => {
+            if (pickListLoading) return <Skeleton>{val}</Skeleton>;
+            return (
+              <Typography variant='body2'>
+                {ensureArray(value)
+                  .map((val) => val.code)
+                  .join(', ')}
```
gigxz:
```suggestion
                {ensureArray(value)
                  .map((val) => val.label || val.code)
                  .join(', ')}
```

And I don't see why `ItemType.Choice` can't have the same render function. Actually we should be able to collapse `case ItemType.OpenChoice` and `case ItemType.Choice`.

src/modules/form/components/viewable/DynamicViewField.tsx
```diff
@@ -0,0 +1,189 @@
...
+          value={
+            [TRUE_OPT, FALSE_OPT].find((o) => o.code === String(value))
+              ?.label || <NotCollectedText variant='body2' />
+          }
+          hasValue={(val) => !isNil(val)}
+        />
+      );
+    case ItemType.String:
+    case ItemType.Text:
+      return <TextContent {...commonProps} />;
```
gigxz:
```suggestion
```

src/modules/form/components/viewable/DynamicViewFields.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+export const isShown = (item: FormItem, disabledLinkIds: string[] = []) => {
+  if (
+    !isEnabled(item, disabledLinkIds) &&
+    item.disabledDisplay !== DisabledDisplay.Protected
+  )
+    return false;
+
+  return true;
+};
```
gigxz: These look like they are copied exactly from `DynamicFormFields`. Can you import them from there, or better yet move them to `formUtil.ts`

gigxz (changes requested):
Looks good so far. Some comments on change handlers, and places to reduce repeated code.

src/modules/form/hooks/useDynamicViewFields.tsx
ZachCMP: Yes that's exactly why. It has to fetch the pick list stuff, then that will cause an update which will set the diabledLinkIds correctly based on the fetched pick list values.

ZachCMP: I've named the handlers "adjustValues" in the components to hopefully make it more clear that they're not meant to update information, but they are functionally just value setters in the end.

gigxz: What do you think of implementing [this suggestion](https://github.com/greenriver/hmis-frontend/pull/214#discussion_r1162118729) so that we can remove all these handlers? The display string change can be handled locally in the display component.

I don't think that the picklist changes could affect the disabledLinkIds, since that should always based on the "code"  and not the "label."

src/modules/form/components/viewable/DynamicViewFields.tsx
ZachCMP: @gigxz So if I don't run `setDisabledLinkIds` with updated values when the pick list values are fetched, then fields will be visible that should not be. I'm all for storing it as state, but I also either need to run `setDisabledLinkIds` with the updated values, or we'll need to make it so that the disabled link IDs initial state can be derived correctly without needing pick list data loaded. I'd definitely prefer the latter if possible.

It looks like maybe the `enabledDependencyMap` part of the logic only gets fired when values are updated, and not actually during the initial state calculation. Though if I call it with old (pre-fetch) values then it doesn't hide the fields that should be hidden, so it looks like it also has to do with the initial values.

What do you think? You know the plumbing in there better than I do

src/modules/form/hooks/useDynamicViewFields.tsx
ZachCMP: See my comment above. I think they definitely do affect the disabledLinkIds currently

src/modules/form/components/viewable/DynamicViewField.tsx
ZachCMP: I moved the logic into the `usePickList` hook as an exported utility function. Does that work for you?

ZachCMP:
@gigxz I pushed fixes for all the feedback that I put a thumbs up reaction on. The change handler question should be the only outstanding issue to be addressed now.

src/modules/form/components/viewable/DynamicViewFields.tsx
gigxz: Oh I see, that's right. For Prior Living Situation on the Intake assessment, we have a conditional that depends on the "groupLabel" of the pick list option, which we only know once we've fetched it. I don't love that, because it presents as the dependent items appearing and then disappearing (both on the view and edit pages). But I don't think that's worth spending more time on. I'll merge this in as is. Thanks!

PR #215, opened 2023-04-11
Reduce number of requests for fetching permissions, use cache

gigxz:
Reduce number of requests made on the view enrollment page & household management page (by 1 request each) by reading fragment directly from the cache.

### DIFF
--- a/src/modules/permissions/PermissionsFilters.tsx
+++ b/src/modules/permissions/PermissionsFilters.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 
 import { useHasPermissions } from './useHasPermissionsHooks';
 
+import apolloClient from '@/providers/apolloClient';
 import {
   ProjectAccess,
   QueryAccess,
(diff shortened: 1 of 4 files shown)


src/modules/permissions/PermissionsFilters.tsx
```diff
@@ -77,11 +79,21 @@ export type ClientPermissionsFilterProps<T> = Omit<
 export const ClientPermissionsFilter: React.FC<
   ClientPermissionsFilterProps<ClientAccess>
 > = ({ id, ...props }) => {
-  const { data, loading } = useGetClientPermissionsQuery({
+  // read ClientAccess from cache if we have it
+  const access = id
+    ? apolloClient.readFragment({
+        id: `ClientAccess:${id}`,
```
gigxz, line 85: Needs https://github.com/greenriver/hmis-warehouse/pull/2962 to work

ZachCMP (approved):
Looks good to me. I like reading the access from the fragment as a way to avoid the query if we already have the info.

PR #221, opened 2023-04-14
184907418 delete client

ZachCMP:
https://www.pivotaltracker.com/story/show/184907418

Depends on backend PR: https://github.com/greenriver/hmis-warehouse/pull/2968


src/modules/client/components/ClientProfileCard.tsx
```diff
@@ -368,26 +369,24 @@ const ClientProfileCard: React.FC<Props> = ({ client, onlyCard = false }) => {
                     : {}),
                 }}
               />
-              <Box sx={{ flexGrow: 1 }}>
+              <Stack sx={{ flexGrow: 1 }} gap={1} mt={2}>
```
ZachCMP, line 372: Just FYI this change was due to the fact that I originally put the delete button here. I decided to leave the change and not revert back since it seems to look fine this way and will handle adding more items to this stack if we ever need to.

ZachCMP:
@gigxz Let me know when you merge the validation dialog stuff and I'll get that pulled into this work

gigxz:
@ZachCMP it's been merged into release-71. let me know if you have any trouble with merge conflicts. thanks!

src/components/dashboard/EditClient.tsx
```diff
@@ -49,11 +52,29 @@ const Profile = () => {
...
       top={STICKY_BAR_HEIGHT + CONTEXT_HEADER_HEIGHT}
       title={
-        <>
-          <Typography variant='h3' sx={{ pt: 0, pb: 4 }}>
+        <Stack direction='row' justifyContent='space-between'>
+          <Typography variant='h3' sx={{ pt: 0, pb: 4 }} flexGrow={1}>
             Edit Client Details
           </Typography>
-        </>
+          {client.id && (
```
gigxz: We shouldn't need this

src/components/dashboard/EditClient.tsx
```diff
@@ -49,11 +52,29 @@ const Profile = () => {
...
+            <div>
+              <RootPermissionsFilter permissions='canDeleteClients'>
```
gigxz: Can you add `canDeleteClients` to the client access fragment so that we don't get the glitchy flashing appearance again here?

src/components/dashboard/EditClient.tsx
```diff
@@ -49,11 +52,29 @@ const Profile = () => {
...
+                  clientId={client.id}
+                  variant='text'
+                  onCompleted={(data) => {
+                    if (
+                      isEmpty(data.deleteClient?.errors) &&
+                      data.deleteClient?.client
+                    ) {
+                      navigate(generateSafePath(Routes.CLIENT_SEARCH));
+                    }
+                  }}
```
gigxz: Nitpick, but I think this is a bit confusing, it made me think we were ignoring the errors. I think it would be cleaner to just call the callback on successful deletion:
```suggestion
                  onSuccess={() => navigate(generateSafePath(Routes.CLIENT_SEARCH))}
```

gigxz (approved):
LGTM. I'll do another pass and test it out once the validation dialog changes are merged in. Thanks!

src/components/dashboard/EditClient.tsx
ZachCMP: I'm not seeing a glitchy flashing appearance, can you describe in more detail what you're seeing that looks glitchy? 

Also `canDeleteClients` is also not a client-based permission, it's a root permission, so it should be fetched on the query access fragment, not on the client access fragment.

ZachCMP:
@gigxz Fixed all the feedback with a thumbs up and merged in the validation dialog changes. Should be ready for that second pass

src/components/dashboard/EditClient.tsx
gigxz: Oh I see I didn't realize it was on the root. Then it shouldn't have that issue. never mind

gigxz (approved):

PR #222, opened 2023-04-14
Upload source maps to Sentry

gigxz:
https://www.npmjs.com/package/@sentry/vite-plugin


.github/workflows/build-image.yml
```diff
@@ -73,6 +73,7 @@ jobs:
           build-args: |
             NODE_VERSION=${{ steps.env.outputs.node_version }}
             GITHASH=${{ steps.env.outputs.git_hash }}
+            SENTRY_AUTH_TOKEN=${{ secrets.SENTRY_AUTH_TOKEN }}
```
gigxz, line 77: Is this correct? How do I get the token up there?

yakloinsteak: That looks correct. See if you can get to https://github.com/greenriver/hmis-frontend/settings/secrets/actions. Since you're an admin in this repo, I think you can. Click on "New repository secret"

PR #223, opened 2023-04-14
Throw error if bad app setting response

gigxz:
Trying to debug a periodic issue I've seen on QA, where the page loads without the app settings (no Okta screen), and you need to hard refresh to get the correct settings. I can't reliably reproduce it.


ttoomey (approved):
Looks good, makes sense to throw unless we get a 200. I could see that causing an issue

PR #232, opened 2023-04-20
MCI clearance; Project layout

gigxz:
Requires https://github.com/greenriver/hmis-warehouse/pull/2983 and https://github.com/greenriver/hmis-warehouse/pull/3005 to fully work


src/modules/form/components/DynamicForm.tsx
```diff
@@ -203,17 +205,19 @@ const DynamicForm = forwardRef(
...
-            severalItemsChanged: handleChangeCallback,
-            errors: errorState.errors,
-            warnings: errorState.warnings,
-            horizontal,
-            pickListRelationId,
-            warnIfEmpty,
-            locked,
-            visible,
-          })}
+          <DynamicFormContext.Provider value={{ getCleanedValues, definition }}>
```
gigxz, line 208: @ZachCMP  adding a provider here so that child components can access the entire form values if necessary. MCI clearance, for example, needs to access a bunch of the current values in order to perform a search against an external system. I expect this to be used rarely.

PR #238, opened 2023-04-27
Closed Client Access

ZachCMP:
https://www.pivotaltracker.com/story/show/184986956

Relies on backend PR https://github.com/greenriver/hmis-warehouse/pull/2998


src/api/operations/client.fragments.graphql
```diff
@@ -33,6 +33,13 @@ fragment ClientName on Client {
   nameSuffix
 }
 
+fragment ClientPermissions on Client {
```
gigxz: Mind if I remove this, since it's not used? I reorganized the fragments to hopefully make things clearer, and part of that was moving all the access fragments into one file and setting them up ON the access objects, not on the parent record.

gigxz (approved):

PR #239, opened 2023-04-27
Handle timeout errors

ZachCMP:
https://www.pivotaltracker.com/story/show/184986956

### DIFF
--- a/src/modules/hmisAppSettings/Provider.tsx
+++ b/src/modules/hmisAppSettings/Provider.tsx
@@ -1,4 +1,5 @@
 import { ReactNode, useEffect, useState } from 'react';
+import { useNavigate } from 'react-router-dom';
 
 import { fetchHmisAppSettings } from './api';
 import { HmisAppSettingsContext } from './Context';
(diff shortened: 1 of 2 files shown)


src/modules/hmisAppSettings/Provider.tsx
```diff
@@ -23,7 +25,22 @@ export const HmisAppSettingsProvider: React.FC<Props> = ({ children }) => {
     if (fetched?.appName) document.title = fetched.appName;
   }, [fetched]);
 
-  if (error) throw error;
+  useEffect(() => {
+    if (error) {
+      const cause = error?.cause as any;
+      if (cause?.error?.type === 'timeout') {
+        // Refetch settings if the session timed out
+        fetchHmisAppSettings().then(setFetched).catch(setError);
```
gigxz, line 33: Why do we ned to fetch it again? If the session is already timed out, wont we just get the same error?

gigxz:
I think is the underlying issue https://github.com/heartcombo/devise/issues/5213
I wouldn't expect the App Settings request to return a timeout, since that endpoint doesn't require authentication.

src/modules/hmisAppSettings/Provider.tsx
ZachCMP: No actually. It will fetch fine the second time. My guess is that the session gets nulled on the frontend after it gets the timeout response and then the next request will have no session which will pass correctly.

Also we have to fetch it again because the app settings aren't returned with the error response so we do need to go and try to fetch them again.

ZachCMP:
Yeah, that seems to be the issue. How would you like to proceed on this?

gigxz:
I don't see a backend fix for it, so this is good. Thanks!

src/modules/hmisAppSettings/Provider.tsx
gigxz: Huh, odd.

PR #246, opened 2023-05-02
185073482 dynamic file details

ZachCMP:
https://www.pivotaltracker.com/story/show/185073482


ZachCMP:
@gigxz I also updated the file preview appearance to look better in the stack. I just did what looked alright to me, so feel free to suggest adjustments if you don't agree with my choices there.

gigxz (approved):

PR #256, opened 2023-05-04
Implement printable assessment

ZachCMP:
https://www.pivotaltracker.com/story/show/184629713

@gigxz This is my first shot at how to do this, very open to suggestions if you disagree with how I've done it here. Overall it was pretty smooth, but I do have one big thing to bring up:

I wanted to put the print trigger higher up in the component tree, but I needed it to fire only when the form was done loading. So I put it into the form and gave it a `hold` prop to handle the loading state. However, the actual dynamic view still needs to do queries to load pick lists. Those queries don't even trigger a re-render of `AssessmentForm`, so unless I use some kind of event listening I can't really get at when those are loaded. That all felt like a rabbit hole, and since the ticket said to timebox this work, I decided not to go down that road. My solution was just to add a configurable timeout and hope that that will work for now, but I don't love that solution long term.

What do you think?

### DIFF
--- a/src/hooks/useIsPrintView.ts
+++ b/src/hooks/useIsPrintView.ts
@@ -1,6 +1,14 @@
 import { useLocation } from 'react-router-dom';
 
-const useIsPrintView = () =>
-  new URLSearchParams(useLocation().search).has('print');
+import useCurrentPath from './useCurrentPath';
+
+import { PRINTABLE_ROUTES } from '@/routes/routes';
+
+const useIsPrintView = () => {
+  const hasPrintParam = new URLSearchParams(useLocation().search).has('print');
+  const isPrintableRoute = PRINTABLE_ROUTES.includes(useCurrentPath() || '');
+
+  return hasPrintParam && isPrintableRoute;
+};
 
 export default useIsPrintView;
--- a/src/hooks/useIsPrintView.ts
+++ b/src/hooks/useIsPrintView.ts
@@ -1,0 +1,6 @@
+import { useLocation } from 'react-router-dom';
+
+const useIsPrintView = () =>
+  new URLSearchParams(useLocation().search).has('print');
+
+export default useIsPrintView;
--- a/src/modules/assessments/components/AssessmentForm.tsx
+++ b/src/modules/assessments/components/AssessmentForm.tsx
@@ -15,6 +15,8 @@ import {
   STICKY_BAR_HEIGHT,
 } from '@/components/layout/layoutConstants';
 import NotFound from '@/components/pages/NotFound';
+import useIsPrintView from '@/hooks/useIsPrintView';
+import usePrintTrigger from '@/hooks/usePrintTrigger';
 import { useScrollToHash } from '@/hooks/useScrollToHash';
 import { hasAnyValue } from '@/modules/errors/util';
 import DynamicForm, {
@@ -88,6 +90,8 @@ const AssessmentForm = ({
     setDialogOpen(false);
     setReloadInitialValues((old) => !old);
   }, []);
+
+  const isPrintView = useIsPrintView();
 
   const { submitHandler, saveDraftHandler, mutationLoading, errors } =
     useAssessmentHandlers({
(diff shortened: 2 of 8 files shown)


src/hooks/useIsPrintView.ts
```diff
@@ -0,0 +1,6 @@
+import { useLocation } from 'react-router-dom';
+
+const useIsPrintView = () =>
+  new URLSearchParams(useLocation().search).has('print');
```
gigxz: It makes me a little nervous to expose this view for routes with `?print` even if we never actually want to show them this way.  Can we incorporate a check for a whitelisted list of "printable" routes?
```
import { PRINTABLE_ROUTES } from '@/routes/routes';


const currentPath = useCurrentPath();

const hasPrintParam = new URLSearchParams(useLocation().search).has('print');

return hasPrintParam && PRINTABLE_ROUTES.includes(currentPath)
```

src/modules/assessments/components/AssessmentForm.tsx
```diff
@@ -165,54 +169,64 @@ const AssessmentForm = ({
 
   const canEdit = enrollment?.access.canEditEnrollments;
 
+  usePrintTrigger({
+    startReady: isPrintView,
+    hold: !enrollment,
```
gigxz: The enrollment will always be present when the AssessmentForm is rendered, so I don't think the `hold` param is doing anything. Am I missing something here?

It looks like there is some remnant code in this component from a previous iteration where enrollment wasn't guaranteed (like the `if (!enrollment) return <NotFound />;` below) but we should be able to remove those things. The `enrollment` prop is required on the AssessmentForm, and the IndividualAssessment always passes it.

src/hooks/useIsPrintView.ts
ZachCMP: Sure! I think that's a good idea

src/modules/assessments/components/AssessmentForm.tsx
ZachCMP: Gotcha, then it probably doesn't need `hold`. I'd like to leave the param itself, since there may be cases where we want it to hold until some loading process completes. Do you agree?

src/modules/assessments/components/AssessmentForm.tsx
```diff
@@ -165,54 +169,64 @@ const AssessmentForm = ({
...
+    timeout: 3000,
```
gigxz, line 173: How about instead of the timeout (or in addition to a very brief timeout), we preload of all the picklists? That way, each DynamicField is just retrieving the picklist form the cache.

I think the timeout was a good call for timeboxing, but I like the idea of "frontloading" the loading state for the pick lists anyway, because it will help us get rid of that glitching thing for Prior Living Situation where it takes a moment for the conditional logic to be applied while the picklist is fetched.


```
// DynamicFormFields.tsx

const { loading } = usePreloadPicklists(itemMap)
if (loading) return loading;

render the stuff
```


```
// usePreloadPicklists.tsx

const pickListTypes = Object.values(itemMap).map(item => item.pickListReference).filter(reference => reference && Object.values<string>(PickListType).includes(reference))

// Run useGetPickListQuery once per pick list type (?), which would get queried in a batch 

return { loading: anyQueriesLoading }
```

src/modules/assessments/components/AssessmentForm.tsx
gigxz: Yeah go for it

src/modules/assessments/components/AssessmentForm.tsx
ZachCMP: @gigxz That sounds ideal to me. Would you like me to do that as part of this work, or do you want to split that into a separate ticket?

ZachCMP:
@gigxz Added the printable routes check and removed both the `hold` usage and the `if (!enrollment) return <NotFound />` lines. Let me know about whether or not to do the preloading.

src/modules/assessments/components/AssessmentForm.tsx
gigxz: Cool. Let's get this merged; I'll make a separate ticket for it and assign it to you. Feel free to do it if you have any extra time for HMIS this sprint, if not no worries, we can plan for it next week. Thanks!!

gigxz (approved):

PR #259, opened 2023-05-08
185052017 support create referral request

ttoomey:
## If merging, please squash
[related wh issue](https://github.com/greenriver/hmis-warehouse/pull/3034)

Create and show referral requests

questions:
* do we need sort / search on referral requests?


gigxz (approved):
We don't need sorting for now, see https://github.com/greenriver/hmis-warehouse/pull/3034#discussion_r1189981267, it is blocked by implementing a pattern for sortable tables https://www.pivotaltracker.com/story/show/185128929. I don't think we need searching either.

PR #260, opened 2023-05-08
185109128 preload picklists

ZachCMP:
https://www.pivotaltracker.com/story/show/185109128

@gigxz Curious to get your take on this one


src/modules/form/hooks/usePreloadPicklists.tsx
```diff
@@ -0,0 +1,79 @@
...
+        Object.values(stabilizedItemMap || {})
+          .map((item) => (item.hidden ? null : item.pickListReference))
+          .filter((reference) => {
+            const isValid =
+              reference &&
+              Object.values<string>(PickListType).includes(reference);
+
+            if (isValid) return true;
+
+            return false;
```
gigxz:
```suggestion
            return isValid;
```

src/modules/form/hooks/usePreloadPicklists.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+const usePreloadPicklists = (
+  itemMap: ItemMap | undefined,
+  relationId?: string
+) => {
+  const [fetchPickList] = useLazyQuery(GetPickListDocument);
+  const [loading, setLoading] = useState(true);
+  const [data, setData] = useState<PickListOptionFieldsFragment[]>([]);
+  // Stabilize the value in case we're not careful about reference equality on the input
+  const stabilizedItemMap = useStabilizedValue(itemMap);
```
gigxz: I don't love this.. why don't we have this hook accept `definition` instead. That should be stable since it's not constructed (unlike the item map).

src/modules/form/hooks/usePreloadPicklists.tsx
```diff
@@ -0,0 +1,79 @@
...
+  GetPickListDocument,
+  PickListOptionFieldsFragment,
+  PickListType,
+} from '@/types/gqlTypes';
+
+const usePreloadPicklists = (
+  itemMap: ItemMap | undefined,
+  relationId?: string
+) => {
+  const [fetchPickList] = useLazyQuery(GetPickListDocument);
```
gigxz: We should still use a network policy that re-fetches the pick lists because their values may have changed (if its an enrollment list or project coc list for example). Previously we were using network-only, but maybe cache-and-network here is more appropriate?

src/modules/form/hooks/usePreloadPicklists.tsx
```diff
@@ -0,0 +1,79 @@
...
+          )
+        );
+      })
+      .catch(() => setData([]))
+      .finally(() => setLoading(false));
+  }, [fetchPickList, pickListTypesToFetch, relationId]);
+
+  useEffect(() => {
+    fetch();
+  }, [fetch]);
```
gigxz, line 71: I'm a bit confused about why we are calling a callback inside a useEffect to achieve this. Why not just do everything in a single useEffect hook?

src/modules/form/components/viewable/DynamicViewField.tsx
```diff
@@ -48,7 +48,6 @@ const DynamicViewField: React.FC<DynamicViewFieldProps> = ({
   const label = noLabel ? null : getLabel(item, horizontal);
 
   const [, pickListLoading] = usePickList(item, pickListRelationId, {
-    fetchPolicy: 'network-only', // Always fetch, because ProjectCoC and Enrollment records change
```
gigxz, line 51: I think this is correct, but it's an issue that you're not _always_ preloading the picklists for DynamicViews. It looks they're only preloaded for assessments (in AssessmentForm). IMO we should just always prefetch the picklists before rendering any kind of Dynamic display, whether it is editable or read-only, because of both the issues mentioned in https://github.com/greenriver/hmis-frontend/pull/256#discussion_r1186146934.

gigxz: As an example, open up a Project > Inventory > click table row. It should pull up a modal with a DynamicView. You'll see the CoC Code section is loading. The idea is to preload such that when the DynamicView appears, there is no secondary loading state.

gigxz (changes requested):
Looks good, but it only pre-loads the picklists for assessments. We should do it for all dynamic views. (example in the comments)

src/modules/form/hooks/usePreloadPicklists.tsx
ZachCMP: I switched to using `client.query` since `useLazyQuery` wasn't working correctly. I'm using `network-only` as the fetch policy to ensure that the picklists are always fetched from network when we load a form.

src/modules/form/hooks/usePreloadPicklists.tsx
ZachCMP: I want to expose the `fetch` function in the return value of the hook, which gives us a way to force-refetch the pick lists should we need to. In order to do that I need to declare it as a const using `useCallback`. Since I also want to call it in this hook on mount and whenever the deps change, I do that by calling the `fetch` function inside of a `useEffect`.

src/modules/form/components/viewable/DynamicViewField.tsx
ZachCMP: Should be fixed now

ZachCMP:
@gigxz All the feedback should be addressed now. Take a look and let me know if anything else is needed here.

gigxz (approved):

PR #268, opened 2023-05-11
185127580-project-households-table

ZachCMP:
https://www.pivotaltracker.com/story/show/185127580

Relies on backend PR: https://github.com/greenriver/hmis-warehouse/pull/3045

### DIFF
--- a/graphql.schema.json
+++ b/graphql.schema.json
@@ -12076,6 +12076,22 @@
             "deprecationReason": null
           },
           {
+            "name": "householdSize",
+            "description": null,
+            "args": [],
+            "type": {
+              "kind": "NON_NULL",
+              "name": null,
+              "ofType": {
+                "kind": "SCALAR",
+                "name": "Int",
+                "ofType": null
+              }
+            },
+            "isDeprecated": false,
+            "deprecationReason": null
+          },
+          {
             "name": "id",
             "description": null,
             "args": [],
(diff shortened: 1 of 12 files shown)


ZachCMP:
@gigxz Just a note on this one:

It was a bit tricky getting the table-within-a-table treatment used for households in the mocks due to the need to line up with the full table headers and the limitation that we have to render all the content for a household in a single row. I decided to just use the `body2` variant for everything with `whiteSpace: 'nowrap'` so that the lines _should_ line up, but I think it's a bit janky. The text selection doesn't work as expected either, which maybe doesn't matter, but does stand out.

I didn't want to go further with it without review, as it's potentially kind of a rabbit hole. If we did want to fix this issue, I see two ways to do it, both of which rely on creating an actual table row for each client:

1. Allow the `render` function to return an array, and if it does, then we will actually render multiple table rows, one for each returned item. We would then render the divider between each group.
2. Alternatively we could provide broader render functions that might let us handle rendering a whole group (something like `renderRow` on the table or something).

Which one I favor depends on whether we think we'll need to do this kind of treatment elsewhere (that's in favor of point 1), and whether or not we'll need more custom row rendering (that's favoring point 2). I suppose we could do both, but that might be overkill.

Anyways, just wanted to raise this point. Maybe we don't need to deal with it now, which I think is fine, but I at least wanted to call it out since I think accessibility might also suffer with this treatment.

gigxz (approved):

gigxz (approved):

gigxz:
Yeah I think this approach is fine, thanks!

gigxz:
only deploy with duplication fix https://github.com/greenriver/hmis-warehouse/pull/3056

PR #272, opened 2023-05-16
185161008 void referral requests

ttoomey:
### please squash merge

Support "cancel referral" button

Misc:
* additional customization params for ConfirmationDialog and DeleteMutationButton
* show apollo errors in dialog
* adjust default fetchPolicy for data table to cache-and-network. Less chance of stale data
* fix typo ProjectReferralRequestsTable component file name

reference:
PT [#185161008](https://www.pivotaltracker.com/story/show/185161008)
warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/3055


src/modules/dataFetching/components/DeleteMutationButton.tsx
```diff
@@ -32,25 +34,24 @@ const DeleteMutationButton = <Mutation, MutationVariables>({
...
   const [deleteRecord, { loading, error }] = useMutation<
     Mutation,
     MutationVariables
-  >(queryDocument, {
-    variables,
-    onCompleted: (result) => {
+  >(queryDocument, { variables });
+  const handleDelete = useCallback(() => {
+    deleteRecord().then((result) => {
+      // fixme: should probably look for errors[] in mutation response
```
gigxz, line 46: Yeah, probably. I can update it to use the same `ErrorState` pattern we have elsewhere. This also doesn't work with  warnings (errors where `severity: warning`), which we will probably want at some point.

src/modules/dataFetching/components/DeleteMutationButton.tsx
```diff
@@ -70,25 +71,30 @@ const DeleteMutationButton = <Mutation, MutationVariables>({
...
-        ) : (
-          <>
-            <Typography>
-              {`Are you sure you want to delete this ${recordName}?`}
-            </Typography>
-            <Typography>This action cannot be undone.</Typography>
-          </>
-        )}
+        <>
+          {error && <ApolloErrorAlert error={error} />}
```
gigxz: This isn't necessary because the `ConfirmationDialog` renders any errors passed in to `errorState`. Won't this result in the error being shown twice?

src/modules/dataFetching/components/GenericTableWithData.tsx
```diff
@@ -65,7 +65,7 @@ const GenericTableWithData = <
   defaultPageSize = DEFAULT_ROWS_PER_PAGE,
   columns,
   recordType,
-  fetchPolicy,
+  fetchPolicy = 'cache-and-network',
```
gigxz: It's intentional that `cache-first` is the default policy for tables. It makes the UI appear a lot faster. For example switching back and forth between the enrollment and assessment tab on the client dashboard, if we use `cache-and-network` the user has to wait for a loading state every time.

It does mean we need to evict the cache when creating a new record or deleting a record, which is kind of annoying, but IMO worth it for the user experience. (It looks like you already added those cache evictions).

ttoomey: I see, my understanding is that the 'cache-and-network' strategy provides the cached data immediately (like cache-first) and then also freshens it from network. The idea being you get fast-loading page initially and fresh data eventually. 

Not sure why that's not working here, I can take a look at it or just revert to the old fetch policy. It could be some interaction with the queryStatus.loading flag that's blocking the UI even though it has data to show.

I mostly added it since a full page reload is slow in dev mode (my guess is that this is a vite/MUI [bundle-size](https://mui.com/material-ui/guides/minimizing-bundle-size/) issue but that's another can of worms)

src/modules/dataFetching/components/DeleteMutationButton.tsx
ttoomey: Ah oops, I somehow missed it was already handled- thank you!

src/modules/dataFetching/components/GenericTableWithData.tsx
ttoomey: Follow-up, yeah it does look like an issue with how we use `status.loading` in GenericTableWithData, I have a fix but it probably should go in another PR (if it's even worth pursuing). I'll revert the change here

gigxz: Oh, interesting. I just misunderstood the loading state, I guess. A separate PR would be great, if its not too much trouble. Or, just an explanation and I can try it out (like do you use the networkStatus to determine which type of loading state it's in, or just only return the loading indicator if `loading && !data` ? I think we show a full loading state when you navigate between pages right now, too. (I'm definitely not attached to that though).

gigxz (approved):

PR #276, opened 2023-05-22
185128929 sort filter table

ZachCMP:
https://www.pivotaltracker.com/story/show/185128929

Relies on backend PR: https://github.com/greenriver/hmis-warehouse/pull/3082

@gigxz This should be ready for you. I tried to follow some conventions that were already in place, and tried to also use the already extant custom form components wherever possible. I'm not sure I got all that right, so please let me know if things need adjusting.

Since I thought there were initially more filters, there is now some unused code, but I left it in place both as a bit of an example, and because other filters may well end up using it.

In terms of usage, you can either provide filters directly to the table or provide a `recordType` and `filterInputType` prop to have it automagically get the filters you want. In the second case you can also provide a `filters` prop to either append or modify the automatic filters by passing and array or a function respectively. That is a different behavior than `columns`, the presence of which prop will totally override anything that was automatically fetched. To do the same thing with filters, you'd have to pass a function and remove the filters you don't want from the automatic ones in that function. That may be totally fine behavior, I just wanted to call out the difference in case that's incorrect.


ZachCMP:
@gigxz The table should auto-detect sort options and default sort option now

gigxz (approved):
looks good

PR #286, opened 2023-05-28
use cache-and-network fetch policy for data tables

ttoomey:
Data tables are served from cache but also freshened. I think this is better than serving potentially stale data but could result in rows blinking in/out shortly after page load. Adjustments to logic around status.loading were needed to make the loading indicators behave


gigxz (approved):
thanks!

PR #287, opened 2023-05-29
185161086 incoming referrals table

ttoomey:
reference
[#185161086](https://www.pivotaltracker.com/story/show/185161086)
https://github.com/greenriver/hmis-warehouse/pull/3099

misc: add check-types script alias


src/modules/projects/components/tables/ProjectsReferralPostingsTable.tsx
```diff
@@ -0,0 +1,78 @@
...
+  ReferralPostingFieldsFragment,
+  ReferralPostingStatus,
+} from '@/types/gqlTypes';
+
+const StatusDisplay: React.FC<{ status: ReferralPostingStatus }> = ({
+  status,
+}) => {
+  return (
+    <Chip
+      label={status.replace(/_status$/, '').replace(/_/g, ' ')}
```
gigxz: We usually do `HmisEnums.ReferralPostingStatus[status]` so that the display value is not coupled to the enum value. I think I'd prefer to keep that pattern.

src/modules/projects/components/tables/ProjectsReferralPostingsTable.tsx
```diff
@@ -0,0 +1,78 @@
...
+  return (
+    <GenericTableWithData<
+      GetProjectReferralPostingsQuery,
+      GetProjectReferralPostingsQueryVariables,
+      ReferralPostingFieldsFragment
+    >
+      queryVariables={{ id: projectId }}
+      queryDocument={GetProjectReferralPostingsDocument}
+      columns={columns}
+      noData='No referral postings'
```
gigxz:
```suggestion
      noData='No referrals'
```

I don't _think_ "posting" needs to be a user-facing term in the hmis.

gigxz (commented):
Looks good! only feedback is to add (and use) enum descriptions as suggested in https://github.com/greenriver/hmis-warehouse/pull/3099

gigxz (approved):

gigxz (approved):

PR #291, opened 2023-05-31
185294347 enrollment filter options

ZachCMP:
https://www.pivotaltracker.com/story/show/185294347

Relies on backend PR: https://github.com/greenriver/hmis-warehouse/pull/3104

### DIFF
--- a/src/components/clientDashboard/enrollments/AllEnrollments.tsx
+++ b/src/components/clientDashboard/enrollments/AllEnrollments.tsx
@@ -92,6 +92,8 @@ const AllEnrollments = () => {
           columns={columns}
           pagePath='client.enrollments'
           fetchPolicy='cache-and-network'
+          showFilters
+          recordType='Enrollment'
         />
       </Paper>
     </>
--- a/src/components/clientDashboard/enrollments/AllEnrollments.tsx
+++ b/src/components/clientDashboard/enrollments/AllEnrollments.tsx
@@ -17,6 +17,7 @@ import { RootPermissionsFilter } from '@/modules/permissions/PermissionsFilters'
 import { ClientDashboardRoutes } from '@/routes/routes';
 import {
   EnrollmentFieldsFragment,
+  EnrollmentSortOption,
   GetClientEnrollmentsDocument,
   GetClientEnrollmentsQuery,
   GetClientEnrollmentsQueryVariables,
@@ -94,6 +95,9 @@ const AllEnrollments = () => {
           fetchPolicy='cache-and-network'
           showFilters
           recordType='Enrollment'
+          filterInputType='EnrollmentsForClientFilterOptions'
+          noSort
+          defaultSortOption={EnrollmentSortOption.MostRecent}
         />
       </Paper>
     </>
--- a/src/modules/hmis/components/HmisFilter.tsx
+++ b/src/modules/hmis/components/HmisFilter.tsx
@@ -6,7 +6,11 @@ import TableFilterItem from '@/components/elements/tableFilters/filters/FilterIt
 import { BaseFilter, FilterType } from '@/modules/dataFetching/types';
 import { HmisEnums } from '@/types/gqlEnums';
 import { GqlInputObjectSchemaType } from '@/types/gqlObjects';
-import { AssessmentSortOption, PickListType } from '@/types/gqlTypes';
+import {
+  AssessmentSortOption,
+  EnrollmentSortOption,
+  PickListType,
+} from '@/types/gqlTypes';
 
 /**
  * Component for dynamically displaying a filter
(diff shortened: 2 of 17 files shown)


src/modules/hmis/components/HmisFilter.tsx
```diff
@@ -57,36 +48,39 @@ export const getDefaultSortOptionForType = (
   recordType: string
 ): string | null => {
   if (recordType === 'Assessment') return AssessmentSortOption.AssessmentDate;
+  if (recordType === 'Enrollment') return EnrollmentSortOption.MostRecent;
```
gigxz, line 51: Seems error-prone that this needs to be kept up-to-date with the backend defaults. I think it would be preferable for the frontend to always pass in the specified sort, so that we can know what the initial sort value is without expecting hard-coded defaults.

src/modules/hmis/components/HmisFilter.tsx
```diff
@@ -57,36 +48,39 @@ export const getDefaultSortOptionForType = (
...
     label: startCase(fieldName),
+    multi: type.kind === 'LIST',
   };
 
   let filter: FilterType<any> | null = null;
 
+  if (inputType === 'ISO8601Date') filter = { ...baseFields, type: 'date' };
+  if (inputType === 'String') filter = { ...baseFields, type: 'text' };
+
   switch (fieldName) {
```
gigxz: This switch seems unnecessary with just one case, can it be removed? If anything, we  would want a switch on the input type, not the field name. But since it depends on both (input type AND field name) then maybe sticking to if-elses is good.

src/components/clientDashboard/enrollments/AllEnrollments.tsx
```diff
@@ -92,6 +92,8 @@ const AllEnrollments = () => {
           columns={columns}
           pagePath='client.enrollments'
           fetchPolicy='cache-and-network'
+          showFilters
+          recordType='Enrollment'
         />
```
gigxz, line 101: This isn't looking quite how I would expect:
1) There are no filters
2) It doesn't make sense to sort on "household id" here since the household ID is not shown (because it is different per row in this context). So, we should hide the sort functionality in this view.

Screen Shot 2023-05-31 at 1 25 34 PM

gigxz: Can you enable sorting and filtering on this table (ProjectClientEnrollmentsTable, I believe)? It is mentioned in the ticket. The ticket also says to update the household table, which you can skip I suppose, since that will require updating the sort and filter options for the household query.


Screen Shot 2023-05-31 at 1 29 16 PM

src/modules/hmis/components/HmisFilter.tsx
ZachCMP: @gigxz The frontend can always pass in the specified sort if needed, this just sets the default if one is not passed in. I take you to be saying that we should not guess at the default for default sort order and always just let the frontend specify it. Is that right?

gigxz: Oh, this actually _sets_ the sort. Nevermind, I misunderstood.

ZachCMP:
@gigxz Addressed all your feedback

gigxz (approved):
Looks good. I made one change to use `localResolvePicklist` which filters out invalid values, since we don't want to show those.

gigxz (changes requested):
Actually there is one more thing @ZachCMP , the date filter doesn't work. Nothing happens when I click apply filters, and it doesn't highlight like anything is applied. The fix wasn't obvious to me, can you take a look?
Screen Shot 2023-06-01 at 9 02 14 AM

ZachCMP:
@gigxz Should be fixed now. I also made it so that if you provide `filters` in the `queryVariables`, it will merge those values with the local filter values held by the table state.

gigxz:
Gotcha. So now if a `filters` object is passed, only those are used (whereas before it was merged with the derived filters). I think that makes sense.

ZachCMP:
@gigxz Yeah, pretty much. In terms of query variables, if you pass a `filters` field, then that will be merged with the local filter values with the latter taking preference. This handles the text search mechanism on the project enrollments page, where we pass the text search through via `queryVariables`.

In terms of filters, it will try to derive the filters by default. If you provide a `filters` prop that is an object, it will completely override the derived filters. If you provide a `filters` prop that is a function, then that function will take the derived filters and return the filters we want to use for the table. Usually this would be to remove a filter from the derived ones, but it could be used to add other filters or modify properties of the derived filters.

Hopefully that makes sense, let me know if you need more clarity there

gigxz (approved):

PR #298, opened 2023-06-01
185276711 referral detail page

ttoomey:
reference
https://github.com/greenriver/hmis-warehouse/pull/3115
https://www.pivotaltracker.com/story/show/185276711

Implement referral details page, household member table for referral

Misc:
* rename some files to correspond to component names
* extract UI into generic components
* build out the TitleCard to support actions. Clean up mess in project referrals

Notes:
* It looks like we want to use body2 for the card-content. I tried to mirror that. 
* ViewCard card title now uses an h5 variant instead of a body1. This seems to be more consistent with other cards


src/api/operations/project.queries.graphql
```diff
@@ -167,6 +167,12 @@ query GetProjectReferralPostings($id: ID!, $limit: Int = 10, $offset: Int = 0) {
   }
 }
 
+query GetReferralPosting($id: ID!) {
```
gigxz, line 170: Can we move the referral-related queries and fragments into their own files? I see we already have `referralRequest.queries`, maybe that could be renamed to `referrals` and include everything referral-related.

(Just to keep things separated for ease of navigating queries, especially since these are not "core hmis")

src/api/operations/referralPosting.fragments.graphql
```diff
@@ -7,3 +7,45 @@ fragment ReferralPostingFields on ReferralPosting {
...
+  score
+  status
+  statusNote
+  statusNoteUpdatedAt
+  statusNoteUpdatedBy
+  statusUpdatedAt
+  statusUpdatedBy
+  unitType {
+    id
+    bedType
```
gigxz: We should be displaying the `description`, not the `bedType`. We don't have a use for the "bed type" field yet, which corresponds to the HUD bed types on the HUD Inventory record.

src/components/elements/CommonLabeledTextBlock.tsx
```diff
@@ -0,0 +1,27 @@
...
+import NotSpecified from './NotSpecified';
+
+interface Props {
+  children: string | null | undefined;
+  title: ReactNode;
+  sx?: TypographyProps['sx'];
+  variant?: TypographyProps['variant'];
+}
+
+export const CommonLabeledTextBlock: React.FC<Props> = ({
```
gigxz, line 13: Thanks for setting these up, long time coming 😬 🙂

src/modules/projects/components/ProjectReferralPostingDetails.tsx
```diff
@@ -0,0 +1,59 @@
...
+  referralPosting,
+}) => {
+  const bedType = referralPosting?.unitType?.bedType;
+  const col1: Array<[string, ReactNode]> = [
+    [
+      'Referral Status',
+      <ReferralPostingStatusDisplay status={referralPosting.status} />,
+    ],
+    ['Referral Project Name', referralPosting.referredFrom],
+    ['Referral Date', referralPosting.referralDate],
```
gigxz: This should be parsed into the display format using `parseAndFormatDate(referralPosting.referralDate)`

src/modules/projects/components/tables/ProjectReferralHouseholdMembersTable.tsx
```diff
@@ -0,0 +1,52 @@
...
+  },
+  {
+    header: 'Birth Date',
+    render: ({ client }: Row) =>
+      client.dob ? parseAndFormatDate(client.dob) : '',
+  },
+  {
+    header: 'Gender',
+    render: ({ client }: Row) =>
+      client.gender.map((g) => HmisEnums.Gender[g]).join(', '),
```
gigxz: We can use the enum display component here

```
                    <MultiHmisEnum
                      values={client.gender}
                      enumMap={HmisEnums.Gender}
                    />
```

src/modules/projects/components/tables/ProjectReferralHouseholdMembersTable.tsx
```diff
@@ -0,0 +1,52 @@
...
+    render: ({ client }: Row) => (
+      <ClientName
+        client={client}
+        linkToProfile={client.access.canViewEnrollmentDetails}
+      />
+    ),
+  },
+  {
+    header: 'Relationship to HoH',
+    render: (row: Row) => HmisEnums.RelationshipToHoH[row.relationshipToHoH],
```
gigxz: Same here, we could use the `HmisEnum` component. Not a huge deal, the main benefit is that the `INVALID` enum would be displayed in red. We have a system whereby numeric values that can't be mapped to the gql enum are resolved as INVALID .

src/modules/projects/components/tables/ProjectReferralHouseholdMembersTable.tsx
```diff
@@ -0,0 +1,52 @@
...
+  },
+  {
+    header: 'Birth Date',
+    render: ({ client }: Row) =>
+      client.dob ? parseAndFormatDate(client.dob) : '',
```
gigxz:
```suggestion
    header: 'DOB / Age',
    render: ({ client }: Row) =>
         <ClientDobAge client={client} reveal />,
```

src/modules/projects/components/tables/ProjectReferralPostingsTable.tsx
```diff
@@ -63,6 +52,16 @@ interface Props {
 export const ProjectReferralPostingsTable: React.FC<Props> = ({
   projectId,
 }) => {
+  const rowLinkTo = useCallback(
```
gigxz, line 56: Can you add `linkTreatment: true,` to the HoH name column definition? Without it, there is no way to tab through the rows, because we use the "linked" cell  as the tab index (even though all cells in the row are clickable). The link treatment also indicates to the user visually that the row is clickable.

gigxz (commented):
Left a few minor suggestions, but overall looks good!

src/modules/projects/components/ProjectReferralPostingDetails.tsx
```diff
@@ -0,0 +1,59 @@
...
+    ['Referred By', referralPosting.referredBy],
+    ['Score', referralPosting.score],
+  ];
+  const col2: Array<[string, ReactNode]> = [
+    ['Referral ID', referralPosting.referralIdentifier],
+    ['Unit Type', bedType ? HmisEnums.InventoryBedType[bedType] : ''],
```
gigxz: This should be the unit type description

src/modules/projects/components/tables/ProjectReferralPostingsTable.tsx
gigxz: This should be done for the referral request table as well

src/api/operations/project.queries.graphql
ttoomey: sure, I had moved the queries already in [185276711-referral-detail-page](https://github.com/greenriver/hmis-frontend/tree/185276711-referral-detail-page)

src/modules/projects/components/tables/ProjectReferralPostingsTable.tsx
ttoomey: One thing with the referral request table is that it doesn't have a rowLinkTo. I added the linkTarget to the cancel button since it's the only clickable element there but that may not be correct

ttoomey:
> Left a few minor suggestions, but overall looks good!

Thanks @gigxz - I think I've addressed all the feedback if you want to take another look

src/modules/projects/components/tables/ProjectReferralPostingsTable.tsx
gigxz: Oh right, I forgot that the referral request table rows aren't clickable. All looks correct now though. The void buttons are tabbable, as are the HoH Name cells on the referral table.

gigxz (approved):
Looks good!

PR #302, opened 2023-06-03
185276764 update referral posting status

ttoomey:
reference
https://www.pivotaltracker.com/story/show/185276764
https://github.com/greenriver/hmis-warehouse/pull/3119

notes:
* currently based off [1739794124](https://github.com/greenriver/hmis-frontend/pull/302#issue-1739794124)
* denialReason values are TBD
* I'm hiding the update form unless the referral is assigned, this may not be right
* denial reason only is relevant for denied_pending status. Not sure how to get conditional forms to happen with our DynamicForm component
* maybe the form should also allow the user to change the denial note too? I added it to the input but didn't add it to the form since it wasn't in spec


src/modules/form/data/referralPosting.json
```diff
@@ -0,0 +1,27 @@
...
+      "fieldName": "status",
+      "pickListReference": "REFERRAL_POSTING_STATUSES"
+    },
+    {
+      "type": "CHOICE",
+      "linkId": "denialReasonId",
+      "required": false,
+      "text": "Reason for Denial",
+      "fieldName": "denialReasonId",
+      "pickListReference": "REFERRAL_POSTING_DENIAL_REASON_TYPES"
```
gigxz: To add conditional:

```
      "enableBehavior": "ALL",
      "enableWhen": [
        {
          "question": "status",
          "operator": "EQUAL",
          "answer_code": "denied_pending_status"
        }
      ]
```

src/modules/form/components/DynamicField.tsx
```diff
@@ -317,8 +317,11 @@ const DynamicField: React.FC<DynamicFieldProps> = ({
       } else {
         inputComponent = (
           <FormSelect
-            // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-            value={currentValue}
```
gigxz, line 321: Eesh yes, there is a bug, but I'm not sure if this is the fix we want.

The call to `getValueFromPickListData` on line 125 is supposed to set the initial value based on the remotely fetched picklist, but it's not being called because the picklist fetch isn't happening anymore since the pick lists are now pre-loaded using `usePreloadPicklists`.

As you can see the `getValueFromPickListData` function has some more complicated logic than simply finding the matching code - it also chooses the initial code, and supports arrays of codes.

A simple fix would be something like updating the usage of `usePickList` hook so that the value gets changed whether the fetch happened or not.

Would you mind digging into this? Otherwise I will create a separate ticket and I can prioritize it for myself this week.

src/modules/projects/components/ProjectReferralPostingForm.tsx
```diff
@@ -0,0 +1,63 @@
...
+  );
+
+  return (
+    <DynamicForm
+      definition={ReferralPostingDefinition}
+      FormActionProps={{
+        submitButtonText: 'Update Referral',
+        discardButtonText: 'Cancel',
+      }}
+      initialValues={referralPosting}
```
gigxz: Sooo - the posting itself can't just be passed in as `initialValues`, it needs to be transformed first. Definitely a confusing API. You'll need to call "createInitialValuesFromRecord" to create the initial form state which is keyed by link id. It also transforms the values (like creating date object, picklist object, etc)

src/modules/form/data/referralPosting.json
gigxz: Sorry, key would be `answerCode`. We don't typically have these on the frontend, but in this case I think its fine.

src/modules/form/data/referralPosting.json
```diff
@@ -0,0 +1,27 @@
+{
+  "item": [
+    {
+      "type": "CHOICE",
+      "linkId": "status",
+      "required": false,
+      "text": "Select Referral Status",
```
gigxz:
```suggestion
      "text": "Referral Status",
```

We don't usually put "select" in the status. It will be added to the placeholder.

src/modules/form/data/referralPosting.json
```diff
@@ -0,0 +1,27 @@
...
+      "fieldName": "status",
+      "pickListReference": "REFERRAL_POSTING_STATUSES"
+    },
+    {
+      "type": "CHOICE",
+      "linkId": "denialReasonId",
+      "required": false,
```
gigxz:
```suggestion
      "required": true,
```

src/modules/form/data/referralPosting.json
```diff
@@ -0,0 +1,27 @@
+{
+  "item": [
+    {
+      "type": "CHOICE",
+      "linkId": "status",
+      "required": false,
```
gigxz:
```suggestion
      "required": true,
```

src/modules/projects/components/ProjectReferralPostingForm.tsx
gigxz: That should fix the issue you saw with DynamicField

src/modules/form/data/referralPosting.json
ttoomey: Make sense, I picked it up from the screenshot of the mock in https://www.pivotaltracker.com/story/show/185276711

src/modules/form/data/referralPosting.json
ttoomey: That's great, thanks!

src/modules/projects/components/ProjectReferralPostingForm.tsx
ttoomey: Got it thanks for explaining. I backed out the previous fix I made to the select component

src/modules/projects/components/ProjectReferralPostingForm.tsx
```diff
@@ -0,0 +1,85 @@
...
+        variables: {
+          id: referralPosting.id,
+          input,
+        },
+      });
+    },
+    [mutate, referralPosting.id]
+  );
+
+  // hook copied from Edit Record
```
gigxz, line 54: This has been moved into a hook on https://github.com/greenriver/hmis-frontend/pull/297, I'll update this to use it once they're both merged in

gigxz (approved):

PR #303, opened 2023-06-05
Add linkcolor

ZachCMP:
https://www.pivotaltracker.com/story/show/185297189

### DIFF
--- a/src/config/theme.ts
+++ b/src/config/theme.ts
@@ -28,12 +28,12 @@ declare module '@mui/material/styles' {
   interface Palette {
     borders: PaletteColor;
     alerts: Record<string, string>;
-    linkColor: string;
+    links: string;
   }
   interface PaletteOptions {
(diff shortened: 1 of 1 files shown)


src/config/theme.ts
```diff
@@ -28,10 +28,12 @@ declare module '@mui/material/styles' {
   interface Palette {
     borders: PaletteColor;
     alerts: Record<string, string>;
+    linkColor: string;
   }
   interface PaletteOptions {
     borders: SimplePaletteColorOptions;
     alerts: { lightWarningBackground?: string };
+    linkColor: string;
```
gigxz: I think `Color` is redundant - since this is a `Palette`, everything in it should be a color, right? Can we just call it `links`?

ZachCMP:
@gigxz Fixed

gigxz (approved):
thanks!

PR #315, opened 2023-06-09
Support creating client within household mgmt; clean up dialogs; bulk services fix

gigxz:


src/components/elements/CommonDialog.tsx
```diff
@@ -0,0 +1,26 @@
...
+const CommonDialog = ({ children, ...props }: DialogProps) => {
+  return (
+    <Dialog {...props}>
+      {children}
+      {props.onClose && (
+        <IconButton
+          aria-label='close'
+          onClick={(e) => props.onClose && props.onClose(e, 'backdropClick')}
+          sx={{
+            position: 'absolute',
```
ttoomey, line 13: It seems possible for the dialog title to flow under the close button. Perhaps there should be some left padding on the title?

src/modules/form/hooks/useFormDialog.tsx
```diff
@@ -1,49 +1,197 @@
...
+  const formRef = useRef<DynamicFormRef>(null);
+  const [errors, setErrors] = useState<ErrorState>(emptyErrorState);
+  const [submitLoading, setSubmitLoading] = useState(false);
+  const closeDialog = useCallback(() => {
+    setDialogOpen(false);
+    setErrors(emptyErrorState);
+    setSubmitLoading(false);
+  }, []);
+
+  const [submitForm, { loading: saveLoading }] = useSubmitFormMutation({
```
gigxz: This is a lot of repeated code. I'm working on a separate branch to pull it into a hook, since its now used in several places.

src/components/elements/CommonDialog.tsx
gigxz: Added!

PR #318, opened 2023-06-12
185383691-missing-useSafeParams

ttoomey:
https://www.pivotaltracker.com/story/show/185383691

* replace useParams with useSafeParams
* linter warning on use of useParams
* adjust useSafeParams to accept generic like useParams (not exactly right but better than nothing)


gigxz (approved):
💯  thanks

PR #321, opened 2023-06-14
185282623 update referral denial

ttoomey:
https://www.pivotaltracker.com/story/show/185282623
https://github.com/greenriver/hmis-warehouse/pull/3159

Note: 
* was not able to include the status note as read-only field in the form, maybe our dynamic forms don't support that. 
* added grid props to dynamic view since it introduces unwanted padding


src/modules/admin/components/AdminReferralPostingDetails.tsx
```diff
@@ -24,14 +29,43 @@ const AdminReferralPostingDetails: React.FC<Props> = ({ referralPosting }) => {
       />,
     ],
     ['Referral ID', referralPosting.referralIdentifier],
-    ['Project Name', referralPosting.referredFrom],
     ['Referral Date', parseAndFormatDate(referralPosting.referralDate)],
-    ['Referred By', referralPosting.referredBy],
+    ['Project Name', referralPosting.referredFrom],
```
gigxz: Re: my other comment about `referredFrom` on the backend PR, I think we want this to be:

```
['Referred From', referralPosting.referredFrom],
['Project Name', <RouterLink to='' openInNew>referralPosting.project.projectName</RouterLink>],
```

Since we're in the admin dashboard context, I think linking to the project page is a good idea. We can use `openInNew` to open in a new tab with a little icon, since it bumps you out of this context.

src/modules/admin/components/AdminReferralPostingForm.tsx
```diff
@@ -0,0 +1,106 @@
...
+      });
+    },
+    [mutate, referralPosting.id, formDefinition]
+  );
+
+  const initialValues = useInitialFormValues({
+    definition: formDefinition,
+    record: referralPosting as unknown as SubmitFormAllowedTypes,
+    // there may be some way to use this to set re-request
+    //localConstants: { resendReferralRequest: 'true' },
```
gigxz: Yeah, you can use a local constant for the conditional logic. Something like:
```
localConstants: { hasReferralRequest: !!referralPosting.referralRequest }
```

and then at the to of the form
```
    {
      "type": "STRING",
      "linkId": "has-referral-request",
      "hidden": true,
      "initial": [
        {
          "initialBehavior": "OVERWRITE",
          "valueLocalConstant": "$hasReferralRequest"
        }
      ]
    },
```

and then the logic changes to 
```
      "fieldName": "resendReferralRequest",
      "enableBehavior": "ALL",
      "enableWhen": [
        {
          "question": "status",
          "operator": "EQUAL",
          "answerCode": "denied_status"
        },
        {
          "_comment": "re-requests are only allowed if the referral came from a request",
          "question": "has-referral-request",
          "operator": "EQUAL",
          "answerBoolean": true
        }
      ]
```

src/modules/admin/components/AdminReferralPostingForm.tsx
```diff
@@ -0,0 +1,106 @@
...
+const AdminReferralPostingForm: React.FC<Props> = ({
+  referralPosting,
+  readOnly = false,
+}) => {
+  const formDefinition = useMemo<FormDefinitionJsonFieldsFragment>(() => {
+    if (referralPosting.postingIdentifier) {
+      // came from LINK, show re-request option
+      return AdminReferralPostingDefinition;
+    } else {
+      // referral request originated locally, hide re-request option
```
gigxz: lets use form logic for that, see below

src/modules/form/data/referralPosting.json
```diff
@@ -22,24 +22,7 @@
...
-        },
-        {
-          "code": "19",
-          "label": "Estimated vacancy no longer available"
-        }
-      ]
+      "pickListReference": "REFERRAL_POSTING_DENIAL_REASON_TYPES"
     },
     {
       "type": "TEXT",
```
gigxz, line 28: Can you mark the status note `required: true`? Re: https://github.com/greenriver/hmis-warehouse/pull/3159#discussion_r1231396343

src/modules/form/data/adminReferralPosting.json
```diff
@@ -0,0 +1,58 @@
...
+          "question": "status",
+          "operator": "EQUAL",
+          "answerCode": "denied_status"
+        }
+      ]
+    },
+    {
+      "type": "TEXT",
+      "linkId": "denialNote",
+      "required": false,
```
gigxz: It looks like the backend validator requires it, so we should mark it as required here. I asked to confirm whether it really should be required.

src/modules/form/data/adminReferralPosting.json
```diff
@@ -0,0 +1,58 @@
+{
+  "item": [
+    {
+      "type": "CHOICE",
+      "linkId": "denialReason",
+      "required": true,
+      "text": "Reason for Denial",
+      "fieldName": "denialReason",
+      "pickListReference": "REFERRAL_POSTING_DENIAL_REASON_TYPES"
```
gigxz: We should be able to use a "local" picklist for the denial reasons. We only really need to fetch them from the backend if the list is dynamic, or if it's grouped.

The way we'd do that is by resolving denialReson as an Enum. https://github.com/greenriver/hmis-warehouse/pull/3159/files#r1231425655
Once that Enum is in the graphql schema, you can pickListReference it here.

src/modules/admin/components/AdminReferralPostingForm.tsx
ttoomey: Oh wow, I don't think I would have figured this out on my own. Thank you!

gigxz: totally, the form stuff needs documenting. no problem

src/modules/admin/components/AdminReferralPostingForm.tsx
```diff
@@ -0,0 +1,100 @@
...
+        },
+      });
+    },
+    [mutate, referralPosting.id, formDefinition]
+  );
+
+  const initialValues = useInitialFormValues({
+    definition: formDefinition,
+    record: referralPosting as unknown as SubmitFormAllowedTypes,
+    localConstants: { hasReferralRequest: !!referralPosting.postingIdentifier },
```
gigxz: There is an ability to "admin assign" from the external system, whereby a new referral is sent to the HMIS even though it is not "fulfilling" any vacancy request. In that case, the Referral and Posting would both have Identifiers, but they would not be linked to a ReferralRequest.

In that scenario, we probably shouldn't show the "re-request" option. Sending a referral request would fail because there is no identifier there. My bad, that detail was not included in the ticket.


```
    localConstants: { hasReferralRequest: !!referralPosting.postingIdentifier && !!referralPosting.referralRequest },
```

ttoomey: got it, I adjusted the condition and query to include the referral request

gigxz (approved):
looks great

PR #322, opened 2023-06-14
Fix issue displaying 0 and false in ReferralPostingDetails

gigxz:
...and show MCI IDs on the household table.


ttoomey (approved):
Looks good!

PR #326, opened 2023-06-16
Updated Unit Management tool; make GenericTable selectable

gigxz:
requires https://github.com/greenriver/hmis-warehouse/pull/3171

Screen Shot 2023-06-16 at 3 19 27 PM


gigxz:
This is a huge PR @ttoomey but I figured getting a second pair of eyes would be a good idea if you have time. I changed the GenericTable component, which is used throughout the app, to be selectable, following the MUI pattern https://mui.com/material-ui/react-table/#sorting-amp-selecting. There are components that can/shoudl be pulled into different files probably.

PR #354, opened 2023-07-11
Support filtering projects by org

gigxz:
Requires https://github.com/greenriver/hmis-warehouse/pull/3257


ZachCMP (approved):
I don't have the full context of what was intended here, maybe some cleanup and adjustment to filter options? Code looks good to me regardless. Let me know if you want me to get filled in on the context and take a closer look

PR #360, opened 2023-07-17
185072207-improve-session-expiration

ttoomey:
https://www.pivotaltracker.com/story/show/185072207
https://github.com/greenriver/hmis-warehouse/pull/3275

* simplify userAuth global context, login/logout are moved into the components that need them
* catch and raise errors on logout to user
* remove unnecessary CSRF calls on get requests
* fetch settings and current user in parallel, cache app settings. Should improve page load time
* replace polling with time-base calculations
* session expiration warning prompt
* adjust eslint to warn on console errors. Change import order to be a warning (seem okay?)


src/modules/auth/components/SessionStatusManager.tsx
```diff
@@ -0,0 +1,72 @@
...
+      .catch(() => reloadWindow());
+  }, []);
+
+  // current session has ended, user clicks continue
+  const handleReload = useCallback(() => {
+    setLoading(true);
+    reloadWindow();
+  }, [setLoading]);
+
+  switch (status) {
```
gigxz, line 29: this looks awesome, thank you for setting up the nice user messaging

src/modules/auth/components/LoginForm.tsx
```diff
@@ -20,18 +25,44 @@ const errorMessage = (error: Error) => {
...
+    (user: HmisUser) => {
+      setUser(user);
+      navigate((state as RouteLocationState)?.prev || '/');
+    },
+    [navigate, state, setUser]
+  );
 
   const handleSubmit = useCallback(
     (event: FormEvent<HTMLFormElement | HTMLDivElement>) => {
       event.preventDefault();
```
gigxz, line 57: I think we are missing a `setLoading(true);` here. Nothing happens when you click submit which feels laggy.

src/modules/auth/components/OneTimePassword.tsx
```diff
@@ -3,19 +3,30 @@ import Box from '@mui/material/Box';
...
-  const { login, loading, error } = useAuth();
+  const [loading, setLoading] = useState<boolean>(false);
 
-  function handleSubmit(event: React.FormEvent<HTMLFormElement>) {
-    event.preventDefault();
-    login({ otpAttempt: otpAttempt });
-  }
+  const handleSubmit = useCallback(
+    (event: FormEvent<HTMLFormElement | HTMLDivElement>) => {
+      event.preventDefault();
```
gigxz, line 20:
```suggestion
      event.preventDefault();
      setLoading(true);
```

.eslintrc.js
```diff
@@ -90,14 +90,19 @@ module.exports = {
...
         ],
-        'newlines-between': 'always',
         alphabetize: {
           order: 'asc',
           caseInsensitive: true,
         },
         warnOnUnassignedImports: false,
       },
     ],
+    'no-console': [
```
gigxz, line 100: There are log statements that I think are convenient to keep around for debugging (in dev). What do you think about dropping them at minification https://github.com/greenriver/hmis-frontend/pull/363 
I think it can co exist with this warning. It is helpful to see the warning in the editor to highlight any logs I may have unintentionally kept around.

gigxz (approved):
This is a huge improvement!

.eslintrc.js
ttoomey: Yeah I added it to help me catch debugging clutter I don't want commit. For spots were we want logging, we could add `// eslint-disable-line` magic comments to keep the linter output clean

I'm a little on the fence about dropping the log statements during minify. I guess as long as there's still some way to log in production it would be fine.

src/modules/auth/components/LoginForm.tsx
ttoomey: Good spot, thanks!

src/modules/auth/components/OneTimePassword.tsx
ttoomey: thanks!

ttoomey (commented):
Thanks @gigxz - I realized that the retry-wrapper I made to the session api functions wasn't working as intended so I backed that out. And I adjusted login so it retries if there's a csrf failure

.eslintrc.js
gigxz: If for some reason we need info logging in prod, we could change it to allow `info`/`warn`/`error` in prod builds (filtering out `debug`/`log`).

gigxz (approved):
Tested the csrf retry on login, looks good to me

PR #371, opened 2023-07-19
Allow moving from AcceptedPending=>DeniedPending

gigxz:
https://www.pivotaltracker.com/story/show/185648474

hacking in a change to the form definition

### DIFF
--- a/src/modules/projects/components/ProjectReferralPostingForm.tsx
+++ b/src/modules/projects/components/ProjectReferralPostingForm.tsx
@@ -1,5 +1,5 @@
-import { useCallback, useState } from 'react';
-
+import { remove } from 'lodash-es';
+import { useCallback, useState, useMemo } from 'react';
 import { emptyErrorState, partitionValidations } from '@/modules/errors/util';
 import DynamicForm, {
   DynamicFormOnSubmit,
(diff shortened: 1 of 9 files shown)


src/modules/projects/components/ProjectReferralPostingForm.tsx
```diff
@@ -34,10 +35,28 @@ export const ProjectReferralPostingForm: React.FC<Props> = ({
     },
     onError: (apolloError) => setErrors({ ...emptyErrorState, apolloError }),
   });
+
+  const definition = useMemo(() => {
+    // Hacky way to drop the "Assigned" option if this posting is in AcceptedPending status.
+    // This lets user change from AcceptedPending=>DeniedPending which is sometimes necessary.
+    const definitionCopy = { ...ReferralPostingDefinition };
```
ttoomey: Because this is a shallow copy, changing `item[0].pickListOptions` will change the original object as well. This can result in some really weird bugs. Recommend we change this.

For small objects that can be serialized, probably the easiest way to get a deep copy is:
```suggestion
    const definitionCopy = JSON.parse(JSON.stringify(ReferralPostingDefinition)) as typeof ReferralPostingDefinition;
```

ttoomey (changes requested):
Looks good! There's just one small bug that should probably get fixed

src/modules/projects/components/ProjectReferralPostingForm.tsx
gigxz: Ah thank you, good catch

PR #381, opened 2023-07-25
adds some possible fixes for nginx freezing up

yakloinsteak:
Sending the HUP signal to nginx fixed things immediately, so this PR is a mix of trying to solve the real problem and having that as the backup plan.

* Limits keep-alives in a few ways since something was occupying all 1024 connections of all the workers when this freezes up. Maybe it was these.
* explicitly set number of workers. Might have been too little CPU given the number of workers.
* run nginx-reload script which gracefully lets connections finish and starts up new workers every hour. This is a backup plan if we didn't actually solve the problem with the other settings.
* fails faster if proxy connections take too long. 30 seconds instead of 60 seconds

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -1,0 +1,49 @@
+#!/bin/sh
+# vim:sw=4:ts=4:et
+
+set -e
+
+entrypoint_log() {
+    if [ -z "${NGINX_ENTRYPOINT_QUIET_LOGS:-}" ]; then
+        echo "$@"
+    fi
+}
+
+/nginx-reload &
+
+if [ "$1" = "nginx" -o "$1" = "nginx-debug" ]; then
+    if /usr/bin/find "/docker-entrypoint.d/" -mindepth 1 -maxdepth 1 -type f -print -quit 2>/dev/null | read v; then
+        entrypoint_log "$0: /docker-entrypoint.d/ is not empty, will attempt to perform configuration"
+
+        entrypoint_log "$0: Looking for shell scripts in /docker-entrypoint.d/"
+        find "/docker-entrypoint.d/" -follow -type f -print | sort -V | while read -r f; do
+            case "$f" in
+                *.envsh)
+                    if [ -x "$f" ]; then
+                        entrypoint_log "$0: Sourcing $f";
+                        . "$f"
+                    else
+                        # warn on shell scripts without exec bit
+                        entrypoint_log "$0: Ignoring $f, not executable";
+                    fi
+                    ;;
+                *.sh)
+                    if [ -x "$f" ]; then
+                        entrypoint_log "$0: Launching $f";
+                        "$f"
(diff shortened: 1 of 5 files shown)


docker/docker-entrypoint.sh
```diff
@@ -0,0 +1,49 @@
...
+
+set -e
+
+entrypoint_log() {
+    if [ -z "${NGINX_ENTRYPOINT_QUIET_LOGS:-}" ]; then
+        echo "$@"
+    fi
+}
+
+/nginx-reload &
```
yakloinsteak, line 12: This is the only change to the file that was present.

schwartzkylej (approved):
LGTM

docker/templates/proxy_headers.conf.template
```diff
@@ -5,9 +5,9 @@ proxy_set_header X-Forwarded-Port $server_port;
 proxy_set_header X-Forwarded-Host $host;
 
 # Defaults to 60 seconds.
-#proxy_connect_timeout       300;
-#proxy_send_timeout          300;
-#proxy_read_timeout          300;
+proxy_connect_timeout       30;
+proxy_send_timeout          30;
+proxy_read_timeout          30;
```
gigxz: Is this going to mess with the cascading timeouts that are needed for the LB? The LB has 55sec and keepalive timeout is 60, and the direction of that difference was important iirc. ("make sure that the keep-alive timeout is greater than the idle timeout settings of your load balancer." [here](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/ts-elb-error-message.html#ts-elb-errorcodes-http503)).

I don't know how the proxy timeout is applied, so feel free to ignore me if this is a non issue. Just wanted to flag the idea just in case.

gigxz (approved):

docker/templates/proxy_headers.conf.template
yakloinsteak: To be honest, I'm not sure, but I had convinced myself that this just needed to be less than the warehouse lb timeout and would "kick out" connections that were stalled more quickly. With nginx doing buffering, I think these values just need to be lower than 65s.

I'll take your advice and put this back since I really don't know for sure, and we'll see if the more restrictive lifetimes of keepalive requests helps by itself. It's a better test to try one thing at a time anyway.

yakloinsteak:
@gigxz I'm going to merge this after CI passes unless you have any more concerns. I increased the time between nginx reloads since I want to try to balance fixing this with masking the problem via those reloads.

gigxz:
Sounds good to me. Thanks Todd

docker/templates/proxy_headers.conf.template
schwartzkylej: The keepalive_timeout is the value that directly correlates with the load balancer idle timeout. Proxy timeouts may contribute to upstream connections though. I would keep them disabled for initial testing.

PR #382, opened 2023-07-25
Occurrence point data collection (MID, DOE, PATH status, Custom, + Unit Assignment)

gigxz:
Merge with https://github.com/greenriver/hmis-warehouse/pull/3305 (See description there)


src/modules/form/util/formUtil.ts
```diff
@@ -149,14 +152,22 @@ export const resolveOptionList = (
...
   if (isPickListOption(value)) return value;
-  if (isDataNotCollected(value)) {
+
+  // Allow convertying object (like a Unit or Project) into a
+  // pick list option; assuming 'id' is the code.
+  if (isTypedObjectWithId(value)) {
+    return { code: value.id };
+  }
+  if (typeof value !== 'string') {
+    console.error(`Can't get option value for ${value}`);
```
ttoomey: Is a string expected? Perhaps this should raise if not

src/modules/enrollment/components/EditableOccurrencePointValue.tsx
```diff
@@ -0,0 +1,63 @@
...
+        Icon={icon === 'calendar' ? DateRangeIcon : EditIcon}
+      >
+        {children}
+      </IconButtonContainer>
+      {renderFormDialog(formDialogArgs)}
+    </>
+  );
+};
+
+const PermissionWrappedEditableOccurrencePointValue = (props: Props) => {
```
ttoomey, line 56: What do you think about trying to standardize on React.FC?
```suggestion
const PermissionWrappedEditableOccurrencePointValue = (props):React.FC<Props> => {
```

src/modules/hmis/hmisUtil.ts
```diff
@@ -371,6 +397,18 @@ export const serviceTypeSummary = (st: ServiceTypeFieldsFragment) => {
   return [st.category, st.name].join(': ');
 };
 
+export const customDataElementValueAsString = (
+  cde: CustomDataElementFieldsFragment
+): string | undefined => {
+  if (cde.value) return customDataElementValue(cde.value, 'for_display');
```
ttoomey, line 403: This is fine, differentiating behavior on cde.value feels a bit fragile though. Logically we would use cde.repeats but that wouldn't make TS happy. Could handle this is with a Union type of CustomDataElement | CustomDataElementRepeated but that seems like a lot of refactoring for little gain.

ttoomey (approved):

src/modules/enrollment/components/EditableOccurrencePointValue.tsx
gigxz: I think I loosely would prefer not to, because 1) it doesn't accept generics, and 2) it feels less readable to me. Are there particular benefits that you like about it? Apparently it also adds a `children` prop implicitly (I had no idea) which could lead to un caught type errors ([link](https://juhanajauhiainen.com/posts/should-you-use-reactfc-to-type-your-components)). I don't feel super strongly about it though, it's not a big deal to treat generic components differently.

src/modules/hmis/hmisUtil.ts
gigxz: Yeah I agree. The GraphQL only returns `value` if its a non-repeating item, though, and `values` if its a repeating item. So its an either-or.

src/modules/enrollment/components/EditableOccurrencePointValue.tsx
ttoomey: Those are good points, there is some controversy with React.FC. I like it because I think it conveys intent well and I find the return type checking to be a helpful safeguard. But if you find it to be less readable, that is a strong point against it. I am in favor of standardizing how we declare components but it doesn't have to be with React.FC, I've just been using it for so long that it looks normal/default to me.

FWIW [I gather that as of Typescript 5.1 the `children` prop is no longer included implicitly](https://dev.to/asciibi/reactfc-is-fixed-5ckn) and the return type is more permissive.

gigxz: Oh neat, okay. Impossible to keep up with all the react updates 🏃‍♀️ 
Neither looks particularly normal to me, so, I'm happy to standardize around React.FC. I don't think its worth going back to change components, but going forward we can use React.FC for non generic components.

ttoomey: Sounds great, thanks! Yeah I totally agree, it's not worth changing existing components. We can add it incrementally going forward. We probably should upgrade to typescript to 5.1 though, it shouldn't be too hard, I can tackle that if you want.

Lacking support for generic components is unfortunate; I suspect those are uncommon(?) in this project so it's probably okay

gigxz: Yeah go for it on the typescript upgrade! We dont have too many generic components. Just a few that are used a lot.

PR #385, opened 2023-07-26
183269204 add cls records

ZachCMP:
https://www.pivotaltracker.com/story/show/183269204

Relies on backend PR: https://github.com/greenriver/hmis-warehouse/pull/3308


gigxz (approved):
Looks great! We also need the ability to edit and to check for permissions, so I'm pushing a commit for that.

gigxz:
Schema is going to need to be re-generated since the cls subsidy casing changed, I think

ZachCMP:
Fixed

PR #387, opened 2023-07-27
adds CSP in reporting-only mode

yakloinsteak:


docker/Dockerfile
```diff
@@ -24,11 +22,33 @@ ENV NODE_OPTIONS '--max-old-space-size=32768'
...
+# Copying files more specifically lets us reuse the build cache and iterate more quickly
+# COPY . .
+# RUN corepack enable \
+#   && yarn install \
+#   && yarn format:check \
+#   && yarn lint:check \
+#   # && yarn test \
+#   && yarn build
+
+RUN mkdir -p public src scripts .husky .yarn .storybook cypress \
```
yakloinsteak: I can revert this, @gigxz if you want, but it was rebuilding if you touched any file at all. It's messier, but caches much better.

docker/templates/default.conf.template
```diff
@@ -15,9 +15,9 @@ server {
         index  index.html index.htm;
         try_files $uri /index.html;
 
-        # application has inline styles
         # use INLINE_RUNTIME_CHECK=false yarn build and nonce / hash to apply a CSP
-        #add_header Content-Security-Policy "default-src 'none'; script-src 'self'; connect-src 'self'; img-src 'self'; font-src fonts.gstatic.com; style-src 'self' fonts.googleapis.com;" always;
+        # Use add_header Content-Security-Policy "......." always;
+        add_header Content-Security-Policy-Report-Only "default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' fonts.googleapis.com 'unsafe-inline'; img-src 'self' data:; font-src 'self' fonts.gstatic.com" always;
```
yakloinsteak, line 20: If all is well after this is deployed, we will circle back and take off the "-Report-Only" part of the header key.

docker/Dockerfile
```diff
@@ -24,11 +22,33 @@ ENV NODE_OPTIONS '--max-old-space-size=32768'
...
 RUN corepack enable \
- && yarn install --offline \
- && yarn format:check \
+ && yarn cache clean \
+ && yarn install
+
+COPY cypress/ cypress
+COPY src/ ./src
+
+RUN yarn format:check \
```
yakloinsteak: I did have to remove the `yarn test` to make progress. I can put it back if you want now that I've been able to minimally test things locally.

docker/Dockerfile
```diff
@@ -24,11 +22,33 @@ ENV NODE_OPTIONS '--max-old-space-size=32768'
...
+COPY .yarn/ .yarn
+COPY .husky/ .husky
+COPY .storybook/ .storybook
+COPY .gitignore .yarnrc .prettierignore .prettierrc.json .nvmrc .eslintrc.js .env.development .env.production babel.config.js cypress.config.ts jest.config.ts package.json tsconfig.json vite.config.js yarn.lock index.html graphql.schema.json codegen.yml .
+
 RUN corepack enable \
- && yarn install --offline \
- && yarn format:check \
+ && yarn cache clean \
+ && yarn install
```
gigxz: I would like to keep this installing from the offline mirror so that we don't run into issues with the yarn package registry being unreliable

docker/Dockerfile
gigxz: it's fine, it doesn't need to run in the docker image. we run it on ci.

docker/Dockerfile
gigxz: I think there is a better way to do this which is just keeping `COPY . .` but adding a `.dockerignore` with node_modules in it. It should make it a lot faster.  Can you try that and see if its good enough for your work setup? (And remove the `yarn cache clean`).

I worry about listing each individual (root-level) file in here, it seems like it will be hard to maintain.

yakloinsteak: TL;DR Sure.
I'll do that, though it reruns yarn install even if your dependencies don't change. Now that I'm done iterating, though, it doesn't matter to me much. I thought it would be just the yarn.lock and package.json file, and then I kept adding things in until it worked (I clearly don't know this ecosystem very well). I think we can have a build cache for node_modules which will solve the problem more cleanly. I'll give up if it takes too long.

yakloinsteak: I'm glad you brought this up. Definitely agree it wasn't maintainable, and what I just pushed I think is a good compromise.

gigxz: Thanks, I think this is much cleaner! Yeah the yarn install will still run, which is bothersome, but it should be relatively quick if all the node_modules are already present.
Thanks for iterating!

gigxz (approved):

yakloinsteak:
@gigxz ping me when this is deployed to QA, and I can look around, but anybody could look at the console and see if I missed something.

PR #390, opened 2023-07-28
Use a custom form for outgoing referrals

gigxz:
Hack together a "custom" form in order to support different pick lists for different elements. There is surely a better way to do this, but this is a quick fix for a custom part of the app. 

Originally I had said this wasn't necessary @ttoomey but we got feedback that it was, and that this workflow will be used more than previously thought.

https://www.pivotaltracker.com/story/show/185706425

requires https://github.com/greenriver/hmis-warehouse/pull/3326 for the note

Screen Shot 2023-07-28 at 4 02 52 PM


src/modules/projects/components/ProjectOutgoingReferralForm.tsx
```diff
@@ -37,43 +48,117 @@ const ProjectOutgoingReferralForm: React.FC<Props> = ({
...
+          itemChanged={({ value }) =>
+            setFormState((old) => ({
+              ...old,
+              selectedProject: value,
+              selectedUnitType: undefined,
+            }))
+          }
+        />
+        {formState.selectedProject && (
+          <DynamicField
```
ttoomey, line 119: Fwiw, it can be nice to wrap conditional form elements in `<Fade>` or another effect to give the user a better indication that the form is changing dynamically. Maybe something for a future UX pass

ttoomey (approved):
Looks good!
Note: I do wonder about updating state on every form change (particularly typing). That would trigger a rerender of the whole form, which could result in poor client-side performance on larger forms. Maybe the rendering improvements in react 18 are helping us out here so it's not a concern.

gigxz:
Yeah React 18 does appear to help out a lot with that, I don't notice any lagginess in the text field on my browser. Since this is a small form I think we're fine.

gigxz:
will merge once https://github.com/greenriver/hmis-warehouse/pull/3326 is merged

PR #399, opened 2023-08-03
Update unit mgmt and pick list eviction

gigxz:


src/utils/cacheUtil.ts
```diff
@@ -1,19 +1,34 @@
...
 import { PickListType, Query } from '@/types/gqlTypes';
 
+/**
+ * Evict picklist of a particular type, with a PARTIAL match on args.
+ *
+ * For example:
+ * evictPickList(AVAILABLE_UNITS_FOR_ENROLLMENT, { projectId, householdId }) evicts picklists for the specific args
+ * evictPickList(AVAILABLE_UNITS_FOR_ENROLLMENT, { projectId }) evicts picklists of that type for that project, regardless of whether householdId was passed
+ * evictPickList(AVAILABLE_UNITS_FOR_ENROLLMENT) evicts ALL picklists of that type
+ */
```
gigxz, line 15: @ttoomey requesting review for this piece, mostly. does it seem reasonable? I haven't used cache.modify before, but this seemed like the only way to achieve the desired behavior.

gigxz:
maybe a better solution has to to with cache keys of PickListOption. The reason I'm needed to evict here is because the secondary label isnt being updated when assignment changes, which maybe is because graphql isnt recognizing it as a meaningful change? look more into that
Screen Shot 2023-08-03 at 10 30 52 AM

gigxz:
Merging this so I can merge and deploy https://github.com/greenriver/hmis-warehouse/pull/3348, since this also contains a fix to make that work

PR #409, opened 2023-08-15
185729503 project services tab

ZachCMP:
https://www.pivotaltracker.com/story/show/185729503

Relies on backend PR: https://github.com/greenriver/hmis-warehouse/pull/3381

### DIFF
--- a/src/api/operations/project.queries.graphql
+++ b/src/api/operations/project.queries.graphql
@@ -94,6 +94,35 @@ query GetProjectHouseholds(
   }
 }
 
+query GetProjectServices(
+  $id: ID!
+  $filters: ServicesForProjectFilterOptions
+  $sortOrder: ServiceSortOption
+  $limit: Int = 10
+  $offset: Int = 0
+) {
+  project(id: $id) {
+    id
+    services(
+      limit: $limit
+      offset: $offset
+      sortOrder: $sortOrder
+      filters: $filters
+    ) {
+      offset
+      limit
+      nodesCount
+      nodes {
+        ...ServiceFields
+        enrollment {
+          id
+          ...EnrollmentFields
+        }
+      }
+    }
+  }
+}
+
 mutation DeleteProject($input: DeleteProjectInput!) {
   deleteProject(input: $input) {
     clientMutationId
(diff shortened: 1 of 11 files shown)


src/modules/projects/components/ProjectServices.tsx
```diff
@@ -0,0 +1,50 @@
...
+                  projectId,
+                })}
+                Icon={LibraryAddIcon}
+              >
+                Record Services
+              </ButtonLink>
+            </Stack>
+          </ProjectPermissionsFilter>
+        }
+      />
```
gigxz: We're actually hiding this feature for now because it's not ready for MVP. I removed it with https://github.com/greenriver/hmis-frontend/pull/413 which I will merge soon
```suggestion
```

src/modules/enrollment/components/dashboardPages/EnrollmentServicesPage.tsx
```diff
@@ -17,7 +17,7 @@ import {
   ServiceFieldsFragment,
 } from '@/types/gqlTypes';
 
-const baseColumns: ColumnDef<ServiceFieldsFragment>[] = [
+export const baseColumns: ColumnDef<ServiceFieldsFragment>[] = [
```
gigxz: can we give this a more descriptive name since its being exported, like SERVICE_COLUMNS

gigxz: Unrelated, but would you mind also combining the category and service type cols? It's too many columns, and sometimes the category and service type are the same. How about combining them like so:

```
  {
    header: 'Service Type',
    render: (service) => {
      const { name, category } = service.serviceType;
      if (name === category) return name;
      return `${category} - ${name}`;
    },
  },
```

src/api/operations/project.queries.graphql
```diff
@@ -94,6 +94,35 @@ query GetProjectHouseholds(
...
+      filters: $filters
+    ) {
+      offset
+      limit
+      nodesCount
+      nodes {
+        ...ServiceFields
+        enrollment {
+          id
+          ...EnrollmentFields
```
gigxz: This queries some related records that we don't need in this instance. Is TS unhappy if you pare it down to the bare bones fields needed?

```
id
entryDate
exitDate
client {
   ...ClientNameDobVet
}
```

ZachCMP:
@gigxz Addressed your feedback and pushed

gigxz (approved):
Thanks Zach!!

PR #414, opened 2023-08-17
Add descriptions to Task List

gigxz:


src/components/clientDashboard/enrollments/EnrollmentReminders.tsx
```diff
@@ -26,29 +34,89 @@ const describeReminder = (reminder: ReminderFieldsFragment): string => {
...
+        ? `${reminder.count} intake assessments have not been submitted.`
+        : notCurrentClient
+        ? `${clientName}'s intake assessment has not been submitted.`
+        : null;
+    case ReminderTopic.ExitIncomplete:
+      return multiple
+        ? `${reminder.count} exit assessments have been started.`
+        : notCurrentClient
+        ? `${clientName}'s exit assessment has been started.`
+        : null;
```
ttoomey, line 74: This is fine. I wonder about use of nested ternaries; It's pretty subjective but I guess I find them harder to read than if/else 
```js
if (multiple) {
  return `${reminder.count} exit assessments have been started.`
} else if (notCurrentClient) {
  return `${clientName}'s exit assessment has been started.`
}
```

ttoomey (approved):
Looks great!

src/components/clientDashboard/enrollments/EnrollmentReminders.tsx
gigxz: Ok good point, I will keep that in mind. Nesting if-elses inside a switch statement feels yucky to me, but I think you're right it's more readable generally.

PR #422, opened 2023-08-24
185872956 new dob ssn treatment

ZachCMP:
https://www.pivotaltracker.com/story/show/185872956


src/modules/client/providers/ClientSsnDobVisibility.tsx
ZachCMP, line 1: Not sure if this is the right place for this, placing it here was just a best guess

src/modules/projects/components/tables/ClientsTable.tsx
ZachCMP, line 1: This file wasn't used anywhere and was the only other place where `CLIENT_COLUMNS` was used. Since it isn't being used, I felt like it didn't make sense to adjust it to work with the new columns. Then considering that it would then be both outdated and unused at that point, I thought it made sense to just remove it. Let me know if you would prefer to keep it

src/utils/context.tsx
ZachCMP, line 1: This is just utilities to be able to create contexts that behave like `useState` without needing to write lots of boilerplate. This is my preferred way to set up contexts like this, as it allows the consumers to consume just the value, just the setter or both. This is helpful for mitigating the potential performance issues that come from using react contexts in my experience.

gigxz:
Looks great! Below is some feedback on the UI and requested changes. I'll take a look at the code once the ui changes are in -- sorry, a bit strapped for time these days 🏃‍♂️ .

1. Can we update the header button style to match the mock
Screen Shot 2023-08-25 at 3 47 33 PM

2. When a user does not have DOB access for any projects (ie they dont have it on the root perm object), the column header should just say "Age" (in the normal style, not bolded)
Screen Shot 2023-08-25 at 3 47 41 PM

3. If the user DOES have some DOB access, just show the "DOB" button. Remove the "(Age)" string

4. Can you make it so the SSN doesnt wrap? Looks like there is ample space for it to go: (EDIT: might be fixed by pulling in release-86, I realized the preferred name column was still present taking up space)

Screen Shot 2023-08-25 at 3 56 03 PM

5. Can you set a fixed width so that the columns don't shift around? Probable would address (4) too

Kapture 2023-08-25 at 16 00 10

6. If the user has no SSN access, we should hide this button. It looks like we already do that in the table-view, so hopefully its not too tricky.
Screen Shot 2023-08-25 at 4 03 28 PM

src/utils/context.tsx
gigxz: Ok, great. Thank you!

src/modules/projects/components/tables/ClientsTable.tsx
gigxz: totally can be removed, thanks

ZachCMP:
@gigxz All your feedback should be addressed now

gigxz (approved):
lgtm!

PR #426, opened 2023-08-25
Implement phone mask

ZachCMP:
https://www.pivotaltracker.com/story/show/185740385


src/components/elements/input/PhoneInput.tsx
```diff
@@ -0,0 +1,48 @@
...
+    const { onChange, ...other } = props;
+    return (
+      <IMaskInput
+        {...other}
+        mask={[
+          {
+            mask: '(000) 000-0000',
+          },
+          {
+            mask: '{+}0` (000) 000-0000',
```
gigxz: Can you update this to allow 2 digits for country code?

gigxz (changes requested):
Looks like some phone numbers aren't being displayed on the client profile. Can you update that to be more permissive? It can continue to apply formatting, but if the phone string doesn't match the expected legnth/chars, we should still display it as-is. We shouldn't be hiding any data.

Screen Shot 2023-08-28 at 8 19 39 AM
Screen Shot 2023-08-28 at 8 19 29 AM

gigxz:
Oh, I just realized that the numbers are actually being persisted in the db with the special characters, like `value: "+1 (802) 999-1111"`. I think we should only store the numeric values (and `+`), and just use the input mask for display purposes. That gives us more flexibility as to how we display them, and makes the data cleaner for export purposes if needed.

Passing the `unmask` prop to the IMaskInput component seems to achieve the desired behavior

ZachCMP:
@gigxz Addressed your feedback. I set it to allow for 2 country codes and also switched the logic to display phone numbers to use the same IMask formatter as the inputs rather than the regex mechanism that was being used before. That should make the input/display of phone numbers consistent. I also adjusted it so that it only persists the numbers. I had to add some extra logic to that to actually persist the `+` symbol, as it needs to be stripped if the number is 10 digits or less otherwise the base value will match the wrong mask pattern. Hopefully that works for you, otherwise I can try and see if I can tune it a bit more.

gigxz (approved):
This looks good, thanks! It still could be more forgiving to display malformed phone numbers, so I'm pushing a fix for that.

PR #433, opened 2023-08-29
185906123 disable autofill

ZachCMP:
https://www.pivotaltracker.com/story/show/185906123

@gigxz It looks like autocomplete was already `off` for the `TextInput` component. However, [Chrome ignores this and will autofill anyway](https://stackoverflow.com/questions/15738259/disabling-chrome-autofill). The proposed solution in that issue is to set autocomplete to a value that Chrome will not recognize, so that's what I did. I used `do-not-autocomplete`, which seemed to work for me. However, I can't guarantee that that would work for any and all password/address/etc. managers. Hopefully this is good enough for now.


gigxz:
@ZachCMP does this work for non-chrome browsers too?

gigxz:
last pass has a specific attribute, can you add that too https://stackoverflow.com/questions/20954944/stop-lastpass-filling-out-a-form and see if any other common password managers have specific attributes for disabling?

This needs to be best effort. We need to avoid users are accidentially saving client's sensitive information - names, addresses, phones, etc --  into their local password managers.

gigxz (changes requested):
more info on which browsers / pw managers this works for

gigxz (approved):

PR #437, opened 2023-08-29
Allow DNC to have meaningful value

ZachCMP:
https://www.pivotaltracker.com/story/show/185906061

@gigxz Looks like the deleted line was introduced in https://github.com/greenriver/hmis-frontend/pull/408. Would removing it like I've done here be problematic for that work?

Also there was actually no issue saving the value, it submits and saves fine, it just doesn't display in the form when you pull it up because the frontend is treating DNC as null in forms due to the line I removed here.


gigxz (commented):
yeah, I think this will have other impacts. esh. I will come back to this and look deeper.. I can take this ticket over

gigxz (approved):

PR #445, opened 2023-09-01
184294397 form progress indicators

ZachCMP:
https://www.pivotaltracker.com/story/show/184294397

@gigxz Putting this here in case you want to pick it up. Here's the two things it still needs that we discussed

- [x] If form is saved or submitted, we should initially show errors/warnings based on required or warn if empty.
- [x] Ensure income from any source error threads correctly. Test other errors to ensure they work


ZachCMP:
@gigxz I fixed the two issues above so this should be ready to review, but there is still a wrinkle that applies to both of those things. Since some of the errors populate only on form submission (income from any sources, for example), I cannot show them on form load if the assessment is already saved like I can do with the warn-if-empty form validations. For certain workflows this is probably okay, but the behavior feels a bit unexpected in certain cases. Most notably if I try to submit I get the errors fine and they show up as expected; But if I then navigate away and then reopen the assessment page the error is gone, which seems kind of unexpected from a user point of view. I don't really see an easy way around this except to do something to persist the errors in the frontend, which feels ugly, or adding a graphql route to validate the form and get the errors, which feels like a lot more scope. So I left the wrinkle in place for now rather than trying to solve it an spend lots more time on it.

gigxz (commented):
This looks like a good start. I played around with it and found several issues. Will discuss at planning when this can be picked up again, and by who

 - [ ] Says "2 errors" when there is only 1 error (entry date missing)
 - [ ] Error state doesn't happen when theres an error about the entry date being invalid (in the future / >20 yr ago /etc). This is after submitting.
 - [ ] Non-cash benefits doesn't appear "complete" even when it is. If the "other" field is disabled (Because Other was not selected) then its always showing a warning in the form stepper. Warnings need to be only shown for **enabled** items. 
 - [ ] Prior Living Situation is going into an error state even though no changes were made to it. probably an issue with erroring about fields that are not currently enabled
 - [ ] Ideally imo the warning states would be present as soon as you open a submitted assessment. Right now they only show up when you unlock to make changes.
 - [ ] Status appears "empty" for Income Benefit even when its filled in with "yes" and values
 - [ ] The "2 changes" text doesn't truly reflect the changes, it only reflects the actions. For example if I add 2 fields and then clear them both, thats still "2 changes" even though nothing changed. I find that distracting/confusing. I think we should just remove the "N changes" text. Its still not ideal that it appears changed even if you undo your changes, but it's better than having a change count thats not right.

PR #453, opened 2023-09-08
185997232-local-store-session 

ttoomey:
https://www.pivotaltracker.com/story/show/185997232
* use session storage instead of local storage. That way our stored data isn't out of sync with session


gigxz (approved):

gigxz:
good find!

PR #465, opened 2023-09-14
Longer caching for fingerprinted assets

gigxz:
Add a new location block for assets, so that they can get a different `Cache-Control` header. Tested on QA. Required some refactoring to maintain all the existing headers.


schwartzkylej (approved):
LGTM

PR #471, opened 2023-09-18
185987568-fix-household-asm-autofill

ttoomey:
## Description

https://www.pivotaltracker.com/story/show/185987568

Fix bug where the autofill assessment options were shown for incorrect household member

Replaced useClientDashboardContext with clientId param. I tucked a id param into the `ClientNameDobVeteranFields` type which seemed convenient. 

## Testing
Find or create a household enrollment where houshold members have previous assessment data. Start a new intake/exit assessment. 
- On the HOH tab, click 'autofill assessment' in sidebar, observe options
- On the non-HOH tab, click 'autofill assessment' in sidebar, observe options (was the same as HOH before)
- the options for each member should be distinct and for the correct client

## Type of change
- [x] Bug fix
- [ ] Dependency update

- [x] My code follows the style guidelines of this project

### DIFF
--- a/src/modules/form/components/DynamicForm.tsx
+++ b/src/modules/form/components/DynamicForm.tsx
@@ -45,6 +45,7 @@ export interface DynamicFormProps
     FormActionProps,
     'disabled' | 'loading' | 'onSubmit' | 'onSaveDraft'
   > {
+  clientId?: string;
   definition: FormDefinitionJson;
   onSubmit: DynamicFormOnSubmit;
   onSaveDraft?: (values: FormValues, onSuccess?: VoidFunction) => void;
@@ -78,6 +79,7 @@ export interface DynamicFormRef {
 const DynamicForm = forwardRef(
   (
     {
+      clientId,
       definition,
       onSubmit,
       onSaveDraft,
@@ -236,6 +238,7 @@ const DynamicForm = forwardRef(
             {renderFields({
               itemChanged: handleChangeCallback,
               severalItemsChanged: handleChangeCallback,
+              clientId,
               errors: errorState.errors,
               warnings: errorState.warnings,
               horizontal,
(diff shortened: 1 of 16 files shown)


gigxz (approved):
lgtm, thanks! good to merge once build type failures are fixed

ttoomey:
> lgtm, thanks! good to merge once build type failures are fixed

Oops! Somehow I missed those type errors 🤦 

I drilled the ClientID down into the picker which was a bit more involved. Hope that's okay.

src/modules/services/hooks/useServiceDialog.tsx
```diff
@@ -190,6 +190,7 @@ export function useServiceDialog({
           )}
           {selectedService && formDefinition && serviceType && (
             <DynamicForm
+              clientId={enrollment?.client.id}
```
gigxz: The service form doesn't need this, right? Since it's only used for the "autofill" pickers? Just wondering.

ttoomey: I guess not, probably not needed on the referral either. I'll pull it out of those components. Thanks!

src/modules/form/components/DynamicForm.tsx
```diff
@@ -45,6 +45,7 @@ export interface DynamicFormProps
     FormActionProps,
     'disabled' | 'loading' | 'onSubmit' | 'onSaveDraft'
   > {
+  clientId?: string;
```
gigxz, line 48: This is okay, but maybe we can add a comment saying that it only gets used if this form is an Assessment, and what it's used for. It's a nonsense prop if the form is being used to edit a Project record, for example.

Another option would be to accept an object like RecordPickerDialogProps, so its flexible. I usually tend toward that approach, but I think it can be confusing in other ways, so I don't feel like either way is better.

ttoomey: Good point, I'll add a comment. I agree that this isn't ideal. Perhaps in the future it could be refactored to provide a more friendly interface that allows for the picker to be composed into just the forms that need it.

gigxz (approved):

PR #488, opened 2023-09-26
185905983-better-apollo-error-message

ttoomey:
## Description
https://www.pivotaltracker.com/story/show/185905983

* add reload button to snack bar
* show improved message for network vs application / server errors
* hide omni search popover when error occurs so it doesn't obscure the snack bar message

## Type of change
- [X] Bugfix

- [x] My code follows the style guidelines of this project

### DIFF
--- a/src/modules/errors/components/ApolloErrorAlert.tsx
+++ b/src/modules/errors/components/ApolloErrorAlert.tsx
@@ -8,40 +8,61 @@ import {
   Snackbar,
 } from '@mui/material';
 
-import { GraphQLError } from 'graphql';
 import { forwardRef, useEffect, useMemo, useState } from 'react';
 
 import { isServerError, UNKNOWN_ERROR_HEADING } from '../util';
 import ApolloErrorTrace from './ApolloErrorTrace';
 
 interface Props2 {
-  graphqlErrors: GraphQLError[];
+  errors: Error[];
+  isNetworkError: boolean;
   alertProps?: AlertProps;
   retry?: VoidFunction;
 }
 
 const BaseAlert = forwardRef<Props2, any>(
-  ({ alertProps, retry, graphqlErrors }, ref) => {
+  ({ alertProps, retry, errors, isNetworkError }, ref) => {
+    let errorMessage = '';
+    if (isNetworkError) {
+      errorMessage = `There was a problem connecting to the network. Please reload the page and try again.`;
+    } else {
+      errorMessage = errors[0]?.message;
+      // replace terse unknown error message with something more helpful
+      if (!errorMessage || errorMessage == UNKNOWN_ERROR_HEADING) {
+        errorMessage = `An error occurred on this page. The error has been reported and will be investigated by our support team. Please reload the page and try again. Contact support if the problem persists.`;
+      }
+    }
+
     return (
       <Alert severity='error' {...alertProps} ref={ref}>
-        <AlertTitle sx={{ mb: 0 }}>
-          {graphqlErrors[0]?.message || UNKNOWN_ERROR_HEADING}
-        </AlertTitle>
-        <ApolloErrorTrace graphqlErrors={graphqlErrors} />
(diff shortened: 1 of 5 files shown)


src/modules/errors/components/ApolloErrorAlert.tsx
```diff
@@ -8,40 +8,61 @@ import {
...
 const BaseAlert = forwardRef<Props2, any>(
-  ({ alertProps, retry, graphqlErrors }, ref) => {
+  ({ alertProps, retry, errors, isNetworkError }, ref) => {
+    let errorMessage = '';
+    if (isNetworkError) {
+      errorMessage = `There was a problem connecting to the network. Please reload the page and try again.`;
+    } else {
+      errorMessage = errors[0]?.message;
+      // replace terse unknown error message with something more helpful
+      if (!errorMessage || errorMessage == UNKNOWN_ERROR_HEADING) {
```
gigxz: This is dependent on the fact that `UNKNOWN_ERROR_HEADING` == [INTERNAL_ERROR_DISPLAY_MESSAGE on the backend](https://github.com/greenriver/hmis-warehouse/blob/58843a72e2db8b1e864ba253d9b0c9bb8d692d32/drivers/hmis/lib/hmis_errors/api_error.rb#L10)?

I was thinking that the backend itself would return the user-facing error message, which is what happens currently. Feels a bit shaky to rely on the strings being equal.

gigxz:
@ttoomey this is a bit of scope creep, but a couple suggestions:
* can we update the [error fallback](https://github.com/greenriver/hmis-frontend/blob/ffa8864467597bb4cc154e58310555433236d6d6/src/modules/errors/components/ErrorFallback.tsx#L11) to show the "contact your administrator" message as well? at this point I don't know if we can access the JSON error from the backend anymore (if its a backend error), but maybe it comes through the args.
* ~~can we update [generic table with data](https://github.com/greenriver/hmis-frontend/blob/ffa8864467597bb4cc154e58310555433236d6d6/src/modules/dataFetching/components/GenericTableWithData.tsx#L203) so it also shows an error toast if the query failed? currently it's just throwing the error which is then caught by an error boundary. there are several other places where we throw as well, but the generic table is just SUCH a widely used component, I think its worth having it's own error display (As opposed to falling back to the full page msg).~~
  * EDIT: hmm, nevermind, I think I may have been confused. It already gets wrapped in its own error boundary, so it shouldn't be doing the full-page thing. let me see why I was seeing that.

LMK if you'd like to discuss!

src/modules/errors/components/ApolloErrorAlert.tsx
```diff
@@ -81,20 +102,35 @@ const ApolloErrorAlert: React.FC<Props> = ({
...
+    ? error.graphQLErrors?.length == 0 && !isServerError(error.networkError)
+    : false;
   if (!error) return;
 
   if (inline) {
-    return <BaseAlert {...props} graphqlErrors={graphqlErrors} />;
+    return (
+      <BaseAlert
+        {...props}
+        graphqlErrors={errors || []}
```
gigxz:
```suggestion
        errors={errors || []}
```

src/modules/errors/components/ApolloErrorAlert.tsx
ttoomey: definitely, I'll pull this into the backend for now. 

Capturing our slack convo: in the future we might consider giving the error object a type and allow the FE to handle the display

ttoomey:
> @ttoomey this is a bit of scope creep, but a couple suggestions:
> 
> * can we update the error fallback

Good catch, I updated the language on the error fallback. Also adjusted treatment a bit and added a reload button.

> * ~can we update so it also shows an error toast if the query failed? currently it's just throwing the error

Fixed this error. Thanks for finding it!

gigxz (approved):
Looks good, thanks!!

PR #489, opened 2023-09-26
pr template

ttoomey:
copied from the warehouse with a few small adjustments


gigxz (approved):
thanks! added a spot to link dependent backend PR, if any

PR #497, opened 2023-09-28
Resolve more enrollment fields to use for population

gigxz:
Fixes a bug where when you create a new assessment, fields on the Enrollment didnt populate correctly. E.g. if you create a new intake on an enrollment that already has a LivingSituation (for whatever reason), it needs to fill.


gigxz:
along with https://github.com/greenriver/hmis-warehouse/pull/3579, fixes the disabling condition bug https://www.pivotaltracker.com/story/show/186143905

PR #510, opened 2023-10-05
Collect address "use", update contact info display to be clearer

gigxz:
Screenshot 2023-10-05 at 9 34 36 AM


gigxz:
Relates to https://github.com/greenriver/hmis-warehouse/pull/3614 but does not need to be merged together

PR #541, opened 2023-10-31
186145876 redirect after login

ttoomey:
## Description
[Fix redirect after okta login](https://www.pivotaltracker.com/story/show/186145876)

clicking a link, such as https://example.com/client/xyz, a user is prompted to login to okta. Then the user is redirected to the original requested location.

I added an allow-list for path patterns we want this behavior for. Not sure if there are urls we should redirect to but seemed like having some filtering would be a good idea.

## Type of change
- [x] Bug fix

## Testing:
* When signed out, visit a restricted url that requires authenication via okta. After authentication you should be redirected to requested path.
* While viewing a restricted url, sign out, then sign in. You should be redirected to the root (/) path.


src/modules/hmisAppSettings/Provider.tsx
```diff
@@ -56,7 +56,7 @@ export const HmisAppSettingsProvider: React.FC<Props> = ({ children }) => {
     setLoading(true);
     return logout()
       .then(() => {
-        reloadWindow();
+        window.location.assign('/');
```
ttoomey, line 59: Navigate to root to avoid saving the current url for redirect after login, assuming that's not desired.

src/routes/AppRoutes.tsx
```diff
@@ -15,24 +16,40 @@ import useAuth from '@/modules/auth/hooks/useAuth';
...
+// allow-list for urls that can be redirected to after sign-in
+function pathForRedirect(arg: string): string | undefined {
+  const path = arg.trim();
+  const idPattern = '[a-zA-Z0-9=]+';
+  const patterns = [
+    `^/client/${idPattern}/.*`,
+    `^/projects/${idPattern}/.*`,
+    `^/projects`,
+    `^/admin/.*`,
+  ];
```
gigxz: why not support redirecting to any url? this seems ok, just wondering.

gigxz: Oh and ideally we would support redirecting to a search route, e.g. `/?textSearch=foo`. I don't think that's secure to store in session storage though.

`organizations/` should be permitted as well

src/modules/hmisAppSettings/Provider.tsx
gigxz: yeah good call, thanks for fixing that

gigxz (approved):
Tested "clicking a link" to log in;  logging out and logging back in; and letting my session expire.

Overall looks great!

Only thing I wonder about is the inability to navigate to search links. I think there's a good argument for not supporting it, given that those urls are not obfuscated. Maybe we should have a convo about that one.

src/routes/AppRoutes.tsx
ttoomey: Oh yeah, seems like we shouldn't expose search terms in the url but that's probably a different issue.

There may be certain paths we might want to prevent redirecting into after login, for example a page nested in a deep workflow or a page that might trigger a side-effect action (maybe something we build in the future).

But maybe it's not worth the bother here. To be less trouble, I can invert this into a deny list and by default we allow redirects to any path after login.

gigxz: Ah okay, I see. We don't have pages like that right now, as far as I know, but we may in the future. I'm happy to leave it either way

PR #545, opened 2023-11-03
Add path details to headers on gql requests

gigxz:
## Description

Add path details to headers on gql requests

These headers are meant to support ~~https://github.com/greenriver/hmis-warehouse/pull/3700~~ https://github.com/greenriver/hmis-warehouse/pull/3720

To test with encoding, set `PUBLIC_PROTECTED_IDS='true'` in your `.env.development`

## Type of change


src/providers/apolloClient.tsx
```diff
@@ -52,6 +56,37 @@ const authLink = setContext(
...
+
+    // Add header for applicable params, if present
+    const { clientId, enrollmentId, projectId } = decodedParams;
+    if (clientId) modifiedHeaders['X-Hmis-Client-Id'] = clientId;
+    if (enrollmentId) modifiedHeaders['X-Hmis-Enrollment-Id'] = enrollmentId;
+    if (projectId) modifiedHeaders['X-Hmis-Project-Id'] = projectId;
+    return { headers: modifiedHeaders };
+  }
+);
+
```
gigxz: This is the important part

src/routes/routes.ts
```diff
@@ -1,3 +1,5 @@
+import { flatMap } from 'lodash-es';
```
ttoomey: I don't think we need lodash for this anymore
https://caniuse.com/?search=flatMap

src/utils/pathEncoding.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { NotFoundError } from '@/modules/errors/util';
+import IdEncoder from '@/modules/hmis/IdEncoder';
+import { HmisEnums } from '@/types/gqlEnums';
+
+export const isIdParam = (key: string) => key.match(/Id$/);
+
+const enumParams: Record<string, string[]> = {
+  formRole: Object.keys(HmisEnums.FormRole),
+};
+
```
ttoomey, line 12: maybe some docs here would be helpful
```suggestion
/**
 * Decodes encrypted ID parameters
 * Converts enum parameter values to uppercase and validates them
 * Throws a NotFoundError for unrecognized enum parameter values
 */
```

src/providers/apolloClient.tsx
```diff
@@ -52,6 +56,37 @@ const authLink = setContext(
...
+  (
+    _,
+    { headers }: { headers: { [key: string]: string } }
+  ): { headers: Record<string, string> } => {
+    const modifiedHeaders: typeof headers = {
+      ...headers,
+      'X-Hmis-Path': window.location.pathname,
+    };
+
+    // Find whith Route matches the current location
```
ttoomey:
```suggestion
    // Find when Route matches the current location
```

ttoomey (approved):
Looks great!

src/routes/routes.ts
gigxz: thanks! fixed

PR #550, opened 2023-11-14
Add path handler to catch spaced urls

ZachCMP:
## Description

Depends on hmis-warehouse PR:

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/components/elements/PathHandler.tsx
+++ b/src/components/elements/PathHandler.tsx
@@ -18,7 +18,7 @@ const PathHandler = ({ children }: { children: ReactNode }) => {
     if (match) {
       const [, corrected] = match;
       attempts.current += 1;
-      navigate({ pathname: corrected, hash, search });
+      navigate({ pathname: corrected, hash, search }, { replace: true });
     }
   }, [location, navigate]);
 
--- a/src/components/elements/PathHandler.tsx
+++ b/src/components/elements/PathHandler.tsx
@@ -6,21 +6,24 @@ const PathHandler = ({ children }: { children: ReactNode }) => {
   const navigate = useNavigate();
   const attempts = useRef(0);
 
-  useEffect(() => {
-    const { pathname, hash, search } = location;
+  const { pathname, hash, search } = location;
+  const match = pathname.match(/^(.*)(%20)+$/);
 
+  useEffect(() => {
     if (attempts.current >= 5) {
       console.error(`Unable to correct route '${pathname}'`);
       return;
     }
 
-    const match = pathname.match(/^(.*)(%20)+$/);
     if (match) {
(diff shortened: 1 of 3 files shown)


ZachCMP:
@gigxz I just added a wrapper component that looks for routes that end with `%20` and redirects them to that same route with the offending characters removed. It uses `replace` to prevent the back button from going back to the incorrect route, and has an attempts counter just to be safe. The attempts counter won't work well if the redirect causes the component to re-mount, but it does work correctly where I've put it, at least based on my testing. Let me know if you need anything else here

src/components/elements/PathHandler.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const location = useLocation();
+  const navigate = useNavigate();
+  const attempts = useRef(0);
+
+  const { pathname, hash, search } = location;
+  const match = pathname.match(/^(.*)(%20)+$/);
+
+  useEffect(() => {
+    if (attempts.current >= 5) {
+      console.error(`Unable to correct route '${pathname}'`);
```
gigxz: I'm not quite following the "attempts", how exactly does this code get hit? Is it trying to make it so the redirect only happens on the first 5 renders? (I couldn't repro this error log)

I think we only really need to do this redirect when someone is loading the page for the first time. So perhaps another approach is would be to only run the `useEffect` once, what do you think of that? seems easier to reason about to me:
```
  useEffect(() => {
    if (match) {
      const [, corrected] = match;
      attempts.current += 1;
      navigate({ pathname: corrected, hash, search }, { replace: true });
    }

    // ignore pathname dependency, we only want to run this once
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, []);

  return children;
```

src/routes/AppRoutes.tsx
```diff
@@ -59,7 +64,11 @@ const PublicRoutes: React.FC = () => {
 const blankRoutes: RouteObject[] = [
   {
     path: '*',
-    element: <></>,
+    element: (
+      <PathHandler>
+        <></>
+      </PathHandler>
```
gigxz: we can skip this, there is nothing being rendered anyway

src/components/elements/PathHandler.tsx
ZachCMP: Technically the attempts should never increment since the page should be redirected on the first try, I just put it there to prevent infinite navigation attempts if, for some reason, the page wasn't able to redirect to a different URL. But yeah, I agree that just running it once is probably cleaner and should solve the problem. I'll make that change

ZachCMP:
@gigxz Addressed your feedback

gigxz (approved):
👍

PR #554, opened 2023-11-19
186406279-hmis-system-e2e-tests-redux

ttoomey:
## Description
Always use the desktop date picker.

The mobile picker causes problems in [e2e testing with capybara](https://github.com/greenriver/hmis-warehouse/pull/3702). My understanding is we don't support mobile devices yet to it should be okay to always use the desktop picker

## Type of change
- [x] Bug fix


gigxz:
Hey @ttoomey sorry for the delay. I think it's nice to use the built-in mobile version for mobile use, even though we haven't focused much on mobile usability yet, we plan to in the future. If it's an issue with the render cycle of MUI choosing which component to use, is it equally solved by doing something like this?
```
const isMobile = useIsMobile();
const DateComponent = isMobile ? MobileDatePicker : DesktopDatePicker
```

It's probably not worth spending too much time on yet, though. I'm also OK with disabling the mobile picker for now, and coming back to the test issue when we are spending dedicated time on making workflows mobile-friendly.

gigxz (approved):

gigxz:
feel free to merge @ttoomey

ttoomey:
> Hey @ttoomey sorry for the delay. I think it's nice to use the built-in mobile version for mobile use, even though we haven't focused much on mobile usability yet, we plan to in the future. If it's an issue with the render cycle of MUI choosing which component to use, is it equally solved by doing something like this?

> ```js
> const isMobile = useIsMobile();
> const DateComponent = isMobile ? MobileDatePicker : DesktopDatePicker
> ```

Good idea, this seems to work. Thanks! Conceptually its similar to what DatePicker does (using media query to figure out which component to use) but this one seems to play well with our chrome container.

PR #558, opened 2023-11-26
186395195 stale object error when submitting household assessments

ttoomey:
## Description

[StaleObjectError when submitting household assessments
](https://www.pivotaltracker.com/story/show/186395195)

[Capybara Tests](https://github.com/greenriver/hmis-warehouse/pull/3767)

* prevent submission of assessment when there are in-flight mutations/dirty forms
* indicate a save is in progress on the form-action bar
* pauses navigation to next assessment tab while save completes
* add some missing accessibility attrs to loading indicators

## Testing
* create a multi-member household and start an intake 
* New function: clicking Save / Submit Assessment on the bottom bar should show a blue progress bar while the form is loading

* go to the last assessment in the household
* make a change the assessment and quickly hit 'next' to go to the "complete entry" tab. You may need to throttle your network connection in chrome
* New function: loading spinner is shown on the page if the previous save is still in progress. Once the saves are complete, the summary page is visible as usual

## Type of change
- [x] Bug fix

### DIFF
--- a/src/components/elements/Loading.tsx
+++ b/src/components/elements/Loading.tsx
@@ -8,6 +8,8 @@ const Loading = () => (
     height='100%'
     data-testid='loading'
     sx={{ p: 10 }}
+    aria-live='polite'
+    aria-busy='true'
   >
     <CircularProgress size={50} />
   </Box>
--- a/src/components/elements/LoadingSkeleton.tsx
+++ b/src/components/elements/LoadingSkeleton.tsx
@@ -1,6 +1,5 @@
 import { Box, Skeleton } from '@mui/material';
 import { BoxProps } from '@mui/system';
-import times from 'lodash-es/times';
 import React from 'react';
 
 interface Props extends BoxProps {
@@ -16,7 +15,7 @@ const LoadingSkeleton: React.FC<Props> = ({
 }) => {
   return (
     <Box aria-live='polite' aria-busy='true' {...props}>
-      {times(count, (i) => (
+      {Array.from({ length: count }, (x, i) => (
         <Skeleton
           animation={animation}
           key={i}
--- a/src/components/elements/LoadingSkeleton.tsx
+++ b/src/components/elements/LoadingSkeleton.tsx
@@ -1,0 +1,32 @@
+import { Box, Skeleton } from '@mui/material';
+import { BoxProps } from '@mui/system';
+import times from 'lodash-es/times';
+import React from 'react';
+
+interface Props extends BoxProps {
+  animation?: 'pulse' | 'wave' | false;
+  count?: number;
+  height?: number;
+}
+const LoadingSkeleton: React.FC<Props> = ({
+  animation = 'wave',
+  count = 6,
+  height = 50,
+  ...props
+}) => {
+  return (
+    <Box aria-live='polite' aria-busy='true' {...props}>
+      {times(count, (i) => (
+        <Skeleton
+          animation={animation}
+          key={i}
+          height={height}
+          variant='rounded'
+          sx={{ mb: 2 }}
+        />
+      ))}
+    </Box>
+  );
+};
+
+export default LoadingSkeleton;
--- a/src/components/elements/table/GenericTable.tsx
+++ b/src/components/elements/table/GenericTable.tsx
@@ -53,7 +53,7 @@ export interface Props<T> {
   headerCellSx?: (def: ColumnDef<T>) => SxProps<Theme>;
   selectable?: 'row' | 'checkbox'; // selectable by clicking row or by clicking checkbox
   isRowSelectable?: (row: T) => boolean;
-  setSelectedRowIds?: (ids: readonly string[]) => void;
+  onChangeSelectedRowIds?: (ids: readonly string[]) => void;
   EnhancedTableToolbarProps?: Omit<
     EnhancedTableToolbarProps<T>,
     'selectedIds' | 'rows'
@@ -110,7 +110,7 @@ const GenericTable = <T extends { id: string }>({
   headerCellSx,
   selectable,
   isRowSelectable,
-  setSelectedRowIds,
+  onChangeSelectedRowIds,
   EnhancedTableToolbarProps,
   filterToolbar,
   renderRow,
@@ -124,7 +124,8 @@ const GenericTable = <T extends { id: string }>({
   );
   const hasHeaders = columns.find((c) => !!c.header);
 
-  const [selected, setSelected] = useState<readonly string[]>([]);
+  // initially undefined so we can early return and avoid state flicker
+  const [selected, setSelected] = useState<string[]>();
 
   const selectableRowIds = useMemo(() => {
     if (!selectable) return [];
@@ -145,18 +146,22 @@ const GenericTable = <T extends { id: string }>({
 
   const handleSelectRow = useCallback(
     (row: T) =>
-      setSelected((old) =>
-        old.includes(row.id) ? without(old, row.id) : [...old, row.id]
-      ),
+      setSelected((old) => {
+        if (!old) return undefined;
+        return old.includes(row.id) ? without(old, row.id) : [...old, row.id];
+      }),
     []
   );
 
   // Clear selection when data changes
   useEffect(() => setSelected([]), [rows]);
-  useEffect(
-    () => setSelectedRowIds && setSelectedRowIds(selected),
-    [selected, setSelectedRowIds]
-  );
+
+  useEffect(() => {
+    if (selected) onChangeSelectedRowIds?.(selected);
+  }, [selected, onChangeSelectedRowIds]);
+
+  // avoid state flicker due to state reset
+  if (!selected) return <Loading />;
 
   if (loading && loadingVariant === 'circular') return <Loading />;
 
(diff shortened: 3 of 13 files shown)


src/components/elements/table/GenericTable.tsx
```diff
@@ -53,7 +53,7 @@ export interface Props<T> {
   headerCellSx?: (def: ColumnDef<T>) => SxProps<Theme>;
   selectable?: 'row' | 'checkbox'; // selectable by clicking row or by clicking checkbox
   isRowSelectable?: (row: T) => boolean;
-  setSelectedRowIds?: (ids: readonly string[]) => void;
+  onChangeSelectedRowIds?: (ids: readonly string[]) => void;
```
ttoomey, line 56: I renamed this to be more idiomatic "onFoo" to distinguish it from local setters

src/components/elements/Loading.tsx
```diff
@@ -8,6 +8,8 @@ const Loading = () => (
     height='100%'
     data-testid='loading'
     sx={{ p: 10 }}
+    aria-live='polite'
+    aria-busy='true'
```
gigxz, line 12: Interesting. I don't totally understand what these do but I created a ticket https://www.pivotaltracker.com/story/show/186561233 for assessing and improving accessibility of loading states

src/modules/assessments/components/household/HouseholdAssessmentTabPanel.tsx
```diff
@@ -27,6 +27,8 @@ interface HouseholdAssessmentTabPanelProps extends TabDefinition {
   navigateToTab: (t: string) => void;
   updateTabStatus: (status: AssessmentStatus, tabId: string) => void;
   assessmentStatus: AssessmentStatus;
+  onInflight: (clientId: string, inFlight: boolean) => void;
+  hasInflight: boolean;
```
gigxz: Consistency on camel-casing for inFlight would be nice. Eg `onInFlight`, `inFlight`, and `hasInFlight`

src/modules/assessments/components/household/HouseholdAssessments.tsx
```diff
@@ -61,11 +61,20 @@ const calculateAssessmentStatus = (
...
   role,
   enrollment,
   assessmentId,
-}: HouseholdAssessmentsProps) => {
+}) => {
+  // inFlights track if there are any outstanding mutations
+  const [inFlights, setInFlights] = useState<Record<string, boolean>>({});
+  const handleInFlight = useCallback((clientId: string, value: boolean) => {
+    setInFlights((c) => ({ ...c, [clientId]: value }));
+  }, []);
```
gigxz: It's possible a client could be enrolled twice in a household. For example if a household member leaves the household and comes back at a later date. In that case, that client would have 2 separate (non-overlapping) Enrollments in the household, each with it's own intake assessment. So we should use `enrollmentId` here, not client id.

src/modules/form/components/SaveSlide.tsx
```diff
@@ -1,28 +1,41 @@
...
+import { Box, LinearProgress, Paper, Slide, SlideProps } from '@mui/material';
 
-type Props = SlideProps;
+interface Props extends SlideProps {
+  loading?: boolean;
+}
 
-const SaveSlide = ({ children, ...props }: Props) => (
+const SaveSlide = ({ children, loading = false, ...props }: Props) => (
   <Slide {...props}>
```
gigxz, line 8: The slide should always be present on the household assessment view, it should be sticky to the bottom of the viewport. On your branch, I'm seeing it at the very bottom of the page instead. I'm not sure which change is causing it though.

src/modules/assessments/components/household/HouseholdSummaryTabPanel.tsx
```diff
@@ -153,6 +156,18 @@ const HouseholdSummaryTabPanel = memo(
...
     );
 
+    if (hasInFlight) {
+      return (
+        <AlwaysMountedTabPanel
+          active={active}
+          key='summary'
+          {...tabPanelA11yProps(id)}
+        >
+          <Loading />
```
gigxz: Can we instead do a Skeleton in place of either the `<TitleCard title='Select Household Members for Submission'>` or in place of the  `<HouseholdSummaryTable />`?

src/modules/assessments/components/AssessmentForm.tsx
```diff
@@ -124,6 +125,13 @@ const AssessmentForm = ({
       },
     });
 
+  const handleDirty = useCallback(
+    (value: boolean) => {
+      onInflight?.(clientId, value);
+    },
+    [onInflight, clientId]
+  );
+
```
gigxz: I'm not totally tracking how the "dirty"=>"inFlight" concepts relate to each other.

Within DynamicForm, the "dirty" state indicates that the user has made some changes to the form state, and has _not_ yet saved them.

"In Flight" to me would indicate that there is a request in flight (perhaps to submit the form). But this seems to indicate that dirty = in flight.

Am I missing something, or are they the same?

Aside: I know the way that household assessments are set up makes this tricky to implement. The fact that the mutations themselves (save/submit) are defined in the `AssessmentForm` component feels wrong at this point. It seems more logical to me that the `HouseholdAssessments` should manage those mutations itself, so it could more easily (1) know which requests are in flight, and (2) add onError/onSuccess callbacks to those mutations (which currently happens in a super roundabout way passing around callbacks to the FormActions component 🙈 ). Or we need some better pattern for doing those things.  Just a thought.

gigxz (commented):
Good progress, thanks for digging into this hairy code! In interacting with the application, the behavior looks correct. I'm a bit confused by the "in flight" concept but maybe it's just a naming thing. Other comments inline ✅

src/modules/assessments/components/AssessmentForm.tsx
ttoomey: Oh yeah, I started by naming it in-flight thinking it just needed to track requests that had not returned but then discovered what I really needed to track was if the form was dirty, otherwise you end up with weird UI flickering on the submit tab since it renders and then fires the mutation. I'll make a pass back through and rename inFlight to dirty, thanks for calling that out.

I agree, I think pulling state and actions up out of the form seems like a good direction when we get a chance to refactor.

src/modules/assessments/components/household/HouseholdAssessments.tsx
ttoomey: Ah I didn't think of that, I'll adjust it. Good catch

src/modules/form/components/SaveSlide.tsx
ttoomey: Thanks! I broke it with this change, should be fixed now

src/modules/assessments/components/AssessmentForm.tsx
```diff
@@ -124,6 +125,13 @@ const AssessmentForm = ({
       },
     });
 
+  const handleDirty = useCallback(
+    (value: boolean) => {
+      onDirty?.(enrollment.id, value);
+    },
+    [onDirty, enrollment.id]
+  );
+
```
gigxz, line 161: I needed the below in addition to handle editing assessments that have already been submitted. The summary panel was ending up in a permanently loading state.

Try a household where some intakes have already been submitted. There are 2 ways:
1. "Unlock" one assessment, make changes, and then change tabs.  It will submit in the background. Then it returns to locked state, rendering a DynamicView. (When that occurs, the dirty state never ends up changing because DynamicForm gets unmounted.)
2. "Unlock" one assessment, make changes, and click save & submit. It will submit and you need to approve warnings if any. Then it returns to locked state, rendering a DynamicView.  (Same issue as above).

```suggestion
  useEffect(() => {
    if (locked) onDirty?.(enrollment.id, false);
  }, [enrollment.id, locked, onDirty]);
```


I'm finding other bugs with the household assessment, it really does need to be revisited more holistically, as discussed. [Ticket](https://www.pivotaltracker.com/story/show/186567726) and [diagram](https://app.diagrams.net/#G1cqGUJjnGAWVmcNi8nl5IkisNLTuh4UWy) so we can prioritize that.

One of the issues I found was a StaleObjectError when you click save on an assessment and then click "next" arrow (or change tabs) before the save completes. The navigation action triggers another save on top of the one that's in flight. That might be another cause of StaleObjectErrors that we are seeing.

ttoomey: * I modified AssessmentForm so that the useAssessmentHandlers callbacks unset the saving state. I think that fixes the first issue. I should add a test case for this partially submitted assessment workflow
* That's a great find with Re the save/next sequence. I was able to reproduce that reliably and got a fix in place. The way I addressed it was to add a 'nextTab' state that pauses navigation until the save completes.
* The above requires tracking both in-flight requests and dirty state on the form. I tried to also add some state to track errors but I couldn't get it behave reliably and was concerned it was taking too long, particularly if we are going to refactor it.

gigxz: Awesome, I think this looks good. Thank you. I share the concern re: time - I think it's worth prioritizing a refactor here, since there are additional features/enhancements that are going to be very difficult (time consuming) to tack on to the current approach.

src/components/elements/LoadingSkeleton.tsx
```diff
@@ -0,0 +1,32 @@
...
+  return (
+    <Box aria-live='polite' aria-busy='true' {...props}>
+      {times(count, (i) => (
+        <Skeleton
+          animation={animation}
+          key={i}
+          height={height}
+          variant='rounded'
+          sx={{ mb: 2 }}
+        />
```
gigxz: Can we do:
```
Array.from({ length: count }, (x, i) => (
        <Skeleton
          animation={animation}
          key={i}
          height={height}
          variant='rounded'
          sx={{ mb: 2 }}
        />
));
```

The test is complaining about the lodash import, maybe its just the style. Trying `import { times } from 'lodash-es';` should also work, pretty sure it tree shakes to only whats imported.

gigxz (approved):
🥇

PR #561, opened 2023-11-29
User access history feature

gigxz:
## Description

1. Adds a very basic User Access History feature. As discussed re:[the ticket](https://www.pivotaltracker.com/n/projects/2591838/stories/186538152), it does not resolve any untrusted information from headers. Per the ticket, we'll need to add post-processing to make the interface more helpful. For now, it's just showing the raw records IDs (eg "Client 4", "Service 6", ...)
2. Adds common components for "contextual" list expansion ([figma](https://www.figma.com/file/F9EHR2IZYuMAlOJ9aOG2IL/OP-%2F-HMIS-%2F-Phase-2-(2024)?type=design&node-id=73-21961&mode=dev))
3. Update the Client Audit History page to use the same collapsible style. Pull out the component for rendering `objectChanges` so it can be used for Enrollment Audit History too.

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/3786

Kapture 2023-11-30 at 10 19 57

## Type of change


src/modules/admin/components/users/UserActionsMenu.tsx
```diff
@@ -0,0 +1,46 @@
...
+  rootAccess: RootPermissionsFragment;
+}
+
+const UserActionsMenu: React.FC<Props> = ({
+  onClickImpersonate,
+  isCurrentUser,
+  userId,
+  rootAccess,
+}) => {
+  const items = [];
```
ttoomey: Small thing, maybe these menu items should be alphabetical? ("Audit User", "Impersonate") Could also be memoized as a micro performance optimization

src/modules/dataFetching/hooks/useUser.tsx
```diff
@@ -0,0 +1,23 @@
+import apolloClient from '@/providers/apolloClient';
+import { UserFieldsFragmentDoc, useGetUserQuery } from '@/types/gqlTypes';
+
+export function useUser(userId?: string) {
```
ttoomey, line 4: Nice! Is the user query particularly slow that we needed to optimize it here?

ttoomey (approved):
Code looks good!

I'm a little unsure about this collapsable list design. I think it is probably fine for an admin interface but I suggest we do some benchmarking before using it more widely to make sure it has acceptable render performance.

Specifically I'd want to check if the vertical expansion triggers expensive layout reflow events (should be visible as "[Violation] Forced reflow while executing JavaScript" in the console in chrome). I'd also want to check for renders and how frequently we are re-rendering every table cell.

If performance is an issue, maybe using a modal rather than expand/collapse could be an alternative approach.

src/modules/dataFetching/hooks/useUser.tsx
gigxz: Its not slow, but what happens without it is:
1. Load table of all users. At this point we have all the user's names.
2. Click on one row. The `UserAuditPage` renders. The page title has the users name in it, so it makes a fetch request for the name, and shows a loading spinner until it returns. (This is the step that's cut out by just checking the fragment).
3. `UserAuditPage` proceeds to load the `UserAccessHistoryTable`, which makes another request to fetch the access history for this user. User sees a loading spinner until it completes.

That flow just bothers me, because the user is faced with 2 loading states, and one of them is for data that they already have in the cache (user's name).

There are other ways it could be resolved, and maybe they are less complex.  We could block the whole page behind a loading spinner until everything is loaded, and batch the requests. Or we could show an intermediate page title like "User Access History" until the name is returned along with everything else.

I prefer the fragment approach just because the UI appears more responsive on navigation.

ttoomey: Thanks for the explanation! The double-loading screen is no-fun. Maybe there's other places we can use this approach to smooth out the loading experience

gigxz:
Good call @ttoomey, I will look into that. There are no forced reflow warnings, but clicking on the "Expand All"  button definitely re renders the whole table - I tried it with a CPU throttle and it was laggy. Lets discuss further, I think, we arrived at this design because of the potential need to view all details at once for scannability (as opposed to viewing one modal at a time).

PR #568, opened 2023-12-04
Aria labeling for grouped input components; household edit controls

gigxz:
## Description

Aria labeling from grouped input components.

The Disability Table in particular didn't have any of the correct labels. Follow https://www.w3.org/TR/WCAG20-TECHS/ARIA9.html to use aria-labelledby for each input in the table. The read-only version of the disability table still needs work.

## Type of change


src/modules/form/components/group/DisabilityTable.tsx
```diff
@@ -169,7 +174,7 @@ const DisabilityTable = ({
                     index & 1 ? undefined : theme.palette.grey[50],
                 }}
               >
-                <TableCell sx={{ width: '250px', py: 3 }}>
+                <TableCell sx={{ width: '250px', py: 3 }} id={disabilityTypeId}>
```
ttoomey: Does this create multiple cells with the same html id? Perhaps this should be `${disabilityTypeId}-${index}`?

ttoomey (approved):
Code looks great! 🎉 
There may be some duplicate ids in the disability table, see note. I'm not sure where that component lives in the interface so I didn't test it.

src/modules/form/components/group/DisabilityTable.tsx
gigxz: Oh gosh good find!! Thank you! It does need to be unique for the labeling to work properly. Just updated to add index

PR #569, opened 2023-12-05
Assessment refactor to pull up change handlers and assessment+definition fetching

gigxz:
#### Bugs fixed
* Fixes a bug where a user could get "stuck" if they attempted to re-submit an assessment and then canceled the warning dialog. Repro steps:
  * Set up household assessments where at least 1 assessment has been Submitted
  * Unlock a submitted assessment
  * Make a change. Make sure some fields are left empty so you get a warning.
  * Click "save & submit"
  * Click "cancel" on the warning dialog
  * Try clicking Next/Prev or Tabs (Bug: nothing happens)
* Fixes a bug where a user could get "stuck" if they saved an assessment with an error. With this change, they still do get stuck, but they get a tooltip explaining that they have to fix errors before proceeding. (needs improvement). Repro steps
  *  Set up a household where at least 1 assessment is In Progress
  *  Update the Entry Date to be in the future (need to use text input, not picker)
  *  Hit "save" OR "next" or "previous" or any tab
  * Error message shows up. 
  * Fix the error by moving the date in range
  * Bug: "next"/"prev" don't do anything. "Save" resolves the error

#### Fetching Assessments and Definitions
- When an Assessment form is rendered, we need to fetch the Assessment (if editing an existing) and the Definition to use for rendering. These fetches are encapsulated in the `useAssessment` hook.
- The `useAssessment` hook was previously used by the `IndividualAssessment` component.
- With this PR, `useAssessment` is instead called by `IndividualAssessmentPage` (for single assmts) and `HouseholdAssessmentTabPanel` (for household assmts).

#### Assessment Mutation Handlers
- The assessment submission "handlers" are responsible for taking the raw form state and transforming it into values for submission (which requires knowledge of the definition), and calling the appropriate mutation. This is all encapsulated in the `useAssessmentHandlers` hook.
- The `useAssessmentHandlers` hook was previously used by the `AssessmentForm` component.
- With this PR, `useAssessmentHandlers` is instead called by `IndividualAssessmentPage` (for single assmts) and `HouseholdAssessmentTabPanel` (for household assmts). This allows each of those components to easily wrap the handlers to specify logic that should occur before/during/after a save occurs, and see the server response directly.
- This should help us with threading errors through the household assessments tab

#### Extra callbacks
- Now that the `useAssessmentHandlers` are pulled up to higher level components, I was able to get rid of some of the weird callbacks-in-callbacks that were around. 

## Type of change
- [x] Bug fix

### DIFF
--- a/src/modules/assessments/components/household/HouseholdAssessmentTabPanel.tsx
+++ b/src/modules/assessments/components/household/HouseholdAssessmentTabPanel.tsx
@@ -1,9 +1,15 @@
 import ArrowBackIcon from '@mui/icons-material/ArrowBack';
 import ArrowForwardIcon from '@mui/icons-material/ArrowForward';
-import { memo, useCallback, useEffect, useRef } from 'react';
+import { memo, useCallback, useEffect, useMemo, useRef } from 'react';
 
+import { useAssessment } from '../../hooks/useAssessment';
+import {
+  AssessmentResponseStatus,
+  useAssessmentHandlers,
+} from '../../hooks/useAssessmentHandlers';
 import IndividualAssessment from '../IndividualAssessment';
 
+import MissingDefinitionAlert from '../MissingDefinitionAlert';
 import AlwaysMountedTabPanel from './AlwaysMountedTabPanel';
 import {
   AssessmentStatus,
@@ -12,6 +18,7 @@ import {
   tabPanelA11yProps,
 } from './util';
 
+import Loading from '@/components/elements/Loading';
 import {
   HouseholdAssessmentFormAction,
   HouseholdAssessmentFormState,
@@ -19,7 +26,7 @@ import {
 import { DynamicFormRef } from '@/modules/form/components/DynamicForm';
 import { FormActionProps } from '@/modules/form/components/FormActions';
 import { FormActionTypes } from '@/modules/form/types';
-import { AssessmentFieldsFragment, FormRole } from '@/types/gqlTypes';
+import { FormRole } from '@/types/gqlTypes';
 
 interface HouseholdAssessmentTabPanelProps extends TabDefinition {
   active: boolean;
@@ -44,7 +51,7 @@ const HouseholdAssessmentTabPanel = memo(
     active,
     navigatingAway,
     id,
-    clientName,
+    // clientName,
     enrollmentId,
     assessmentId,
     client,
@@ -60,7 +67,7 @@ const HouseholdAssessmentTabPanel = memo(
     onFormStateChange,
     formState,
   }: HouseholdAssessmentTabPanelProps) => {
-    // console.debug('Rendering assessment panel for', clientName);
+    // if (active) console.debug(clientName, formState);
 
     const formRef = useRef<DynamicFormRef>(null);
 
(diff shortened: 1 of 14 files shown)


src/modules/assessments/components/household/HouseholdAssessmentTabPanel.tsx
```diff
@@ -70,23 +77,14 @@ const HouseholdAssessmentTabPanel = memo(
...
-          onFormStateChange(enrollmentId, 'saveCompleted');
-          // TODO: Update tab status to 'error' if error?
-          // console.debug(`Saved ${clientName}!`);
-          if (!assessmentId) {
-            // This was a NEW assessment; we need to re-fetch to get it
-            updateTabStatus(AssessmentStatus.Started, id);
-            refetch();
-          }
-        });
+        formRef.current.SaveIfDirty();
```
gigxz, line 87: We don't need these weird callbacks anymore, because we specify them in the `onCompletedMutation` defined below. We have a standard callback for form submission, so it will happen the same way whether the form is submitted from an explicit "save" or whether it's from a `SaveIfDirty` background save. This still needs improvement, but hopefully this is moving in the right direction.

src/modules/assessments/components/household/HouseholdAssessmentTabPanel.tsx
```diff
@@ -98,95 +96,162 @@ const HouseholdAssessmentTabPanel = memo(
...
+    const {
+      definition,
+      assessment,
+      loading: definitionLoading,
+      assessmentTitle,
+      formRole,
+    } = useAssessment({
+      enrollmentId,
+      assessmentId,
+      formRoleParam: role as unknown as FormRole,
```
gigxz, line 111: This should be cleaned up to all use AssessmentRole

src/modules/assessments/hooks/useAssessmentHandlers.tsx
```diff
@@ -178,7 +207,7 @@ export function useAssessmentHandlers({
     errors,
   } as {
     submitHandler: DynamicFormOnSubmit;
-    saveDraftHandler: (values: FormValues) => void;
+    saveDraftHandler: (values: FormValues) => void; // wrong
```
gigxz: ^

gigxz:
This branch is currently deployed to QA

src/modules/assessments/components/household/HouseholdAssessmentTabPanel.tsx
```diff
@@ -98,95 +96,180 @@ const HouseholdAssessmentTabPanel = memo(
...
+        assessmentId,
+        assessmentLockVersion: assessment?.lockVersion,
+        onCompletedMutation,
+      });
+
+    // disabling nav buttons while loading, and if there are errors
+    const disableNavigation = useMemo(
+      () => mutationLoading || formState.errors,
+      [formState.errors, mutationLoading]
+    );
```
ttoomey: micro optimization, maybe skip the memoization for simple calculations that return a primitive (so same object identity)
```suggestion
    const disableNavigation = mutationLoading || formState.errors 
```

ttoomey (approved):
This looks great! Nice work!!

PR #581, opened 2023-12-11
Add auto-exit UI elements

ZachCMP:
## Description

Add a UI for editing Auto Exit Configs

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/3817
Depends on frontend PR: https://github.com/greenriver/hmis-frontend/pull/582

## Type of change


ZachCMP:
@gigxz Take a look at this and let me know what you think about it. Because the form is small, I decided it would be best in a dialog rather than its own page, and I decided to use react-router routing to control the dialog state. I don't think we use this patten elsewhere, so I'm happy if you'd rather I do it a different way. Also let me know if there are certain conventions that I should be using that I'm not currently using.

src/modules/admin/components/autoExit/AutoExitDialog.tsx
```diff
@@ -0,0 +1,83 @@
...
+                  props.onClose({}, 'escapeKeyDown');
+              }}
+            />
+          ) : (
+            <StaticForm<
+              CreateAutoExitConfigMutation,
+              CreateAutoExitConfigMutationVariables
+            >
+              role={StaticFormRole.AutoExitConfig}
+              initialValues={{ lengthOfAbsenceDays: 30 }}
```
gigxz: Lets leave this empty, the user should be required to input the length of days

src/modules/form/data/index.ts
```diff
@@ -1,4 +1,5 @@
 import adminReferralPosting from '@/modules/form/data/adminReferralPosting.json';
+import autoExitConfig from '@/modules/form/data/autoExitConfig.json';
```
gigxz: This is erroring because the file doesn't exist

src/modules/admin/components/autoExit/AutoExitTable.tsx
```diff
@@ -0,0 +1,126 @@
...
+const columns: ColumnDef<AutoExitConfigFieldsFragment>[] = [
+  {
+    header: 'Length of Absence',
+    render: ({ lengthOfAbsenceDays }) => <>{lengthOfAbsenceDays} Days</>,
+  },
+  {
+    header: 'Project Type',
+    render: ({ projectType }) => (
+      <HmisEnum enumMap={HmisEnums.ProjectType} value={projectType} />
+    ),
```
gigxz:
```suggestion
        render: ({ projectType }) =>
      projectType ? (
        <HmisEnum enumMap={HmisEnums.ProjectType} value={projectType} />
      ) : (
        <NotCollectedText>Any Project Type</NotCollectedText>
      ),
```

src/modules/admin/components/autoExit/AutoExitTable.tsx
```diff
@@ -0,0 +1,126 @@
...
+    header: 'Project',
+    render: ({ project }) =>
+      project ? (
+        <RouterLink
+          to={generateSafePath(Routes.PROJECT, { projectId: project.id })}
+        >
+          {project.projectName}
+        </RouterLink>
+      ) : (
+        <NotCollectedText variant='body2'>No Project</NotCollectedText>
```
gigxz: I think this wording is a bit more descriptive. The variant should be inherited so isn't needed explicitly
```suggestion
        <NotCollectedText>Any Project</NotCollectedText>
```

src/modules/admin/components/autoExit/AutoExitTable.tsx
```diff
@@ -0,0 +1,126 @@
...
+      organization ? (
+        <RouterLink
+          to={generateSafePath(Routes.ORGANIZATION, {
+            organizationId: organization.id,
+          })}
+        >
+          {organization.organizationName}
+        </RouterLink>
+      ) : (
+        <NotCollectedText variant='body2'>No Organization</NotCollectedText>
```
gigxz:
```suggestion
        <NotCollectedText>Any Organization</NotCollectedText>
```

gigxz (commented):
Looks good! Comments inline.
I think the modal is OK, I probably will introduce a modal for Static Forms that has the action buttons as part of the modal chrome, like we have elsewhere, but that should be easy to plug in when I do. (Adding it for the sake of editing form instances).

src/routes/routes.ts
```diff
@@ -22,6 +22,9 @@ const adminDashboardRoutes = {
   CONFIGURE_SERVICES: 'services',
   CONFIGURE_SERVICE_CATEGORY: 'services/:serviceCategoryId',
   CONFIGURE_FORM_RULES: 'form-rules',
+  CONFIGURE_AUTO_EXIT: 'auto-exit',
+  CONFIGURE_AUTO_EXIT_EDIT: 'auto-exit/:autoExitId/edit',
+  CONFIGURE_AUTO_EXIT_CREATE: 'auto-exit/new',
```
gigxz: These can be removed?

src/routes/protected.tsx
```diff
@@ -642,6 +643,15 @@ export const protectedRoutes: RouteNode[] = [
               </RootPermissionsFilter>
             ),
           },
+          {
+            path: `${AdminDashboardRoutes.CONFIGURE_AUTO_EXIT}/*`,
```
gigxz: ?
```suggestion
            path: AdminDashboardRoutes.CONFIGURE_AUTO_EXIT,
```

ZachCMP: @gigxz Addressed all your feedback. I removed the extra paths. I had been using those to control the modal state. So clicking on the new button would go to `auto-exit/new`, which would open the modal for a new config, and clicking a row would go to `auto-exit/:id/edit`, which would open the modal for an edit modal. This was determined by the `useMatchPath` react router hook to determine if those routes were active. To make that work, the base route has to be `auto-exit/*`, so it would render the page even if the route isn't an exact match. I gather from your PR review that that pattern seemed unintuitive, so I switched to just using normal react state to control it. Let me know if you want anything else for this one.

gigxz: Ohh okay I didn't realize that. Sorry, I thought they were leftover routes if you had first had the editing on a different page or something. But yeah, I think it's good to not introduce a new routing pattern for a behavior-pattern that we usually implement with state.

gigxz (approved):

PR #584, opened 2023-12-14
186612957 stale object error

ZachCMP:
## Description

Fixes the stale object error when switching relationship to HOH fields in the household table

## Type of change
- [x] Bug fix

### DIFF
--- a/src/api/operations/household.queries.graphql
+++ b/src/api/operations/household.queries.graphql
@@ -18,6 +18,10 @@ mutation UpdateRelationshipToHoH($input: UpdateRelationshipToHoHInput!) {
           client {
             id
           }
+          enrollment {
+            id
+            lockVersion
+          }
         }
       }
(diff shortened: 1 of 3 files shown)


ZachCMP:
@gigxz Looks like this was caused by two things:

1. The relationship to HOH dropdown was not disabled while an update is loading, allowing users to initiate multiple requests with the same `lockVersion`, which causes this.
2. The `enrollment.lockVersion` field was not being returned in the mutation body, preventing it from updating in the frontend correctly, meaning that all requests would be sent with the same lock version. This causes all requests after the first to fail.

I addressed this by disabling the input during mutation loading and fetching the `lockVersion` for all enrollments in the household as part of the mutation response. This seems to manage the `lockVersion` correctly and prevent these errors.

gigxz (approved):
👍

PR #593, opened 2023-12-21
186557580 json form editor

ZachCMP:
## Description

- Form definition editor
- Ability to create new form definitions
- Ability to update attributes of existing form definitions
- Ability to delete form definitions

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/3853

## Type of change


ZachCMP:
@gigxz Probably went a bit outside the scope of the issue here, but it felt worthwhile to add full CRUD operations for form definitions alongside the form definition JSON editor, as we would need those eventually and it aligned with this work.

src/routes/routes.ts
```diff
@@ -23,6 +23,7 @@ const adminDashboardRoutes = {
   CONFIGURE_SERVICE_CATEGORY: 'services/:serviceCategoryId',
   FORMS: 'forms',
   VIEW_FORM: 'forms/:formId',
+  EDIT_FORM: 'forms/edit/:formId',
```
gigxz: for consistency with other routes, it would be `forms/:formId/edit`

gigxz (approved):
looks good to me!!

src/api/operations/form.fragments.graphql
```diff
@@ -121,10 +121,12 @@ fragment FormDefinitionMetadata on FormDefinition {
   role
   title
   cacheKey
+  identifier
 }
 
 fragment FormDefinitionFields on FormDefinition {
   ...FormDefinitionMetadata
+  rawDefinition
```
gigxz: Oh whoops, we don't want this here. This will case the raw JSON definition to be resolved alongside the parsed definition every time a form is loaded, which is unnecessary. Can you set up a different fragment that's used for the admin functions? (Or infer the type based on the query responses)

ZachCMP:
@gigxz Addressed your feedback

PR #602, opened 2024-01-03
View all Case Notes on Client; compose form dialog hooks

gigxz:
## Description

1. Add the ability to view Case Notes on the client dashboard. The tab is currently disabled because we need an API change in order to tell if the Case Notes feature is enabled for any projects.
2. Compose the "view" and "edit" form dialog hooks into one hook that handles both plus the associated state

## Type of change


src/modules/form/hooks/useViewEditRecordDialogs.tsx
```diff
@@ -0,0 +1,159 @@
...
+  evictCache?: VoidFunction;
+  localConstants?: LocalConstants;
+  maxWidth?: Breakpoint | false;
+}
+
+/**
+ * Helper for dialogs for Viewing and Editing/Deleting records.
+ * Only works for records that are submitted using SubmitForm.
+ * Often used alongside a table component.
+ */
```
gigxz, line 27: Just roughly pulling out a bunch of repeated logic into this hook.

src/modules/hmis/hmisUtil.ts
```diff
@@ -369,6 +369,24 @@ export const assessmentDescription = (
 };
 
 export const eventReferralResult = (e: EventFieldsFragment) => {
+  if (e.probSolDivRrResult) {
+    if (e.probSolDivRrResult == NoYesMissing.Yes) {
```
ttoomey: suggest always using strict equality '===' (maybe should be a linter rule)

ttoomey (approved):
Looks good!

src/modules/hmis/hmisUtil.ts
gigxz: whoops good call, added eqeqeq linter

PR #605, opened 2024-01-10
Support custom client detail forms (for pathways)

gigxz:
## Description

Add a "Custom Fields" section to the client dashboard. It gets populated if there are any active forms with role `CLIENT_DETAIL`. This can be used to collect or display arbitrary data on the client dashboard.

The card is basically the same as the `EnrollmentDetails` card on the Enrollment Dashboard, so I did some refactoring to allow them to share code.

Editing these fields requires `can_edit_clients` perm for the client. Viewing these fields requires `can_view_clients` perm for the client.

Screenshot 2024-01-10 at 3 55 00 PM

### How to test

see instructions on backend pr

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/3898

## Type of change

### DIFF
--- a/src/components/clientDashboard/ProfileLayout.tsx
+++ b/src/components/clientDashboard/ProfileLayout.tsx
@@ -36,9 +36,7 @@ const ProfileLayout: React.FC<Props> = ({ client, notices = [] }) => {
         )}
         <Grid item md={12} lg={canViewEnrollments ? 6 : 8}>
           <Stack gap={2}>
-            <Box>
-              <ClientProfileCard client={client} />
-            </Box>
+            <ClientProfileCard client={client} />
             <ClientCustomDataElementsCard client={client} />
           </Stack>
         </Grid>
--- a/src/components/clientDashboard/ProfileLayout.tsx
+++ b/src/components/clientDashboard/ProfileLayout.tsx
@@ -1,6 +1,7 @@
-import { Alert, AlertProps, AlertTitle, Box, Grid } from '@mui/material';
+import { Alert, AlertProps, AlertTitle, Box, Grid, Stack } from '@mui/material';
 import { isEmpty } from 'lodash-es';
 
+import ClientCustomDataElementsCard from '@/modules/client/components/ClientCustomDataElementsCard';
 import ClientEnrollmentCard from '@/modules/client/components/ClientEnrollmentCard';
 import ClientProfileCard from '@/modules/client/components/ClientProfileCard';
 import { ClientFieldsFragment } from '@/types/gqlTypes';
@@ -34,7 +35,12 @@ const ProfileLayout: React.FC<Props> = ({ client, notices = [] }) => {
           </Grid>
         )}
         <Grid item md={12} lg={canViewEnrollments ? 6 : 8}>
-          <ClientProfileCard client={client} />
+          <Stack gap={2}>
+            <Box>
+              <ClientProfileCard client={client} />
+            </Box>
+            <ClientCustomDataElementsCard client={client} />
+          </Stack>
         </Grid>
         {canViewEnrollments && (
           <Grid item md={12} lg={6}>
--- a/src/components/elements/CommonDetailGrid.tsx
+++ b/src/components/elements/CommonDetailGrid.tsx
@@ -1,12 +1,12 @@
 import { Grid, Typography } from '@mui/material';
 
-import React, { ReactNode } from 'react';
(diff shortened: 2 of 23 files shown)


src/modules/form/util/formUtil.ts
```diff
@@ -1420,3 +1420,32 @@ export const itemDefaults = {
   hidden: false,
   repeats: false,
 };
+
+export const parseOccurrencePointFormDefinition = (
```
gigxz, line 1424: just moved from component file

src/modules/form/components/OccurrencePointForm.tsx
```diff
@@ -1,75 +1,61 @@
...
+/**
+ * This component renders the values from the `record` into the `definition` and displays
+ * it as a read-only view. If any of the `definition` items are not read-only, it shows
+ * a pencil icon button. When clicked, it brings up a modal that renders a form according to
+ * the `definition`, and submits changes using SubmitForm.
+ *
+ * This is currently primarily used for viewing/editing Custom Data Elements that are associated with
+ * a Client or an Enrollment, on the Client Dashboard and Enrollment Dashboard respectively.
+ */
+const OccurrencePointForm: React.FC<OccurrencePointFormProps> = ({
```
gigxz, line 50: This component probably has a poor name, but I couldn't come up with a good name for the component. I stuck with occurrence point form since thats currently _how_ we're using it.

gigxz: I realize that probably means this component is doing too much..

src/components/clientDashboard/ProfileLayout.tsx
```diff
@@ -34,7 +35,12 @@ const ProfileLayout: React.FC<Props> = ({ client, notices = [] }) => {
           </Grid>
         )}
         <Grid item md={12} lg={canViewEnrollments ? 6 : 8}>
-          <ClientProfileCard client={client} />
+          <Stack gap={2}>
+            <Box>
+              <ClientProfileCard client={client} />
+            </Box>
```
ttoomey: is Box needed?

src/modules/form/components/OccurrencePointForm.tsx
```diff
@@ -1,75 +1,61 @@
...
-  enrollment,
+/**
+ * This component renders the values from the `record` into the `definition` and displays
+ * it as a read-only view. If any of the `definition` items are not read-only, it shows
+ * a pencil icon button. When clicked, it brings up a modal that renders a form according to
+ * the `definition`, and submits changes using SubmitForm.
+ *
+ * This is currently primarily used for viewing/editing Custom Data Elements that are associated with
+ * a Client or an Enrollment, on the Client Dashboard and Enrollment Dashboard respectively.
+ */
```
ttoomey, line 49: Nice component documentation! 🙌

src/components/elements/CommonDetailGrid.tsx
```diff
@@ -0,0 +1,57 @@
...
+          border: 'unset',
+        },
+      }}
+    >
+      {children}
+    </Grid>
+  );
+};
+
+export const CommonDetailGridItem: React.FC<{
```
ttoomey, line 33: This could probably could use some adjustment for accessibility. Maybe labeled by? Or[ aria table rows? ](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/rowheader_role)

ttoomey (approved):
Looks good!

src/components/elements/CommonDetailGrid.tsx
gigxz: Good call, thanks!

src/components/clientDashboard/ProfileLayout.tsx
gigxz: Yeah because the ClientProfileCard was actually a fragment of 2 components 🤔  this was just a quick way to use the `Stack` so I didn't have to touch that other component too much. But it would probably be a better solution to wrap the `ClientProfileCard` in a div itself, I can make that change.

src/components/elements/CommonDetailGrid.tsx
gigxz: Went with labelled-by since the grid isn't always a table, with a small enough screen it floats so the values are below the labels.
I also added aria labels to the edit buttons which were missing them

PR #611, opened 2024-01-18
186854441-fix-session-expiry

ttoomey:
## Description

If requests to fetch user/settings fail due to authorization error, try to reload the page. Reload only happens once to protect from infinite reloads if something unexpected happens.

Also returns more appropriate HttpError for api failures 

[Server error after session expiration
](https://www.pivotaltracker.com/story/show/186854441)

](https://github.com/greenriver/hmis-warehouse/pull/3919)

## Type of change
- [x] Bug fix


gigxz (approved):
LGTM!
As mentioned on https://github.com/greenriver/hmis-warehouse/pull/3919 it would be preferable if users didn't need to go through a refresh when navigating back to an expired session (since we already had the UI with a modal "your session expired click to log in"). But, I think this code makes sense to include regardless.

PR #612, opened 2024-01-18
React-hook-forms refactor of DynamicForm

ZachCMP:
## Description

Refactors the DynamicForm component to use `react-hook-form` and does some splitting up of components to make composing things easier. Couple notes:

1. This is currently draft, there is some cleanup left to do before I'd merge this, but I wanted to put it up for QA/reactions before proceeding with that
2. I've replaced all DynamicForm component with the refactored one by just making `DynamicForm` export the `RefactorForm` instead of the normal dynamic form. If there are instances in the app where components are using sub-components of DynamicForm or some such, those might not be handled right.
3. Due to the way that `react-hook-form` handles field names, the format `4.12.4` is troublesome, since it interprets that as a path (i.e. `{ '4': { '12': { '4': value } } }`). I've decided to deal with that by converting the periods to commas ('safe' format), and storing the values that way, converting them back to periods ('clean' format) where needed. This isn't ideal, but I don't see a way around it unless we can configure RHF to use a different delimiter for pathing. Using periods is typically the convention for path flattening in every case I've seen.
4. My immediate goal with this work was primarily to port over the current functionality, so there's definitely more that could be done in terms of enabling composition and handling certain workflows, as well as more work that could be done on refactoring the form logic code generally beyond what I've done here.

## Type of change


src/modules/form/components/refactor/useFormDefinitionHandlers.tsx
```diff
@@ -0,0 +1,186 @@
...
+  definition: FormDefinitionJson;
+  initialValues?: T;
+  localConstants?: LocalConstants;
+  errors?: ValidationError[];
+}
+
+export const getSafeLinkId = (linkId: string) =>
+  String(linkId).replace(/\./g, ',');
+export const getCleanedLinkId = (linkId: string) =>
+  String(linkId).replace(/,/g, '.');
```
ttoomey: This seems like a reasonable work-around but it's too bad we need it. I wonder if we could change the link-ids in form definitions so that they no longer include dots. That would mean either we translate the fields in the form definition in graphql or perhaps a db migration.

Could we use some other replacement character? A Link id of `1,2,3` suggests that it's CSV formatted data rather than an id. Maybe use kebab-case `-` or double underscore `__` in place of dots?

src/modules/form/components/refactor/RefactorField.tsx
```diff
@@ -0,0 +1,482 @@
...
+      }}
+    />
+  );
+};
+
+const MAX_INPUT_AND_LABEL_WIDTH = 600; // allow label to extend past input before wrapping
+export const MAX_INPUT_WIDTH = 500;
+const FIXED_WIDTH_MEDIUM = 350;
+const FIXED_WIDTH_SMALL = 200;
+const FIXED_WIDTH_X_SMALL = 100;
```
ttoomey: Looks like this component has quite a few responsibilities
* renders the correct input component for the item 
* determines the width of the component
* fetches pick-list options 
* sets up event handlers 
* validation / warnings (warn if empty, etc)

I know we decided we didn't want to touch the input components themselves for this round of work, so perhaps this is all we can do with that constraint. I'm fine with that; maybe we can plan come back for another round of refactoring when we have more space.

For future consideration, I'd suggest creating wrapper components for each input type that can manage these responsibilities.

src/modules/form/components/refactor/RefactorField.tsx
```diff
@@ -0,0 +1,482 @@
...
+      }
+      console.warn(
+        'Unable to render component for object type. Link ID:',
+        item.linkId
+      );
+      return <></>;
+
+    default:
+      console.warn('Unrecognized item type:', item.type);
+      return <></>;
```
ttoomey: suggest throwing an exception in these cases

src/modules/form/components/refactor/RefactorFormLayout.tsx
```diff
@@ -0,0 +1,26 @@
...
+const RefactorFormLayout = ({
+  errors,
+  children,
+  saveButtons,
+}: RefactorFormLayoutProps) => {
+  return (
+    <>
+      <Grid container direction='column' spacing={2}>
+        {errors && <Grid item>{errors}</Grid>}
+        {children}
```
ttoomey: Small thing but I wonder if we can get away from the pattern of components that require children to be wrapped in `<Grid item>`. Maybe it should be this component's responsibility to wrap the children in Grid item?

```jsx
    <>
      <Grid container direction='column' spacing={2}>
        {errors && <Grid item>{errors}</Grid>}
        {React.Children.map(children, (child) => (<Grid item>{child}</Grid>))}
      </Grid>
      {saveButtons}
    </>
```

package.json
```diff
@@ -98,10 +98,12 @@
     "react-dom": "^18.2.0",
     "react-dropzone": "^14.2.3",
     "react-error-boundary": "^3.1.4",
+    "react-hook-form": "^7.49.2",
     "react-i18next": "^12.0.0",
     "react-imask": "^7.1.3",
     "react-pdf": "^6.2.2",
     "react-router-dom": "^6.8.0",
+    "recoil": "^0.7.7",
```
ttoomey: looks like recoil didn't end up being needed

ttoomey (commented):
Looking great! I left a few thoughts/ suggestions. 

There's some extra space under some input components now, probably dependent components getting rendered

Screenshot_2024-01-22_at_12_50_36_PM

src/modules/form/components/refactor/useFormDefinitionHandlers.tsx
ZachCMP: I think the dots correspond to HUD spec, so I'm unsure how easy/okay it would be to replace them with another character at the backend or API level. I agree with the sentiment that the comma isn't ideal, and I'm open to the option of another replacement character. Though I think using any character that could possibly be part of a logical ID isn't the best, since it's possible it would be used in link IDs innocently and then cause issues or need special handling. For that reason I don't think using `-` or `_` is a good idea. My logic for using a comma is: 1) It reads as related to a period, so it's maybe more clear what character it's escaping; and 2) I would never think to use a comma in an ID, so I can't see one being used in a link ID in a future form definition.

In terms of picking a replacement for a comma, I see a couple of options:

1. Use some other character that wouldn't typically be included in IDs, maybe: `;`, `:`, `|`, `^`, `~`, `%`, etc.
2. Use an escape sequence of some kind: `:dot:` or some such.
3. Hash or base64 the link IDs. This would do it, and would probably make this a non-issue for any other reserved character issues, but probably not ideal since the internal form state would be hard to read.

What do you think?

ttoomey: > For that reason I don't think using - or _ is a good idea

I imagine double or triple underscore would be safe (we'd leave single underscores alone and transform with `s/\./__/ and s/__/\./`).

Here's another idea:
These are HUD defined values so maybe we prefix them all with 'hud'? That would have the extra benefit of allowing us to target these properties with selectors (currently these ids that start with numeric chars are hard to use in js/css).

I'm thinking something like `hud-1-2-3` would be ideal, it looks nice and is not too hard to transform back to the original value with something like `/^hud-(\d+)-(\d+)-(\d+)/`. The 'hud' prefix should make it safe to use simple hyphens or underscores in place of dots.

ZachCMP: I changed it to `__`

src/modules/form/components/refactor/RefactorField.tsx
ZachCMP: Yeah, this is definitely one that I want to revisit in a future iteration. I consider this out of scope for now as I see this pass as primarily trying to rewire state and mechanics to use RHF. I've done some splitting out of components to that end, but I'd like to tackle these deeper input component pieces as their own piece of work.

src/modules/form/components/refactor/RefactorFormLayout.tsx
ZachCMP: Agreed, I think the parents should be responsible to wrapping the children as much as possible. However, in this case the "children" is actually just the single form layout component, so I don't think we want to do this here.

package.json
ZachCMP: Yup, removed

ZachCMP:
@ttoomey Addressed your feedback and pushed some updates. The gaps should mostly be gone now unless the group components are trying to wrap their children in some other component. This was another consequence of inverting the disabled/autofill control. It was rendering groups where all the children were disabled, so I added additional checks in the wrapper to hide the group if all of its children are also hidden.

src/modules/form/components/refactor/useFormDefinitionHandlers.tsx
gigxz: https://www.pivotaltracker.com/story/show/186922374

src/modules/form/data/service.json
```diff
@@ -0,0 +1,588 @@
+{
```
gigxz: Can you move it into the `test/__mocks__` dir since it is not used in the applciation

src/test/__mocks__/pickLists.ts
```diff
@@ -0,0 +1,379 @@
+import { PickListOption } from '@/types/gqlTypes';
+
+export const pickListData: Record<string, PickListOption[]> = {
```
gigxz: You should be able to pull these from the gqlEnums file

yarn.lock
```diff
@@ -10817,6 +10827,13 @@ recast@^0.23.1:
     source-map "~0.6.1"
     tslib "^2.0.1"
 
+recoil@^0.7.7:
```
gigxz: I dont think this should be here, canyou run yarn and recommit

gigxz:
Conditional logic weirdness repro:
1. Open an intake assessment in a permanent housing project
2. Set Prior Living values:
Screenshot 2024-01-26 at 10 19 08 AM
3. Submit, re-open the assessment for editing
4. Clear out the Length of Stay value. There is an empty gray bubble:

Screenshot 2024-01-26 at 10 18 03 AM



The previous implementation had logic for groups, where if all children in a group were hidden, then the group component would be hidden as well. I think we need to keep that logic. I think it's missing which is what is causing the above.

gigxz:
Disabling condition had some logic where the Overall Disabling Condition would be fixed to `Yes` and not editable if it was set from a system rule. I'd be happy to look at that later, I think it was internal to the disability table logic. So I wouldn't worry too much about it now, just want to flag it so it doesnt get forgotten.

src/modules/form/data/service.json
ZachCMP: Used for a story so I could test different form defs. I'll probably remove most of the ones I added since I don't need them beyond development

src/test/__mocks__/pickLists.ts
ZachCMP: I actually don't think that I can. I need the `groupLabel` and `groupCode`, which I don't think are included in that file. But I'll probably end up removing this since I don't think I'll end up leaving it in storybook beyond the dev on this feature.

ZachCMP:
Yeah, the group components need some additional work. They assume they'll be passed `values` but that's not really true any more, so I need to find a workaround. I could pull the RHF logic into them, which would probably be best performance, but I could also wrap them in a `Watcher` or something and just pass the values from there and leave them in the old pattern. I'll see what makes most sense when I get there.

ZachCMP:
@gigxz I pushed fixes for the conditional logic. Could you take a look when you get a sec and see if it fixes it for you? I expanded the amount of things that the dependent wrapper listens to and what values it considers.

gigxz:
Yep that fixed the issue on PLS, thanks!

ZachCMP:
@gigxz @ttoomey This should be pretty close to ready to merge, so I'm taking if off draft status. I replaced all of the existing DynamicForm components with the new set of components and tweaked the behavior of everything to work with the new paradigm. Broadly speaking, I've tried to keep the RHF-dependent stuff in the base components and leave `DynamicField` and `DynamicGroup` with the values and setter paradigm they had before so we can use them outside of a RHF context. The component doing most of the heavy lifting in that regard is `DynamicFormField`, which handles making the RHF mechanics work with the value/setter paradigm.

I also changed the exports of DynamicForm around so there's a version that takes `handlers`–the RHF-based mechanics returned from `useFormDefinitionHandlers`, and a version that provides its own handlers. I've made the one that provides its own handlers the default export since that's the current paradigm, but the other is available if you want to declare the handlers at a higher level and pass them into the form (i.e. for assessments and such). I've left the `useImperativeHandle` bits in, since I wasn't sure how those pieces were being used and I didn't want to break anything.

One thing to note is that this removes the `useDynamicFields` hook that implemented the old react state mechanics. `DynamicForm` was the primary consumer of that hook, but it was also used by `DynamicView` and `BulkAdd`. I reworked DynamicView to use the new mechanics instead, as well as the BulkAdd component, but it doesn't look like that latter is used anywhere at the moment. It's used by `BulkAddServices`, but that component doesn't seem to be imported by anything currently. Regardless, some extra QA around those components would be helpful to make sure they're working well too. A quick look in storybook seems to indicate that at least the DynamicView is working, but I don't know every place it's used in the app and I don't have a utility to see a viewable version of the forms in the JSON form editor.

Take a look and let me know if there's anything else you'd like to see on this.

src/modules/form/components/AutofillFormItemWrapper.tsx
```diff
@@ -0,0 +1,62 @@
...
+}
+
+const AutofillFormItemWrapper: React.FC<Props> = ({
+  handlers,
+  item,
+  children,
+  getDependentLinkIds,
+}) => {
+  const { autofillInvertedDependencyMap, getAutofillValueForField } = handlers;
+  const name = (
```
ttoomey: suggest rename to dependentLinkIds

src/modules/form/components/DependentFormItemWrapper.tsx
```diff
@@ -0,0 +1,75 @@
...
+    () =>
+      compact(
+        getAllChildLinkIds(item, { onlyQuestions: false }).map(
+          (e) => itemMap[e]
+        )
+      ),
+    [item, itemMap]
+  );
+
+  const name = useMemo(() => {
```
ttoomey: suggest rename to dependentLinkIds

src/modules/form/components/DynamicFormField.tsx
```diff
@@ -0,0 +1,276 @@
...
+      });
+    },
+    [handlers]
+  );
+  const severalItemsChanged: SeveralItemsChangedFn = useCallback(
+    ({ values, type }) => {
+      Object.entries(values).forEach(([linkId, value]) =>
+        handlers.methods.setValue(getSafeLinkId(linkId), value, {
+          shouldDirty: type === ChangeType.User,
+        })
```
ttoomey: Maybe dry this up and use itemChanged so we get parseInt/Float?
```suggestion
        itemChanged({linkId, value, type})
```

src/modules/form/components/DynamicFormField.tsx
```diff
@@ -0,0 +1,276 @@
...
+  bulk?: boolean;
+  visible?: boolean;
+  pickListArgs?: PickListArgs;
+  nestingLevel: number;
+  item: FormItem;
+  props?: OverrideableDynamicFieldProps;
+  renderFn?: (children: ReactNode) => ReactNode;
+}
+
+const DynamicFormField: React.FC<Props> = ({
```
ttoomey, line 40: My understanding is the stack here is something like this. Does that seem right?
```js
DynamicFormField(renderFn)(
  // renderItemWithWrappers
  DependentFormItemWrapper(
    AutofillFormItemWrapper(
      //itemGroup, table etc
      (item) => item.children.map (
        ValueWrapper(DynamicField)
      )
    )
  )
)
```

ttoomey (approved):
Looks good to me!! Great work 🙌

src/modules/form/components/DynamicFormField.tsx
ZachCMP: @ttoomey Yeah, that's right. These are the responsibilities:

- `DynamicFormField` handles wiring up the form item to the form state handlers. It will use `renderItemWithWrappers` to wrap the children. It will only apply the wrappers that are appropriate, so it won't wrap a component with an autofill wrapper if the form item has no autofill logic.
- `DependentFormItemWrapper` handles hiding/showing the component based on the `enableWhen` logic. It does that by watching anything that the component depends on and updating the rendering when they change.
- `AutofillFormItemWrapper` does the same thing for autofill logic, and it handles updating the form values with the autofilled values when they are autofilled. It also uses `dirty` to only autofill if the field is not dirty, and does not mark the field dirty when it autofills.
- `ValueWrapper` pulls `value` out of the form state, allowing us to keep the value/setter pattern for the actual `DynamicField` and `DynamicGroup` components. So this lets us keep that pattern but isolate the re-renders per-field.

ZachCMP:
@ttoomey Addressed your feedback and pushed

src/modules/form/components/DynamicFormField.tsx
ttoomey: Thanks for the write up! The description is helpful to get oriented in the form mechanics. Maybe the above could become inline comments/docs? (if it's not already there)

ZachCMP:
@gigxz Anything else you need from me on this?

src/modules/form/components/DynamicFormField.tsx
```diff
@@ -0,0 +1,272 @@
...
+          {...fieldProps}
+        />
+      ) : (
+        <ValueWrapper name={getSafeLinkId(item.linkId)} handlers={handlers}>
+          {(value) => (
+            <DynamicField
+              value={value}
+              key={item.linkId}
+              item={item}
+              itemChanged={itemChanged}
```
gigxz: Note to self:

ValueWrapper provides value for this form item from RHF state; `itemChanged` uses `setValue` internally to update the form state. This lets us use all the existing input components and `DynamicField` without modification.

(As opposed to adjusting DynamicField, or the input components themselves, to consume the RHF context)

src/modules/errors/components/ApolloErrorTrace.tsx
```diff
@@ -25,8 +25,9 @@ const ApolloErrorTrace: React.FC<Props> = ({ errors }) => {
     >
       {errors.map((e) => (
         <CommonUnstyledList key={JSON.stringify(e)}>
-          {(e?.backtrace || []).map((line) => (
-            <li>{line}</li>
+          {(e?.backtrace || []).map((line, idx) => (
```
ttoomey, line 28: fixes warning about keys (trying to reduce chatter in the console)

yarn.lock
```diff
@@ -8139,10 +8139,10 @@ react-dropzone@^14.2.3:
     file-selector "^0.6.0"
     prop-types "^15.8.1"
 
-react-hook-form@^7.52.1:
-  version "7.52.1"
-  resolved "https://registry.yarnpkg.com/react-hook-form/-/react-hook-form-7.52.1.tgz#ec2c96437b977f8b89ae2d541a70736c66284852"
-  integrity sha512-uNKIhaoICJ5KQALYZ4TOaOLElyM+xipord+Ha3crEFhTntdLvWZqVY49Wqd/0GiVCA/f9NjemLeiNPjG7Hpurg==
+react-hook-form@^7.54.2:
```
ttoomey, line 9709: this is optional, seemed like this release was just bug fixes and non-breaking changes so worth doing

src/modules/hmis/components/HmisEnum.tsx
```diff
@@ -29,7 +29,7 @@ interface Props extends TypographyProps {
 const HmisEnum = ({ value, enumMap, noData, ...props }: Props) => {
   const [label, color] = getLabelAndColor(enumMap, value);
   return (
-    <Typography variant='body2' color={color} {...props}>
+    <Typography variant='body2' color={color} component='span' {...props}>
```
ttoomey, line 32: fixes invalid dom nesting warning

ttoomey:
@gigxz I've got this merged and the tests are passing. I'm going to take a another pass at reviewing it

src/modules/external/mci/components/MciClearance.tsx
```diff
@@ -296,23 +296,39 @@ const MciClearanceWrapper = ({
...
-  // currentMciAttributes gets re-calculated any time one of the dependent values changes (because of enableWhen dependency)
-  const [clientId, externalIds, currentMciAttributes] = useMemo(() => {
-    if (!getValues || !definition) return [];
-    const values = getValues();
-    let byKey = createHudValuesForSubmit(values, definition);
+  const { getValuesForSubmit, methods } = props.handlers;
+  const { getValues } = methods;
+
+  // watch specific form fields.
+  // FIXME: I suspect that this should translate MCI_CLEARANCE_FIELDS into form fields instead of hard-coding them
```
ttoomey: This seems to work but I should probably clean this up a bit.

src/modules/form/components/client/addresses/SimpleAddressInput.tsx
```diff
@@ -27,7 +27,11 @@ const SimpleAddressInput: React.FC<Props> = ({
 
   useEffect(() => {
     if (values.length === 0) {
-      onChange([createInitialValue()]);
+      // setTimeout is needed because if the inputs were not previously rendered, they are not yet registered. So the
+      // timeout allows them to register before we change the value here
+      setTimeout(() => {
+        onChange([createInitialValue()]);
+      });
```
ttoomey: I think the other way to fix this would be to descend into all the inputs for hidden groups and register them when the form is displayed. Maybe that would be a more robust solution? I think otherwise this problem would occur anytime we show fields and write initial values in a useEffect

src/modules/form/hooks/useFormDefinitionHandlers.tsx
```diff
@@ -0,0 +1,213 @@
...
+  initialValues?: T;
+  localConstants?: LocalConstants;
+  errors?: ValidationError[];
+  viewOnly?: boolean;
+}
+
+export const getSafeLinkId = (linkId: string) =>
+  String(linkId).replace(/\./g, '__');
+export const getCleanedLinkId = (linkId: string) =>
+  String(linkId).replace(/__/g, '.');
```
ttoomey: I don't think these are needed anymore as we have cleaned up our link id format. I guess they aren't hurting anything but probably safe to remove(?)

src/modules/form/hooks/useFormDefinitionHandlers.tsx
```diff
@@ -0,0 +1,213 @@
...
+      const values = mapKeys(methods.getValues(), (v, k) =>
+        getCleanedLinkId(k)
+      );
+      if (!autofillInvertedDependencyMap[linkId]) return;
+      const shouldAutofill = autofillValues({
+        item,
+        values,
+        itemMap,
+        localConstants,
+        viewOnly, // TODO: Handle this
```
ttoomey: I'm not sure what needs to be done here

src/modules/errors/components/ApolloErrorTrace.tsx
gigxz: thanks!

src/modules/form/hooks/useFormDefinitionHandlers.tsx
gigxz: Yes, I think it's a good idea to remove them. All the forms have been updated to have a valid format

src/modules/form/hooks/useFormDefinitionHandlers.tsx
gigxz: It appears handled to me.
Related PR https://github.com/greenriver/hmis-frontend/pull/588/commits/5c7955c248dde71053de0c60728f4a881fb1afc0 added support for `autofill_readonly` attribute to specify whether the field should autofill even when form is being viewed as "read only".

One such example is Monthly Total Income. To reproduce, I did:
* Fill out and submit assessment
* In the rails console, manually change MonthlyTotalIncome to a different value (aka make it so incomes don't add up correctly)
* Reopen the submitted assessment. It correctly shows the "wrong" value ($400) rather than the auto-calculated value. Screenshot below. This is what we expect. (This comes up for data correction, it's beneficial to see the **real** stored value when viewing. Re-submitting the form will fix the calculation).

Screenshot 2025-01-13 at 11 14 38 AM

src/modules/form/components/AutofillFormItemWrapper.tsx
```diff
@@ -0,0 +1,60 @@
...
+  children,
+}) => {
+  const linkId = getSafeLinkId(item.linkId);
+  // dependentLinkIds are referenced from the 'autofill_when' property of item
+  // example:
+  //   link_id: item.id,
+  //   autofill_values: [ { autofill_when: [{ question: dependentLinkIds[0] }] } ],
+  const { autofillInvertedDependencyMap, getAutofillValueForField } = handlers;
+  const dependentLinkIds =
+    autofillInvertedDependencyMap[item.linkId].map(getSafeLinkId);
```
gigxz:
```suggestion
  const dependentLinkIds = (
    autofillInvertedDependencyMap[item.linkId] || []
  ).map(getSafeLinkId);
```


This fixes the following bug. I haven't dug into it too much beyond this. Why did we think that the `item.linkId` would definitely be present in `autofillInvertedDependencyMap`?

Repro:
* Add a new member to a household. They should have incomplete status.
* Fill out intake assessment for new member, hit "save"
* Go to the summary/submission tab in the household workflow
* Submit the intake for the new member
* Navigate back to the intake for the new member

Without this fix I was seeing an error:
Screenshot 2025-01-13 at 11 11 30 AM

ttoomey: Thanks! I totally forgot to check the household enrollment form. I'll fix it up

gigxz: Thanks!
@ttoomey I think I misrepresented this a bit. I am seeing this error when viewing most **submitted** assessment (tried: a local custom assessment, a hud intake, a hud annual), regardless of whether it's in the individual or household view.

The Intake Form Definition I'm using is the one that gets seeded by default in a development environment and for `CLIENT=qa_hmis` -- that's the one that has _all_ the HUD questions enabled. Maybe you're using something different? 

Assessments that I _can_ open successfully:
* Any work-in-progress assessments (HUD and custom)
* A tiny submitted "housing needs assessment" I have locally -- the difference may be that it doesn't have conditionals or autofills (Not sure, but I can track down the differences if you have trouble reproducing)

src/modules/form/util/formUtil.tsx
```diff
@@ -1375,11 +1388,12 @@ export const getDependentItemsDisabledStatus = ({
 }: GetDependentItemsDisabledStatus) => {
```
gigxz: Poking around at this because I'm looking at a difference with the DisabilityTable. Maybe I'm missing something obvious but I'm not seeing that `getDependentItemsDisabledStatus` is used anymore? (Or maybe it already wasn't?)

src/modules/form/components/group/DisabilityTable.tsx
```diff
@@ -52,7 +52,7 @@ const DisabilityTable = ({
       ?.linkId;
   }, [item]);
```
gigxz, line 57: Not sure where this comment will belong, but placing it here. There is a difference in something between QA and this branch:

* Select Yes to Physical Disability
* Select Yes to Physical Disability is a Disabling Condition
* Clear out the value for Physical Disability
  * **QA**: this results in "Physical Disability is a Disabling Condition" being cleared out as well
  * **This Branch**:  "Physical Disability is a Disabling Condition" appears to remain as Yes.

Pretty sure this doesn't stem from logic within the DisabilityTable component (which mostly deals with the "overall disabling condition" at the bottom), but rather something about "autofill dependencies" . Let me know if you'd like me to spend some more time tracking it down or pairing!

ttoomey: good find! I've patched this behavior in ~~DependentFormItemWrapper~~ DynamicFormField (d2a32a047d3c1a715b833b2fc1f20931d8eb757c)

src/modules/form/components/AutofillFormItemWrapper.tsx
ttoomey: Yes, I was testing with definitions from a different client env to exercise the MCI component, thanks for the reminder on using QA definitions. Your fix to autofillInvertedDependencyMap seems to corect the crash

src/modules/form/hooks/useFormDefinitionHandlers.tsx
ttoomey: Will do, I'll leave them for the moment; removing them may cause some code-churn; I want to address the more immediate bugs first

src/modules/form/util/formUtil.tsx
ttoomey: good catch! It was used previously but I've removed it

src/modules/form/components/DependentFormItemWrapper.tsx
```diff
@@ -0,0 +1,118 @@
...
+      return [true, null];
+
+    // Hide if all of this item's children are disabled
+    if (childItems.length === 0) return [false, null];
+    const childrenHidden = childItems.every(
+      (child) =>
+        isItemDisabled(child) &&
+        child.disabledDisplay === DisabledDisplay.Hidden
+    );
+    return [childrenHidden, dependantValues];
```
gigxz: I'm confused at the return of `dependantValues` here which doesn't seem to be used?

src/modules/form/components/DependentFormItemWrapper.tsx
```diff
@@ -0,0 +1,118 @@
...
+  ]);
+
+  // If the element has become disabled, clear the value
+  const { getValues, setValue } = handlers.methods;
+  useEffect(() => {
+    if (!isDisabled) return;
+    if (isNil(getValues(linkId))) return;
+
+    setValue(linkId, null, { shouldDirty: false });
+  }, [isDisabled, getValues, linkId, setValue]);
```
gigxz: Suggestion to not clear value if item has `DisabledDisplay.ProtectedWithValue`. This matches current behavior. (Tested)
```suggestion
  // If the element has become disabled, clear the value (if appropriate)
  const { getValues, setValue } = handlers.methods;
  useEffect(() => {
    if (!isDisabled) return;
    // If item is configured with "ProtectedWithValue", the value should remain even when the item is disabled.
    if (item.disabledDisplay === DisabledDisplay.ProtectedWithValue) return;
    if (isNil(getValues(linkId))) return;

    setValue(linkId, null, { shouldDirty: false });
  }, [isDisabled, getValues, linkId, setValue, item.disabledDisplay]);
```

Desired:
Kapture 2025-01-14 at 09 41 41

Bug:
Kapture 2025-01-14 at 09 43 53


---


The "disabled display" functionality is not documented in the API, but it is described in the form builder:

Screenshot 2025-01-14 at 9 39 19 AM



Form items used in gif example:
```
        {
          "text": "test2",
          "type": "STRING",
          "bounds": [
            {
              "id": "bound_0b0fc13f",
              "type": "MAX",
              "severity": "error",
              "value_number": 6
            },
            {
              "id": "bound_962ccc5e",
              "type": "MIN",
              "severity": "error",
              "value_number": 3
            }
          ],
          "link_id": "test_2",
          "mapping": {
            "custom_field_key": "customassessment_test_2"
          },
          "required": false,
          "read_only": false,
          "helper_text": "min/max",
          "warn_if_empty": false,
          "disabled_display": "HIDDEN"
        },
        {
          "text": "number enabled when previous text is present",
          "type": "INTEGER",
          "link_id": "number_2",
          "mapping": {
            "custom_field_key": "customassessment_number_2"
          },
          "required": false,
          "read_only": false,
          "enable_when": [
            {
              "operator": "EXISTS",
              "question": "test_2",
              "answer_boolean": true
            }
          ],
          "helper_text": "disabled display = protected with value",
          "warn_if_empty": false,
          "enable_behavior": "ALL",
          "disabled_display": "PROTECTED_WITH_VALUE"
        },
```

src/modules/form/components/DependentFormItemWrapper.tsx
ttoomey: The intention was to ensure isDisabled is re-evaluated when dependantValues change; however including in the dependency array without referencing it from the function is a linter error. Returning the dependency lets us include it without needing magic comments. 

But this isn't super clear; I could:
1) add a comment
2) remove this memoization (maybe we can rely on isolated renders to keep this from being called too frequently)
3) rewrite it to be clearer using useState/useEffect

src/modules/form/components/DependentFormItemWrapper.tsx
ttoomey: Thanks, that was very helpful. I think I've located where the original branch was trying to handle this case (src/modules/form/components/DynamicField.tsx:97). But that doesn't seem to work due to props-spreading overrides.

Perhaps I should back out my fix and fix this bug instead? This would correct the display without actually nulling-out the form value; I'm a little unsure which is better
 
```diff
--- a/src/modules/form/components/DynamicFormField.tsx
+++ b/src/modules/form/components/DynamicFormField.tsx
@@ -1,5 +1,5 @@
 import { Grid } from '@mui/material';
-import { isEmpty, pick } from 'lodash-es';
+import { isEmpty, omit, pick } from 'lodash-es';
 import React, { cloneElement, ReactNode, useCallback } from 'react';
 
 import {
@@ -240,14 +240,14 @@ const DynamicFormField: React.FC<Props> = ({
               key={item.linkId}
               item={item}
               itemChanged={itemChanged}
-              disabled={isDisabled}
+              disabled={isDisabled || fieldProps?.disabled}
               errors={getFieldErrors(item)}
               horizontal={horizontal}
               pickListArgs={pickListArgs}
               warnIfEmpty={warnIfEmpty}
               // Needed to support referencing local constants in expression evaluation (DynamicDisplay)
               localConstants={localConstants}
-              {...fieldProps}
+              {...omit(fieldProps, ['disabled'])}
               inputProps={{
                 ...fieldProps?.inputProps,
                 ...buildCommonInputProps({
```

src/modules/form/components/client/addresses/SimpleAddressInput.tsx
ttoomey: doesn't actually appear to be registration related. Not totally sure why this doesn't work but dropping for a cleaner defaultValue solution

src/modules/form/components/AutofillFormItemWrapper.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  const { setValue, getValues, control } = handlers.methods;
+
+  // subscribe to changes in the dependent field values
+  useWatch({ control: control, name: dependentLinkIds });
+  // not memoized, we rely on useWatch to re-render here when dependentLinkIds change
+  const autofillValue = getAutofillValueForField(item);
+
+  // Listen to see if this field has been edited by the user
+  const { isDirty } = useFormState({ control: control, name: linkId });
```
ttoomey: there was a bug here; isDirty is not the way to know if a single form field is diry

src/modules/form/components/AutofillFormItemWrapper.tsx
```diff
@@ -43,10 +43,12 @@ const AutofillFormItemWrapper: React.FC<Props> = ({
...
 
   // Listen to see if this field has been edited by the user
-  const { isDirty } = useFormState({ control: control, name: linkId });
+  const { dirtyFields } = useFormState({ control: control, name: linkId });
+  const isDirty = !!dirtyFields[linkId];
 
   useEffect(() => {
     // Don't autofill this field if it's been edited (i.e. is dirty)
+    // for example, we automatically set the radio choice for Income Source to true if any of the income fields are non-zero
     if (isDirty) return;
```
gigxz: Note-to-self: do we need to differentiate between user-modified dirty fields vs system-modified (initial values, autofills)? see if behavior matches in those cases

src/modules/form/components/DynamicFormField.tsx
```diff
@@ -0,0 +1,292 @@
...
+              );
+            }}
+            values={values}
+            itemChanged={itemChanged}
+            severalItemsChanged={severalItemsChanged}
+            visible={visible}
+            debug={
+              import.meta.env.MODE === 'development'
+                ? (keys?: string[]) => {
+                    const sectionValues = keys ? pick(values, keys) : values;
```
gigxz: An adjustment I made locally so I could use the "Debug" button on the form during local development. `values` is stale here otherwise
```suggestion
                    const currentValues = handlers.getValues();
                    const sectionValues = keys
                      ? pick(currentValues, keys)
                      : currentValues;
```

src/modules/form/components/DynamicFormField.tsx
```diff
@@ -0,0 +1,292 @@
...
+              // Needed to support referencing local constants in expression evaluation (DynamicDisplay)
+              localConstants={localConstants}
+              {...omit(fieldProps, ['disabled'])}
+              inputProps={{
+                ...fieldProps?.inputProps,
+                ...buildCommonInputProps({
+                  item,
+                  values,
+                  localConstants: localConstants || {},
+                }),
```
gigxz, line 129: There is a sneaky bug here. The reason `buildCommonInputProps` accepts values is because it calculates the `min` and `max` props to pass, based on the `bound`(s) specified on the item. Those min/max values can be based on other questions that have changing values. That functionality breaks here since we pass the stale `values` rather than `handler.getValues()`.


There is probably a way to optimize the renders a bit more, but, this does fix the functionality for me:

```suggestion
              inputProps={{
                ...fieldProps?.inputProps,
                ...buildCommonInputProps({
                  item,
                  values: handlers.getValues(),
                  localConstants: localConstants || {},
                }),
```


Example:
Kapture 2025-01-16 at 09 46 44

```json
        {
          "text": "Q1",
          "type": "INTEGER",
          "link_id": "number_2",
          "mapping": {
            "custom_field_key": "customassessment_number_2"
          },
          "required": false,
          "read_only": false,
          "enable_when": [
            {
              "operator": "EXISTS",
              "question": "test_2",
              "answer_boolean": true
            }
          ],
          "helper_text": "disabled display = protected with value",
          "warn_if_empty": false,
          "enable_behavior": "ALL",
          "disabled_display": "PROTECTED_WITH_VALUE"
        },
        {
          "text": "Q2",
          "type": "INTEGER",
          "bounds": [
            {
              "id": "bound_95078fbc",
              "type": "MAX",
              "question": "number_2",
              "severity": "error"
            }
          ],
          "link_id": "q_2",
          "mapping": {
            "custom_field_key": "customassessment_q_2"
          },
          "helper_text": "max = value of Q1",
          "disabled_display": "HIDDEN"
        },
```

src/modules/form/components/DynamicFormField.tsx
```diff
@@ -0,0 +1,292 @@
...
+        }
+
+        return group;
+      }
+
+      const itemComponent = item.readOnly ? (
+        <Grid item key={item.linkId}>
+          <DynamicViewField
+            item={item}
+            value={values[item.linkId]}
```
gigxz: Assumption here that `values` is unchanging, but we do pass that `itemChanged` below for some specific use case ... one of those tiny complicated things we have added over time. Note-to-self to come back and test that scenario, to see if behavior regressed due to "stale" values being passed now.

src/modules/form/components/DynamicFormField.tsx
```diff
@@ -0,0 +1,292 @@
...
+  warnIfEmpty = false,
+  horizontal = false,
+  visible = true,
+  pickListArgs,
+  item,
+  nestingLevel,
+  props: fieldProps,
+  renderFn,
+}) => {
+  const values = handlers.getValues();
```
gigxz: Can you add a comment about how this `values` contains the form values as they were when this component was last rendered, but does not necessarily represent the _current_ form state (since there may have been other input changes that didn't trigger a re-render of this component). Unlike "use watch" . (If I got that right). Something I know will/has tripped me up :)

src/modules/form/components/DynamicFormField.tsx
```diff
@@ -0,0 +1,292 @@
...
+                <AutofillFormItemWrapper handlers={handlers} item={item}>
+                  {(autofillValue) => {
+                    return isCurrency
+                      ? formatCurrency(autofillValue || 0)
+                      : autofillValue || 0;
+                  }}
+                </AutofillFormItemWrapper>
+              );
+            }}
+            values={values}
```
gigxz: Note to self: look more closely at this interaction with `FormCard` and any other DynamicGroup that looks at `values` prop. We already have the special case for DisabilityTable below that passes the current values.

Maybe this could be simplified by passing the group the "values" of just the child link ids within that group. Or have the group watch what it needs separately via context.

src/modules/form/components/DependentFormItemWrapper.tsx
gigxz: Ah I see, thanks. I think adding a comment would be great

gigxz (commented):
This is looking great. Leaving a batch of in-progress review comments- several of these are places I plan to come back to (hopefully today) to test/QA more closely.

src/modules/form/components/DynamicFormField.tsx
gigxz: Alright yeah, we do need the value wrapper here, to make behavior consistent with the main branch. It would also make this consistent with `DynamicViewFields` which also wraps `DynamicViewField` elements in a ValueWrapper to get the "actual" value.

Example:
* Q1 is a numeric field
* Q2 is a numeric field marked as "read only", with an autofill rule stating that it should autofill to value `500` if the value of Q1 is greater than 5. (Found a bug in the form builder when trying to configure this btw,  https://github.com/greenriver/hmis-frontend/pull/1027)

Screenshot 2025-01-16 at 5 41 24 PM

--

Suggestion that fixes it (tested):

```diff
+           // Even though this field is marked as "read only", it needs the current value because it could change via autofill, or through the `adjustValue` fn which may change a value to have a readable label based on a remote pick list
+          <ValueWrapper name={getSafeLinkId(item.linkId)} handlers={handlers}>
+            {(value) => (
              <DynamicViewField
                item={item}
+                value={value}
                disabled={isDisabled}
                horizontal={horizontal}
                pickListArgs={pickListArgs}
                // Needed because there are some enable/disabled and autofill dependencies that depend on PickListOption.labels that are fetched (PriorLivingSituation is an example)
                adjustValue={itemChanged}
                // Needed to support referencing local constants in expression evaluation (DynamicDisplay)
                localConstants={localConstants}
                {...fieldProps}
              />
+            )}
+          </ValueWrapper>
```

src/modules/form/components/DynamicFormField.tsx
ttoomey: I've adjust the implementation to use RHF useWatch instead

src/modules/form/components/DynamicFormField.tsx
ttoomey: should be fixed now, I improved error tracking in NumericInput

src/modules/form/hooks/useFormDefinitionHandlers.tsx
ttoomey: thanks, I've removed the TODO comment

src/modules/form/hooks/useFormDefinitionHandlers.tsx
ttoomey: I've removed them now

src/modules/form/components/DynamicFormField.tsx
ttoomey: should be fixed now

ttoomey:
@gigxz - per our conversation on Friday, I did some additional refactoring to try and squash bugs and simplify form interactions.

gigxz:
Awesome thanks @ttoomey  I'll do another QA pass!

src/modules/form/hooks/useDynamicFormContext.tsx
```diff
@@ -1,15 +1,21 @@
...
+  definition: FormDefinitionJson;
+  itemMap: ItemMap;
+  localConstants: LocalConstants;
+  viewOnly: boolean;
+  autofillInvertedDependencyMap: LinkIdMap;
+  disabledDependencyMap: LinkIdMap;
 };
 
-export const DynamicFormContext = createContext<DynamicFormContextType>({});
+export const DynamicFormContext = createContext<DynamicFormContextType>(
```
ttoomey, line 16: This context makes it convenient to access static form configuration. I find the context makes it easier to make smaller composable hooks since I don't need to pass around a bag of form props.

src/modules/form/hooks/rhf/useDynamicFieldStatus.tsx
```diff
@@ -0,0 +1,89 @@
...
+ * Computes the status of a field
+ * result.disabled is evaluated from the item's enableWhen configuration
+ * result.hidden is evaluated by:
+ * 1  The item is configured to be hidden
+ * 2. The item is disabled and configured to be hidden when disabled
+ * 3. All of the items children are hidden (they are disabled and configured to be hidden when disabled
+ *
+ * Attempts to isolate re-renders by watching only the items dependencies and those of the child items
+ *
+ */
```
gigxz, line 26: This is much easier to understand and a helpful comment, thank you!

src/modules/form/hooks/rhf/useDynamicFieldWatchValues.tsx
```diff
@@ -0,0 +1,22 @@
+import { useMemo } from 'react';
+import { FieldValues, useFormContext, useWatch } from 'react-hook-form';
+
+export const useDynamicFieldWatchValues = (watchFields: string[]) => {
+  const { control } = useFormContext();
+  const valueArray = useWatch({
+    control,
+    name: watchFields,
+  });
```
gigxz:
```suggestion
export const useDynamicFieldWatchValues = (watchFields?: string[]) => {
  const { control } = useFormContext();
  const valueArray = useWatch({
    control,
    name: watchFields || [],
  });
```
I noticed a performance issue, and this fixes it. Passing `undefined` here causes fields that have no dependencies to watch ALL other fields.

Underlying type issue is that `LinkIdMap` doesn't allow for undefined value, but the `autofillInvertedDependencyMap` doesn't contain all link ids (or have non-undefined values for all link ids).

So either:
* Update map(s) to include all link ids mapped to `[]` when no dependencies (this would be an adjustment to buildAutofillDependencyMap I think, and maybe the inver logic?), or
* Add secondary type like `PartialLinkIdMap = Record<string, string[] | undefined>;` to use for autofill map which we know contains only partial set of keys

Second option seems simpler to me.

ttoomey: Thanks for finding and fixing that; I agree that improving the types is probably simplest, I'll do that

src/modules/form/components/DynamicFormField.tsx
gigxz: adjustment proposed https://github.com/greenriver/hmis-frontend/pull/1032

gigxz:
Rebased to 150 and opened https://github.com/greenriver/hmis-warehouse/pull/5097 to run system tests again

ttoomey:
@gigxz made a few fixes to address the QA issues (see thread  in [6043](https://github.com/open-path/Green-River/issues/6043)).

src/modules/form/hooks/rhf/useEnrichedFormData.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const { data: picklistValues, loading } = usePreloadPicklists({
+    definition: definition,
+    pickListArgs,
+  });
+
+  const [defaultValues] = useState<DefaultValues<T>>(() => {
+    const newValues: Record<string, any> = {
+      ...getInitialValues(definition, localConstants),
+      ...cloneDeep(initialValues),
+    };
```
gigxz, line 47: Things seem to be working well during my testing, so I'm inclined to leave this as proposed, but I'm not totally understanding why we added this here. 

Previously the parent component of the Form was responsible for generating the initial values (Eg AssessmentForm)-- and they still do-- but now we also re-generate the initial values values and then overlay the parent-provided initialValues on top of it. 

It does make more sense to me for the initial value calculation to live here. Just calling out that the calculation occurs elsewhere too.

src/modules/form/components/DynamicForm.tsx
```diff
@@ -167,132 +98,89 @@ const DynamicForm = forwardRef(
...
+    localConstants: props.localConstants,
+    viewOnly: false,
+  });
+  if (loading || !defaultValues) {
+    return <Loading />;
+  }
+  return (
+    <DynamicFormWithHandlers
+      defaultValues={defaultValues}
+      {...props}
```
gigxz, line 177: `initialValues` is no longer used by any child components now that we are using "defaultValues". Suggest not passing it on? (to avoid confusion of why it's there in the first place)


```diff
   onDirty?: (value: boolean) => void;
-  initialValues?: InitialValues;
   localConstants?: LocalConstants;
   variant?: 'standard' | 'without_top_level_cards';
   errorFilter?: ErrorFilterFn;
@@ -65,7 +64,6 @@ export const DynamicFormWithoutHandlers = forwardRef<
       onSubmit,
       onSaveDraft,
       onDirty,
-      initialValues,
       errors: errorState,
       localConstants,
       handlers,
@@ -158,12 +156,12 @@ const DynamicFormWithHandlers = forwardRef<
   );
 });

-// load remote data (picklists) to augment form data
-
+// load remote data (picklists) to augment form data (initialValues => defaultValues)
 const DynamicFormEnrichedDataLoader = forwardRef<
   DynamicFormRef,
-  DynamicFormProps
+  DynamicFormProps & { initialValues?: InitialValues }
 >((props, ref) => {
+  const { initialValues, ...rest } = props;
   const { defaultValues, loading } = useEnrichedFormData({
     pickListArgs: props.pickListArgs,
     definition: props.definition,
@@ -177,7 +175,7 @@ const DynamicFormEnrichedDataLoader = forwardRef<
   return (
     <DynamicFormWithHandlers
       defaultValues={defaultValues}
-      {...props}
+      {...rest}
       ref={ref}
```

gigxz (approved):
this looks like a huge improvement. thanks for all your work @ttoomey. a couple comments below, but overall I say it's ready to move to QA.


system test run against 152: https://github.com/greenriver/hmis-warehouse/pull/5097

src/modules/form/hooks/rhf/useEnrichedFormData.tsx
ttoomey: Oh interesting I didn't realize the caller was also performing this. I just extracted it from `useFormDefinitionHandlers`, definitely open to removing it if it's redundant.

src/modules/form/components/DynamicForm.tsx
ttoomey: good point, I'll remove it

src/modules/form/hooks/rhf/useEnrichedFormData.tsx
gigxz: Oh okay. Let's leave it because I'm not totally sure if its redundant, maybe it is for assessments but not other form types .

gigxz (approved):
yay

PR #619, opened 2024-01-24
Support multiple custom assessments

gigxz:
## Description

**Bug**: if there are multiple custom assessments, you can't start them. Each one shows the same definition.

This is implemented as a routing change. `formDefinitionId` is added to the route for custom assessments.
I'm proposing splitting out the assessment route so we have separate routes for new individual assessment vs existing assessment.

depends on https://github.com/greenriver/hmis-warehouse/pull/3948

## Type of change
- [x] Bug fix


src/routes/routes.ts
```diff
@@ -56,7 +56,9 @@ const enrollmentDashboardRoutes = {
   EVENTS: 'events',
   CE_ASSESSMENTS: 'ce-assessments',
   CUSTOM_CASE_NOTES: 'case-notes',
-  ASSESSMENT: 'assessments/:formRole/:assessmentId?',
+  ASSESSMENT: 'assessments/:formRole/:assessmentId?', // Legacy
+  VIEW_ASSESSMENT: 'assessments/:formRole/:assessmentId',
+  NEW_ASSESSMENT: 'assessments/:formRole/new/:formDefinitionId?',
```
gigxz: intake and exit: want to be able to just link to it regardless of whether it exists or not. act as shortcuts to redirect based on whether assmt exists or not

`/intake`
`/exit`


new assessment: `/assessments/new/:formDefinitionId`
existing assessment: `/assessments/:assessmentId`

specify `household: bool` on the definition(?), get rid of the "is household x" logic from the frontend

gigxz: @ttoomey I ended up implementing it like this:
```
INTAKE: 'intake', // Go to intake (render household view, or navigate to View/Create individual assessment appropriate)
  EXIT: 'exit', // Go to intake (render household view, or navigate to View/Create individual assessment as appropriate)
  VIEW_ASSESSMENT: 'assessments/:assessmentId', // View/Edit individual assessment
  NEW_ASSESSMENT: 'assessments/new/:formDefinitionId', // Create new individual assessment
```

Re: rendering the household view directly for intake/exit: the main reason I did that was ease of implementation... it seemed difficult to try to use the cache for the form definition after fetching it, since it has a weird cache key. 

I think what I have here all works, and seems slightly better.. I did split out the queries so we have a dedicated query for _assessment_ form definition lookup, since it is different from looking up other forms.

I haven't done self-review or serious QA yes, I will re request you after that, but feel free to look before. 🙇‍♀️

src/components/clientDashboard/enrollments/EnrollmentReminders.tsx
```diff
@@ -132,30 +132,26 @@ const rowLinkTo = ({
 }: ReminderFieldsFragment): string => {
   switch (topic) {
     case ReminderTopic.AnnualAssessment:
-      return generateSafePath(EnrollmentDashboardRoutes.ASSESSMENT, {
+      // FIXME
```
gigxz: https://github.com/greenriver/hmis-frontend/pull/621

src/components/clientDashboard/enrollments/EnrollmentReminders.tsx
```diff
@@ -132,30 +132,26 @@ const rowLinkTo = ({
...
+      return generateSafePath(EnrollmentDashboardRoutes.ASSESSMENTS, {
         clientId: client.id,
         enrollmentId: enrollment.id,
-        formRole: AssessmentRole.Annual,
       });
     case ReminderTopic.AgedIntoAdulthood:
-      return generateSafePath(EnrollmentDashboardRoutes.ASSESSMENT, {
+      // FIXME
```
gigxz: https://github.com/greenriver/hmis-frontend/pull/621

src/modules/assessments/hooks/useRelatedAnnualAssessments.tsx
```diff
@@ -78,46 +66,10 @@ export function useRelatedAnnualAssessments({
...
+        path: generateSafePath(EnrollmentDashboardRoutes.VIEW_ASSESSMENT, {
           clientId: client.id,
           enrollmentId: enrollment.id,
           assessmentId: assessment.id,
-          formRole: FormRole.Annual,
-        }),
-      });
-    });
-
-    // Next, add all other members that DON'T have Annuals within that target range
```
gigxz, line 90: We have another [ticket](https://www.pivotaltracker.com/story/show/186384658) around removing this because itis considered too confusing. So I am just  doing that here.

src/modules/assessments/components/household/HouseholdAssessmentTabPanel.tsx
```diff
@@ -99,19 +95,32 @@ const HouseholdAssessmentTabPanel = memo(
...
+      useAssessment(assessmentId);
+
+    const definition = useMemo(() => {
+      if (assessmentId && !assessment) return;
+
+      // If we are loading an existing Assessment, always prefer to use
+      // the FormDefinition that was resolved on the Assessment. This could
+      // be important if it's an older WIP assessment that was saved using a certain
+      // form. (It should be re-opened using the same form).
+      const chosenDefinition = assessment?.definition || mainFormDefinition;
```
gigxz, line 108: This feels a bit sneaky. We have the "main" form definition that we fetch before rendering the household's assessments. But if one of the HHMs has a WIP assessment that was started using a different form definition, we should use that one when rendering that panel. This won't come up yet with the current approach, but it may in the future.

src/modules/assessments/components/household/HouseholdAssessments.tsx
```diff
@@ -70,7 +71,7 @@ const calculateAssessmentStatus = (
 const HouseholdAssessments: React.FC<Props> = ({
   role,
   enrollment,
-  assessmentId,
+  formDefinition,
```
gigxz, line 74: I was able to reduce the number of queries when loading the HH Assessment page. It was fetching each form def separately. It still loads them multiple times potentially because existing assessments will resolve their own form def on the assessment record. but at least it is fewer queries

src/routes/routes.ts
```diff
@@ -57,7 +57,10 @@ const enrollmentDashboardRoutes = {
   EVENTS: 'events',
   CE_ASSESSMENTS: 'ce-assessments',
   CUSTOM_CASE_NOTES: 'case-notes',
-  ASSESSMENT: 'assessments/:formRole/:assessmentId?',
+  INTAKE: 'intake', // Go to intake (render household view, or navigate to View/Create individual assessment appropriate)
+  EXIT: 'exit', // Go to intake (render household view, or navigate to View/Create individual assessment as appropriate)
```
ttoomey:
```suggestion
  EXIT: 'exit', // Go to exit (render household view, or navigate to View/Create individual assessment as appropriate)
```

src/modules/assessments/components/household/HouseholdAssessments.tsx
gigxz: note to self... we can also get rid of the "always mounted tab panel" approach

src/components/clientDashboard/enrollments/NewAssessmentPage.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { useEnrollmentDashboardContext } from '@/components/pages/EnrollmentDashboard';
+import NotFound from '@/components/pages/NotFound';
+import useSafeParams from '@/hooks/useSafeParams';
+import IndividualAssessmentPage from '@/modules/assessments/components/IndividualAssessmentPage';
+import MissingDefinitionAlert from '@/modules/assessments/components/MissingDefinitionAlert';
+import useAssessmentFormDefinition from '@/modules/assessments/hooks/useAssessmentFormDefinition';
+import { EnrollmentDashboardRoutes } from '@/routes/routes';
+
+/**
+ * Renders blank assessment(s), for an individual or a household.
```
gigxz: fix, no household

src/components/clientDashboard/enrollments/AssessmentPage.tsx
```diff
@@ -1,53 +1,61 @@
...
-    enrollment.householdSize > 1 &&
-    isHouseholdAssesmentRole(role)
-  );
-};
+import { applyDefinitionRulesForClient } from '@/modules/form/util/formUtil';
+import { EnrollmentDashboardRoutes } from '@/routes/routes';
+import { useGetAssessmentQuery } from '@/types/gqlTypes';
 
+/**
+ * Renders existing assessment(s), for an individual or a household
```
gigxz: fix, not for household

src/modules/assessments/components/household/HouseholdAssessments.tsx
gigxz: I'm not sure if this was the right approach, maybe it would be better if each tab fetched their own thing and it was cached since they're the same. Ill do a separate PR for the tab panel mounting change

src/components/clientDashboard/enrollments/NewAssessmentPage.tsx
```diff
@@ -0,0 +1,51 @@
...
+ */
+const NewAssessmentPage = () => {
```
ttoomey: maybe call this `NewIndividualAssessmentPage` to differentiate it from the household?

src/components/clientDashboard/enrollments/AssessmentPage.tsx
ttoomey: Not sure if adjusting the name would be too much of a rabbit hole but maybe this component could be called "IndividualAssessmentPage" to be more clear? However we already have a component with that name.

A convention I've seen in other projects is to call something a "Page" if it is addressable from a URL (mounted from the router). If we use that convention here then this component is a page but our current "IndividualAssessmentPage" is not; it configures the form behaviors and renders the form. Maybe that component would be better named as something like "IndividualAssessmentFormController".

Feel free to ignore this suggestion but I wonder we could rename the components this way:
IndividualAssessmentPage => IndividualAssessmentFormController (or something)
AssessmentPage => IndividualAssessmentPage

ttoomey (approved):
Looks great, thank you!! 🙌

src/components/clientDashboard/enrollments/AssessmentPage.tsx
gigxz: agreed that makes sense to me, updated!

PR #620, opened 2024-01-25
Add client alert cards to client profile and enrollment pages

martha:
## Description

Adds Client Alerts to the Client Profile page.

https://www.pivotaltracker.com/n/projects/2591838/stories/184132729

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/3946

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/config/theme.ts
```diff
@@ -74,6 +82,12 @@ export const baseThemeDef: ThemeOptions = {
     },
     alerts: {
       lightWarningBackground: '#fffde0',
+      highPriorityErrorBackground: '#FF7575',
```
martha: @gigxz a couple of questions about the colors that I wanted to discuss with you. You mentioned `lightWarningBackground` on slack, but I think it's not actually the same hex value as the one in the Figma files. A few of the other colors were present already in the theme -- e.g. the high priority border and body background were versions of the `theme.palette.error.main` with some transparency -- but most were hex values that weren't present yet anywhere in the codebase, as far as I could find. I'm not really sure what approach to take here with adding the colors -- naming them, where to put them, trying to consolidate similar colors? (e.g. the border color and the header background color for the high and medium priority alerts are slightly different values in Figma.) Let me know what you think, or if we should loop in Ash?

src/modules/client/components/ClientAlerts.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+const ClientAlertCard: React.FC<ClientAlertCardProps> = ({
+  alert,
+  client,
+  shouldShowClientName = false,
+}) => {
+  const priority = alert.severity || 'low';
+
+  const priorityColors: { [index: string]: any } = {
+    high: {
```
martha: Another question about colors - should this dictionary mapping alert priorities to their respective colors be moved into the theme file, instead of living here?

src/modules/client/components/ClientAlerts.tsx
```diff
@@ -0,0 +1,116 @@
...
+    low: {
+      header: theme.palette.alerts.lowPriorityHeaderBackground,
+      body: theme.palette.alerts.lowPriorityBodyBackground,
+      border: theme.palette.alerts.lowPriorityHeaderBackground,
+    },
+  };
+
+  return (
+    <Alert
+      sx={{
```
martha: Here's a screenshot of what this looks like (high priority version)
Screenshot 2024-01-25 at 2 34 11 PM

src/modules/client/components/ClientAlerts.tsx
```diff
@@ -0,0 +1,116 @@
...
+  client: ClientFieldsFragment;
+  shouldShowClientName?: boolean;
+}
+
+const ClientAlertCard: React.FC<ClientAlertCardProps> = ({
+  alert,
+  client,
+  shouldShowClientName = false,
+}) => {
+  const priority = alert.severity || 'low';
```
martha: Question about language -- I used the word "severity" in the database/backend (following from the figma document "Initial Discovery" section, I think), but the frontend language uses "priority" instead. Should I add another db migration to rename this column (or edit my existing yet-unmerged PR)?

martha: Also, should severity/priority become an enum, and if so, where should we enforce that?

martha: Lastly, I'm informally setting a default here of "low," but should that be done somewhere more formally on the backend?

src/components/clientDashboard/ProfileLayout.tsx
```diff
@@ -1,50 +1,33 @@
...
           </Stack>
         </Grid>
-        {canViewEnrollments && (
-          <Grid item md={12} lg={6}>
-            <ClientEnrollmentCard client={client} />
-          </Grid>
-        )}
+        <Grid item md={12} lg={6}>
+          <Stack gap={2}>
+            {canViewClientAlerts && <ClientAlerts client={client} />}
```
martha: Is it ok that right now Alerts are attached to Clients and fetched in the same query? Or should they be more like Enrollments and fetched in a separate query?

martha: I saw in the Figma file "This should not be shown if the alert feature is not enabled for the installation" -- are there other examples of installation-wide feature switches in our codebase that I can check out to understand the best way to do this?

src/api/operations/household.fragments.graphql
```diff
@@ -35,6 +26,22 @@ fragment HouseholdClientFields on HouseholdClient {
   }
 }
 
+fragment HouseholdClientFieldsClient on Client {
```
martha: This fragment name is extremely awkward-- I'm trying to reflect that it's the `Client` fragment internal to the `HouseholdClientFields` fragment. Is this ok or does this name spaghetti indicate too much code spaghetti?

src/modules/client/components/ClientAlerts.tsx
```diff
@@ -0,0 +1,150 @@
...
+    <TitleCard
+      title={title}
+      headerVariant='border'
+      headerTypographyVariant='body1'
+    >
+      <Box sx={{ m: 2 }}>
+        {clientAlerts.length === 0 &&
+          `${alertContext} has no alerts at this time`}
+        <Stack gap={2}>
+          {clientAlerts.map((ca) => (
```
martha: Should alerts perhaps be ordered by their priority, then by date?

src/modules/client/components/ClientAlerts.tsx
gigxz: All good Qs! I think I answered all of that here https://github.com/greenriver/hmis-warehouse/pull/3946#discussion_r1467747598

Whether to rename the DB column is up to you, I think it's fine to leave it

src/components/clientDashboard/ProfileLayout.tsx
gigxz: Since we have a special permission for viewing client alerts, we'll just use that permission as a feature toggle. That's usually what we do. So it should really read, "This should not be shown if the user does not have permission to view client alerts for this client"

src/modules/client/components/ClientAlerts.tsx
gigxz: There is also a placeholder in the PR description template for linking to the backend PR - that is always helpful for me as a reviewer!

src/components/clientDashboard/ProfileLayout.tsx
gigxz: I think same query is fine here

src/modules/client/components/ClientAlerts.tsx
martha: Thanks! I linked it.

src/modules/enrollment/components/dashboardPages/HouseholdPage.tsx
```diff
@@ -13,6 +14,10 @@ const HouseholdPage = () => {
     clientId: string;
   };
 
+  const [householdMembers, { loading: householdMembersLoading, error }] =
```
martha: Do you think this is a good approach -- moving the householdMembers query one level up so that HouseholdMemberTable doesn't perform it anymore (and the results can be shared between HouseholdMemberTable and the alerts on the Enrollments page)? Or should HouseholdMemberTable be more ambidextrous and capable of either making the request itself or receiving the household members as props?

src/components/clientDashboard/ProfileLayout.tsx
martha: Thanks, that makes sense!

src/modules/enrollment/components/dashboardPages/HouseholdPage.tsx
gigxz: Let's check in about it next week. From a quick glance, I think it relates to the spaghetti-ness of the new fragment name.

If you load up the Enrollment dashboard from scratch, you should see that all the graphql queries get batched in one request:
Screenshot 2024-01-26 at 4 44 02 PM


`GetEnrollmentWithHousehold` is this one thats loading all the members and all the Enrollment details. `GetEnrollmentReminders` loads the "household tasks" reminder card that shows up on the right.

My inclination would be to have the alert card function similarly to the household task card in this case. Meaning, the "Household Alert" card would be responsible for fetching what it needs. It could use a query that _just_ loads the household members' alerts. (A "new" query from the frontend perspective, but using the existing backend household query).  Imo that makes the code a lot more readable -- we can take advantage of GraphQL batching to help us separate concerns.

I know that is sort of contradictory to what I said about the Client Dashboard... on the Client dash, it makes more sense to me to load it on the Client because it's just the one record.

Those are just some initial thoughts! I really don't have a clear-cut answer on this type of question,. and it comes up a lot

martha: Yeah I agree with everything you said, thanks for steering me in the right direction!

martha: (I'm converting this PR back to draft mode for now haha)

src/modules/client/components/ClientAlerts.tsx
gigxz: Yeah that probably makes sense. Thanks!

src/config/theme.ts
gigxz: > You mentioned lightWarningBackground on slack, but I think it's not actually the same hex value as the one in the Figma files

Good to know, we should update it in figma probably. We should use the existing value in the theme

> A few of the other colors were present already in the theme -- e.g. the high priority border and body background were versions of the theme.palette.error.main with some transparency -- but most were hex values that weren't present yet anywhere in the codebase, as far as I could find. I'm not really sure what approach to take here with adding the colors -- naming them, where to put them, trying to consolidate similar colors? (e.g. the border color and the header background color for the high and medium priority alerts are slightly different values in Figma.) Let me know what you think, or if we should loop in Ash?


Yeah good question, lets consult with Ash about the usage. For reference:

Screenshot 2024-01-29 at 9 47 44 AM
Screenshot 2024-01-29 at 9 47 39 AM

src/modules/client/components/clientAlerts/ClientAlertCard.tsx
```diff
@@ -0,0 +1,80 @@
...
+  clients: ClientWithAlertFieldsFragment[];
+  alertContext: AlertContext;
+  loading?: boolean;
+}
+
+const ClientAlertCard: React.FC<ClientAlertCardProps> = ({
+  clients,
+  alertContext = AlertContext.Client,
+}) => {
+  const clientAlerts: ClientAlertProps[] = [];
```
martha: @gigxz  I'm curious what you think about this approach, where the ClientAlertCard receives a list of clients and then iterates thru them creating a `ClientAlertCardProps` for each one which it passes in to the ClientAlert component. I guess an alternative would be to have the ClientAlertCard be less smart and simply accept a list of alerts instead of a list of clients, so this processing would have already happened farther up the component tree? [resolving as outdated]

src/modules/client/components/ClientAlerts.tsx
martha: If the list of alerts is shown in a context where multiple household members' alerts are present, should they be also ordered by household member? So all the alerts for Person A would appear first (ordered internally by priority then date), then all the alerts for Person B? Or should the whole list of alerts just be ordered by priority + date regardless of the clients they're associated with?

martha: Another related question - I'm assuming we should not show expired alerts at all, and I'm guessing expired alerts should be filtered out already by the time we get here -- e.g. the backend query should just not return expired alerts. Is that right? Should I modify the backend query to not return expired alerts ever, or just not by default but there's some param that allows it if specifically requested?

martha: resolved at standup! We agreed that high priority alerts should appear highest -- regardless if whether it's a multi-client household and different client alerts get jumbled together. Let me know if you agree that it then makes sense to sort the alerts here, rather than in the backend (which I think would require restructuring our graphql queries somewhat)

martha: I updated https://github.com/greenriver/hmis-warehouse/pull/3946 to prevent returning expired alerts in the graphql query

src/modules/form/hooks/useFormDialog.tsx
```diff
@@ -146,6 +147,7 @@ export function useFormDialog<T extends SubmitFormAllowedTypes>({
             ) : formDefinition ? (
               <Grid container spacing={2} sx={{ mb: 2, mt: 0 }}>
                 <Grid item xs>
+                  {props.preFormComponent}
```
martha, line 150: Flagging that we wanted to find a better name instead of `preFormComponent`

src/modules/client/components/clientAlerts/ClientAlertCard.tsx
```diff
@@ -0,0 +1,113 @@
...
+    if (AlertPriority[a.alert.priority] === AlertPriority[b.alert.priority]) {
+      return Date.parse(a.alert.createdAt) - Date.parse(b.alert.createdAt);
+    } else {
+      return AlertPriority[b.alert.priority] - AlertPriority[a.alert.priority];
+    }
+  });
+
+  const title = `${alertContext} Alerts (${clientAlerts.length})`;
+
+  if (!shouldRenderFrame) {
```
martha: After the most recent changes, enabling us to show client alerts in the enrollment dialog box, `ClientAlertCard` now kinda has 2 different 'jobs' -- and it doesn't always actually appear as a card -- and it does contain a little bit of repeated code. (lines 60-68 are almost identical to lines 96-106)

I thought about a couple of different ways to refactor this and didn't really decide that any of the ones I explored was actually better than the code as it exists now, but I would certainly be open to restructuring if you have suggestions!

martha:
since this PR is already getting large, I'll move it out of draft mode, ~and plan to make a separate PR with the dialogbox for creation of client alerts.~ (since working more on the dialog involved refactoring some of the work here, I've included the dialog in this PR.)

src/modules/client/components/clientAlerts/ClientAlertCard.tsx
martha: scratch this question for now -- after some more work on the dialog modal a refactor is even more needed so I'll take another stab at it.

src/modules/client/components/clientAlerts/CreateClientAlertDialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+} from '@/types/gqlTypes';
+
+interface ClientAlertDialogProps extends DialogProps {
+  client: ClientWithAlertFieldsFragment;
+}
+export const CreateClientAlertDialog: React.FC<ClientAlertDialogProps> = ({
+  client,
+  ...props
+}) => {
+  const clientName = clientBriefName(client);
```
martha, line 28: Just checking it's appropriate to use `clientBriefName` rather than `clientNameAllParts` here?

src/modules/client/components/clientAlerts/CreateClientAlertDialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+                input: {
+                  clientId: client.id,
+                  ...values,
+                } as ClientAlertInput,
+              };
+            }}
+            getErrors={(data) => data.createClientAlert?.errors || []}
+            onCompleted={(data) => {
+              if (!data?.createClientAlert?.errors?.length && props.onClose)
+                props.onClose({}, 'escapeKeyDown');
```
martha, line 58: This is copied from Auto Exit Config, but I wasn't sure why we pretend `escapeKeyDown` is the 'cause' of closing the dialog when it was really submission of the form?

src/modules/client/components/clientAlerts/CreateClientAlertDialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+            }}
```
martha, line 71: Also, here we need to re-fetch the client so that the newly created alert shows up right away. It looks like Apollo has some built-in functionality for this available in react hooks but I wasn't quite understanding how to incorporate it; maybe we can pair on this tomorrow?

martha: Two other topics I'd like to pair on tomorrow:
- I've picked 'CHOICE' as the type for the priority field [here](https://github.com/greenriver/hmis-warehouse/pull/3952/files#diff-cd3fe7155a7e32a7811cb8434c97e51b7f2aa8694788e5a1c72e9516720fb1ec), and it shows up as radio buttons, whereas other 'CHOICE' types I'm seeing in the json form builder show up as dropdowns -- is this because there are only 3 options? (also, Ash indicated we should add colored circles next to the priorities in the dropdown, but since it is a dynamically built form, I am not sure how to do this)
- The `priority` and `note` fields should be required, and in the json definition linked above they are, but it's not taking effect in the same way that it does for AutoExit - so I'd like to investigate this together and figure out which part I am missing

gigxz:
Just flagging (for myself) that this shouldn't be merged until https://github.com/greenriver/hmis-warehouse/pull/3952 is merged because it expects the client alert mutations to be present

src/modules/client/components/clientAlerts/CreateClientAlertDialog.tsx
gigxz: Yep!

src/modules/client/components/clientAlerts/ClientAlert.tsx
```diff
@@ -0,0 +1,90 @@
...
+  parseAndFormatDateTime,
+} from '@/modules/hmis/hmisUtil';
+import {
+  ClientAlertFieldsFragment,
+  ClientWithAlertFieldsFragment,
+} from '@/types/gqlTypes';
+
+export interface ClientAlertProps {
+  alert: ClientAlertFieldsFragment;
+  client: ClientWithAlertFieldsFragment;
```
gigxz: This confused me a little. This component only needs the `client` in order to show the client name. It doesn't _need_ to have the alerts hanging off the client. Changing the type to `client: ClientNameFragment` would be clearer. Typescript should be happy with it because the `ClientWithAlertFieldsFragment` includes the name fragment.

src/modules/client/components/clientAlerts/ClientAlert.tsx
```diff
@@ -0,0 +1,90 @@
...
+  shouldShowClientName?: boolean;
```
gigxz: `showClientName` or `hideClientName` would be more consistent with other props in the codebase

src/components/clientDashboard/ProfileLayout.tsx
```diff
@@ -1,50 +1,32 @@
...
-          <Grid item md={12} lg={6}>
-            <ClientEnrollmentCard client={client} />
-          </Grid>
-        )}
+        <Grid item md={12} lg={6}>
+          <Stack gap={2}>
+            <ClientAlertProfileWrapper client={client} />
+            {canViewEnrollments && <ClientEnrollmentCard client={client} />}
+          </Stack>
+        </Grid>
```
gigxz: There is some other tweaking needed here, it's minor but we want the profile card to take up more space if the user doesn't have access to view enrollments _or_ alerts. That's why we check up on line `18` and change the width of the first column depending on the access levels.

It's a little wordy, but something like this should work:
```
  const { canViewEnrollmentDetails, canViewClientAlerts } = client.access;
  const hasRightColumn = canViewEnrollmentDetails || canViewClientAlerts;
  return (
    <Box data-testid='clientProfile'>
      <Grid container spacing={2}>
        <Grid item md={12} lg={hasRightColumn ? 6 : 8}>
          <Stack gap={2}>
            <ClientProfileCard client={client} />
            <ClientCustomDataElementsCard client={client} />
          </Stack>
        </Grid>
        {hasRightColumn && (
          <Grid item md={12} lg={6}>
            <Stack gap={2}>
              {canViewClientAlerts && (
                <ClientAlertProfileWrapper client={client} />
              )}
              {canViewEnrollmentDetails && (
                <ClientEnrollmentCard client={client} />
              )}
            </Stack>
          </Grid>
        )}
        {/* disabled "quick actions" card because the only action was Enroll in Project, which
          we are disabling for now #185750557 */}
        {/* <ClientActionsCard client={client} /> */}
      </Grid>
    </Box>
  );
```

martha: great - thanks for this catch!

src/modules/client/components/clientAlerts/ClientAlert.tsx
martha: I figured maybe it's less confusing to have the ClientAlert card just receive a string.

The `ClientWithAlertFieldsFragment` is still used in a couple of other places, though - should I get rid of those and replace them all with just `Client`?

gigxz: It makes sense to use `ClientWithAlertFieldsFragment` only when you need the `client` to have alerts on it, if that makes sense. Yeah you could also just accept a `clientName` string, I think thats also a good option.

src/api/operations/household.fragments.graphql
```diff
@@ -21,6 +21,9 @@ fragment HouseholdClientFields on HouseholdClient {
     externalIds {
       ...ClientIdentifierFields
     }
+    alerts {
+      ...ClientAlertFields
+    }
```
gigxz, line 26: Is this still needed? I wouldn't think so, now that you use a separate query to fetch the household alerts

src/modules/client/components/clientAlerts/ClientAlert.tsx
```diff
@@ -0,0 +1,86 @@
...
+        {showClientName && (
+          <Typography variant='body2' sx={{ pb: 1 }}>
+            {clientName}
+          </Typography>
+        )}
+        <Typography variant='body1' sx={{ pb: 1 }}>
+          {alert.note}
+        </Typography>
+        <Typography variant='body2' sx={{ fontSize: '12px' }}>
+          Created by {alert.createdBy?.name} on{' '}
```
gigxz: Handle case where we don't have a name
```suggestion
          Created by {alert.createdBy?.name || 'Unknown'} on{' '}
```

src/modules/client/components/clientAlerts/getClientAlerts.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { clientBriefName } from '@/modules/hmis/hmisUtil';
+import { ClientWithAlertFieldsFragment } from '@/types/gqlTypes';
+
+enum AlertPriority {
+  high = 3,
+  medium = 2,
+  low = 1,
+}
+
+export function getClientAlerts(
```
gigxz: I like that this is factored out, but I would probably make it a hook and memoize the alerts so they are not re-calculated on every render.

```
function useClientAlerts(clients, shouldShowClientName) {

    const clientAlerts = useMemo(() => {
          // code to construct and sort the Client Alerts

     }, [clients, shouldShowClientName]}) // only re-run when these change
    

   return { clientAlerts }
}
```

gigxz: Another way to do this would be to make the hook responsible for fetching the data, if any. For example:

`useClientAlerts ` hook accepts arguments `{ householdId?: string, client?: ClientWithAlertFieldsFragment, showClientName?: boolean }`
The hook would be responsible for:
1. Fetch alerts for household if `householdId` was provided (use `skip` to skip query otherwise)
2. If the `client` was provided, use the alerts on it
3. Return a properly sorted array with type `ClientAlertFieldsFragment[]`

That type of conditional-fetching-hook is okay, I think, it makes the component code a lot cleaner. The `useFormDefinition` is similar.

Let me know if you want to chat/pair through these ideas

src/modules/client/components/clientAlerts/ClientAlertWrappers.tsx
```diff
@@ -0,0 +1,71 @@
...
+  const clients = household.householdClients.map((c) => c.client);
+  const canViewClientAlerts = clients.some((c) => c.access.canViewClientAlerts);
+  if (!canViewClientAlerts) return;
+
+  const clientAlerts = getClientAlerts(clients, clients.length > 1);
+
+  return (
+    <ClientAlertCard
+      alertContext={AlertContext.Household}
+      clientAlerts={clientAlerts}
```
gigxz, line 47: It feels a bit off to me that we're technically passing a prop type here. (Actually an array of Prop types)

Usually when we pass a prop object, we capitalize the prop name (like `ButtonProps={buttonProps}`).

Maybe it makes sense to add an intermediate type like `ClientAlertType`, so you can pass an array of that instead of an array of props.

martha: Yes this felt a bit weird to me too, good idea!

src/api/operations/household.fragments.graphql
martha: Hm, it seems this is indeed still needed bc in the enrollment modal, the `client` we have access to can be either a `ClientFieldsFragment` or `RecentHouseholdMember` (which is based on `HouseholdClientFieldsFragment`)

src/modules/client/components/clientAlerts/getClientAlerts.tsx
martha: @gigxz thank you for the suggestions! I like the idea of the cleaner code that this could achieve but I wasn't quite able to get it working -- see https://github.com/greenriver/hmis-frontend/compare/martha/client-alerts...martha/client-alert-hooks for my attempts, but yes I would love to pair or chat more to get a better sense of how to put this together.

src/modules/client/components/clientAlerts/CreateClientAlertDialog.tsx
```diff
@@ -0,0 +1,77 @@
...
+            getVariables={(values) => {
+              return {
+                input: {
+                  clientId: client.id,
+                  ...values,
+                } as ClientAlertInput,
+              };
+            }}
+            getErrors={(data) => data.createClientAlert?.errors || []}
+            onCompleted={(data) => {
```
martha, line 54: Somehow I must still be missing something related to getting required fields working properly -- they now do appear with the little 'Required' red label, but on submitting the form, it still makes a network call (instead of erroring because the form's invalid, like the Auto Exit Config form does).

gigxz: Ah, my bad. I was thinking of the "record editing forms" which all use the `SubmitForm` mutation. For _those_ types of forms, it's sufficient to specify requiredness in the FormDefinition because the mutation actually validates against the form ([code](https://github.com/greenriver/hmis-warehouse/blob/256a2a28946c1e773c17d76fe0198bcd6f8afbff/drivers/hmis/app/graphql/mutations/submit_form.rb#L70)).

Since this is a static form that uses a custom mutation, it's a little different. I think you'll need the following changes:
* Make `note` input optional on the `ClientAlertInput`. This mutation is accepting the user input directly, so we need to  allow that to be empty (or else GraphQL will reject the request)
* Validate input in the mutation:
      ```
      errors = HmisErrors::Errors.new
      errors.add :note, :required if input.note.blank?
      errors.add :priority, :required unless input.priority
      return { errors: errors } if errors.any?
      ```


That should let you see user-facing validation when you submit with those fields empty. It does make a server request.

src/modules/client/components/clientAlerts/CreateClientAlertDialog.tsx
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <CommonDialog maxWidth='sm' fullWidth {...props}>
+      <DialogTitle>Create Alert for {clientName}</DialogTitle>
+      <DialogContent>
+        <Box mt={2}>
+          <Alert sx={{ mb: 2 }} severity='info'>
+            Reminder: All staff who can view {clientName} will be able to see
+            this alert
+          </Alert>
+          <StaticForm<
```
gigxz, line 39: I noticed that I'm able to put an expiration date in the past. It looks like the "local constants" aren't passing through to the form. (Thats what lets us say `$today` on the minimum value bound) Can you add `localConstants={localConstants}` [here](https://github.com/greenriver/hmis-frontend/blob/e7fbbaae3ecad78b026d93ef61dc3f00844222bc/src/modules/form/components/StaticForm.tsx#L83) to fix that? Thanks!

src/modules/client/components/clientAlerts/getClientAlerts.tsx
gigxz: Ok great! Lets. 
This one is annoying, you just need to cheat to make typescript happy:
> variables: { id: householdId || '' },
>     skip: !householdId,

src/modules/client/components/clientAlerts/CreateClientAlertDialog.tsx
martha: cool! Thank you.

src/modules/client/components/clientAlerts/CreateClientAlertDialog.tsx
martha: Yes! Good catch

src/config/theme.ts
```diff
@@ -27,15 +38,29 @@ declare module '@mui/material/styles' {
     cardTitle?: React.CSSProperties;
   }
 
+  interface AlertPriorityColorOptions {
```
gigxz, line 41: Nice!! This looks great.

src/config/theme.ts
```diff
@@ -277,7 +313,63 @@ const createThemeOptions = (theme: Theme) => ({
...
+                width: '100%',
+              },
+              '& .MuiAlert-body': {
+                padding: theme.spacing(2),
+              },
+            }),
+          };
+        },
       },
+      variants: [
```
gigxz: I was playing around with this a bit to see if we can define all these styles in the `variants` section rather than the `root` -- since they are all styles that are specific to one variant. I got it working like this. The `props` key specifies which props should be present for the styles to be applied, and the `style` specifies the styles. There is a way to return sx using `unstable_sx` which I used here.

What do you think? Although there is some repetition, it feels cleaner to me because it's not part of the `root` object -- so we can be sure that it doesn't affect the default alerts.

```
      variants: [
        {
          // styles that are shared across all withHeader alerts
          props: { variant: 'withHeader' },
          style: theme.unstable_sx({
            padding: 0,
            color: 'black',
            '& .MuiAlertTitle-root': {
              padding: theme.spacing(2),
              margin: 0,
              fontWeight: 800,
            },
            '& .MuiAlert-message': {
              padding: 0,
              width: '100%',
            },
            '& .MuiAlert-body': {
              padding: theme.spacing(2),
            },
          }),
        },
        {
          // styles for LOW priority
          props: { variant: 'withHeader', color: 'low' },
          style: theme.unstable_sx({
            borderColor: theme.palette.alerts.low.header,
            backgroundColor: theme.palette.alerts.low.background,
            '& .MuiAlertTitle-root': {
              backgroundColor: theme.palette.alerts.low.header,
              color: theme.palette.getContrastText(
                theme.palette.alerts.low.header
              ),
            },
          }),
        },
        {
          // styles for MEDIUM priority
          props: { variant: 'withHeader', color: 'medium' },
          style: theme.unstable_sx({
            borderColor: theme.palette.alerts.medium.header,
            backgroundColor: theme.palette.alerts.medium.background,
            '& .MuiAlertTitle-root': {
              backgroundColor: theme.palette.alerts.medium.header,
              color: theme.palette.getContrastText(
                theme.palette.alerts.medium.header
              ),
            },
          }),
        },
      ],
```

gigxz: By the way, the `theme` variable referenced here is defined on line `121`. It is the default MUi theme overlayed with any customer-specific values from the backend. The end of the file shows how they are composed

src/modules/client/components/clientAlerts/ClientAlert.tsx
```diff
@@ -0,0 +1,52 @@
...
+};
+
+interface ClientAlertProps {
+  clientAlert: ClientAlertType;
+}
+const ClientAlert: React.FC<ClientAlertProps> = ({ clientAlert }) => {
+  const alert = clientAlert.alert;
+  const clientName = clientAlert.clientName;
+  const showClientName = clientAlert.showClientName;
+  const priority = alert.priority || 'low';
```
gigxz:
```suggestion
  const { alert, clientName, showClientName, priority } = clientAlert;
```

There is no need to set a default for the `priority` since it's non-nullable from graphql

src/modules/client/components/clientAlerts/ClientAlert.tsx
```diff
@@ -0,0 +1,52 @@
...
+      <Box className='MuiAlert-body'>
+        {showClientName && (
+          <Typography variant='body2' sx={{ pb: 1 }}>
+            {clientName}
+          </Typography>
+        )}
+        <Typography variant='body1' sx={{ pb: 1 }}>
+          {alert.note}
+        </Typography>
+        <Typography variant='body2' sx={{ fontSize: '12px' }}>
```
gigxz: Setting a font-size manually is a code smell - we should usually be able to match existing typography, or else something is off with the designs.

It was hidden away in figma, but I can tell this is a `caption` variant
Screenshot 2024-02-02 at 9 30 19 AM



```suggestion
        <Typography variant='caption'>
```

src/modules/client/components/clientAlerts/ClientAlertCard.tsx
```diff
@@ -0,0 +1,56 @@
...
+          <Box
+            sx={{
+              p: 2,
+              backgroundColor: '#fafafa',
+              color: '#00000060',
+              borderRadius: 1,
+              display: 'flex',
+              justifyContent: 'space-between',
+            }}
+          >
```
gigxz: Using a box (div) is totally fine, but I usually use `Stack` for something like this: https://mui.com/material-ui/react-stack/


Here's my suggestion
```
          <Stack
            direction='row'
            justifyContent='space-between'
            alignItems='center'
            p={1}
            sx={{
              backgroundColor: (theme) => theme.palette.grey[50],
              color: 'text.secondary',
              borderRadius: 1,
            }}
          >
```

The Grey 50 and secondary text come from figma:

Screenshot 2024-02-02 at 9 42 00 AM

src/modules/client/components/clientAlerts/ClientAlert.tsx
```diff
@@ -0,0 +1,52 @@
+import { Alert, AlertTitle, Box, Typography } from '@mui/material';
```
gigxz, line 1: (Not related to this line)

I remembered I did add a hack for fixing incorrect ordering for the generated dropdowns.
Try adding `ClientAlertPriorityLevel: ['low'],` to [this object](https://github.com/greenriver/hmis-frontend/blob/e7fbbaae3ecad78b026d93ef61dc3f00844222bc/scripts/generateEnumDescriptions.js#L20-L23) and re-running codegen. That should re-order it correctly!

src/modules/household/components/elements/AddToHouseholdButton.tsx
```diff
@@ -76,8 +78,9 @@ const AddToHouseholdButton = ({
         </Button>
       </ButtonTooltipContainer>
       {renderFormDialog({
-        title: <>Enroll {clientName}</>,
+        title: <>Enroll {clientBriefName(client)}</>,
         submitButtonText: `Enroll`,
+        preFormComponent: <ClientAlertEnrollmentWrapper client={client} />,
```
gigxz: Do we really need this wrapper component? There are a lot of small components here and I wonder if some of them can be consolidated. The logic is already encapsulated in the hook, so we might as well just call that hook in this component, I think.

Eg:

```
const { clientAlerts } = useClientAlerts({ client }, false);

...

{renderFormDialog({
        title: <>Enroll {clientName}</>,
        title: <>Enroll {clientBriefName(client)}</>,
        submitButtonText: `Enroll`,
        preFormComponent: {clientAlerts.length > 0 && <ClientAlertStack clientAlerts={clientAlerts} />},

```

src/modules/client/hooks/useClientAlerts.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+    const clientAlerts = clients.flatMap((c) =>
+      c.alerts.map((a) => {
+        return {
+          alert: a,
+          clientName: clientBriefName(c),
+          showClientName: showClientName,
+        };
+      })
+    );
```
gigxz: I think this looks good! Would prefer multi-char variable names though:

```suggestion
    const clientAlerts = clients.flatMap((client) =>
      client.alerts.map((alert) => ({
        alert,
        clientName: clientBriefName(client),
        showClientName,
      }))
    );
```

gigxz (commented):
Awesome @martha!! I left a couple small comments but lgtm after that 👍 💯

src/config/theme.ts
martha: Thanks @gigxz this is great, I agree it's a lot cleaner!

src/modules/client/components/clientAlerts/ClientAlertCard.tsx
martha: Nice :) thank you

gigxz (approved):
👏

PR #623, opened 2024-01-29
186786001 spdat

ttoomey:
## Description

* support helper text/html on choice options
* support numeric value / score on choice options
* support simple interpolation in display component

[support assessment score
](https://www.pivotaltracker.com/story/show/186786000)
[support spdat
](https://www.pivotaltracker.com/story/show/186786001)

](https://github.com/greenriver/hmis-warehouse/pull/3954)
## Type of change


src/components/elements/input/RadioGroupInput.tsx
```diff
@@ -145,34 +146,40 @@ const RadioGroupInput = ({
...
+          {options.map(({ code, label, helperText }) => (
+            <Fragment key={code}>
+              <FormControlLabel
+                data-testid={`option-${code}`}
+                disabled={props.disabled}
+                value={code}
+                aria-label={label || code}
+                onClick={(e) => onClickOption(e, code)}
+                control={
+                  <ControlComponent
```
gigxz: For accessibility this relies on being rendered inside the label. But we should probably link up the helper text to it too. I think something like`aria-labelledby="main-label-element helper-text-element>"` here?

src/modules/form/util/formUtil.ts
```diff
@@ -453,6 +454,22 @@ export const shouldEnableItem = ({
   }
 };
 
+const isPicklistOption = (
```
gigxz: We already have a type guard function called isPickListOption -- did you miss that, or did you need a separate one for the fragment?

gigxz (commented):
Looks good! a couple small comments

src/components/elements/input/RadioGroupInput.tsx
ttoomey: Good suggestion, I did a little work to improve accessibility of the input component; I also extracted the option out into its own component.
* removed rid of the enclosing MUI Form Group (it didn't seem to be needed)
* add fieldset and legend elements for screen readers
* add aria-labelledby for both label and helperText. (I suspect aria-describedby would be more accurate but I couldn't get that work with voiceover)

src/modules/form/util/formUtil.ts
ttoomey: ah I missed that, thanks

gigxz (approved):
Looks good to me, thanks!

PR #628, opened 2024-02-01
Enable deletion of ClientAlerts

martha:
## Description

This PR adds a deletion button to Client Alerts in some contexts (only on Profile and Enrollment Dashboard pages, not in the Enrollment modal).

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/3972

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/config/theme.ts
```diff
@@ -341,6 +345,9 @@ const createThemeOptions = (theme: Theme) => ({
                 theme.palette.alerts.low.header
               ),
             },
+            '& .MuiAlert-deleteIcon': {
+              color: theme.palette.alerts.low.icon || 'rgba(0, 0, 0, 0.54)',
```
gigxz: Is this needed? I don't see any difference when disabling this, it looks like this is the mui default anyway? If it's just slightly different, I would prefer removing this to keep the theme simpler.

src/modules/dataFetching/components/DeleteMutationButton.tsx
```diff
@@ -28,6 +28,8 @@ interface DeleteMutationButtonProps<MutationVariables> {
   confirmationDialogContent?: ReactNode;
   verb?: string;
   deleteIcon?: boolean;
+  onlyIcon?: boolean;
```
gigxz, line 31: The `children` prop should be made optional, since it's ignored in this case.

src/modules/household/components/elements/AddToHouseholdButton.tsx
```diff
@@ -62,7 +62,7 @@ const AddToHouseholdButton = ({
   const { openFormDialog, renderFormDialog } =
     useFormDialog<SubmittedEnrollmentResultFieldsFragment>(memoedArgs);
 
-  const { clientAlerts } = useClientAlerts({ client }, false);
+  const { clientAlerts } = useClientAlerts({ client }, false, false);
```
gigxz: This is OK but for improved readability I would change it to accept an object:
`useClientAlerts({ client, householdId,  showClientName, showDeleteButton })`
so this becomes `useClientAlerts({ client, showClientName: false, showDeleteButton: false })` (unless false are the default?)

src/config/theme.ts
martha: The `'rgba(0, 0, 0, 0.54)'` color here is indeed the mui default. I added the icon color as an option on the theme in order to get the effect specified by Ash in figma ([here](https://www.figma.com/file/F9EHR2IZYuMAlOJ9aOG2IL/OP-%2F-HMIS-%2F-Phase-2-(2024)?type=design&node-id=622%3A2869&mode=design&t=ihkylfJD9cxM7VpK-1)) where the high priority icon is a different color from the others, for better contrast against the red background.

I think we could do away with it if we use `getContrastColor` (but then the icons would be black/white instead of this nice grey), or we could go back to using a paler red for high priority alerts (we switched to this one in order to be consistent with `theme.palette.error.main`).

src/modules/household/components/elements/AddToHouseholdButton.tsx
martha: False is the default for both, but I agree with you that it's clearer if these are in an object.
There's already an interface defined `ClientAlertParams` that I just added the two booleans onto, let me know if you like that approach? I also removed them when they are false -- so the default is just used -- although, I don't know if this technicality matters, but they will actually be `undefined` instead of `false` if not provided because I don't think we can give default values on a typescript interface.

src/config/theme.ts
gigxz: Ah gotcha, thanks. 

So these class overrides are not needed for `low` and `medium` colors, so we can remove those from the theme.

For the high priority, I would just do specify a rule here saying that high-priority alerts should use this color for icon buttons inside the dark-colored header:
```
            '.MuiAlertTitle-root .MuiIconButton-root': {
              color: theme.palette.alerts.high.icon,
            },
```
that way we don't need to be passing `className`s around. Does that make sense?

src/modules/household/components/elements/AddToHouseholdButton.tsx
gigxz: Yep. That should be fine

src/config/theme.ts
martha: Yes that's much better!

gigxz (approved):
looks good!

PR #629, opened 2024-02-02
Add user edit history audit page

martha:
## Description

We already can audit users' views of clients and enrollments; this adds the ability to view users' edit events too.
https://www.pivotaltracker.com/n/projects/2591838/stories/186570782

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/3976

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/admin/components/users/ClientAccessSummaryTable.tsx
```diff
@@ -59,7 +59,7 @@ const ClientAccessSummaryTable: React.FC<Props> = ({
       pagePath='user.clientAccessSummaries'
       noData='No access history'
       filterInputType='ClientAccessSummaryFilterOptions'
-      recordType='clients accessed'
+      recordType='accessed client'
```
martha: I updated this from 'clients accessed' to 'accessed client' (and the below, 'enrollments accessed' to 'accessed enrollment') because I noticed that when no records are returned, such as when the filter is very narrow, the pluralization happening [here](https://github.com/greenriver/hmis-frontend/blob/83d419411011e56b2941a80304a1c00787e12231/src/modules/dataFetching/components/GenericTableWithData.tsx#L337-L339) was causing a weird typo, "clients accesseds".

Let me know if you think this change will cause any other unforeseen consequences though, as I can see that `recordType` is used for more things than just text -- such as sort and filter options. I couldn't find usages of these strings anywhere else in the codebase, so I thought it would probably be ok.

src/modules/admin/components/users/UserAuditHistory.tsx
```diff
@@ -0,0 +1,113 @@
...
+import RelativeDateTableCellContents from '@/modules/hmis/components/RelativeDateTableCellContents';
+import { auditActionForDisplay } from '@/modules/hmis/hmisUtil';
+import {
+  UserAuditEventFilterOptions,
+  GetUserAuditEventsDocument,
+  GetUserAuditEventsQuery,
+  GetUserAuditEventsQueryVariables,
+} from '@/types/gqlTypes';
+
+type AuditHistoryType = NonNullable<
```
martha, line 16: There's a lot of duplicated code here across `ClientAuditHistory`, `EnrollmentAuditHistory`, and  now `UserAuditHistory`. It can't be fully factored out because of typescript -- they each use a different `Get<Something>AuditEventsQuery` in the `AuditHistoryType` -- but maybe it could be factored out in smaller ways, such as centrally defining the contents of the items in `columns`?

src/modules/admin/components/users/UserAuditPage.tsx
```diff
@@ -15,14 +17,28 @@ import useDebouncedState from '@/hooks/useDebouncedState';
...
 import ClientAccessSummaryTable from '@/modules/admin/components/users/ClientAccessSummaryTable';
 import EnrollmentAccessSummaryTable from '@/modules/admin/components/users/EnrollmentAccessSummaryTable';
+import UserAuditHistory from '@/modules/admin/components/users/UserAuditHistory';
 import { useUser } from '@/modules/dataFetching/hooks/useUser';
 import ClientTextSearchInput from '@/modules/search/components/ClientTextSearchInput';
 import CommonSearchInput from '@/modules/search/components/CommonSearchInput';
 import { AdminDashboardRoutes } from '@/routes/routes';
 
-type EntityType = 'clients' | 'enrollments';
+type PageEntityType = 'access' | 'edits';
```
martha: I'm not happy with this name `PageEntityType`, it would be much better to change it to something clearer, let me know if you have a good suggestion! I will keep thinking on it too

src/modules/admin/components/users/UserAuditHistory.tsx
martha: Ok, I took a stab at reducing repetition but with a couple open Q's left in todos in the `auditHistoryColumnDefs.tsx` file.
1. maybe there would be a way to achieve this 'upstream', as it were, by using a smarter graphql fragment?
2. relatedly: the changes I've made still contain some repetitive code, it's just repetitive in a different way -- maybe this is better (because we can now easily change the code for any audit history table, including formatting/styling, without changing it in 3 places), but maybe it's actually worse and could lead to more confusion later on (because updates to the graphql schema for audit history types won't be reflected in the manually defined interface)? What do you think?

src/modules/admin/components/users/ClientAccessSummaryTable.tsx
gigxz: great, thank you. This should really be the `paginationItemName` prop, not `recordType`. I think if `paginationItemName` is missing we try to infer it from `recordType`, but the record type ha another purpose. That should probably be cleaned up at some point

src/modules/admin/components/users/UserAuditPage.tsx
```diff
@@ -73,64 +116,86 @@ const UserAuditPage: React.FC<Props> = ({ entityType }) => {
 
   return (
     <>
-      <PageTitle title={`Access History for ${user.name}`} />
+      <PageTitle title={`${user.name} Audit History`} />
+      <Alert severity='warning' sx={{ mb: 2 }}>
+        Data on this page may be delayed by up to an hour. Recent user activity
+        may not be immediately visible.
+      </Alert>
```
gigxz: This is only true for the User Access history, since it requires some periodic processing of activity logs. The User Edits show up right away. So can we move this alert down into the card?

src/modules/admin/components/users/UserAuditHistory.tsx
martha: With the more recent changes I'm especially unsure of this refactor -- please let me know if you think I should revert it or keep thinking about some other way the code could be more dry and less brittle!

martha:
See screenshot for how custom fields look. I tried editing a project, but have not yet seen those changes reflected, so I need to investigate why that is. 
Screenshot 2024-02-06 at 1 35 45 PM

martha:
.. Never mind, in fact auditing project edits seems to be working fine -- see screenshot -- although I noticed that when I create a new organization, that does not show up among my user edits. I think this is expected based on the code we have now, but is it good?
Screenshot 2024-02-06 at 2 57 03 PM

src/modules/admin/components/users/UserAuditPage.tsx
gigxz: Maybe `UserAuditType`?

gigxz: Or `UserHistoryType`

src/modules/admin/components/users/UserAuditPage.tsx
```diff
@@ -36,9 +52,13 @@ const toggleItems: ToggleItem<EntityType>[] = [
 ];
 
 interface Props {
-  entityType: EntityType;
+  pageEntityType: PageEntityType;
+  accessEntityType?: AccessEntityType;
 }
-const UserAuditPage: React.FC<Props> = ({ entityType }) => {
+const UserAuditPage: React.FC<Props> = ({
```
gigxz, line 38: This component is feeling pretty bloated, as it is responsible for the state of both toggles.

I would recommend creating a new component `UserAccessHistory` that renders this (red) part of the page
Screenshot 2024-02-07 at 3 40 11 PM


Then you would have a separation of concerns like this:
* UserAuditPage: renders the page title and toggle, owns the Audit History Type toggle state
* UserAuditHistory: renders "User Edits" card content
* UserAccessHistory: renders "User Access" card content. Owns the Client-vs-Enrollment toggle state.

src/modules/admin/components/users/UserAuditHistory.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const { userId } = useSafeParams() as { userId: string };
+  // On the user audit history page, we don't need to show the user column, since the value will always be the same
+  const columns = auditHistoryColumns.filter((column) => column.key !== 'user');
+
+  return (
+    <>
+      <Alert severity='warning' sx={{ m: 3 }}>
+        Data on this page may be delayed by up to an hour. Recent user activity
+        may not be immediately visible.
+      </Alert>
```
gigxz: This alert is actually supposed to go on the other card ("user access"). User edit history should be available immediately.

src/modules/audit/components/auditHistoryColumnDefs.tsx
```diff
@@ -0,0 +1,161 @@
...
+    name: string;
+  } | null;
+  clientId?: string;
+  clientName?: string;
+  projectId?: string;
+  projectName?: string;
+  enrollmentId?: string;
+}>;
+
+export const auditHistoryColumns: ColumnDef<AuditHistoryNode>[] = [
```
gigxz: Nice, thanks for pulling this out into its own file. 
I think the most convenient way I have found for managing the columns definitions so far is to specify shared columns [in an object](https://github.com/greenriver/hmis-frontend/blob/ea9fbca1ec3893fbed44cd8eee5c3f4f644ccb07/src/modules/projects/components/tables/ProjectClientEnrollmentsTable.tsx#L56).

When done like that, there are 2 benefits:
1. different tables can easily pick and choose which columns they want and in which order. they can also insert their own "custom " columns in whichever location they want. [like this!](https://github.com/greenriver/hmis-frontend/blob/ea9fbca1ec3893fbed44cd8eee5c3f4f644ccb07/src/modules/bedNights/components/ProjectEnrollmentsTableForBedNights.tsx#L58-L82)
2. we can use a union type to support different fragments when there are slight variations

Your union type would be: 
```
export type AnyAuditEvent =
  | ClientAuditEventFieldsFragment
  | EnrollmentAuditEventFieldsFragment
  | UserAuditEventFieldsFragment;
```
I think that should make typescript happy. The Client and Project column definitions are specific to the User Audit page, at least for now.

src/modules/audit/components/auditHistoryColumnDefs.tsx
```diff
@@ -0,0 +1,161 @@
...
+      if (projectId) {
+        return (
+          <Link
+            href={generateSafePath(Routes.PROJECT, { projectId: projectId })}
+          >
+            {projectName}
+          </Link>
+        );
+      }
+      return projectName; // Should never get here, but just in case
```
gigxz: lets remove as its definitely dead code

src/modules/audit/components/auditHistoryColumnDefs.tsx
```diff
@@ -0,0 +1,161 @@
...
+    width: '180px',
+    render: ({ projectName, clientId, enrollmentId, projectId }) => {
+      if (!projectName) return;
+      if (clientId && enrollmentId) {
+        // Link to the enrollment
+        return (
+          <Link
+            href={generateSafePath(Routes.ENROLLMENT_DASHBOARD, {
+              enrollmentId: enrollmentId,
+              clientId: clientId,
```
gigxz: nit:
```suggestion
              enrollmentId,
              clientId,
```

src/modules/audit/components/auditHistoryColumnDefs.tsx
```diff
@@ -0,0 +1,161 @@
...
+    key: 'projectName',
+    header: 'Project Name',
+    width: '180px',
+    render: ({ projectName, clientId, enrollmentId, projectId }) => {
+      if (!projectName) return;
+      if (clientId && enrollmentId) {
+        // Link to the enrollment
+        return (
+          <Link
+            href={generateSafePath(Routes.ENROLLMENT_DASHBOARD, {
```
gigxz: We always use RouterLink for internal links! that uses the react-router-dom component
```suggestion
          <RouterLink
            to={generateSafePath(Routes.ENROLLMENT_DASHBOARD, {
```

src/api/operations/audit.fragments.graphql
```diff
@@ -33,3 +33,26 @@ fragment EnrollmentAuditEventFields on EnrollmentAuditEvent {
...
+  projectId
+  projectName
+  user {
+    id
+    name
+  }
+  trueUser {
+    id
+    name
+  }
```
gigxz: Technically don't need to resolve these here, right?

src/modules/admin/components/users/UserAuditPage.tsx
martha: Oh good thinking. Thanks 🙏

src/modules/admin/components/users/UserAuditHistory.tsx
martha: Sorry I missed this - thanks!

src/modules/audit/components/auditHistoryColumnDefs.tsx
martha: Thanks for this suggestion. I think it's much better

gigxz (approved):
looks good!

PR #630, opened 2024-02-05
Make project dashboard default to enrollments

martha:
## Description

If the user has permission to view enrollments on a project, then navigating to the project should default to bringing up the enrollments page. Otherwise, if they don't have permission, then the current behavior (bringing up the overview page) should be maintained.
https://www.pivotaltracker.com/n/projects/2591838/stories/186877672

### Testing instructions
Here's the list that I used to help me test the different cases- I'll paste it here in case it's helpful to others testing and so you can make sure I haven't missed any:
- [ ] Clicking a project from the projects search
  - [ ] With permission to view enrollments on that project
  - [ ] Without permission to view enrollments on that project
- [ ] Clicking a project from an organization’s list of projects
  - [ ] With permission to view enrollments on that project
  - [ ] Without permission to view enrollments on that project
- [ ] Clicking a project from an enrollment
  - [ ] With permission to view enrollments on that project
  ~Without permission: I believe this cannot happen; if you're already viewing an enrollment on the project then you must have permission to view enrollments on that project~

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/components/accessWrappers/ProjectRoute.tsx
```diff
@@ -0,0 +1,44 @@
...
+import Loading from '../elements/Loading';
+import NotFound from '../pages/NotFound';
+
+import { ProjectPermissions } from '@/modules/permissions/types';
+import { useHasProjectPermissions } from '@/modules/permissions/useHasPermissionsHooks';
+import { useProjectDashboardContext } from '@/modules/projects/components/ProjectDashboard';
+import { ensureArray } from '@/utils/arrays';
+import { generateSafePath } from '@/utils/pathEncoding';
+
+/**
```
martha, line 12: This file is copied and only lightly modified from `EnrollmentProjectRoute.tsx`. It's also very similar to a few of the other files in that `accessWrappers` directory. Let me know if you would like me to dedicate some time to reducing repeated code here?

src/routes/protected.tsx
```diff
@@ -131,7 +132,17 @@ export const protectedRoutes: RouteNode[] = [
         path: Routes.PROJECT,
         element: <ProjectDashboard />,
         children: [
-          { path: '', element: <Navigate to='overview' replace /> },
+          {
+            path: '',
+            element: (
+              <ProjectRoute
+                permissions={['canViewEnrollmentDetails']}
```
martha, line 139: I had more questions about which is the correct permission to use here. I gathered a list of all the permissions that I thought might be relevant:

- Permissions on the project (`ProjectAccessFieldsFragment`)
  - `canEditEnrollments`
  - `canViewEnrollmentDetails`
  - (No such thing as simply `canViewEnrollments`)
- Other possibly relevant permissions on the HMIS admin dashboard
  - `canViewLimitedEnrollmentDetails` - this does not seem to be attached to a project, but if someone has this permission overall, should they be directed to a project enrollment page, or to the overview?
  - `canViewOpenEnrollmentSummary` which I think is associated with the Root permissions - not associated with a particular project

gigxz: This is correct, `canViewEnrollmentDetails` is needed to view the Project Enrollments page. `canViewLimitedEnrollmentDetails` would not give someone access to view the full enrollments, it only lets you see limited enrollment details on the enrollment tab on the client dashboard.

gigxz (approved):

src/components/accessWrappers/ProjectRoute.tsx
gigxz: looks good to me

PR #639, opened 2024-02-06
Add phone input field that accepts only one number

martha:
## Description

The normal phone input component (currently in use on the default New Client form) enables adding multiple phone numbers, each with their own type (home, cell, etc.) and note.

For emergency contacts, I think we don't want to be able to add multiple phone numbers, but we do want a phone-validating input component, so it's useful to reuse this Phone component.

**Open Question**
Is it desirable to be able to input the phone type and any notes (see screenshot below)? This is built-in behavior with the Phone component, but we can also disable it.
Screenshot 2024-02-06 at 4 10 24 PM

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/3987

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/modules/form/components/DynamicField.tsx
+++ b/src/modules/form/components/DynamicField.tsx
@@ -31,6 +31,7 @@ import DatePicker from '@/components/elements/input/DatePicker';
 import LabeledCheckbox from '@/components/elements/input/LabeledCheckbox';
 import NoYesMissingCheckbox from '@/components/elements/input/NoYesMissingCheckbox';
 import NumberInput from '@/components/elements/input/NumberInput';
+import PhoneInput from '@/components/elements/input/PhoneInput';
 import RadioGroupInput from '@/components/elements/input/RadioGroupInput';
 import SsnInput from '@/components/elements/input/SsnInput';
 import TextInput from '@/components/elements/input/TextInput';
@@ -38,7 +39,6 @@ import YesNoRadio from '@/components/elements/input/YesNoRadio';
 import Uploader from '@/components/elements/upload/UploaderBase';
 import MciClearance from '@/modules/external/mci/components/MciClearance';
 import SimpleAddressInput from '@/modules/form/components/client/addresses/SimpleAddressInput';
-import PhoneInput from '@/modules/form/components/client/phones/PhoneInputGroup';
 import { INVALID_ENUM, parseHmisDateString } from '@/modules/hmis/hmisUtil';
 import { Component, FormItem, InputSize, ItemType } from '@/types/gqlTypes';
 
@@ -238,7 +238,12 @@ const DynamicField: React.FC<DynamicFieldProps> = ({
           <PhoneInput
             id={linkId}
             value={value || ''}
-            onChange={onChangeValue}
+            onChange={onChangeEvent}
+            sx={{
+              width,
+              maxWidth: MAX_INPUT_AND_LABEL_WIDTH,
+              '.MuiInputBase-root': { maxWidth: MAX_INPUT_WIDTH },
+            }}
             {...commonInputProps}
           />
         );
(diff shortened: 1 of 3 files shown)


src/modules/form/components/DynamicField.tsx
```diff
@@ -232,6 +233,17 @@ const DynamicField: React.FC<DynamicFieldProps> = ({
...
+      if (item.component === Component.Phone) {
+        return (
+          <PhoneInput
+            id={linkId}
+            value={value || ''}
+            onChange={onChangeValue}
+            {...commonInputProps}
+          />
+        );
+      }
```
gigxz: This is going to override the behavior on line `439` which we don't want.
The component naming is unfortunate, `PHONE` really renders 3 fields (number, type, note) which is more than what is needed here.

Can you add a new component type, maybe called `PHONE_NUMBER`, that just renders the `PhoneInput` (not the `PhoneInputGroup`?

martha: Sounds good!

I think it doesn't override the behavior on line 439, because that's inside of a `case ItemType.Object:` block, while this is inside of a `case ItemType.String:` block. Am I right to think that means they don't conflict?

Either way, I could still see that it would be desirable to have another way to differentiate these two. I'm happy to make a new `PHONE_NUMBER` component if you think that is the clearest way.

src/modules/form/components/client/phones/MultiPhoneInput.tsx
```diff
@@ -7,7 +7,7 @@ import { PhoneInputType } from '../types';
 import { useRenderLastUpdated } from '../useRenderLastUpdated';
 import { createInitialValue } from '../util';
 
-import PhoneInput from './PhoneInputGroup';
+import PhoneInputGroup from './PhoneInputGroup';
```
martha, line 10: I'm not sure if this is what had happened here, but I noticed that when I asked my IDE to auto-import `PhoneInput`, it imported it from `PhoneInputGroup` but renamed as `PhoneInput` -- which led to my confusion of thinking that the phone type dropdown and notes text field are built-in to the `PhoneInput` component, when they are actually not, they are only built-in to the `PhoneInputGroup` component.

Anyway, when I searched the codebase for usages of `PhoneInput` I noticed the same import-rename shenanigans going on here, so I replaced it for more clarity -- let me know if you think this is an ok change to make.

src/modules/form/components/client/emails/EmailInput.tsx
```diff
@@ -1,16 +1,16 @@
 import { Grid, Stack } from '@mui/material';
 
 import { getRequiredLabel } from '../../RequiredLabel';
-import { PhoneInputType } from '../types';
+import { EmailInputType } from '../types';
 
 import TextInput from '@/components/elements/input/TextInput';
 
-const PhoneInput = ({
+const EmailInput = ({
```
martha, line 8: See my comment below for more backstory, but when I was searching the codebase for usages of `PhoneInput`, I noticed this and thought it was probably a typo from an incomplete copy/paste?

gigxz: yep  thanks for fixing!

src/modules/form/components/client/phones/MultiPhoneInput.tsx
gigxz: thanks! yeah thats great

src/modules/form/components/DynamicField.tsx
martha: With the import statement updated to pull in a real `PhoneInput` component and not a `PhoneInputGroup` deceptively renamed (see my comment [here](https://github.com/greenriver/hmis-frontend/pull/639/files#r1481842844)), this is now working correctly, aka just showing an input text box with phone validation rather than the extra dropdown and notes field.

The MultiPhone input component also still works as expected, which I think confirms that this doesn't conflict with line 439. That means the current behavior is as follows:
- if type is STRING and component is PHONE, then show a single phone input text box
- if type is OBJECT and component is PHONE, then show a multi-phone input where each phone number also has a type and notes box.

I am still happy to make a new component type to differentiate these more clearly, but I think it may not be necessary. Let me know your preference!

gigxz: Gotcha, yeah I think that makes sense. Lets merge as is

gigxz (approved):

PR #644, opened 2024-02-08
Add a signature component to the Dynamic Form

martha:
## Description

This PR adds a frontend component for e-signatures in dynamic form fields.
Let's discuss the whole feature more as I still have a couple of open questions:
- what will the context for this signature component be?
- how should the data be stored after the signatures are submitted?
- testing out correct error behavior - I need to make better use of `commonContainerProps` and `commonInputProps`

but, anyway, here's a screenshot of how it looks now!
Screenshot 2024-02-08 at 4 36 44 PM

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/3997

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/form/types.ts
```diff
@@ -82,6 +82,7 @@ export interface DynamicInputCommonProps {
   placeholder?: string;
   breakpoints?: BreakpointOverrides;
   ariaLabelledBy?: string;
+  themeColor?: string;
```
martha: Maybe this is an abuse of `DynamicInputCommonProps`, since it's not really common to anything else? This is to allow one signature component to have a blue color and the other to have purple (or whatever the installation's theme primary and secondary colors are), see lines 26-28 of `SignatureEnvelope.tsx`

src/modules/form/components/DynamicField.tsx
```diff
@@ -233,6 +234,24 @@ const DynamicField: React.FC<DynamicFieldProps> = ({
...
+          checked: false,
+        };
+        return (
+          <Signature
+            id={linkId}
+            value={value || initialValue}
+            onChange={onChangeValue}
+            whoseSignature={item.text || 'Signature'}
+            {...inputProps}
+            {...commonInputProps}
```
martha: Many of the other components returned in this big switch statement pass in `sx` having to do with width; I didn't insert it here because I couldn't see if it was making a difference, but I didn't test on a ton of different screen widths; should it be added here too?

src/modules/form/components/group/variants.tsx
```diff
@@ -0,0 +1,27 @@
+import { Draw, SvgIconComponent } from '@mui/icons-material';
```
ttoomey: suggest adding a new entry for signatures to `src/components/elements/SemanticIcons.tsx` - then this would become:
```js
import { SignatureIcon } from '@/components/elements/SemanticIcons';
```

src/modules/form/components/group/FormCard.tsx
```diff
@@ -88,6 +93,11 @@ const FormCard: React.FC<Props> = ({
     size: 'small',
     sx: { height: 'fit-content' },
   };
+
+  // todo @martha this is a bit messy
+  let variantStyles: FormVariantStylesProps = {};
+  if (item.variant) variantStyles = getFormGroupVariantStyles(item.variant);
```
ttoomey: maybe cleaner with a ternary?
```ts
const variantStyles: FormVariantStylesProps =  item.variant ? getFormGroupVariantStyles(item.variant) : {}
```

src/modules/form/components/group/variants.tsx
```diff
@@ -0,0 +1,27 @@
...
+    case Variant.Signature:
+      return {
+        icon: Draw,
+        titleProps: { color: 'primary', variant: 'h3', fontWeight: 600 },
+        subtitleProps: { variant: 'body1' },
+        themeColorFn: (i: number) =>
+          i % 2 === 0
+            ? theme.palette.primary.main
+            : theme.palette.secondary.main,
+      };
```
ttoomey: perhaps it would be cleaner to do this with [css nth-child selector](https://developer.mozilla.org/en-US/docs/Web/CSS/:nth-child)?

src/modules/form/components/group/FormCard.tsx
```diff
@@ -162,7 +183,28 @@ const FormCard: React.FC<Props> = ({
           }}
         >
           {renderChildItem &&
-            item.item?.map((childItem) => renderChildItem(childItem))}
```
ttoomey, line 165: Rather than wrapping renderChildItem in an additional Box in FormCard, I wonder if it would be easier and more flexible to add another variant to the form definition (say 'ACCENT' or 'HIGHLIGHT') for these inner group components. Then QuestionGroup could handle rendering the appropriate border, much as it does for `isConditionalGroup`. I think that gets us out of passing around groupHeaderProps around too. It could also allow you to have children of a SIGNATURE group that don't have this accent treatment.

Happy to chat about this approach is it would be helpful.

ttoomey (commented):
looking good!

src/modules/form/components/group/FormCard.tsx
martha: Good plan! Thank you. I think the component is actually not a `QuestionGroup` but a `HorizontalGroup` so I added it there, but let me know if you think I should add handling of these variant styles to the `QuestionGroup` as well?

src/modules/form/components/group/FormCard.tsx
```diff
@@ -140,6 +154,12 @@ const FormCard: React.FC<Props> = ({
           </Stack>
         )}
 
+        {item.helperText && (
```
martha: I'd like to ask for feedback/pointers about accessibility for the signature text inputs. On Friday we had discussed using `fieldset` and `legend`, but I'm a bit unsure about how to fit this in with the generic `TextInput` component. I thought about using `helperText` for the aria label, but `helperText` is already in use for the grey label beneath the text input box, like this:
Screenshot 2024-02-13 at 10 21 31 AM

What do you suggest? Should we add a new form field attribute `accessibilityText`? or re-use `helperText`? or something else?

ttoomey: Good question,  I'm didn't mean to suggest that the helperText on the group should be used on the input components themselves, sorry for the confusion. I was imagining that a fieldset would be a good way to improve accessibility of the inner group component using something like the following structure:
```jsx
<fieldset>
  <legend>Head of household</legend>
  <TextInput label="Full Name">
  <TextInput label="Today's date">
</fieldset>
```
So the group component would render fieldset/legend elements instead of divs/h3s. A screen reader should then be able to describe that the focused "Full Name" is for head of house household and distinguish it from the case manager full-name below.

I don't know if helper text within a card needs special treatment itself. A simple p tag might be enough; could test it it in a screen reader and add aria-describedby if needed. Looking at our current current implementation, the encapsulating element for these cards probably should be a `<section/>` rather than a div. This might be scope creep though...

Happy to discuss on slack

martha: Thanks! Let me know if this is along the lines of what you had in mind?

ttoomey (approved):
Great work, thanks!! 🙌

src/modules/form/components/group/FormCard.tsx
ttoomey: Yes, this looks great to me, thank you!!

PR #645, opened 2024-02-12
Fix bug that caused filter disappearance

martha:
## Description

Mary Ann pointed out when testing the User Audit feature in QA that it lacked filter capabilities. I also noticed that after my changes, the filter capability disappeared from the client and enrollment access summary tables! 🐞 

It turned out to be because all three of these tables require a `recordType` prop in order to render filtering, so this PR addresses that problem. The Client and Enrollment access tabs are restored to their former filter functionality. The User Audit (edit history) filter function looks like this: 

Screenshot 2024-02-12 at 10 01 13 AM

Question: is adding an "on or after date" part of the filtering requirements? Any other filter requirements to meet?

PT issue: https://www.pivotaltracker.com/n/projects/2591838/stories/186570782/comments/240192255

## Type of change
- [x] Bug fix
- [ ] Dependency update


gigxz (approved):
good find, thanks!

PR #649, opened 2024-02-14
Improve mobile usability for critical paths

martha:
## Description

This PR improves mobile usability for things that street outreach teams are likely to need.
Figma: https://www.figma.com/file/F9EHR2IZYuMAlOJ9aOG2IL/OP-%2F-HMIS-%2F-Phase-2-(2024)?type=design&node-id=520-773&mode=design&t=HY9uhOq646rtsKII-0
PT: https://www.pivotaltracker.com/n/projects/2591838/stories/184151814

## Type of change
- [x] Bug fix
- [ ] Dependency update


src/components/elements/input/TextInput.tsx
```diff
@@ -24,9 +24,7 @@ interface Props extends Partial<Omit<TextFieldProps, 'error' | 'variant'>> {
 export type TextInputProps = Props & DynamicInputCommonProps;
 
 export const horizontalInputSx: SxProps<Theme> = {
-  flexDirection: 'row',
   justifyContent: 'space-between',
```
martha, line 28: - I removed `flexDirection: 'row'` here because it's overridden by the added `direction={{ xs: 'column', sm: 'row' }}` on line 122.
- I'm less sure about removing `alignItems: 'center'`; on the Income input component (the one I was focusing on here since it's used for the Intake assessment) it looks like this directive wasn't taking effect until the flex direction switched to `column` for the mobile screen widths, and we don't want it there either, so I just removed it. But maybe this overrides desired styles elsewhere?

src/modules/admin/components/users/UserAccessHistory.tsx
```diff
@@ -88,7 +88,6 @@ const UserAccessHistory = ({ accessEntityType }: Props) => {
             label='Search client access'
             value={search}
             onChange={setSearch}
-            helperText={null}
```
martha: Removed this helperText from `ClientTextSearchInput` and moved it one level up into `ClientTextSearchForm` because of this conversation: https://www.figma.com/file/F9EHR2IZYuMAlOJ9aOG2IL?node-id=1650:25045&mode=design#700990401

src/modules/assessments/components/AssessmentAutofillButton.tsx
```diff
@@ -0,0 +1,21 @@
+import { Button } from '@mui/material';
```
martha, line 1: Nothing new in this file, the button is just factored out into its own component so it can be used by both the `AssessmentForm` (where it's shown on mobile above the other input fields), and the `AssessmentFormSideBar` (which is only shown on desktop).

src/modules/enrollment/components/dashboardPages/EnrollmentOverview.tsx
```diff
@@ -59,7 +59,7 @@ const EnrollmentOverview = () => {
     <>
       <PageTitle title='Enrollment Overview' />
       <Grid container spacing={4}>
-        <Grid item xs={7}>
+        <Grid item md={8} xs={12}>
```
martha, line 62: The grid item's width was previously 7 instead of 8 here, meaning that the two side-by-side uneven-width components didn't fill the whole 12 available columns. I switched it to 8 on the idea that this was probably a mistake, but let me know if there was actually an important reason for it!

src/components/elements/input/TextInput.tsx
gigxz: Hmm. I'm seeing that it makes a difference here. Are there situations where we don't want the label + input to be center-aligned when displayed horizontally? If thats the case, maybe this styling should be moved into the `InputGroup` (screenshotted below) which definitely wants to keep them centered.

Screenshot 2024-02-23 at 9 40 48 AM
Screenshot 2024-02-23 at 9 40 42 AM

gigxz: We also sometimes have labels wrapping like this, probably  worth looking at. (Demonstrated by editing `src/test/__mocks__/mockFormDefinition.json`, please feel free to make changes in that file if its helpful to show different variations in storybook!)
Screenshot 2024-02-23 at 9 47 20 AM

src/modules/assessments/components/AssessmentAutofillButton.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { SxProps } from '@mui/system';
+import ButtonTooltipContainer from '@/components/elements/ButtonTooltipContainer';
+
+export default function AssessmentAutofillButton(props: {
+  onClick: VoidFunction;
+  sx?: SxProps;
+}) {
```
gigxz: We use the below convention for components in the project. (Not because its better than other ways of declaring components, but we just want to keep it consistent).


```
const AssessmentAutofillButton: React.FC<Props> = (props) => ...


export default AssessmentAutofillButton
```

src/modules/enrollment/components/dashboardPages/EnrollmentOverview.tsx
gigxz: This looks good to me, thanks!

src/modules/form/components/group/DisabilityTable.tsx
```diff
@@ -141,18 +143,50 @@ const DisabilityTable = ({
...
+
+  function getDisabilityDropdowns(
+    cellItem: FormItem,
+    idx: number,
+    disabilityTypeLabelId: string
+  ) {
+    return renderChildItem(cellItem, {
+      noLabel: true,
+      inputProps: {
+        label: null,
```
gigxz: On mobile, it looks like we lose the label for the Disabling Condition questions (the second column in the table).
I think we need to adjust it to remove these` noLabel`/`label: null` for mobile, so that we have both labels on the inputs. 

For mobile, we'd also need to remove the typography that renders the leftmost column on desktop, or else we get double-labeling. Does that make sense?

src/modules/form/components/group/DisabilityTable.tsx
```diff
@@ -141,18 +143,50 @@ const DisabilityTable = ({
...
+        placeholder: idx === 0 ? 'Select Status' : 'Select Disabling Condition',
+        // Reads as "Physical Disability Status" or "Physical Disability Disabling Condition"
+        ariaLabelledBy: `${disabilityTypeLabelId} ${
```
gigxz: This shouldn't be included for mobile-view, if you take my suggestion above, since the inputs would be directly labeled.

src/modules/form/components/group/FormCard.tsx
```diff
@@ -111,7 +111,11 @@ const FormCard: React.FC<FormCardProps> = ({
         >
           {/* Card title */}
           {item.text && (
-            <Stack justifyContent='space-between' direction='row'>
+            <Stack
+              justifyContent='space-between'
+              direction={{ xs: 'column', sm: 'row' }}
+              sx={{ mb: 2 }}
```
gigxz: This looks to add quite a bit of spacing on desktop, more than necessary imo. How about: `sx={{ mb: { xs: 2, sm: 0 } }}` (I didn't know you could do that until just now! handy)

src/modules/projects/components/tables/ProjectEnrollmentsTable.tsx
```diff
@@ -40,9 +40,9 @@ const ProjectEnrollmentsTable = ({
   return (
     <>
       <Box py={2} px={3} sx={{ borderBottom: 1, borderColor: 'divider' }}>
-        <Stack direction='row' gap={4}>
+        <Grid container direction='row' spacing={4}>
```
gigxz: Nice! I don't think we need as much spacing between the rows though, how about `rowSpacing={2} columnSpacing={4}`

gigxz (commented):
sick

looks good!!! a few small suggestions but looks good to merge after those

src/modules/form/components/group/FormCard.tsx
martha: Much better! Thank you

src/modules/form/components/group/DisabilityTable.tsx
martha: Yes, this makes sense to me. I guess it's possible that the `rowItem.text` and `cellItem.text` could actually be different coming from the json form, do you think it's safe to assume they won't be? Or should we add some check here where we display them both if they are actually different, but only one if they're the same?

src/components/elements/input/TextInput.tsx
martha: Oh! I see. This is supposed to align items to the center vertically. And when the horizontal input group actually is horizontal, it is really doing that. The problem on mobile is that the horizontal group ends up vertical anyway to fit the space (from my changes to the Stack's direction on what's now line 126), and it ends up like this:
Screenshot 2024-02-23 at 4 56 28 PM
(I entirely overlooked the vertical centering while I was trying to solve the problem of the horizontal centering -- sorry and thank you for pointing it out to me!)

Anyway, it seems to me now like the most sensible thing to do is not to pass the `horizontalInputSx` prop at all if the screen is so small as to make the component actually vertical. That's what I've implemented here, let me know if you think it's good. Here are the results:

On mobile: 
Screenshot 2024-02-23 at 5 01 56 PM

On desktop:
Screenshot 2024-02-23 at 5 02 05 PM

(The storygraph still does not look that good on mobile, I think bc there are some other components not included in the Intake Assessment that are still taking up fuller width and causing the whole screen to have horizontal scroll. If you agree, I'd say that is probably out of scope for this ticket)

gigxz (approved):

PR #654, opened 2024-02-17
187020495 time of day input

ttoomey:
## Description

[#187020495
](https://www.pivotaltracker.com/story/show/187020495)

](https://github.com/greenriver/hmis-warehouse/pull/4031)

details
* upgrade `mui/x-date-pickers` from 5 beta to 6
* support time of day picker
* support generic formulas for calculated values using `jsep` to show formatted duration between two times. Supported for autofill values and interpolated display text
* adjusted yarn dedup strategy to resolve CI failures around the safe-buffer dep

## Testing

I've been testing with this fragment
```json
{
  "item": [
    {
      "item": [
        {
          "text": "Assessment Date",
          "type": "DATE",
          "link_id": "date-1",
          "mapping": {
          },
          "required": true,
          "assessment_date": true
        },
        {
          "text": "Start Time",
          "type": "TIME_OF_DAY",
          "link_id": "start_date",
          "required": true
        },
        {
          "text": "End Time",
          "type": "TIME_OF_DAY",
          "link_id": "end_date",
          "required": true,
          "bounds": [
            {
              "id": "min-end-timee",
              "type": "MIN",
              "question": "start_date"
            }
          ]
        },
        {
          "text": "Duration: ${FORMAT_DURATION(value)}",
          "type": "DISPLAY",
          "link_id": "duration_minutes",
          "autofill_values": [
            {
              "formula": "ABS(end_date - start_date)",
              "autofill_when": [],
              "autofill_behavior": "ANY",
              "autofill_readonly": false
            }
          ]
        }
      ],
      "text": "Date Test",
      "type": "GROUP",
      "link_id": "text-group"
    }
  ]
}
```
## Type of change
- [x] Dependency update


src/components/elements/input/TextInput.tsx
```diff
@@ -49,7 +49,7 @@ const TextInput = ({
   const htmlId = id || generatedId;
 
   let width = inputWidth;
-  if (!width && inputProps.inputMode === 'numeric') {
+  if (!width && inputProps.inputMode === 'numeric' && !sx) {
```
ttoomey, line 52: skip this hard-coded width if we are supplying sx, this prevents the width from changing on the input when it's focused in the mui pickers.

src/config/theme.ts
```diff
@@ -204,11 +204,6 @@ const createThemeOptions = (theme: Theme) => ({
         },
       },
     },
-    MuiDatePicker: {
```
ttoomey, line 207: this doesn't seem to be used in the new pickers version

src/modules/form/util/formUtil.ts
```diff
@@ -865,6 +876,14 @@ export const buildAutofillDependencyMap = (
     item.autofillValues.forEach((av) => {
       if (viewOnly && !av.autofillReadonly) return;
 
+      if (av.formula) {
```
gigxz, line 879: I agree that "formula" is something that gets evaluated at the "autofill" stage.
I think the `formula`  field should probably be on the `AutofillValue` object, not the `FormItem` object. 
Something like 
```
autofill: [
  {
      value_formula: 'x + y + 17',
      autofill_when: <condition>
   },
   // more autofill conditions can go here. The first one who's "when" condition evalutes to true will be rendered
]
```


This would have a few benefits imo
1. Clearer how the formula is used 
2. Easier to expand "formula" for other use-cases later. For example, we might want to evaluate a formula on the `InitialValue` object to specify the initial value for a field. Or, on a `ValueBound` to specify that a max bound is ExitDate + 3 days (currently implemented with `ValueBound.offset`). Each of those objects defines a distinct behavior. A "formula" is more of a tool that could be used for each of those behaviors. Wdyt? I'm not sure I am finding the right words to describe it, but there's something there.
3. Autofill formula can be conditionally applied as above


There is one issue with it:
* `autofill_when` is non-nullable, so you need to specify a condition. In this case we always want it to evaluate (I think?). This has been a minor annoyance I am aware of. We should fix it so there is an easy way to specify an autofill value that should "always" run (or rather, alway run when it's dependent values change**).


**Edit to add: Is the formula aware of "dependencies" from the autofilled item to the other item values that the formula depends on? If they don't get into the autofill dependency map, I think there will be render issues. Relates to my other question about variables that I don't fully understand yet.

src/modules/form/util/expressions/formula.ts
```diff
@@ -0,0 +1,70 @@
...
+import { parseNumber } from '@/utils/numbers';
+
+const ensureNumeric = (value: any): number => {
+  if (parseNumber(value) === undefined)
+    throw new ExpressionEvaluationError('Non numeric result');
+  return value;
+};
+
+type EvalContext = Map<string, any>;
+const evaluate = (ast: any, context: EvalContext): number | undefined => {
```
gigxz, line 13: This is cool and very flexible!
Can you explain a little bit how the variables get evaluated? I'm getting a little lost at that part. In your example `end_date - start_date` where do those values get read from?

ttoomey: Sure! So the formula expression is processed in two stages. It is pretty much the same as how we handle `sumQuestions`, I just copied that plumbing.

The first stage simply collects all the variable names referenced by the expression statement (evaluator encounters `ast.type === "Identifier"`). Then each variable referenced by the formula is aded as a dependency. That's this code in formUtils:
```js
for (const id of collectExpressionReferences(av.formula)) { addDependency(id); }
// for "end_date - start_date",  this is: addDependency("end_date"); addDependency("start_date")
```

In the second stage the values are read and operations are performed to return the result. We construct a new context map to provide the current values to the evaluator. This includes all variables (`Object.keys(values)`) which is probably too broad and could be narrowed. The `evaluateFormula` function only uses the values referenced in the expression from the context. Note, instead of constructing a map, we could use the values object directly; the map provides more safety and control in exchange for overhead.
```js
const context = new Map();
for (const key of Object.keys(values)) {
  const value = numericValueForFormValue(values[key]);
  if (!isNil(value)) context.set(key, value);
}
return evaluateFormula(av.formula, context);
// context would be new Map([['end_date', 120], ['start_date', 60], ... other keys in the values object])
```

src/modules/form/util/formUtil.ts
ttoomey: >I think the formula field should probably be on the AutofillValue object, not the FormItem object.

Sorry, I'm probably misunderstanding but I think the formula [is already on autofillValue](https://github.com/greenriver/hmis-warehouse/pull/4031/files#diff-8ef987cb4091bd67da4667c6c74ff8ba133bb9e5f5f03ae7cd621f8fcd85564aR20). Maybe we can huddle about it this week; I'm sure there's plenty of room for improvement in my implementation 😅 

> Edit to add: Is the formula aware of "dependencies" from the autofilled item to the other item values that the formula depends on

I think so, or at least it should have the same dependency behavior as sum_questions

gigxz: Whoops 🤦 my bad. Got it

src/modules/form/components/DynamicField.tsx
```diff
@@ -304,6 +309,21 @@ const DynamicField: React.FC<DynamicFieldProps> = ({
           />
         </InputContainer>
       );
+    case ItemType.TimeOfDay:
```
gigxz, line 328: This needs to be added to storybook but I can do it as part of https://github.com/greenriver/hmis-frontend/pull/657 after this merges, so we dont need to deal with merge conflicts

gigxz: Can you open storybook on this branch? I can't, I get errors about the `@mui/utils` package. It's fixed by running `yarn add @mui/material`, which does some other minor upgrades: https://github.com/greenriver/hmis-frontend/pull/659

gigxz: Git is having trouble with the diff it seems, but I just ran that yarn add command

src/modules/form/util/expressions/formula.ts
gigxz, line 13: I get `N/A` no matter what if the linkIds have dashes in them. We use lots of dashes in IDs so I think that is an issue. This is what I tested in storybook:
```
        {
          "type": "TIME_OF_DAY",
          "linkId": "time-1",
          "text": "Start time"
        },
        {
          "type": "TIME_OF_DAY",
          "linkId": "time-2",
          "text": "End time"
        },
        {
          "text": "Duration: ${FORMAT_DURATION(value)}",
          "type": "DISPLAY",
          "linkId": "duration_minutes",
          "autofillValues": [
            {
              "formula": "ABS(time-1 - time-2)",
              "autofillWhen": [],
              "autofillBehavior": "ANY",
              "autofillReadonly": false
            }
          ]
        },
```

src/modules/form/components/DynamicDisplay.tsx
```diff
@@ -19,25 +20,22 @@ const SeverityMap: Record<string, AlertColor> = {
...
-    if (viewOnly && !isNil(item.readonlyText)) stringValue = item.readonlyText;
-    if (isNil(stringValue)) return undefined;
+    let displayValue = item.text;
+    if (viewOnly && !isNil(item.readonlyText)) displayValue = item.readonlyText;
+    if (isNil(displayValue)) return undefined;
 
-    return { __html: DOMPurify.sanitize(interpolate(stringValue, { value })) };
+    return {
+      __html: DOMPurify.sanitize(
+        evaluateTemplate(displayValue, new Map([['value', value]]))
```
gigxz, line 36: This is nice

gigxz (changes requested):
Nice work!! I haven't done extensive testing but the intended use-case (time duration) looks good to me. 
Requested changes are:
1. Fix package issue with storybook (https://github.com/greenriver/hmis-frontend/pull/659)
2. Fix problem identifying link ids in formula

src/modules/form/util/expressions/formula.ts
gigxz: 👍  awesome.

ttoomey: Oh good catch, thanks! This requires some thought. Allowing operators as part of an identifier is not really compatible with this (or most) parsers. 

Possible solutions:
* We could just "not do that" - if we want to use link ids in an expression, they can't have hyphens (or other operator chars).
* we could wrap each variable reference in a function with an escaped string (for json). that looks like: `"formula": "ABS($(\"end-date\") - $(\"start-date\"))"`. This starts to get hard to read and adds complexity/processing overhead but it should work
* We could technically reconfigure the parser with different operator and identifier chars
* In the form refactor branch Zach built a utility that translates invalid link ids so they work with RHF. We could try and extract and adapt that solution.

Long term I'd love for us to move away from hyphens, dots, etc. Supporting those invalid chars seems limiting without providing much benefit. Ideally our link ids would also be valid JS identifiers. If we want to rip off the bandaid, I think we could write a data migration to update all to form definitions.

What do you think?

src/modules/form/components/DynamicField.tsx
ttoomey: Huh, I'd kind of forgotten storybook was even here tbh. Storybook does seem to work for me locally on this branch, not sure what the difference is.

Thanks for the update/fix in https://github.com/greenriver/hmis-frontend/pull/659 but I do want to call out that the MUI upgrade from 5.11 to 5.15 includes breaking changes (check the MUI [github releases pages going back to 5.12](https://github.com/mui/material-ui/releases?page=1)). I haven't reviewed those release notes yet but wanted to make sure that was okay

gigxz: Hmm okay. If its working for you, lets not merge my PR. I'll keep plugging away at the build error I was getting on this branch

gigxz: Huh ok good to know. I'll review the release notes to see

src/modules/form/util/expressions/formula.ts
gigxz: honestly I think we should just go ahead and migrate them. I added a ticket here: https://www.pivotaltracker.com/story/show/187087981

src/modules/form/components/DynamicField.tsx
gigxz: After doing a full yarn cache cleanout and reinstalling, its working for me now. I will close that PR, and we can update those packages separately. thanks!

gigxz (approved):
LGTM provided we will migrate linkIds so that we can use formulas with all items

PR #656, opened 2024-02-20
Add time duration input component for dynamic forms

martha:
## Description

This adds a `TimeDurationInput` component to the frontend.

I'm not 100% sure of the approach and would like to discuss more. As long as we can receive an already-filled-in-value, and the minutes/hours should be shown to the user but we should only store minutes on the backend, then some kind of minutes/hours translation math is necessary, but the approach I've taken here does behave somewhat unexpectedly if the user does something unexpected (like try to enter more than 60 minutes.) The nice aspect to this approach is that by doing the translation in real time we can keep just one source of truth for the value's state -- rather than storing it in two places, which seems very messy.

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4038
PT issue: https://www.pivotaltracker.com/story/show/187028540

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/components/elements/input/TimeDurationInput.tsx
```diff
@@ -0,0 +1,65 @@
...
+    (hours?: string | number, minutes?: string | number) => {
+      const h = getIntValue(hours);
+      const m = getIntValue(minutes);
+      return h * 60 + m;
+    }
+  );
+
+  return (
+    <>
+      {label}
```
gigxz: So I think we want to wrap all of this in a [FormControl](https://mui.com/material-ui/api/form-control/) -- so you have the input label, the 2 inputs, and the FormHelperText.

Currently if I add `helperText` to the item it is showing up twice, once under each input field.

src/components/elements/input/TimeDurationInput.tsx
```diff
@@ -0,0 +1,65 @@
...
+      <Stack direction='row' gap={1}>
+        <NumberInput
+          {...props}
+          value={hours}
+          onChange={(e) => {
+            if (onChange)
+              onChange(hoursMinutesToValue(e.target.value, minutes));
+          }}
```
gigxz: I would suggest `onChange={onHoursChange}` (and similar for minutes), and defining each callback using `useCallback` to memoize

src/components/elements/input/TimeDurationInput.tsx
```diff
@@ -0,0 +1,65 @@
...
+    return v;
+  };
+
+  const hoursMinutesToValue = memoize(
+    (hours?: string | number, minutes?: string | number) => {
+      const h = getIntValue(hours);
+      const m = getIntValue(minutes);
+      return h * 60 + m;
+    }
+  );
```
gigxz: `memoize` here is imported from lodash and I'm not sure how it behaves. I think you're looking for react `useMemo`?

src/components/elements/input/TimeDurationInput.tsx
```diff
@@ -0,0 +1,65 @@
...
+import NumberInput from '@/components/elements/input/NumberInput';
+import { TextInputProps } from '@/components/elements/input/TextInput';
+
+type Props = { value?: number; onChange?: (val: number | null) => void } & Omit<
+  TextInputProps,
+  'value' | 'onChange'
+>;
+
+const TimeDurationInput = ({ value, onChange, label, ...props }: Props) => {
+  const [hours, minutes] = useMemo(() => {
```
gigxz: I got into a weird state where the hours were always prefixed with a "0" and I couldn't make it go away. I think it happens if you do this:
1. select minutes field, type a number
2. select hour field (which is now 0) and try to type a number

If we switch to using `0` as placeholders (See other comment), I think we should just leave hours as empty if there are no hours selected. That should get rid of the funkiness.

src/components/elements/input/TimeDurationInput.tsx
```diff
@@ -0,0 +1,65 @@
...
+      const h = getIntValue(hours);
+      const m = getIntValue(minutes);
+      return h * 60 + m;
+    }
+  );
+
+  return (
+    <>
+      {label}
+      <Stack direction='row' gap={1}>
```
gigxz: `alignItems='center'` to get the colon in the middle

src/components/elements/input/TimeDurationInput.tsx
```diff
@@ -0,0 +1,65 @@
...
+        />
+        <Typography variant='body1'>:</Typography>
+        <NumberInput
+          {...props}
+          value={minutes}
+          onChange={(e) => {
+            if (onChange) onChange(hoursMinutesToValue(hours, e.target.value));
+          }}
+          label={null}
+          placeholder='Minutes'
```
gigxz: I don't love this as placeholder, because it's not clear that these are minutes/hours if you're viewing a form that's already filled out.

My suggestion would be to make the placeholder `0` for both, and also add an `endAdornment` (a MUI thing) that specifies hours vs minutes. I played around with this, I think it looks decent:
```
          inputWidth={85}
          InputProps={{
            endAdornment: (
              <Typography
                variant='subtitle2'
                color='text.secondary'
                component='span'
              >
                hours
              </Typography>
            ),
            ...props.InputProps, // need to make sure we pass this so parent component can send any other InputProps
          }}

```
Screenshot 2024-02-20 at 4 30 44 PM


Screenshot 2024-02-20 at 4 31 28 PM

src/types/gqlTypes.ts
```diff
@@ -1130,6 +1130,8 @@ export enum Component {
   SignatureGroup = 'SIGNATURE_GROUP',
   /** SSN input component */
   Ssn = 'SSN',
+  /** Time duration component with hours and minutes */
+  TimeDuration = 'TIME_DURATION',
```
gigxz: The component really only collects *minutes* duration. It doesn't let you select a number of days for example. So lets call it `MINUTES_DURATION`.

src/components/elements/input/TimeDurationInput.tsx
```diff
@@ -0,0 +1,65 @@
...
+      const m = getIntValue(minutes);
+      return h * 60 + m;
+    }
+  );
+
+  return (
+    <>
+      {label}
+      <Stack direction='row' gap={1}>
+        <NumberInput
```
gigxz: Both of these inputs will need a `aria-label`, since they are technically  separate inputs they can't both be labeled by the same label. I think something like `aria-label={`${label} (hours)`}`, so the aria labels become `Time Spent (hours)` and `Time Spent (minutes)` (for example).

gigxz (commented):
Looking good!

I like this approach, and agree that we should do the fanciness on the frontend and only send the Minutes number to the backend.

src/components/elements/input/TimeDurationInput.tsx
martha: Oh thank you! Good point. I had to add `helperText` in explicitly at the bottom, copying an example from `CheckboxGroupInput` and `RadioGroupInputOption`. But I think it looks ok now.

src/components/elements/input/TimeDurationInput.tsx
martha: yes, I didn't really understand the behavior of `useCallback`, and was trying to find a way to memoize the results of a function outside of React's purview -- but I agree with your approach here, it's better!

src/components/elements/input/TimeDurationInput.tsx
martha: Yes, this is better! I also added ` || ''` to the value prop for each of the inputs, so it should render as blank instead of 0 when there is no value for minutes or hours.

src/components/elements/input/TimeDurationInput.tsx
martha: This is nice! Thanks for the suggestion. 🎉

src/components/elements/input/TimeDurationInput.tsx
martha: All good, tho since `label` is a `ReactNode` here we have to do some magic with casting in order to get the text prop from it - let me know if you think it looks ok?

martha:
Overall I still feel like this approach might be a little unexpected for users if they try to type in numbers above 60 in the minutes field. A better UX might be to not change anything under the user's fingers, but wait until focus shifts away, or even until after the form is submitted, to do any minutes/hours translation math. But this is some more effort since the components are fully controlled components. Let me know if you think this is worth pursuing.

gigxz (approved):
Looks good!!
I agree it could be a minor improvement to adjust some of the calculation to happen `onBlur` (like if the user types 90 minutes => 1h30m) but I don't think it's necessary.

PR #664, opened 2024-02-23
Neaten assessment form side navigation

martha:
## Description

This should fix the appearance of the awkward empty box as shown in screenshots here: https://www.pivotaltracker.com/n/projects/2591838/stories/186976233/comments/240342546

Let me know if the refactor of the logic from `DeleteAssessmentButton.tsx` to here looks okay to you.

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/modules/assessments/components/AssessmentFormSideBar.tsx
+++ b/src/modules/assessments/components/AssessmentFormSideBar.tsx
@@ -14,6 +14,7 @@ import FormStepper from '@/modules/form/components/FormStepper';
 
 import { EnrollmentDashboardRoutes } from '@/routes/routes';
 import {
+  AssessmentRole,
   EnrollmentFieldsFragment,
   FormDefinitionFieldsFragment,
   FullAssessmentFragment,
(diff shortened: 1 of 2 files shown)


src/modules/assessments/components/AssessmentFormSideBar.tsx
```diff
@@ -60,6 +61,32 @@ const AssessmentFormSideBar: React.FC<Props> = ({
     [enrollment, navigate]
   );
 
+  const shouldShowDeleteButton = (assessment: FullAssessmentFragment) => {
```
gigxz: This feels funky to me. It's a function that is defined inside the component that accepts a variable `assessment` that "shadows" the outer variable `assessment` that is a prop. But we always pass the `assessment`(prop) in as the `assessment` var to the function.

I would suggest a different pattern.. Here are some options, I think I prefer the first:

1) Contain this logic inside a useMemo hook. It only needs to re-run when the assessment changes (as opposed to on every render)
```
const showDeleteAssessmentButton = useMemo(() => {
    if (!assessment) return false;

   // full logic for whether deletion is allowed goes here
   return true
}, [assessment]);
```

2) Move this `shouldShowDeleteButton ` function outside of the react component. There is no need for it to be re-defined and called every render, if it just a function evaluating some conditions about an assessment record.

gigxz (commented):
Look good to me! small suggestion on react style

src/modules/assessments/components/AssessmentFormSideBar.tsx
martha: Thanks -- this is much better! :)

src/modules/assessments/components/AssessmentFormSideBar.tsx
```diff
@@ -79,35 +105,45 @@ const AssessmentFormSideBar: React.FC<Props> = ({
...
+                    clientId: assessment.enrollment.client.id,
+                    enrollmentId: assessment.enrollment.id,
+                    assessmentId: assessment.id,
+                  }
+                )}
+              >
+                Print
+              </PrintViewButton>
+            )}
+            {showDeleteAssessmentButton && assessment && (
```
martha, line 131: I added `&& assessment` here to keep typescript happy, even though it's always true from `showDeleteAssessmentButton`. The other thing we could do is pass `assessment as FullAssessmentFragment`. Let me know if you prefer one over the other!

gigxz: I usually take the approach you have here to keep typescript happy when necessary

gigxz (approved):

PR #665, opened 2024-02-23
Make AutoExit and AutoEnter generic into ProjectConfig

martha:
## Description

Here is a first pass at the frontend changes that would be needed for generifying the existing AutoExitConfig UI to be able to handle all ProjectConfigs including AutoEnterConfigs and future types of project configs.

As discussed this will need some more design input!

PT issue: https://www.pivotaltracker.com/story/show/187091393
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4055/files

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/admin/components/projectConfig/ProjectConfigTable.tsx
```diff
@@ -62,9 +65,24 @@ const columns: ColumnDef<AutoExitConfigFieldsFragment>[] = [
...
   },
+  {
+    header: 'Config Options',
+    render: ({ configOptions }) => {
+      if (!configOptions) return;
+      const parsedOptions = JSON.parse(configOptions);
+
+      return Object.keys(parsedOptions).map((key) => {
+        return (
+          <span key={key}>
```
martha: this whole thing will need design review of course, but this is a particular place in the code where we may need some thoughts from Ash on how to display the newly generified key/value pairs of config options

src/types/gqlObjects.ts
```diff
@@ -2783,7 +2753,7 @@ export const HmisObjectSchemas: GqlSchema[] = [
         type: {
           kind: 'NON_NULL',
           name: null,
-          ofType: { kind: 'SCALAR', name: 'ISO8601DateTime', ofType: null },
+          ofType: { kind: 'SCALAR', name: 'ISO8601Date', ofType: null },
```
gigxz, line 2756: Hmm looks like these got accidentally reverted by a PR that theron merged recently. So we can keep them, these changes should go back into 106

src/modules/admin/components/projectConfig/ProjectConfigTable.tsx
```diff
@@ -62,9 +65,24 @@ const columns: ColumnDef<AutoExitConfigFieldsFragment>[] = [
...
+      });
+    },
+  },
 ];
 
-const AutoExitTable = ({
+const ProjectConfigTable = ({
   setSelectedId,
 }: {
   selectedId?: string | null;
```
gigxz: I know this is legacy but it doesn't look like `selectedId` is used at all. This is more of a "click" than a selection action. I would suggested renaming the `setSelectedId` prop to  `onClickRow`, which is a bit more descriptive of what it's actually doing. The caller would do `onClickRow={setSelected}`.

src/modules/admin/components/projectConfig/ProjectConfigPage.tsx
```diff
@@ -2,36 +2,36 @@ import AddIcon from '@mui/icons-material/Add';
...
       />
       <Paper>
-        <AutoExitTable selectedId={selected} setSelectedId={setSelected} />
+        <ProjectConfigTable selectedId={selected} setSelectedId={setSelected} />
       </Paper>
-      <AutoExitDialog
+      <ProjectConfigDialog
         open={!!selected}
-        autoExitId={selected || ''}
+        projectId={selected || ''}
```
gigxz: Can we rename prop to `configId` or `projectConfigId`? I got confused thinking it was a project

src/modules/admin/components/projectConfig/ProjectConfigDialog.tsx
```diff
@@ -0,0 +1,96 @@
...
+const ProjectConfigDialog: React.FC<ProjectDialogProps> = ({
+  projectId,
+  ...props
+}) => {
+  const { data } = useGetProjectConfigsQuery();
+
+  const project = useMemo(
+    () => data?.projectConfigs?.nodes?.find((aec) => aec.id === projectId),
+    [data, projectId]
+  );
```
gigxz: This seems off to me - we already have the record from the table, so we do we need to query (even if it just hits the cache) to find it again.

Because we need these values in the dialog, I would switch to passing around the selected row (the Project Config reciord) rather than hyst the id. Eg:

ProjectConfigTable: `handleRowClick={onClickRow}`
ProjectConfigPage:
```
<ProjectConfigTable onRowClick={setSelected} />
{selected && <ProjectConfigDialog
        open={!!selected}
        config={selected}
        onClose={() => setSelected(null)}
      />}
```

gigxz: I see this was a pre-existing thing but I think its worth the change!

src/modules/admin/components/projectConfig/ProjectConfigTable.tsx
gigxz: Agreed. Lets just change the `span` to a `Stack` for now so that multiple attributes would  stack vertically.

gigxz:
Looks great @martha!! comments above

src/modules/admin/components/projectConfig/ProjectConfigDialog.tsx
martha: Yeah I agree with this change! Does [this](https://github.com/greenriver/hmis-frontend/pull/665/files#diff-9c32d267577bb1b277d2efa915f845e0e37674ff78cc132700dceb75fbab361aR20) look correct to you for opening the form when it's a new config? Previously we were just passing the string 'new' but that now of course makes typescript unhappy.
```
onClick={() => setSelected({} as ProjectConfigFieldsFragment)}
```

src/modules/admin/components/projectConfig/ProjectConfigPage.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const [selected, setSelected] = useState<ProjectConfigFieldsFragment | null>(
+    null
+  );
+  return (
+    <>
+      <PageTitle
+        title='Project Configs'
+        actions={
+          <Button
+            onClick={() => setSelected({} as ProjectConfigFieldsFragment)}
```
gigxz: Hrm no I think this feels off too. Nothing is really "selected" here, its just that we want the dialog to open. I think we are missing state to track whether the dialog is open or not.

```
const [open, setOpen] = useState(false)

...
onClick={() => setOpen(true)


```

You will just need to pass `open` and `onClose` to the dialog properly.

src/modules/admin/components/projectConfig/ProjectConfigDialog.tsx
gigxz: oops response at https://github.com/greenriver/hmis-frontend/pull/665#discussion_r1508238149

gigxz (approved):
Thank you!!

PR #669, opened 2024-02-26
Add bounds to form definition story

martha:
## Description

This arose out of trying to debug a problem with character bounds. PT issue: https://www.pivotaltracker.com/story/show/187116631

The issue turns out to be because of an Apollo caching problem, so, it's not reproduced here, but I'll put up the PR anyway so we can add it and make our stories slightly more useful.

I also considered just adding bounds on the existing string and text inputs here, rather than making separate ones. Let me know if you'd like me to implement this change. 

## Type of change
- [ ] Bug fix
- [x] Code clean-up / housekeeping (adds storybook detail)
- [ ] Dependency update

### DIFF
--- a/src/test/__mocks__/mockFormDefinition.json
+++ b/src/test/__mocks__/mockFormDefinition.json
@@ -14,6 +14,20 @@
           "helperText": "This is helper text"
         },
         {
+          "type": "STRING",
+          "required": false,
+          "linkId": "string-2",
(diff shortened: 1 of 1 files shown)


gigxz:
Ahhh caching on the value bound "id"! that makes sense. good find. And thanks for adding this.

gigxz (approved):

PR #672, opened 2024-02-29
Enable rich text labels in dynamic forms

martha:
## Description

This enables rich text in Dynamic Form field labels.
PT issue: [186869594](https://www.pivotaltracker.com/story/show/186869594)

Design work around this is still ongoing. My understanding of the issue is that form field labels are semi-bold by default right now; but, removing the semi-boldness default (which seems sensible/perhaps necessary as a default for a rich-text field) has wider impacts for the text hierarchy on the page. In this PR, the semi-boldness is not modified, but rich text is just enabled on top of it. In the Storybook mock, you can see how it will look to add bold on top of semi-bold:
Screenshot 2024-02-29 at 9 34 37 AM

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/components/elements/CommonHtmlContent.tsx
+++ b/src/components/elements/CommonHtmlContent.tsx
@@ -4,7 +4,7 @@ import { useMemo } from 'react';
 
 interface Props extends TypographyProps {
   children: string;
-  component: React.ElementType;
+  component?: React.ElementType;
 }
 const CommonHtmlContent: React.FC<Props> = ({
   children,
--- a/src/components/elements/CommonHtmlContent.tsx
+++ b/src/components/elements/CommonHtmlContent.tsx
@@ -1,14 +1,20 @@
 import { Typography, TypographyProps } from '@mui/material';
+import DOMPurify from 'dompurify';
+import { useMemo } from 'react';
 
 interface Props extends TypographyProps {
   children: string;
 }
 const CommonHtmlContent: React.FC<Props> = ({ children, ref, ...props }) => {
+  const purifiedHtml = useMemo(() => {
+    return DOMPurify.sanitize(children);
+  }, [children]);
+
   return children ? (
     <Typography
-      {...props}
       component='div'
-      dangerouslySetInnerHTML={{ __html: children }}
+      {...props}
+      dangerouslySetInnerHTML={{ __html: purifiedHtml }}
     />
   ) : null;
 };
--- a/src/components/elements/CommonHtmlContent.tsx
+++ b/src/components/elements/CommonHtmlContent.tsx
@@ -4,15 +4,21 @@ import { useMemo } from 'react';
 
 interface Props extends TypographyProps {
   children: string;
+  component: React.ElementType;
 }
-const CommonHtmlContent: React.FC<Props> = ({ children, ref, ...props }) => {
+const CommonHtmlContent: React.FC<Props> = ({
+  children,
+  ref,
+  component,
+  ...props
+}) => {
   const purifiedHtml = useMemo(() => {
     return DOMPurify.sanitize(children);
   }, [children]);
 
   return children ? (
     <Typography
-      component='div'
+      component={component || 'div'}
       {...props}
       dangerouslySetInnerHTML={{ __html: purifiedHtml }}
     />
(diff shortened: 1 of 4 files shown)


src/components/elements/CommonHtmlContent.tsx
```diff
@@ -1,14 +1,20 @@
...
+  const purifiedHtml = useMemo(() => {
+    return DOMPurify.sanitize(children);
+  }, [children]);
+
   return children ? (
     <Typography
-      {...props}
       component='div'
-      dangerouslySetInnerHTML={{ __html: children }}
+      {...props}
```
martha, line 22: ~I moved `props` under `component` here because the `RequiredLabel` passes in `component='span'`. I think it makes sense to keep `div` as default but allow it to be overridden.~

src/components/elements/CommonHtmlContent.tsx
```diff
@@ -1,14 +1,20 @@
 import { Typography, TypographyProps } from '@mui/material';
+import DOMPurify from 'dompurify';
+import { useMemo } from 'react';
 
 interface Props extends TypographyProps {
   children: string;
 }
 const CommonHtmlContent: React.FC<Props> = ({ children, ref, ...props }) => {
+  const purifiedHtml = useMemo(() => {
```
martha, line 15: @ttoomey do you think it's a good idea to move this purification step inside of the `CommonHtmlContent` as I have done here? Or better to keep it outside, and have the components that want this call it separately as was the case before?

src/modules/form/components/DynamicDisplay.tsx
```diff
@@ -51,17 +46,18 @@ const DynamicDisplay: React.FC<Props> = ({
           severity={SeverityMap[item.component as keyof typeof SeverityMap]}
           sx={{ mt: 2 }}
         >
-          <div dangerouslySetInnerHTML={html} />
+          <CommonHtmlContent>{html}</CommonHtmlContent>
```
martha, line 49: Changes in this file aren't related to rich-text labels, but just using the `CommonHtmlContent` component for consistency

src/components/elements/CommonHtmlContent.tsx
martha: I made `component` an explicit prop to try to solve this build failure on the `tsc` step:
```
Error: src/modules/form/components/RequiredLabel.tsx(17,63): error TS2322: Type '{ children: string; component: string; border?: ResponsiveStyleValue<number | "medium" | "none" | (string & {}) | "inset" | "hidden" | "-moz-initial" | "inherit" | "initial" | "revert" | "revert-layer" | ... 188 more ... | undefined> | ((theme: Theme) => ResponsiveStyleValue<...>); ... 363 more ...; onTransitionEndC...' is not assignable to type 'IntrinsicAttributes & Props'.
  Property 'component' does not exist on type 'IntrinsicAttributes & Props'.
```

I was surprised about this though, because the `CommonHtmlContent`'s `Props` extends `TypographyProps`, so I would expect `component` to be allowed, and adding it explicitly is a bit redundant. Let me know if you think I've taken an OK approach here or we should change things around.

(Also, I noticed that running `yarn tsc` locally with the failing code resulted in not this error but a couple of other errors in `apolloClient`, `ApolloErrorAlert`, and `GenericTableWithData` - is this true for others too? Anything to worry about?)

ttoomey: > I was surprised about this though, because the CommonHtmlContent's Props extends TypographyProps, so I would expect component to be allowed, and adding it explicitly is a bit redundant. Let me know if you think I've taken an OK approach here or we should change things around.

Thanks for calling this out, that is odd, it doesn't seem like declaring the component prop should necessary. 

There might be something funny going on with our typescript dependency; I see we have several versions of typescript installed. Seems like we should do some investigation,[ added issue for it here](https://www.pivotaltracker.com/story/show/187142123)

src/components/elements/CommonHtmlContent.tsx
ttoomey: That's a nice cleanup, thanks!!

ttoomey (approved):
Looks great, thank you!

src/components/elements/CommonHtmlContent.tsx
gigxz: Strange, thanks for flagging!

gigxz (approved):
great!

PR #673, opened 2024-02-29
Add spacing between client alert and form errors

martha:
## Description

This fixes a bug where there was no spacing between the client alert and the form errors in the enrollment dialog.
PT issue: [187117705](https://www.pivotaltracker.com/n/projects/2591838/stories/187117705)

After this change:
Screenshot 2024-02-29 at 11 58 39 AM

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/modules/form/hooks/useFormDialog.tsx
+++ b/src/modules/form/hooks/useFormDialog.tsx
@@ -152,9 +152,14 @@ export function useFormDialog<T extends SubmitFormAllowedTypes>({
             {definitionLoading ? (
               <Loading />
             ) : formDefinition ? (
-              <Grid container spacing={2} sx={{ mb: 2, mt: 0 }}>
+              <Grid
+                container
+                direction='column'
+                spacing={2}
+                sx={{ mb: 2, mt: 0 }}
+              >
+                <Grid item>{props.preFormComponent}</Grid>
                 <Grid item xs>
-                  {props.preFormComponent}
                   <DynamicForm
                     ref={formRef}
                     definition={formDefinition.definition}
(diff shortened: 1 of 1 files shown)


src/modules/form/hooks/useFormDialog.tsx
```diff
@@ -152,9 +152,14 @@ export function useFormDialog<T extends SubmitFormAllowedTypes>({
             {definitionLoading ? (
               <Loading />
             ) : formDefinition ? (
-              <Grid container spacing={2} sx={{ mb: 2, mt: 0 }}>
+              <Grid
```
martha, line 155: I _think_ changing this grid's `direction` prop should be all right, because in the previous version of the code it has only one grid item within it, and the child `DynamicForm` component renders its own grid internally. But I know this hook is used in lots more contexts than just the one I'm looking at, so let me know if you think the change has a possibility of unintended side effects, and I can revert this change and instead add a `Stack` that separates the pre form component from the form itself.

gigxz: Yeah this makes me a little nervous just because it's used so widely. Can we just add margin below the preform component instead? Eg `mb` on the `ClientAlertStack` component

gigxz: Yep this looks good to me (deleted my previous comment). The only issue I am seeing is that there is additional padding added when there is no preformcomponent.

src/modules/form/hooks/useFormDialog.tsx
```diff
@@ -152,9 +152,14 @@ export function useFormDialog<T extends SubmitFormAllowedTypes>({
...
+                container
+                direction='column'
+                spacing={2}
+                sx={{ mb: 2, mt: 0 }}
+              >
+                <Grid item>{props.preFormComponent}</Grid>
```
gigxz:
```suggestion
                {props.preFormComponent && (
                  <Grid item>{props.preFormComponent}</Grid>
                )}
```

gigxz (approved):

PR #675, opened 2024-03-04
Suppress some console noise

martha:
## Description

Lowest priority;
I was noticing some console noise on the Dynamic Form (both in storybook and when these components are used in their natural habitats), and since I actively use the console while developing, and had a bit of extra time last week, I thought it would be worth while to dig in a little and see if I could solve a few of them.

Let me know if you think it looks okay, or, I'm fine to drop this PR entirely if you think it's not worthwhile!

The changes in TableGroup suppress the error:
```
Warning: validateDOMNesting(...): <tr> cannot appear as a child of <table>. Add a <tbody>, <thead> or <tfoot> to your code to match the DOM tree generated by the browser.
```

The other changes related to aria labelling suppress the errors:
```
Warning: Invalid ARIA attribute `ariaLabel`. Did you mean `aria-label`?
```

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/components/elements/input/GenericSelect.tsx
+++ b/src/components/elements/input/GenericSelect.tsx
@@ -34,6 +34,7 @@ const GenericSelect = <
   ...rest
 }: GenericSelectProps<T, Multiple, Creatable>) => {
   const { placeholder, ...inputProps } = textInputProps || {};
+  const { ariaLabel, ...restOfProps } = rest;
   // Show a loading indicator if we have a value but the picklist is still loading
   const startAdornment =
     rest.loading && hasMeaningfulValue(value) ? (
@@ -65,12 +66,13 @@ const GenericSelect = <
             ...inputProps.inputProps,
           }}
           disabled={rest.disabled}
+          aria-label={ariaLabel}
           // Only render placeholder if no values are selected
           placeholder={hasMeaningfulValue(value) ? undefined : placeholder}
           label={label}
         />
       )}
-      {...rest}
+      {...restOfProps}
     />
   );
 };
--- a/src/components/elements/input/GenericSelect.tsx
+++ b/src/components/elements/input/GenericSelect.tsx
@@ -20,6 +20,7 @@ export interface GenericSelectProps<
   > {
   label?: ReactNode;
   textInputProps?: TextInputProps;
+  ariaLabel?: string;
 }
 
 const GenericSelect = <
@@ -31,10 +32,10 @@ const GenericSelect = <
   label,
   textInputProps,
   options,
+  ariaLabel,
   ...rest
 }: GenericSelectProps<T, Multiple, Creatable>) => {
   const { placeholder, ...inputProps } = textInputProps || {};
-  const { ariaLabel, ...restOfProps } = rest;
   // Show a loading indicator if we have a value but the picklist is still loading
   const startAdornment =
     rest.loading && hasMeaningfulValue(value) ? (
@@ -72,7 +73,7 @@ const GenericSelect = <
           label={label}
         />
       )}
-      {...restOfProps}
+      {...rest}
     />
   );
 };
--- a/src/components/elements/input/GenericSelect.tsx
+++ b/src/components/elements/input/GenericSelect.tsx
@@ -65,9 +65,9 @@ const GenericSelect = <
             ...params.inputProps,
             value: rest.loading ? '' : params.inputProps.value,
             ...inputProps.inputProps,
+            'aria-label': ariaLabel,
           }}
           disabled={rest.disabled}
-          aria-label={ariaLabel}
           // Only render placeholder if no values are selected
           placeholder={hasMeaningfulValue(value) ? undefined : placeholder}
           label={label}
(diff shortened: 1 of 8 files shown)


src/components/elements/input/GenericSelect.tsx
```diff
@@ -65,6 +67,7 @@ const GenericSelect = <
             ...inputProps.inputProps,
           }}
           disabled={rest.disabled}
+          aria-label={ariaLabel}
```
gigxz: Ah thanks @martha , I also fixed this on a branch I'm working on https://github.com/greenriver/hmis-frontend/pull/676/files but I passed it through a separate prop so that the `aria-label` gets applied to the actual `input`, not the surrounding div. Aria label should go on interactive elements. Can you update to pass it through input props?

src/components/elements/input/LabeledCheckbox.tsx
```diff
@@ -56,6 +56,8 @@ const LabeledCheckbox = ({
     [onChange]
   );
 
+  const { ariaLabel, ...restOfProps } = props;
```
gigxz: Suggest moving to  `ariaLabel` line 28, we already destructure some other props there

src/components/elements/input/LabeledCheckbox.tsx
```diff
@@ -71,7 +73,8 @@ const LabeledCheckbox = ({
             ...labelSx,
           }}
           onChange={onChange}
-          {...props}
+          aria-label={ariaLabel}
```
gigxz: We only really need aria label on the checkbox for some reason the checkbox label is hidden. And in that case it should be passed to the `Checkbox` so it ends up on the `input` component. I don't think there is any point in adding an `aria-label` to a label like this.

src/components/elements/input/RadioGroupInput.tsx
```diff
@@ -118,7 +119,8 @@ const RadioGroupInput = ({
           }),
           ...sx,
         }}
-        {...props}
+        aria-label={ariaLabel}
```
gigxz, line 122: same here, not sure if it's in the right place. Is it on the element with the `radiogroup` role? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/radiogroup_role

src/modules/form/components/group/TableGroup.tsx
```diff
@@ -128,23 +129,25 @@ const TableGroup = ({
...
-                  {
-                    noLabel: true,
-                    inputProps: { ariaLabelledBy: tableHeaderInfo[index].id },
-                  }
-                )}
-              </TableCell>
-            ))}
-          </TableRow>
-        ))}
+        <TableBody>
```
gigxz, line 132: thanks!

src/modules/form/components/FormSelect.tsx
```diff
@@ -143,7 +145,8 @@ const FormSelect = <Multiple extends boolean | undefined>({
       groupBy={isGrouped ? (option) => option.groupLabel || '' : undefined}
       isOptionEqualToValue={isOptionEqualToValue}
       value={value}
-      {...props}
+      aria-label={ariaLabel}
```
gigxz: also has a dup on https://github.com/greenriver/hmis-frontend/pull/676/files#diff-2df75c89af9883b0513486e14ca3c47dc1b55e5ed219867025d72d005c6b4d63 that passes it to the input

src/components/elements/input/RadioGroupInput.tsx
martha: Yes, I think so! 
Screenshot 2024-03-04 at 10 27 00 AM

src/components/elements/input/TextInput.tsx
```diff
@@ -64,10 +65,15 @@ const TextInput = ({
         !props.multiline && e.key === 'Enter' && e.preventDefault()
       }
       autoComplete={formAutoCompleteOff}
+      aria-label={ariaLabel}
```
gigxz: I don't think we want the aria label in two places for the same input

src/components/elements/input/TextInput.tsx
```diff
@@ -64,10 +65,15 @@ const TextInput = ({
...
       {...props}
       sx={sx}
       inputProps={{
-        'aria-label': hiddenLabel ? String(label) : undefined,
```
gigxz, line 70: I think this can just be changed to `'aria-label': hiddenLabel ? ariaLabel || String(label) : undefined,`. It shouldn't be necessary to set aria-label "just because" -- if the input has a label, the TextField should already take care of making it accessible (eg the Label relating appropriately to the Input)

gigxz (commented):
whoops sorry @martha I had a couple unsubmitted comments

src/components/elements/input/TextInput.tsx
martha: That makes sense, yeah I was worried about dropping `ariaLabel` somewhere if it was provided so I probably erred on the side of propagating it too much.

PR #676, opened 2024-03-04
Bulk Services / Bed Nights v2

gigxz:
## V2 Bulk Services & Bed Nights

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4087

## How to Test
* BED NIGHTS: go to a night-by-night shelter project page
* BULK SERVICES: mark one or more CUstomServiceTypes as "supports_bulk_assignment" in the database, and go to the project page for a project that supports that service
* COC SELECTION: add multiple ProjectCoC records to the project
* UNIT ASSIGNMENT: add units to the project. when you enroll someone, they should be assigned to the next available unit
see also [qa instructions](https://www.pivotaltracker.com/story/show/187011913/comments/240472290)

## Type of change


src/hooks/useSearchParamState.ts
```diff
@@ -0,0 +1,136 @@
+import { isDate } from 'date-fns';
+import { isNil } from 'lodash-es';
+import { useSearchParams } from 'react-router-dom';
+import { formatDateForGql, parseHmisDateString } from '@/modules/hmis/hmisUtil';
+
+// adapted from https://github.com/jschwindt/react-use-search-params-state/tree/main
```
gigxz, line 7: check in -- should we add this as an npm package to give proper credit

src/modules/dataFetching/components/GenericTableWithData.tsx
```diff
@@ -360,14 +364,17 @@ const GenericTableWithData = <
       )}
       <Box sx={containerSx}>
         <GenericTable<RowDataType>
-          loading={loading && !data}
+          // if loadingVariant was passed explicitly, show the loading state for refetches.
+          // otherwise only show the initial loading state
+          loading={loadingVariant ? loading : loading && !data}
+          loadingVariant={loadingVariant}
```
gigxz: janky, revisit

src/modules/services/components/bulk/AssignServiceButton.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const { bulkAssign, bulkRemove, loading, apolloError } =
+    useBulkAssignMutations();
+
+  // Locally disable the button while the table refetches, AFTER this specific mutation has completed
+  const [localDisabled, setLocalDisabled] = useState(false);
+
+  // When the entire table has finished re-fetching, remove the loading state for this button.
+  useEffect(() => {
+    if (!tableLoading) setLocalDisabled(false);
+  }, [tableLoading]);
```
gigxz, line 38: Some funkiness here. `useBulkAssignMutations` uses apollo `refetchQueries` to specify that the table query should always refetch after a mutation completes. During that refetch, we want to keep this button disabled.

src/routes/routeUtil.ts
```diff
@@ -15,3 +16,39 @@ export const locationFromDefaultOrLogin = (location: Location): boolean => {
...
   return !!location.state?.fromLoginRedirect;
 };
+
+export const PREV_SEARCH_KEY = 'prev';
+export const prevSearchParam = () =>
+  `${PREV_SEARCH_KEY}=${encodeURIComponent(
+    window.location.pathname + window.location.search
+  )}`;
+
+export const backToFromSearchParam = (
```
gigxz: 🚨 this seems to be working great, but just want to flag it is new and kinda complicated. Constructs a `To` based on the `prev` search param if its present, with the ability to inject search params.

gigxz:
Alright @ttoomey @martha  - ready for review. I apologize for the enormity! I tried to call out some interesting parts in the comments.

This is up on QA (in a few minutes) as well.

( tagging @martha as secondary QA, so please don't feel the need to go through everything!)

src/api/operations/bedNight.queries.graphql
```diff
@@ -1,52 +0,0 @@
-query GetProjectEnrollmentsForBedNights(
```
martha, line 1: Is this being deleted because bed nights will only be used through bulk services from now on?

src/hooks/useSearchParamState.ts
martha: I guess the pros are giving proper credit and being able to benefit from improvements to that repo. But the con is adding a dependency to something that seems to be maintained by just one person, and isn't very widely used. Could this be a security concern as well?

We could look at whether there are other npm packages that are more popular and by extension maybe more trustworthy, maybe [this](https://www.npmjs.com/package/use-query-params?activeTab=readme) provides similar functionality?

src/modules/services/components/bulk/AssignServiceButton.tsx
martha: This seems sensible. Maybe it would be good to add what you've written here in the github PR comment as a comment in the code as well?

src/routes/routeUtil.ts
martha: This is cool! I would just say that the name `backToFromSearchParams` adds some cognitive load, haha! Even after understanding that `To` is the type of the object being returned, I had to concentrate to parse this, I think due to the `to` and `from` words right next to each other in the name. Also I wasn't sure `back` is needed in the name, maybe it's a synonym for `prev` and we should use that instead for consistency? maybe `prevUrlFromSearchParam` is the clearest, but on the other hand, that doesn't indicate that you are returning a type `To` instead of a type `URL`. Maybe `toPrevUrlFromSearchParam` although that's getting verboser.

Either way I also think it would be good to add a comment similar to what you've written here in gh, explaining what this is doing since the name is likely to end up a bit hard to parse either way

martha (commented):
Amazing ‼️

src/api/operations/bedNight.queries.graphql
gigxz: Yep

src/modules/services/components/bulk/BulkServicesTable.tsx
```diff
@@ -0,0 +1,198 @@
...
+        defaultFilters={defaultFilters}
+        // TODO: add user-facing filter options for enrolled clients and bed night date. No filter options for now.
+        filters={(f) => omit(f, 'project', 'organization')}
+        filterInputType='ClientFilterOptions'
+        defaultSortOption={
+          searchTerm
+            ? ClientSortOption.BestMatch
+            : ClientSortOption.LastNameAToZ
+        }
+        EnhancedTableToolbarProps={{
```
ttoomey, line 189: When you've scrolled into the table this top header isn't visible. Then when selecting a checkbox it's not clear why all the row buttons are disabled. What if we made this bar stick to the top (or bottom) so it's always visible to the user? Or maybe animate-in even?

src/routes/routeUtil.ts
ttoomey: I think this is totally fine for now but I'd want to think hard about adopting this approach elsewhere. 

I think it's great to put navigational, filter, and pagination params into the URL; it's a better user experience imo. However I'm a little unsure about using the URL for workflow state that is internal / ephemeral.

An alternative might be something like react router location state that wouldn't expose this to the user:

```
navigate({pathname: bednights}, {state: })
```

src/modules/services/types.ts
```diff
@@ -0,0 +1,21 @@
+import { sub } from 'date-fns';
+
+export type ServicePeriod = {
+  start: Date;
+  end: Date;
+};
+
+const today = new Date();
```
ttoomey, line 8: I think there's a bug here. If the day rolls over while the app is open, today will not update. Not sure how much of a problem it would be but it might impact staff working late at a shelter. This probably wants to be a hook which would let the date at least update when components mounts. Ideally it would be a useCurrentDate hook.

maybe something like this (not tested):
```ts
import { endOfDay, differenceInMilliseconds } from 'date-fns';
import { useState, useEffect } from 'react';

const timeUntilMidnight = () => {
  const now = new Date();
  const midnight = endOfDay(now);
  return differenceInMilliseconds(midnight, now);
};

const useCurrentDate = () => {
  const [value, setValue] = useState(new Date());

  useEffect(() => {
    let timeout = setTimeout(function tick() {
      setValue(new Date());
      timeout = setTimeout(tick, timeUntilMidnight());
    }, timeUntilMidnight());

    return () => clearTimeout(timeout);
  }, []);

  return value;
};

export default useCurrentDate;
```

src/config/theme.ts
```diff
@@ -250,6 +250,7 @@ const createThemeOptions = (theme: Theme) => ({
       },
       styleOverrides: {
         root: {
+          whiteSpace: 'nowrap',
```
ttoomey: This means button text won't wrap, right? Is this to handle wide buttons on mobile? I wonder if it would better to allow the text to wrap when needed and adjust the container to better manage the layout. For example, using flex-wrap:wrap on the container and flex-shrink:0 on the buttons.

If button text can't wrap, you could have buttons that overflow the container causing weird horizontal scroll.

src/hooks/useSearchParamState.ts
ttoomey: So the intention is to centralize casting query params from strings. It also provides a properly typed setter for params. That seems reasonable.

I expect we will need to modify this code, for example adding encrypted fields. So it's reasonable to keep it here for us to hack on.

The use-query-params lib is more widely used but it looks like it may not have an active maintainer (1 yr since last commit).

src/hooks/useSearchParamState.ts
```diff
@@ -0,0 +1,137 @@
...
+  });
+  return allUrlParams;
+};
+
+const useSearchParamsState = (paramsDefinition: SearchParamsStateType) => {
+  const [searchParams, setSearchParams] = useSearchParams();
+
+  const values = getValues(paramsDefinition, searchParams);
+
+  const setValues = (newValues: Record<string, any>) => {
```
ttoomey: memoize?
```suggestion
  const setValues = useCallback((newValues: Record<string, any>) => {
```

src/hooks/useSearchParamState.ts
```diff
@@ -0,0 +1,137 @@
...
+      allUrlParams[key] = value;
+    }
+  });
+  return allUrlParams;
+};
+
+const useSearchParamsState = (paramsDefinition: SearchParamsStateType) => {
+  const [searchParams, setSearchParams] = useSearchParams();
+
+  const values = getValues(paramsDefinition, searchParams);
```
ttoomey: memo?
```suggestion
  const values = useMemo(() => getValues(paramsDefinition, searchParams), [paramsDefinition, searchParams]);
```

src/modules/services/components/bulk/BulkServicesPage.tsx
```diff
@@ -0,0 +1,271 @@
...
+    );
+  }, [project, navigate, serviceTypeIdProp, setFilterParams]);
+
+  const stepThreeTitle = `Assign ${pluralize(serviceTypeName, 2)}`;
+
+  return (
+    <>
+      <PageTitle title={title} />
+      <Grid container rowSpacing={1}>
+        <Grid item sm={12} md={8} lg={8} xl={4}>
```
ttoomey: Do we need to encapsulate the steps in a Grid? This is subjective but I think it looks strange with the step cards being different widths. Maybe simpler to let them all be 100% width, take out the grid, and just throw margin bottom on each card? Cuts some of the indentation out this component too, which would be nice

src/modules/services/components/bulk/BulkServicesPage.tsx
```diff
@@ -0,0 +1,271 @@
...
+                    setFilterParams({ serviceTypeId: option?.code })
+                  }
+                  label='Service Type'
+                  bulk
+                />
+              )}
+              <DatePicker
+                value={serviceDate}
+                onChange={(date) => setFilterParams({ serviceDate: date })}
+                max={new Date()}
```
ttoomey, line 162: might be another place for a hypothetical useCurrentDate()

src/hooks/useSearchParamState.ts
```diff
@@ -0,0 +1,137 @@
...
+        allUrlParams[key] = [allUrlParams[key], value];
+      }
+    } else {
+      allUrlParams[key] = value;
+    }
+  });
+  return allUrlParams;
+};
+
+const useSearchParamsState = (paramsDefinition: SearchParamsStateType) => {
```
ttoomey, line 108: right now, the values returned from this hook are `any`. If we are going to use this utility more broadly, we should probably fix up the types to match the paramsDefinition using generics. Something for a future pass

src/modules/services/components/bulk/ServiceDateRangeSelect.tsx
```diff
@@ -0,0 +1,195 @@
...
+          renderOption={renderOption}
+          disableClearable
+        />
+        <Button
+          color='error'
+          variant='outlined'
+          onClick={handleClear}
+          disabled={disabled || !selected}
+        >
+          Clear Results
```
ttoomey: I think something like "Reset Search" is more clear

src/modules/services/components/bulk/BulkServicesPage.tsx
```diff
@@ -0,0 +1,271 @@
...
+                  onSearch={(value) => setFilterParams({ searchTerm: value })}
+                  onClearSearch={() => setFilterParams({ searchTerm: null })}
+                  label={null}
+                  placeholder='Client Name, DOB, SSN or ID'
+                  helperText='Search includes all of HMIS'
+                  ClearButtonProps={{
+                    color: 'error',
+                    startIcon: null,
+                    variant: 'outlined',
+                    children: 'Clear Results',
```
ttoomey: There may be no results found so I think something like "Reset Search" is more clear

src/modules/services/components/bulk/AssignServiceButton.tsx
ttoomey: would it work to just disable the buttons when tableLoading is true?
```jsx
<LoadingButton disabled={disabled || localDisabled || tableLoading} ... />
```

ttoomey (commented):
This is awesome! This is a huge piece of work and it's come together really well! 🙌 

I left a few suggestions and ideas but there's nothing that should be a merge blocker

src/modules/services/components/bulk/BulkServicesTable.tsx
gigxz: I noticed the same issue. I will add a tooltip to the disabled buttons so at least you can get a message there.
[commented in figma](https://www.figma.com/file/F9EHR2IZYuMAlOJ9aOG2IL?node-id=2088:66116&mode=design#738730183), we can add a separate story.

src/modules/services/types.ts
gigxz: Good catch, thanks

src/config/theme.ts
gigxz: Removed this change

src/modules/services/types.ts
gigxz: Although, I don't know. I don't think we want that behavior for this page. If they reach midnight while they are using the application to record bednights, they are still going to want to record it for the same date.. you don't want to record bed nights for 3/14 at 1am on 3/14 - you would still be recording for 3/13 at that point. See what I mean?

src/modules/services/components/bulk/ServiceDateRangeSelect.tsx
gigxz: I agree

src/modules/services/components/bulk/AssignServiceButton.tsx
gigxz: > would it work to just disable the buttons when tableLoading is true? 

that would disabled all the buttons in the table. we only want to disable the buttons that just performed a mutation

src/modules/services/components/bulk/BulkServicesPage.tsx
gigxz: Design has staggered cards like this. But I feel like the complexity isnt really worth it.. removed

src/routes/routeUtil.ts
gigxz: Cool,  thanks both. I updated to use react-router-dom state for the household navigation piece: https://github.com/greenriver/hmis-frontend/pull/676/commits/ec0e90833ef288efaaec4c0ba295cf5e1af64f9e

src/hooks/useSearchParamState.ts
gigxz: leaving this as is. per theron's other comment, too, if we use it going forward we need to make it type safe. [ticket for that](https://www.pivotaltracker.com/story/show/187258786)

src/modules/services/types.ts
ttoomey: > If they reach midnight while they are using the application to record bednights, they are still going to want to record it for the same date

That makes sense and it should be fine as-is for now. However I think that it would be preferable to implement the logic explicitly in the component with an offset (the default date rolls over at 6am or something). That means if you reload the page after midnight, you'd get yesterday's date as desired.

gigxz: 👍  https://www.pivotaltracker.com/story/show/187263616

PR #679, opened 2024-03-04
187040557 external forms

ttoomey:
## Description

[PT Issue](https://www.pivotaltracker.com/story/show/187040557)

](https://github.com/greenriver/hmis-warehouse/pull/4037)

Note: @martha is the primary author on this PR

## Testing

To test locally:
* see the backend PR for specific testing instructions
* visit https://hmis-warehouse.dev.test/hmis_external_api/external_forms/tchc/prevention_screening
* form submissions should appear in the HMIS 
* give your roles the 'can_manage_external_form_submissions' permission
* add a form rule to enable the prevention form on a given project (or org-wide)
* The submissions table in the interface does not initially include the columns (first name, last name). To make these show up, you may need to set show_in_summary=true on the relevant cdeds

## Type of change


src/api/operations/externalFormSubmission.queries.graphql
```diff
@@ -0,0 +1,17 @@
+query projectExternalFormSubmissions(
```
martha: Suggestion -- what do you think of renaming this to `GetProjectExternalFormSubmissions` for consistency with other graphql queries?

ttoomey: sounds good, I adjusted the query name

gigxz:
can you change the base to release-108 at some point? thanks!

src/modules/projects/components/ProjectExternalFormSubmissions.tsx
```diff
@@ -0,0 +1,68 @@
...
+      {loading && <Loading />}
+      {!loading && pickList && pickList.length > 1 && (
+        <FormSelect
+          sx={{ mb: 2 }}
+          options={pickList || []}
+          value={selectedFormType}
+          onChange={(e, val) =>
+            setSelectedFormType(isPickListOption(val) ? val : null)
+          }
+          label={getRequiredLabel('Select form type', false)}
```
gigxz:
```suggestion
          label={getRequiredLabel('Form Type', false)}
```
We usually would just do "Form Type" rather than including "Select" in the label. Maybe the designs were different, but that's the pattern we have elsewhere, so thats what I'd suggest!

src/routes/protected.tsx
```diff
@@ -326,6 +327,14 @@ export const protectedRoutes: RouteNode[] = [
               </ProjectEditRoute>
             ),
           },
+          {
+            path: ProjectDashboardRoutes.EXTERNAL_FORM_SUBMISSIONS,
+            element: (
+              <ProjectEditRoute>
```
gigxz: This will make it so the page automatically resolves `<NotFound />` if user lacks `canManageExternalFormSubmissions` for this project. 
```suggestion
              <ProjectEditRoute permissions={['canManageExternalFormSubmissions']}>
```

src/modules/projects/components/tables/ProjectExternalSubmissionsTable.tsx
```diff
@@ -0,0 +1,198 @@
...
+  const submissionValues = useMemo(() => {
+    if (!selected) return {};
+    const itemMap = getItemMap(selected.definition.definition, true);
+    const submissionValues: FormValues = {};
+    Object.keys(itemMap).forEach((key) => {
+      const pickListOptions = itemMap[key].pickListOptions;
+      const customFieldKey = itemMap[key].mapping?.customFieldKey;
+      if (!customFieldKey) return;
+
+      const value = customDataElementValueForKey(
```
gigxz, line 94: This could be a single value or an array or values, depending on whether the custom field `repeats` or not. (That would be like a multi-select dropdown).

To accommodate that (untested):
```
# if item has a picklist, convert value to PickListOption(s) so we can display the readable label
if (item.pickListOptions) {
  if (Array.isArray(value)) {
     submissionValues[key] = value.map(v => getOptionValue(v, item);
  } else {
     submissionValues[key] = getOptionValue(value, item);
  }
} else {
   submissionValues[key] = value;
}
```

src/modules/form/hooks/useStaticFormDialog.tsx
```diff
@@ -90,6 +92,7 @@ export function useStaticFormDialog<
             }}
           >
             <Grid container spacing={2} sx={{ mb: 2, mt: 0 }}>
+              {beforeFormComponent && <Grid item>{beforeFormComponent}</Grid>}
```
gigxz, line 95: `beforeFormComponent` needs to go in the useCallback dependency array

src/api/operations/externalFormSubmission.fragments.graphql
```diff
@@ -0,0 +1,13 @@
+fragment ExternalFormSubmissionFields on ExternalFormSubmission {
+  id
+  submittedAt
+  spam
+  status
+  notes
+  definition {
+    ...FormDefinitionFields
```
gigxz: This gave me pause, because I don't think there's anywhere else where we load the full FormDefinition JSON on each row in a table. This is going to return a lot of data that is repeated, because we're only ever showing a table for one form definition.

This is really just an optimization, but I think it's worth doing. The way I would go about it is:
* In the `ProjectExternalFormSubmissions` component, fetch the FormDefinition when an identifier is selected. I think this would require adding a new query `external_form_definition(identifier:)`  since we currently only do lookups by role or id.
* Stop resolving definition on each row item
* Replace calls to `selected.definition.definition` with the single definition that was passed to `ProjectExternalSubmissionsTable` as a prop

gigxz (commented):
Looks great @martha! see comments inline

src/api/operations/externalFormSubmission.fragments.graphql
martha: @gigxz  sounds good, I created a PR here for the backend change to add that query: https://github.com/greenriver/hmis-warehouse/pull/4158/files

The frontend is still giving me some trouble ("Uncaught Invariant Violation: Missing field 'cacheKey' while extracting keyFields from {"identifier":"tchc_prevention_screening","__typename":"FormDefinition"}") altho the `cacheKey` does appear to exist ... I'll keep poking at it!

ttoomey: Hi @martha - we configure apollo to use definition.cacheKey for FormDefinitions (see the typePolicy in src/providers/apolloClient.tsx:150). Maybe the cacheKey isn't being selected in your query? Happy to pair if it would be helpful

martha: thanks @ttoomey for the help!! :)

src/modules/projects/components/tables/ProjectExternalSubmissionsTable.tsx
martha: I tested this out by locally setting one of the dropdowns to have `repeats` as true and then adding an extra CDE value for it in the db. It looked good to me!

src/modules/projects/components/ProjectExternalFormSubmissions.tsx
```diff
@@ -0,0 +1,68 @@
...
+          onChange={(e, val) =>
+            setSelectedFormType(isPickListOption(val) ? val : null)
+          }
+          label={getRequiredLabel('Form type', false)}
+          placeholder='Select a form type'
+        />
+      )}
+
+      {formType && (
+        <Paper sx={{ p: 2 }}>
```
gigxz: This isn't consistent with patterns elsewhere in the app - we almost always have no padding on table cards
```suggestion
        <Paper>
```

gigxz (approved):
looks great!!

ttoomey:
Ah oops, this got lost. Thought I merged this in a few days ago. Merging in now

gigxz:
whoops thanks!

PR #680, opened 2024-03-05
Only show existing CustomDataElements in table columns

gigxz:
## Description

Adapt to new behavior from the backend (https://github.com/greenriver/hmis-warehouse/pull/4086) whereby only CustomDataElements _with value(s)_ are resolved.

The TableSummary custom data element columns will only be shown if any rows **on this page** have a value for them.

Easiest way to test is to use the Story.

## Type of change
- [x] Bug fix


src/modules/hmis/hmisUtil.ts
```diff
@@ -612,25 +613,40 @@ export const dataUrlForClientImage = (
...
-    .filter((cde) => cde.displayHooks.includes(DisplayHook.TableSummary))
-    .map((cde) => ({
-      header: cde.label,
-      key: cde.key,
+  function generateColumnDefinition(cded: CustomDataElementFieldsFragment) {
+    return {
+      header: cded.label,
+      key: cded.key,
       render: (row: RowType) => {
         const thisCde = row.customDataElements.find(
```
martha, line 629: I wonder if it would be possible/desirable to change the api so that `row.customDataElements` returns a map instead of an array?

martha (approved):

src/modules/hmis/hmisUtil.ts
ttoomey: I don't think Maps, Sets or other js specific structures are supported by graphql natively. You could probably implement some kind of custom scalar to support it but I'm guessing that would be tricky

ttoomey (approved):
LGTM, thanks for adding stories!

PR #681, opened 2024-03-05
Add client fields as local constants on assessments

martha:
## Description

PT issue https://www.pivotaltracker.com/story/show/187136045
This change allows us to pull in info about the client -- DOB, name, and SSN -- to autofill in assessments.

I have a few questions:
- Is pulling SSN in here like this okay? The client is from the enrollment, so the most obvious way to do it (and what I've implemented here) is to add client's ssn on the enrollment fragment. But of course I can think of lots of reasons not to do that, and it's a huge change! Is there a better way?
- Also related to SSN, should it be auto-filled but hidden unless the user clicks, like on the client profile?
- And, should we do any checking of the `canViewFullSsn` or `canViewPartialSsn` before passing the ssn around?
- Just to make sure, it looks like the UHA form has space for a lot of household members, but this change is only putting the info for the head of household into local constants. Is this okay?

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/api/operations/client.fragments.graphql
+++ b/src/api/operations/client.fragments.graphql
@@ -80,6 +80,12 @@ fragment ClientNameDobVet on Client {
   veteranStatus
 }
 
+fragment ClientNameDobSsn on Client {
+  ...ClientName
+  ssn
+  dob
+}
+
 fragment ClientNameObjectFields on ClientName {
   id
   first
--- a/src/api/operations/client.fragments.graphql
+++ b/src/api/operations/client.fragments.graphql
@@ -80,12 +80,6 @@ fragment ClientNameDobVet on Client {
   veteranStatus
 }
 
-fragment ClientNameDobSsn on Client {
-  ...ClientName
-  ssn
-  dob
-}
-
 fragment ClientNameObjectFields on ClientName {
   id
   first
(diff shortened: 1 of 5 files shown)


src/modules/assessments/components/AssessmentForm.tsx
```diff
@@ -147,9 +148,14 @@ const AssessmentForm: React.FC<Props> = ({
       entryDate: enrollment.entryDate,
       exitDate: enrollment.exitDate,
       projectName: enrollment.project.projectName,
+      clientFirstName: client.firstName,
```
martha, line 151: I just realized that only ssn and dob were requested in the ticket, should I remove the name fields?

src/api/operations/client.fragments.graphql
```diff
@@ -80,6 +80,12 @@ fragment ClientNameDobVet on Client {
   veteranStatus
 }
 
+fragment ClientNameDobSsn on Client {
```
gigxz: Interesting, we don't usually take this approach of defining a fragment that is not used in any queries. I think we should either (1) define this type in the code, ` type ClientNameDobSsn = ClientNameFragment & {ssn?: string, dob?:string}` or, (2)  add `... ClientNameDobSsn` fragment to the AllEnrollmentDetails.client fields

src/modules/assessments/components/AssessmentForm.tsx
gigxz: We can keep them, thanks

gigxz:
> Is pulling SSN in here like this okay? The client is from the enrollment, so the most obvious way to do it (and what I've implemented here) is to add client's ssn on the enrollment fragment. But of course I can think of lots of reasons not to do that, and it's a huge change! Is there a better way?

I think that is appropriate

> Also related to SSN, should it be auto-filled but hidden unless the user clicks, like on the client profile?

That would be nice to have but out of scope I think

> And, should we do any checking of the canViewFullSsn or canViewPartialSsn before passing the ssn around?

No need, the ssn value will be null if user lacks access, or partial if thats what the user can see

> Just to make sure, it looks like the UHA form has space for a lot of household members, but this change is only putting the info for the head of household into local constants. Is this okay?

Yep

gigxz (approved):

PR #684, opened 2024-03-05
Wrap page title with actions on mobile

martha:
## Description

This PR wraps the PageTitle action buttons around onto the next line on small screens. It also removes padding on the whole page for mobile. This is a big change that affects every screen! 😮
https://www.pivotaltracker.com/n/projects/2591838/stories/187166502

## Type of change
[//]: # 'remove options that are not relevant'
- [ ] Bug fix
- [x] New feature (adds functionality)
- [ ] Code clean-up / housekeeping
- [ ] Dependency update

## Checklist before requesting review
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [x] My code includes comments and/or descriptive variable names to help other engineers understand the intent (or not applicable)
- [x] My code follows the style guidelines of this project (eslint)
- [x] I have updated the documentation (or not applicable)
- [x] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

PR #687, opened 2024-03-07
Fix padding for page containers on mobile

martha:
## Description

This is a follow-up PR from this comment: https://github.com/greenriver/hmis-frontend/pull/684#discussion_r1513612926
Related to this ticket: https://www.pivotaltracker.com/story/show/187166502

I looked around at a few options for places where we could solve this padding problem for dashboards, and I think it's probably best to add this common padding to `DashboardContentContainer`, which is used in common by the Client, Admin, Enrollment, and Project dashboards.

Then, I added the same change to PageContainer.tsx, which seems to be used in a few other places -- I just noticed it because of toggling back and forth between the All Projects tab and the Projects dashboard while testing.

Let me know what you think! There may be other places in the app that want similar updates too?

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/components/layout/dashboard/DashboardContentContainer.tsx
+++ b/src/components/layout/dashboard/DashboardContentContainer.tsx
@@ -116,8 +116,8 @@ const DashboardContentContainer: React.FC<Props> = ({
           )}
           <Box
             sx={{
-              py: { xs: 2 },
-              px: { xs: 0, sm: 1, lg: 4 },
+              py: 2,
+              px: { xs: 1, sm: 3, lg: 4 },
               maxWidth: `${maxPageWidth}px`,
             }}
           >
--- a/src/components/layout/dashboard/DashboardContentContainer.tsx
+++ b/src/components/layout/dashboard/DashboardContentContainer.tsx
@@ -116,7 +116,8 @@ const DashboardContentContainer: React.FC<Props> = ({
           )}
           <Box
             sx={{
-              py: 2,
+              pt: 2,
+              pb: 8,
               px: { xs: 1, sm: 3, lg: 4 },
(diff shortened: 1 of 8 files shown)


gigxz:
Some suggestions on https://github.com/greenriver/hmis-frontend/pull/690

src/components/layout/dashboard/DashboardContentContainer.tsx
```diff
@@ -116,8 +116,9 @@ const DashboardContentContainer: React.FC<Props> = ({
           )}
           <Box
             sx={{
-              py: { xs: 2 },
-              px: { xs: 0, sm: 1, lg: 4 },
+              pt: 2,
+              pb: 8,
```
martha, line 120: `pb=8` here because of the 2 that existed before from `py=2` plus `pb=6` that I removed from the children

src/modules/assessments/components/household/HouseholdAssessments.tsx
```diff
@@ -269,8 +269,8 @@ const HouseholdAssessments: React.FC<Props> = ({
           top: STICKY_BAR_HEIGHT + CONTEXT_HEADER_HEIGHT,
           backgroundColor: 'white',
           px: { sm: 3, lg: 5 },
-          ml: { xs: -2, sm: -3, lg: -4 },
-          mt: { xs: -1, sm: -2 },
+          ml: { xs: -1, sm: -3, lg: -4 },
```
martha, line 272: This nav bar for household assessments still needs more work to be mobile friendly, but here's how it looks now:
Screenshot 2024-03-08 at 2 34 41 PM

gigxz (approved):
nice ⚡

PR #688, opened 2024-03-07
Sketch a possible solution to DynamicForm readonly bug

martha:
## Description

PT issue: https://www.pivotaltracker.com/story/show/187136045
Specific comment in the deep comment section on that issue that gave rise to this PR: https://www.pivotaltracker.com/story/show/187136045/comments/240506104

Currently, the Dynamic Forms aren't displaying an individual component within an editable form as read-only. This is because in [useDynamicFields](https://github.com/greenriver/hmis-frontend/blob/5a89dcddcac5be5b9fabc9900994418acc66c3c2/src/modules/form/hooks/useDynamicFields.tsx#L265), it's not decided case-by-case whether to display each item field as view-only, but rather, it's decided whether to use the top-level components DynamicViewFields or DynamicFormFields based on whether the whole form is viewOnly or not.

Here was my thought around this sketch of a possible solution: We want to keep the current structure (recursive from the root using different render functions depending on whether the whole form is view-only or not). We want to add additional functionality on top of that, which is that the dynamic form fields should be able to render an individual dynamic view field if the item itself is read only, but (based on my understanding of current usage) this doesn't have to be in a nested/recursive way.

Separately, also, this PR introduces a small change to render readonly SSNs with dashes in them.

## Type of change
- [x] Bug fix
- [ ] Dependency update


src/modules/form/components/viewable/DynamicViewItem.tsx
```diff
@@ -0,0 +1,55 @@
+import { Grid } from '@mui/material';
```
martha: This new component is extracted out from the deleted code in DynamicViewFields.tsx, to avoid repeated code - but maybe it's not worth it

src/modules/form/components/viewable/DynamicViewField.tsx
```diff
@@ -114,6 +114,16 @@ const DynamicViewField: React.FC<DynamicViewFieldProps> = ({
       );
     case ItemType.Text:
     case ItemType.String:
+      if (item.component === Component.Ssn) {
+        return (
+          <TextContent
+            {...commonProps}
+            renderValue={(val) =>
+              `${val.slice(0, 3)}-${val.slice(3, 5)}-${val.slice(5, 9)}`
```
gigxz: I think we can skip this behavior, it would need more special casing to support partial SSN values

src/modules/form/components/viewable/DynamicViewItem.tsx
gigxz: Thank you for taking initiative on this one @martha ! Hmm, I think I prefer not to add a new factored-out component (`DynamicViewItem`). There wouldn't be much repeated code if we do it like this in `DynamicFormFields.tsx`:

```
    const itemComponent = item.readOnly ? (
      <DynamicViewField
        item={item}
        value={values[item.linkId]}
        disabled={isDisabled}
        nestingLevel={nestingLevel}
        horizontal={horizontal}
        pickListArgs={pickListArgs}
        // Needed because there are some enable/disabled and autofill dependencies that depend on PickListOption.labels that are fetched (PriorLivingSituation is an example)
        adjustValue={itemChanged}
        {...props}
      />
    ) : (
      // render DynamicField in the same way we do now
    );

  // ...  renderFn code is already present
```

gigxz: I added a quick story to test out how this looks. Can you commit it to `DynamicForm.stories.tsx`?

```
# Render the DynamicForm with values, modified such that every item is marked "readOnly" in the FormDefinition
export const WithValuesAsReadOnly = Template.bind({});
WithValuesAsReadOnly.args = {
  definition: modifyFormDefinition(
    formDefinition,
    (item) => (item.readOnly = true)
  ),
  initialValues: ViewStory.args?.values,
  errors: emptyErrorState,
};

```

gigxz (commented):
thank you 🙌 ! a couple small suggestions, but otherwise looks great

gigxz (approved):

PR #689, opened 2024-03-07
Add basic external submissions table

martha:
## Description

PT ticket: https://www.pivotaltracker.com/story/show/187040557
This PR sets up the skeleton for the external form submissions reviewing interface.
Filtering, sorting, and opening the submissions in a modal to edit them/mark as reviewed isn't included yet.

Question: I got custom data elements working using the show_in_summary attribute and it looks pretty cool! To do this, I just went in the DB for custom data element definitions and manually set show_in_summary to true on a few columns. Is there a better way to do this?

## Type of change
[//]: # 'remove options that are not relevant'
- [ ] Bug fix
- [x] New feature (adds functionality)
- [ ] Code clean-up / housekeeping
- [ ] Dependency update

## Checklist before requesting review
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [x] My code includes comments and/or descriptive variable names to help other engineers understand the intent (or not applicable)
- [x] My code follows the style guidelines of this project (eslint)
- [x] I have updated the documentation (or not applicable)
- [x] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

PR #691, opened 2024-03-08
Create dialog for External Submissions review

martha:
## Description

This PR adds graphql schema updates, and code for the dialog box that should open when a row of the form submissions table is clicked.

This PR builds on #689. ~Still not included are sorting/filtering functionalities and a delete button + mutation.~

Also, I have been noticing some bugs with field values that aren't displayed in the form. I'm still investigating this, but I think there are two separate issues going on:
1. boolean fields are coming back as "1"/"0" instead of "yes"/"no"
2. some fields are not getting saved successfully as CDEs, I'm seeing this on `dob` and `explanation_of_non_payment_of_rent` among others

Depends on hmis-warehouse PR:

## Type of change
- [ ] Bug fix
- [ ] Dependency update


ttoomey (approved):
Looks good!

> Also, I have been noticing some bugs with field values that aren't displayed in the form. 

Thanks for calling this out. Probably related to storing everything as a string. It may be that the shortest path is to adjust the definitions to use the right form type and process the CDE's accordingly, right now all values end up in cde.value_string.

gigxz (approved):

gigxz (approved):
looks good to me!

src/modules/projects/components/tables/ProjectExternalSubmissionsTable.tsx
```diff
@@ -0,0 +1,192 @@
...
+}) => {
+  const getColumnDefs = useCallback((rows: ExternalFormSubmissionFields[]) => {
+    const customColumns = getCustomDataElementColumns(rows);
+    return [
+      {
+        header: 'Status',
+        linkTreatment: false,
+        render: (s: ExternalFormSubmissionFieldsFragment) => (
+          <Chip
+            label={capitalize(s.status)}
```
ttoomey: maybe these chips could be a little less heavy on the page
```suggestion
            variant='outlined'
            size='small'```

ttoomey (approved):
Looks great, thanks!

src/modules/projects/components/tables/ProjectExternalSubmissionsTable.tsx
martha: Thanks @ttoomey , the `small` size is indeed closer to the mocks so thank you for pointing that out! I also added the `outlined` variant for reviewed submissions. And I noticed the color of the chip for reviewed submissions should be `text.secondary`, so I've added it here.

I added it as `sx` after spending a little time trying to get the Chip to accept `text.secondary` as a color prop, using `ChipPropsColorOverrides`. I ran into issues with this -- similar to some of the problems described [here](https://stackoverflow.com/questions/69481467/mui-error-cannot-read-properties-of-undefined-reading-drawer) and in some of the comments [here](https://stackoverflow.com/questions/64932595/how-do-i-extend-the-color-palette-from-mui-with-typescript) -- basically Chip color properties can't be quoted and they require a contrast color to be defined. Anyway, I still think this is possible but decided not to proceed, because I think we lose some of the benefits if it has to be `textSecondary` instead of `text.secondary` because of that quote issue. Let me know what you think.

Here's how it looks for now:
Screenshot 2024-03-14 at 1 54 49 PM

ttoomey: Thanks Martha, that looks good! In the future maybe we can extract this into a common component that wraps Mui's Chip (CommonChip or something).

PR #693, opened 2024-03-11
Add delete button to external form review

martha:
## Description

This adds a delete button and mutation to the external form submission review interface.

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4105
PT https://www.pivotaltracker.com/story/show/187040557

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/projects/components/ProjectSubmissionsDialog.tsx
```diff
@@ -54,6 +58,24 @@ const ProjectSubmissionsDialog: React.FC<Props> = ({
             definition={submission.definition.definition}
           />
         </Box>
+        <DeleteMutationButton<
```
martha, line 61: I wanted to call out that this delete button isn't shown in the correct place in the model, compared to the [mock](https://www.figma.com/file/F9EHR2IZYuMAlOJ9aOG2IL/OP-%2F-HMIS-%2F-Phase-2-(2024)?type=design&node-id=1883-14891&mode=design&t=dG81G2oIzUSMl81h-0), because of using the static form. Let me know if you think I should incorporate an option for a delete button, perhaps in the FormActionProps?

ttoomey: Thanks for calling this out, it does seem kind of odd to place it above the form. It's probably good enough for MVP though.

I'm also not sure about adding the button to the form actions since it makes it less clear what primary action the user should take.

Perhaps something to discuss on our design call; it would be nice to find a general treatment for these actions. One idea is to use a menu on the index page. In this case each submission row could have a menu with [Review Submission, Delete Submission]. We use this pattern on the admin->users page. This reduces visual clutter and scales pretty well when there are many actions

ttoomey (approved):
Looks good!

src/modules/projects/components/ProjectSubmissionsDialog.tsx
gigxz: Can you use `useStaticFormDialog` instead of adding this new dialog? It seems similar, you would probably need to add a prop to allow you to inject a ReactNode above the StaticForm (so you can show the DynamicView). 

For the delete button, you just pass it in `otherActions` prop to the `renderFormDialog` function returned by that hook.

martha: Yes, much better, thank you! I made that update [here](https://github.com/greenriver/hmis-frontend/pull/691)

PR #697, opened 2024-03-15
Improve label widths in dynamic form

martha:
## Description

[PT#187186982](https://www.pivotaltracker.com/story/show/187186982)

I got a little mired in the many different settings for widths, minWidths, and maxWidths on dynamic form fields. I found that the code already accommodates for the idea that the maxWidth of the labeled input is not the same as the width of the input itself. However, this wasn't consistently applied.

In this PR, I've tried to keep the existing approach, apply it consistently, and improve on it by avoiding using an explicit css classname ('.MuiInputBase-root') in favor of passing in `sx` on `InputProps`. I also tried to remove width-related code that wasn't doing anything, to avoid future confusion -- I'll point out in comments a few spots where this was the case.

I have an overarching question: Given that sometimes we have forgotten to add the correct width logic, should we hoist this up and make it live somewhere more central, like in the `InputContainer` for example, which seems to be used pretty consistently? `InputContainer` has its own maxWidth and minWidth applied, but they appear to be getting clobbered by the child values. (e.g. I debugged and saw that max width on the `InputContainer` was set to 574, but child components were taking on width of 600px.)

The easiest way to test is to check the DynamicForm storybook's new "With Extra Long Labels" variant. A few things I didn't fix in this PR, because they're not input labels so could be considered out of scope:
- Readonly fields (DynamicViewFields)
- Alerts
- Titles of item groups

However these might be easy fixes to pull in and I'm happy to do so if you suggest it!

## Type of change
- [x] Bug fix
- [ ] Dependency update


src/components/elements/input/RadioGroupInput.tsx
```diff
@@ -117,7 +117,6 @@ const RadioGroupInput = ({
               borderRadius: 1,
             },
           }),
-          ...sx,
```
martha, line 120: I checked usages of `RadioGroupInput` and found that it is only used in `DynamicField.tsx` and `YesNoRadio.tsx`, which in turn is only used in `DynamicField.tsx`, so I think this change is safe

src/modules/form/components/DynamicField.tsx
```diff
@@ -245,7 +247,6 @@ const DynamicField: React.FC<DynamicFieldProps> = ({
               sx={{
                 width,
                 maxWidth: MAX_INPUT_AND_LABEL_WIDTH,
-                '.MuiInputBase-root': { maxWidth: MAX_INPUT_WIDTH },
```
martha, line 248: I removed this because it's not used -- the SSN component's input widths are set directly in the component

src/modules/form/components/DynamicField.tsx
```diff
@@ -218,6 +218,7 @@ const DynamicField: React.FC<DynamicFieldProps> = ({
                 })
```
martha: I can't leave a comment on line 204, but I was wondering if you remember @gigxz  the reason for adding `+ 100` here for display types?

src/modules/form/components/DynamicField.tsx
```diff
@@ -294,7 +301,7 @@ const DynamicField: React.FC<DynamicFieldProps> = ({
             <MinutesDurationInput
               value={isNil(value) ? '' : value}
               onChange={onChangeValue}
-              inputWidth={width}
```
martha, line 297: `inputWidth` was not being used here, since MinutesDurationInput sets the width of the input fields directly

src/modules/form/components/DynamicField.tsx
```diff
@@ -306,8 +313,9 @@ const DynamicField: React.FC<DynamicFieldProps> = ({
             onChange={onChangeEvent}
             horizontal={horizontal}
             currency={item.type === ItemType.Currency}
-            inputWidth={width}
+            inputWidth='120px'
```
martha: For `NumberInput`, the `inputWidth` was not getting passed down, and instead the width was getting set [here](https://github.com/greenriver/hmis-frontend/blob/release-108/src/components/elements/input/TextInput.tsx#L52-L55)  in the `TextInput` component. However, that is skipped now that we added the `sx` prop to this component, so I set it manually here. Let me know if I should take a different approach -- I saw that adding ` && !sx` in the linked snippet was a recent change

src/modules/form/components/DynamicField.tsx
```diff
@@ -364,11 +374,12 @@ const DynamicField: React.FC<DynamicFieldProps> = ({
             placeholder={placeholder}
             textInputProps={{
               horizontal,
-              sx: {
-                maxWidth: MAX_INPUT_AND_LABEL_WIDTH,
-                '.MuiInputBase-root': { maxWidth: MAX_INPUT_WIDTH },
-              },
+              InputProps: { sx: { width } },
```
martha: Touching this code made me wonder, what was behind the original desire to have 600px wide labels but 500px wide inputs in general -- so that these Select components are the exception to that rule? Maybe something to check in with Ash about

src/modules/form/components/DynamicForm.stories.tsx
```diff
@@ -17,7 +17,7 @@ export default {
   argTypes: { label: { control: 'text' } },
   decorators: [
     (Story) => (
-      <Box sx={{ width: 800 }}>
```
martha, line 20: I removed this so that we can test mobile usability more easily

src/components/elements/input/PhoneInput.tsx
```diff
@@ -47,10 +47,10 @@ const TextMaskCustom = forwardRef<HTMLInputElement, CustomProps>(
   }
 );
 
-const PhoneInput = (props: TextInputProps) => {
+const PhoneInput = ({ InputProps, ...props }: TextInputProps) => {
   return (
     <TextInput
-      InputProps={{ inputComponent: TextMaskCustom as any }}
+      InputProps={{ inputComponent: TextMaskCustom as any, ...InputProps }}
       {...props}
```
gigxz, line 55: Mini scope-creep, but this feels like one where we should fix the input width. Similar to SSN, this is input accepts a fixed number of characters. (With some variability based on the presence of a country code).

```
      placeholder='(XXX) XXX - XXXX'
      inputWidth='155px'
```
What do you think of that?

src/components/elements/input/RadioGroupInput.tsx
gigxz: 👍  this is better, anyway. I think if we have an outer `sx` prop it is most intuitive if it applies to the outermost component.

src/modules/form/components/DynamicField.tsx
```diff
@@ -258,11 +259,13 @@ const DynamicField: React.FC<DynamicFieldProps> = ({
...
             value={value || ''}
             onChange={onChangeEvent}
-            sx={{
-              width,
-              maxWidth: MAX_INPUT_AND_LABEL_WIDTH,
-              '.MuiInputBase-root': { maxWidth: MAX_INPUT_WIDTH },
+            InputProps={{
+              sx: {
+                width,
+                maxWidth: MAX_INPUT_WIDTH,
```
gigxz: see other comment re: fixed with of phone input

gigxz:
The story is helpful for reviewing, thanks!

I think it's worth fixing the issue with `InputContainer`, so that max width for the inputs can be set there, instead of in each individual component.

It looks like a made some vague attempt to allow us to set the max width using grid breakpoints, which still seems like it might be a decent idea? Happy to discuss further. just wanted to leave that overall comment about InputContainer, I think that would be ideal.

src/test/__mocks__/mockFormDefinition.json
```diff
@@ -570,7 +595,8 @@
             {
               "linkId": "clientSignature",
               "type": "STRING",
-              "text": "Client Name"
+              "text": "Client Name",
+              "size": "MEDIUM"
```
martha: I added size MEDIUM to the signatures here to get the text input boxes for the name portion of the signature to appear the correct size, now that by default text inputs are set as much wider.

src/components/elements/input/CheckboxGroupInput.tsx
```diff
@@ -57,6 +67,7 @@ const CheckboxGroupInput: React.FC<CheckboxGroupInputProps> = ({
             sx={{
               color: disabled ? 'text.disabled' : 'text.primary',
               '&.Mui-focused': { color: 'text.primary' },
+              ...labelSx,
```
martha, line 70: I added `labelSx` here, instead of just using `maxWidth` like everywhere else, in order for the label to actually be smaller than the width of the whole component, since these are horizontal radio buttons.

martha (commented):
Ok, @gigxz  I think this is ready for re-review with a new approach using `DynamicInputCommonProps`, but still with a few open questions.
- see my comment about `labelSx` -- I wonder if we should just apply this philosophy everywhere instead of using `maxWidth` at all? something like `labelWidth` and `inputWidth`?
- The disability table, in particular, looks bad when labels are super long and at least one of the elements in the table is a display instead of a dropdown. I couldn't track down exactly why one column of the table was getting so much narrower, so maybe we could pair on this, or de-prioritize as less important
- Open question for design tomorrow: Original reason to have the width of a wide input (e.g. text input or select) differ from the width of the label? (500 vs. 600 px?)

src/modules/form/components/DynamicField.tsx
martha: For components like phone (and ssn and etc.), should we still accept a width from the form definition? Should it override the normal width of that component?

src/test/__mocks__/mockFormDefinition.json
martha: Actually, I'd like to ask for advice about how to better handle this, as well as the Table input group, both of which are now showing fields narrower than they used to be. Should it be the form definition's responsibility to set a width for these, or should it happen by default somewhere else?

src/modules/form/types.ts
```diff
@@ -83,6 +82,8 @@ export interface DynamicInputCommonProps {
   placeholder?: string;
   breakpoints?: BreakpointOverrides;
   ariaLabelledBy?: string;
+  maxWidth?: number | string;
+  inputWidth?: number;
```
gigxz: Comments here would be helpful! Is my understanding correct?

```suggestion
  maxWidth?: number | string; // maximum width of form element (including the label)
  inputWidth?: number; // fixed with of input element (does not affect label)
```

src/test/__mocks__/mockFormDefinition.json
```diff
@@ -195,6 +211,15 @@
           "pickListReference": "NoYesReasonsForMissingData",
           "helperText": "This is helper text"
         },
+        {
+          "type": "CHOICE",
+          "component": "RADIO_BUTTONS",
+          "linkId": "radio-1",
+          "text": "Choice Option as Radio that Repeats",
```
gigxz, line 218: Thanks for adding these!

src/test/__mocks__/mockFormDefinition.json
gigxz: Yeah, good question. The Signature component expects to contain a text field as its first element. IMO we can count on that, and the parent Signature component should be responsible for setting the styles of its children. I tried that, adding [here](https://github.com/greenriver/hmis-frontend/blob/50e8bf943084dc236a1a65086072735ef51a30b6/src/modules/form/components/group/Signature.tsx#L51):
```
          <Grid
            container
            direction='row'
            columnGap={viewOnly ? 6 : 2}
            sx={{
              # Use max medium-width for name input, which is expected to be first
              '.HmisForm-inputContainer:first-of-type': {
                flex: 1,
                maxWidth: FIXED_WIDTH_MEDIUM,
              },
            }}
          >
            {renderChildItem &&
              item.item?.map((childItem) => renderChildItem(childItem))}
          </Grid>
```

That approach relies on CSS selectors, which is a pattern we use in some places. Another option would be to pass some arguments to `renderChildItem` for the first item, to specify the width. CSS seems easier to understand to me.

gigxz: And yeah agreed, the `TableGroup` might need some tweaking.. I suggest just removing the [initial](https://github.com/greenriver/hmis-frontend/blob/50e8bf943084dc236a1a65086072735ef51a30b6/src/modules/form/components/group/TableGroup.tsx#L95) so it goes full screen. It looks a little goofy for small tables (like in storybook) but it will be more flexible to accomodate large tables.

gigxz (commented):
A few small comments about the components you  called out, otherwise looks great! Lets re target to 109

gigxz:
> The disability table, in particular, looks bad when labels are super long and at least one of the elements in the table is a display instead of a dropdown. I couldn't track down exactly why one column of the table was getting so much narrower, so maybe we could pair on this, or de-prioritize as less important

I think we can safely ignore this, the Disability Table will not be very configurable since its only used for HUD fields. (TBD on whether the Form Editor will completely "lock" groups of HUD elements like this, so they cant be edited. There are certain aspects of HUD Data Element collection that shouldnt be configurable because it will mess with data quality)

src/test/__mocks__/mockFormDefinition.json
martha: Ok great, thank you for both of these! The Signature component still looks wacky when labels are extremely long, but that seems like a rare use case anyway.

src/modules/form/types.ts
martha: Yep exactly!

PR #700, opened 2024-03-19
Add ability to create, rename, and delete Service Types

martha:
## Description

PT issue: https://www.pivotaltracker.com/n/projects/2591838#

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4128

This PR adds UI elements around creation, deletion, and renaming of custom service types. We decided to leave custom service categories out of the UI for now, though the mutations are there for us to use later.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/admin/components/services/NewServiceTypeButton.tsx
```diff
@@ -0,0 +1,108 @@
...
+      setDialogOpen(false);
+    },
+  });
+
+  const {
+    data: pickListData,
+    loading: pickListLoading,
+    error: pickListError,
+  } = useGetPickListQuery({
+    variables: { pickListType: PickListType.AllServiceCategories },
```
martha, line 48: I was wondering if we will need to create a new pick list for only non-HUD service categories?

gigxz: Hmm good question. I think this is okay, afaik there is no logical reason why custom services couldn't be inside a group with other hud services.

src/modules/admin/components/services/NewServiceTypeButton.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+const NewServiceTypeButton = () => {
+  const [dialogOpen, setDialogOpen] = useState(false);
+  const [name, setName] = useState('');
+  const [supportsBulkAssignment, setSupportsBulkAssignment] = useState(false);
+  const [serviceCategory, setServiceCategory] = useState<PickListOption | null>(
+    null
+  );
+
+  const [createServiceType, { error, loading }] = useMutation<
```
gigxz: You should be able to use `useCreateServiceTypeMutation` here. It is generated into `gqlTypes.ts`

src/modules/admin/components/services/ServiceTypeDetailPage.tsx
```diff
@@ -20,30 +45,83 @@ const ServiceTypeDetailPage = () => {
     variables: { id: serviceTypeId },
   });
 
+  const navigate = useNavigate();
+
+  const [editDialogOpen, setEditDialogOpen] = useState(false);
+  const [editedName, setEditedName] = useState<string>();
+
+  const [renameServiceType, { error: renameError, loading: renameLoading }] =
+    useMutation<RenameServiceTypeMutation, RenameServiceTypeMutationVariables>(
```
gigxz: Same here on using generated hook useRenameServiceTypeMutation

gigxz (approved):
looks good!

PR #722, opened 2024-03-29
Add search capability to Organization table

martha:
## Description

PT ticket: https://www.pivotaltracker.com/story/show/186494330
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4203

Here's the new search bar:
Screenshot 2024-03-29 at 4 54 00 PM

I'd like to ask design about the medium-height search bar (consistent with Projects search bar) next to the small-height Add Organization button -- looks a bit odd to me. Should we make the Organization search bar small-height so it looks normal next to the button, though it will be inconsistent with Projects?

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/components/pages/AllProjects.tsx
+++ b/src/components/pages/AllProjects.tsx
@@ -138,20 +138,20 @@ const AllProjects = () => {
           </Grid>
         )}
         {viewMode === 'organizations' && (
-          <RootPermissionsFilter permissions={['canEditOrganization']}>
-            <Grid item xs={12}>
-              <Stack gap={2} direction={isMobile ? 'column' : 'row'}>
-                <CommonSearchInput
-                  label='Search Organizations'
-                  name='search organizations'
-                  placeholder='Search by Organization Name or ID'
-                  value={search}
-                  onChange={setSearch}
-                  fullWidth
-                  size='medium'
-                  searchAdornment
-                  clearAdornment
-                />
+          <Grid item xs={12}>
+            <Stack gap={2} direction={isMobile ? 'column' : 'row'}>
+              <CommonSearchInput
+                label='Search Organizations'
+                name='search organizations'
+                placeholder='Search by Organization Name or ID'
+                value={search}
+                onChange={setSearch}
+                fullWidth
+                size='medium'
+                searchAdornment
+                clearAdornment
+              />
+              <RootPermissionsFilter permissions={['canEditOrganization']}>
                 <Box sx={{ width: 'fit-content', pt: isMobile ? 0 : 3.5 }}>
                   <ButtonLink
                     data-testid='addOrganizationButton'
@@ -162,9 +162,9 @@ const AllProjects = () => {
                     Add Organization
                   </ButtonLink>
                 </Box>
-              </Stack>
-            </Grid>
-          </RootPermissionsFilter>
+              </RootPermissionsFilter>
+            </Stack>
+          </Grid>
         )}
         <Grid item xs={12}>
           <Paper data-testid='allProjectsTable'>
@@ -174,6 +174,7 @@ const AllProjects = () => {
                 GetProjectsQueryVariables,
                 ProjectAllFieldsFragment
               >
+                key='projectTable'
                 queryVariables={{
                   filters: { searchTerm: debouncedSearch || undefined },
                 }}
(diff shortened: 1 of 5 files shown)


src/components/pages/AllProjects.tsx
```diff
@@ -171,21 +190,27 @@ const AllProjects = () => {
...
-                columns={ORGANIZATION_COLUMNS}
-                rowLinkTo={organizationRowLink}
-                noData='No organizations'
-                pagePath='organizations'
-                showFilters
-                recordType='Organization'
-                defaultPageSize={25}
-                queryVariables={{}}
-              />
+              <Box>
```
martha: This extra `<Box>` was a quick and dirty solution to a bug I'd like to discuss more, both to improve my own React knowledge, and because I suspect there is a cleaner way to solve it. Here's the issue:

React seems to treat both `<GenericTableWithData>` components as the same component, regardless of `viewMode`, so it doesn't re-render when switching from Project to Organization. The `defaultFilters` param may change when `viewMode` changes, but because it is not used directly, but instead is passed in as the default value for the `filterValues` state [here](https://github.com/greenriver/hmis-frontend/blob/efbcb35416ec43fc23767d900980fea6c5ed8307/src/modules/dataFetching/components/GenericTableWithData.tsx#L178), changing the default value after the component has been rendered doesn't cause the `filterValues` state to change.

You can reproduce the issue locally by commenting out the `<Box>` wrapper on line 193 and 213. You should see an error when switching from Projects to Organizations: "Variable $filters of type OrganizationFilterOptions was provided invalid value for status (Field is not defined on OrganizationFilterOptions)". Inspecting in the chrome console, you can see that `filterValues` state in `GenericTableWithData` still includes the `status` attribute.

Anyway, adding a `Box` here basically tricks React, I guess, because if the dom tree is different for the 2 different modes then React will do a full re-render and treat them as separate components.

(edit: this may not be helpful to anyone but me, but for posterity, [here](https://codesandbox.io/p/sandbox/friendly-silence-8gzykg?file=%2Fsrc%2FApp.js) is a link to a sandbox I made to minimally reproduce the issue so I could understand what was going on)

src/components/pages/AllProjects.tsx
```diff
@@ -136,14 +140,29 @@ const AllProjects = () => {
...
-              <ButtonLink
-                data-testid='addOrganizationButton'
-                to={generateSafePath(Routes.CREATE_ORGANIZATION)}
-                Icon={AddIcon}
-                leftAlign
-              >
-                Add Organization
-              </ButtonLink>
+              <Stack gap={2} direction={isMobile ? 'column' : 'row'}>
+                <CommonSearchInput
```
gigxz: This search only shows up if the user has permission to edit organizations, because of the `RootPermissionsFilter` wrapper. We should show it in all cases. The wrapper only needs to hide the Add Organization button

src/components/pages/AllProjects.tsx
gigxz: Huh, thats a weird one. The issue goes away if you add `key='projectTable'` and `key='organizationTable'` to the two tables, that must help React figure out that they are different components. I'm not sure why it couldn't figure it out, though.

src/components/pages/AllProjects.tsx
```diff
@@ -171,21 +190,27 @@ const AllProjects = () => {
...
+                  OrganizationType
+                >
+                  queryDocument={GetOrganizationsDocument}
+                  columns={ORGANIZATION_COLUMNS}
+                  rowLinkTo={organizationRowLink}
+                  noData='No organizations'
+                  pagePath='organizations'
+                  showFilters={false}
+                  defaultFilters={undefined}
+                  filters={undefined}
```
gigxz: These can be removed, since they are the defaults. It would be nice to have consistency with the Project table, but I agree it probably makes sense to just hide this since there are no available sort or filter options.

src/components/pages/AllProjects.tsx
martha: Ah thank you, that is much cleaner 👍

src/components/pages/AllProjects.tsx
martha: Thanks 👍 I added these when I was trying to debug the issue with React thinking the 2 tables were the same component, and forgot to remove

gigxz (approved):

PR #723, opened 2024-04-02
Prevent access to Edit Client route without canViewCLientName perm

martha:
## Description

PT ticket: https://www.pivotaltracker.com/story/show/186494363
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4214

This PR implements Elliot's suggestion to prevent access to the client edit route unless the user has permission to view client names. This prevents bugs where the user without permission might try to edit a client and actually change their name in the system to "Client <id>".

Question: I've hidden the edit button if the user doesn't have access to view the client name. Is there any other functionality we should remove? e.g. should you not be able to add a client if you don't have this permission?

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/modules/client/components/ClientProfileCard.tsx
+++ b/src/modules/client/components/ClientProfileCard.tsx
@@ -415,7 +415,7 @@ const ClientProfileCard: React.FC<Props> = ({ client }) => {
               <Stack sx={{ flexGrow: 1, maxWidth: '300px' }} gap={1} mt={2}>
                 <ClientPermissionsFilter
                   id={client.id}
-                  permissions={['canEditClient']}
+                  permissions='canEditClient'
                 >
                   <ButtonLink
                     data-testid='editClientButton'
--- a/src/modules/client/components/ClientProfileCard.tsx
+++ b/src/modules/client/components/ClientProfileCard.tsx
@@ -415,8 +415,7 @@ const ClientProfileCard: React.FC<Props> = ({ client }) => {
               <Stack sx={{ flexGrow: 1, maxWidth: '300px' }} gap={1} mt={2}>
                 <ClientPermissionsFilter
                   id={client.id}
-                  permissions={['canEditClient', 'canViewClientName']}
-                  mode='all'
+                  permissions={['canEditClient']}
                 >
                   <ButtonLink
                     data-testid='editClientButton'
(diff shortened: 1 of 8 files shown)


src/modules/enrollment/components/EnrollmentQuickActions.tsx
```diff
@@ -30,7 +30,9 @@ const EnrollmentQuickActions = ({
     enabledFeatures.includes(DataCollectionFeatureRole.Service) &&
     enrollment.access.canEditEnrollments;
 
-  const canEditClient = enrollment.client.access.canEditClient;
+  const canEditClient =
+    enrollment.client.access.canEditClient &&
+    enrollment.client.access.canViewClientName;
```
gigxz:
```suggestion
    enrollment.client.access.canViewClientName; # Prevent UI-access to edit client if user lacks name access, to avoid situation where masked name ("Client X") is submitted. See follow-up #187358025
```

gigxz: Want to clarify that we're just preventing the user from accessing it in the UI. It's not actually restricted on the mutation.

src/modules/client/components/ClientProfileCard.tsx
```diff
@@ -415,7 +415,8 @@ const ClientProfileCard: React.FC<Props> = ({ client }) => {
               <Stack sx={{ flexGrow: 1, maxWidth: '300px' }} gap={1} mt={2}>
                 <ClientPermissionsFilter
                   id={client.id}
-                  permissions='canEditClient'
+                  permissions={['canEditClient', 'canViewClientName']}
```
gigxz: lets add a comment here too about why `canViewClientName` is included

gigxz: Actually.. instead of scattering these around, we could also choose to just keep the change to `src/routes/protected.tsx` with a comment there @martha. This is just a safeguard not an expected scenario. Either way is fine with me, but if we have it in multiple places lets comment in all of them. thanks

gigxz:
>  should you not be able to add a client if you don't have this permission?

I don't think that restriction is necessary. You can still add a client with a name, if you have that perm.

src/modules/client/components/ClientProfileCard.tsx
martha: I was wondering about that too. I agree, since this is just a safeguard, I'll remove the checks to avoid unnecessary complexity and keep the safeguard in only 1 place

gigxz (approved):
thanks!

PR #725, opened 2024-04-02
Reduce noise in sentry

gigxz:
## Description

Stop sending duplicate and unnecessary errors to sentry [PT](https://www.pivotaltracker.com/n/projects/2591838/stories/187277496)

## Type of change

### DIFF
--- a/src/providers/apolloClient.tsx
+++ b/src/providers/apolloClient.tsx
@@ -3,7 +3,6 @@ import {
   ApolloLink,
   HttpOptions,
   InMemoryCache,
-  ServerError,
   from,
 } from '@apollo/client';
 import { BatchHttpLink } from '@apollo/client/link/batch-http';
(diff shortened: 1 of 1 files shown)


src/providers/apolloClient.tsx
```diff
@@ -125,25 +124,27 @@ const errorLink = onError(({ operation, graphQLErrors, networkError }) => {
         user: sentryUser(),
       })
     );
-  }
-
-  if (networkError) {
+  } else if (networkError) {
```
gigxz, line 127: We were double-sending before because errors woudl have both the graphql error and the network error. Change to else if to avoid that

src/providers/apolloClient.tsx
```diff
@@ -125,25 +124,27 @@ const errorLink = onError(({ operation, graphQLErrors, networkError }) => {
...
       const { statusCode, result } = networkError;
       console.error('[Server error]', statusCode, result);
-      Sentry.captureException(statusCode, {
-        user: sentryUser(),
-        extra: { result },
-      });
+
+      if (statusCode === 401) {
+        // May mean that session was invalidated on the server. No need to send to Sentry.
+        dispatchSessionTrackingEvent(undefined);
```
gigxz, line 134: I don't think we need to notify sentry if someones session became invalid

martha (approved):

PR #726, opened 2024-04-03
Fix minor issues with client image tile and uploader

martha:
## Description

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4222
PT issue: https://www.pivotaltracker.com/story/show/187354988

This PR makes the following minor UI fixes around client images:
- If you are not allowed to edit clients, the client image tile no longer appears as a button, preventing users clicking around from getting the 'access denied' error mentioned in the ticket.
- If you are not allowed to view client photos, you shouldn't see the image tile at all.
- If uploading the photo fails, the UI displays an error instead of hanging on the loading screen. (Hopefully this isn't user-facing, but since uploading fails locally without s3 set up, this is just to make developer experience a little more helpful)

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/components/elements/upload/UploaderBase.tsx
+++ b/src/components/elements/upload/UploaderBase.tsx
@@ -231,7 +231,11 @@ const Uploader: React.FC<UploaderProps> = ({
             setCurrentUpload(res);
             onUpload(res, file);
           })
-          .then(() => setLoading(false));
+          .then(() => setLoading(false))
+          .catch((error) => {
+            setLoading(false);
+            setError(error.message);
+          });
       }
     },
     [uploadFile, onUpload]
--- a/src/modules/client/components/ClientProfileCard.tsx
+++ b/src/modules/client/components/ClientProfileCard.tsx
@@ -360,6 +360,8 @@ const ClientProfileCard: React.FC<Props> = ({ client }) => {
     variables: { id: client.id },
   });
 
+  const [canViewClientPhoto] = useHasRootPermissions(['canViewClientPhoto']);
+
   const [canViewSsn] = useHasRootPermissions([
     'canViewFullSsn',
     'canViewPartialSsn',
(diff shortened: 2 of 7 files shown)


src/modules/client/components/ClientProfileCard.tsx
```diff
@@ -287,54 +289,58 @@ const ClientCardImage = ({
...
-            className='overlay'
-            sx={{
-              ...overlaySx,
-              backgroundColor: 'rgba(0,0,0, 0.5)',
-            }}
-          >
-            <Chip
-              label='Update Photo'
-              icon={<EditIcon />}
+      {client.access.canEditClient ? (
```
martha, line 292: GH is displaying this as a more complicated diff than it really is; the only difference is wrapping in the `{client.access.canEditClient ? (...) : ...` clause.

src/modules/client/components/ClientProfileCard.tsx
```diff
@@ -377,14 +385,13 @@ const ClientProfileCard: React.FC<Props> = ({ client }) => {
               <Skeleton
                 variant='rectangular'
                 sx={{
-                  height: 150,
-                  width: 150,
-                  mr: 1,
+                  height: size,
```
martha: This is a small aesthetic change, to make the loader the same size as the image tile, so that the UI doesn't jump around once the image has loaded. I'm guessing it was the original intention but then one got changed w/out the other being updated - but please let me know if there is a reason for it, I'm happy to revert!

src/components/elements/upload/UploaderBase.tsx
```diff
@@ -231,7 +231,11 @@ const Uploader: React.FC<UploaderProps> = ({
             setCurrentUpload(res);
             onUpload(res, file);
           })
-          .then(() => setLoading(false));
+          .then(() => setLoading(false))
+          .catch((error) => {
+            setLoading(false);
+            setError(error.message);
+          });
       }
```
gigxz, line 239: What's the expected behavior here? When I try to upload an image, the `CreateDirectUploadMutation` succeeds but the "save" button remains disabled. This is probably just because I don't have s3 setup locally, but just wanted to check.

src/modules/client/components/ClientProfileCard.tsx
```diff
@@ -269,6 +269,8 @@ const ClientCardImage = ({
 
   if (!client) return <ClientCardImageElement />;
 
+  if (!client.access.canViewClientPhoto) return;
```
gigxz, line 272: We should do a design pass at some point, it seems like the edit button could flow to the right maybe. There are some installations that may never want to use photos, and now those installations can just turn off this permission, which is great. 👍

gigxz: Although, it looks like the loading box still appears momentarily, and then disappears. Lets move this line up above 270 so that doesn't happen

martha: Hm, moving it above line 270 means the client could be undefined. But I've solved it one level up by checking the root permissions, here: https://github.com/greenriver/hmis-frontend/pull/726/files#diff-f2bf835ebbcb6e05582acd008cd571838fa17cdafc8a48b58a85b24ae41541c1R386-R400

src/components/elements/upload/UploaderBase.tsx
martha: For me, before this change, trying to upload an image causes a 500 error in the console (because of not having s3 locally setup) but the modal continues spinning on the "loading" animation after the error.

With this change, after it errors out the modal shows "Response not successful: Received status code 500" and the save button is still disabled.
Screenshot 2024-04-05 at 4 34 14 PM

gigxz (approved):

PR #733, opened 2024-04-05
Add assessments to project dashboard

martha:
## Description

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4234
PT ticket: https://www.pivotaltracker.com/story/show/187372090

This PR adds a new Assessments tab on the Project dashboard where we can view all assessments for the given project.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/api/operations/project.queries.graphql
```diff
@@ -123,6 +123,34 @@ query GetProjectServices(
...
+  $sortOrder: AssessmentSortOption = ASSESSMENT_DATE
+  $filters: AssessmentFilterOptions = null
+) {
+  project(id: $id) {
+    id
+    assessments(
+      limit: $limit
+      offset: $offset
+      sortOrder: $sortOrder
+      filters: $filters
```
martha, line 139: Wanted to note that filtering by 'Custom Assessment' type didn't work how I expected. It returned no results despite the presence of some custom assessments. @gigxz is this a bug or due to my misunderstanding?

This issue seemed to impact the existing assessments tables (client, enrollment) as well as the newly built one, so, if it's a bug I'll create a new PT bugfix ticket.

src/modules/assessments/assessmentColumns.tsx
```diff
@@ -0,0 +1,44 @@
...
+} = {
+  date: {
+    header: 'Assessment Date',
+    render: (a) => <AssessmentDateWithStatusIndicator assessment={a} />,
+  },
+  type: {
+    header: 'Assessment Type',
+    render: (a) => formRoleDisplay(a),
+  },
+  linkedType: {
```
martha: I did some cleanup here to follow the pattern from e.g. CLIENT_COLUMNS which I remember discussing on some recent PR. I wanted to check what you think about this `linkedType` column. It's the easiest but probably least sophisticated way to deal with the problem of different assessment tables having slightly different requirements (the 'type' column values should have link treatment on some tables, but not on others). Maybe there is a better/prettier way to do this?

src/modules/assessments/assessmentColumns.tsx
martha, line 1: I would include this in `util.ts` in this directory, converting it to a `.tsx` file, if you'd consider that a good idea

src/api/operations/project.queries.graphql
gigxz: It's a bug - pt story https://www.pivotaltracker.com/story/show/187400755
Thanks!

src/modules/assessments/assessmentColumns.tsx
gigxz: This seems fine to me!

src/modules/assessments/util.ts
```diff
@@ -42,3 +44,15 @@ export const generateAssessmentPath = (
     assessmentId: assessment.id,
   });
 };
+
+export const assessmentRowLinkTo = (
+  record: ClientAssessmentType | ProjectAssessmentType,
+  clientId?: string
```
gigxz: This should be required since it's needed for the path

src/modules/assessments/assessmentColumns.tsx
gigxz: Yep that sounds like a good idea

gigxz (approved):
🌟

PR #737, opened 2024-04-11
Minor storybook upgrade

gigxz:
## Description

Closes https://github.com/greenriver/hmis-frontend/pull/735

Minor upgrade only

## Type of change
- [x] Dependency update


ttoomey (approved):
This is probably fine. But I do worry a little about such a massive diff for a minor bump to storybook; there is some risk of creating dependency-related regressions.

Maybe we should check in on our approach to dependency management at some point

ttoomey (changes requested):
Oh sorry I missed the build failures, changing approval status

gigxz:
The diff is just because of the offline mirror,  we can definitely revisit that though. Not sure how reliable the registry is nowadays

gigxz:
This was a really weird one, some packaged caused `Popper` types to change (not sure what) and I had to upgrade react types slightly to make it happy. I believe the underlying issue was https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/69006

ttoomey (approved):
Thanks for tracking down the build issue!

I was thinking that ideally bumping storybook up a minor version wouldn't require upgrading so many packages (~600). This might be because we are using yarn-deduplicate with the 'highest' strategy and it's causing churn. Or maybe it's that storybook just has massive dependencies. Something we can think about in the future, maybe we should move away from 'yarn classic' if it's too hard to mange

gigxz:
Storybook definitely has massive dependencies, since it does transpilation it needs babel and everything (so those packages are not shared with prod, but they may be shared with jest which also transpiles).

Also I used upgrade-interactive so I wonder if that made it worse 👀

gigxz:
But yeah def should revisit package manager tool at some point!

PR #740, opened 2024-04-12
Upgrade node js to 20.x

gigxz:
## Description

* Upgrade node to the next version. No particular reason we just hadn't updated in a while https://nodejs.org/en/about/previous-releases
* Remove yarn binary. If I understand correctly, we should already all be using the one that gets downloaded with `corepack enable`

To test locally, perform steps 2 and 3 here: https://github.com/greenriver/hmis-frontend?tab=readme-ov-file#install-node-npm-and-yarn

## Type of change
- [x] Dependency update


ttoomey (approved):
Great, thanks for updating! I didn't see any breaking changes called out in the release notes from 18->20 so we should be good

PR #743, opened 2024-04-22
Use network-only on merge search

gigxz:
## Description

* Use network-only policy, to avoid bug where client that has already been merged shows up as a possible merge candidate
* Show errors in toast instead of taking over the whole page

[GH issue](https://github.com/open-path/Green-River/issues/5929)

**How to test:** the bug could be reproduced as follows:
* Enable throttling in your browser
* Go to Client => merges => new
* Search for a client
* Merge the client. Make sure you choose a client that was created MORE RECENTLY than the current client, so they get "merged in" to the current client.
* As soon as the modal closes, click "search" again, and click "merge" and "confirm" quickly to merge the same client again (you have to be fast)
* Previously this would result in an error. With this change, you should not see the old client in the results at all.

## Type of change
- [x] Bug fix


martha (approved):
Wonderful, looks good!

PR #744, opened 2024-04-23
Enable filtering by specific custom assessment type

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/5918

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4277

This PR makes frontend changes needed to support filtering Assessment tables by both role and form definition identifier.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/hmis/components/HmisFilter.tsx
```diff
@@ -69,6 +69,7 @@ const FILTER_NAME_TO_PICK_LIST = {
   project: PickListType.Project,
   appliedToProject: PickListType.Project,
   organization: PickListType.Organization,
+  assessmentType: PickListType.AssessmentTypes,
```
gigxz: Hmph so this code is a little confusing. This map is meant to set the default pick list based on the name of the filter. It assumes that the filter names are descriptive, and always basically mean the same thing. (E.g a filter called `project` is assumed to always use the `PROJECT` pick list).

In this case, what you added to the map is not being used, because the name of the filter option is actually just `type`. If you changed this to `type` you can get rid of all the `getAssessmentTypeFilter` stuff as far as I can tell, because [this code](https://github.com/greenriver/hmis-frontend/blob/5a399036bf41f688bb44eea018c702d7e5ec64a6/src/modules/hmis/components/HmisFilter.tsx#L104-L110) will automagically set up the filter list for you.

The problem with that, of course, is `type` is not descriptive. I think it may be worth [changing the filter name](https://github.com/greenriver/hmis-warehouse/blob/219cf8afe2c33817ba91961aaa399ec208fc17df/drivers/hmis/app/graphql/types/hmis_schema/assessment.rb#L15) to `assessment_type` so it has a more descriptive user-facing title, and we can use this auto-filter approach. It would be a breaking change, but I don't think it should cause problems because it is a tucked-away filter that is only used on-demand. 

Does that make sense?

The alternative would be to leave it as-is, and make sure `getAssessmentTypeFilter` gets added to Client Assessments and anywhere else in the future that lists assessments. It's OK but it doesn't follow the pattern we use elsewhere.

gigxz: Hmm actually, I am overlooking the "pickListArgs" piece. The filter cannot infer the project id for filters, that needs to be passed explicitly (which you currently do with the `filters` prop). 

I don't think we have a clean way to do that. I'm going to post these review comments because I need to head about but I think it requires some further thinking on an API for filtering that works for dynamic pick lists with arguments. We can check in tomorrow!

gigxz:
The Client Assessments page is showing a text box right now, it should have the same behavior

src/modules/hmis/components/HmisFilter.tsx
martha: Thanks!! I sketched a possible solution here: https://github.com/greenriver/hmis-frontend/commit/8d21b0ac8e27c415242eab6587e2083b2f32c562, adding a new prop `filterArgs` (could be renamed to `filterPickListArgs`?) on `GenericTableWithData`. `filterArgs` maps filter names to their pick list args, and is passed all the way down to the `getFilterForType` function.

I would still need to update `type` to `assessmentType` as you said in your comment above. What do you think of this approach?

gigxz: That sounds reasonable @martha. I see that your approach allows you to pass different pick list args to different filters, for example, pass `projectId` to one pick list but not to another. I don't think we really need that, I can't think of a use case, but there may be one. My only other comment would be to add typing (use existing `PickListArgs`) and use the filterPickListArgs naming for clarity.

--

This is certainly scope creep from the original ticket, but, I think some light refactoring would help us here.. It feels wrong that the `GenericTableWithData` component is responsible for the logic of generating the filter definition object based on the filter type (eg `AssessmentsForProjectFilterOptions`). It's making it so we need to keep adding more filter-related props to the table component. That component is already responsible for a lot, so it feels like something that could be pulled out.


I'm thinking about something like the below. What do you think? This would be more of a developer ergonomics change that anything else. it would make it a bit simpler to implement the new pick list requirement imo.

```jsx
const MyTable () => {
   // returns memoized filter object `TableFilterType<FilterOptionsType>`
   // this hook would encapsulate the logic thats currently in the allFieldFilters fn
   const filters = useFilters({
      type: 'AssessmentsForProjectFilterOptions', // filter input type from schema
      omit: ['searchTerm'], // optional: skip some filters
      pickListArgs: { projectId }, // optional: pick list args to be applied to all PickList filter items
     // could add other args to hook as needed, without messing with GenericTable component
    })

   // filters can be modified as needed

   return (
     <GenericTableWithData...
        filters={filters} // filter object, that may have been constructed manually or generated from schema
        defaultFilters={ {status: Status.OPEN } } // default  filter *values* still accepted
        // filterInputType - deprecate. we could still keep it around and GenericTableWithData could use the useFilter hook internally, or, we could refactor all 33 usages
     />
   )
}
```

src/modules/dataFetching/components/GenericTableWithData.tsx
```diff
@@ -108,16 +110,16 @@ function allFieldColumns<T>(recordType: string): ColumnDef<T>[] {
 }
 
 function allFieldFilters<T>(
-  recordType: string,
-  filterInputType: string
+  filterInputType: string,
```
martha: Excised `recordType` as it was not used

src/modules/hmis/filterUtil.ts
martha, line 1: I found the `HmisFilter` component itself was not used anywhere, so I removed this, and moved this to be a util file.

src/modules/hmis/components/HmisFilter.tsx
martha: Discussed and decided to attempt the refactor, timeboxed to ~2-3 hours. Result of my efforts here: https://github.com/greenriver/hmis-frontend/pull/749

src/modules/projects/components/ProjectAssessments.tsx
```diff
@@ -77,7 +75,8 @@ const ProjectAssessments = () => {
           pagePath='project.assessments'
           recordType='Assessment'
           showFilters
-          filters={(filters) => omit(filters, 'project', 'projectType')}
+          filterPickListArgs={{ projectId: projectId }}
```
gigxz:
```suggestion
          filterPickListArgs={{ projectId }}
```

gigxz:
This looks good and lets put the refactor on this branch!
@martha can you merge in the most recent 114 to this at some point. Just want to make sure we don't overwrite anything with schema changes from other branches.

gigxz (approved):
looks good!

PR #745, opened 2024-04-23
Fix intake bug for changing household size

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/5722
This PR evicts the HoH's Enrollment from the Apollo cache, so that an out-of-date household size is not used when determining how to render the Intake Assessment page.

Link to code where that householdSize check is happening: https://github.com/greenriver/hmis-frontend/blob/c37b83c10d89e6bf897e65d8241ad8c0a3eb1c44/src/modules/assessments/components/IntakeAssessmentPage.tsx#L29

A few other edge cases I checked:
- Will this problem manifest in a 3-member household where 2 clients with completed assessments and the 3rd is removed from the household? I thought this might be a problem since, in the solution in this PR, I'm only evicting the cache for the current client's enrollment, instead of checking all enrollments on the household. It doesn't turn out to be an issue, because the code linked above in IntakeAssessmentPage checks if householdSize > 1. In other words, I think this bug only manifests when household size goes from 2 to 1, not 3 to 2.
- Will the problem manifest if we _add_ a client to the household? For example going from 1 to 2? I thought of this because I was checking other usages of `refetchHousehold`, the success callback for when a member is removed from a household. It's also used when updating HoH (but I figured this would not be a problem since the household size is not impacted) and when adding a member. I checked both the case where the HoH intake assessment is already completed and where it has not been completed yet. Both scenarios did not manifest the bug. The cache seems to be already updating somewhere, although I did not quite find where -- but you can see on the Enrollment Details page where the section heading changes from “Enrollment Details” (how it appears for single-person household) to “Client Name Enrollment Details”.

## Type of change
- [x] Bug fix
- [ ] Dependency update

- [x] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue - See bug reproducing steps in GH issue linked above

### DIFF
--- a/src/modules/assessments/components/household/HouseholdSummaryTabPanel.tsx
+++ b/src/modules/assessments/components/household/HouseholdSummaryTabPanel.tsx
@@ -94,7 +94,7 @@ const HouseholdSummaryTabPanel = memo(
           refetch();
 
           tabs.forEach(({ clientId, enrollmentId }) => {
-            // Fully refetch the Enrollment for fields like householdSize
+            // Fully refetch the Enrollment
             cache.evict({
               id: `Enrollment:${enrollmentId}`,
             });
--- a/src/modules/assessments/components/household/HouseholdSummaryTabPanel.tsx
+++ b/src/modules/assessments/components/household/HouseholdSummaryTabPanel.tsx
@@ -97,7 +97,7 @@ const HouseholdSummaryTabPanel = memo(
             // Make it so the "Assessments" table on the Enrollment page gets refetched next time
             cache.evict({
               id: `Enrollment:${enrollmentId}`,
-              fieldName: 'assessments',
+              // fieldName: 'assessments',
             });
             // Make it so the "Assessments" table on the Client page gets refetched next time
             cache.evict({
--- a/src/modules/assessments/components/household/HouseholdSummaryTabPanel.tsx
+++ b/src/modules/assessments/components/household/HouseholdSummaryTabPanel.tsx
@@ -94,10 +94,9 @@ const HouseholdSummaryTabPanel = memo(
           refetch();
 
           tabs.forEach(({ clientId, enrollmentId }) => {
-            // Make it so the "Assessments" table on the Enrollment page gets refetched next time
+            // Fully refetch the Enrollment for fields like householdSize
             cache.evict({
               id: `Enrollment:${enrollmentId}`,
-              // fieldName: 'assessments',
             });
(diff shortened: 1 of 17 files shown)


src/modules/dataFetching/components/GenericTableWithData.tsx
```diff
@@ -108,7 +108,7 @@ function allFieldColumns<T>(recordType: string): ColumnDef<T>[] {
 }
 
 function allFieldFilters<T>(
-  recordType: string,
+  recordType: string, // < not needed
```
martha: Made this change in another PR here: https://github.com/greenriver/hmis-frontend/pull/749

src/modules/assessments/components/household/HouseholdSummaryTabPanel.tsx
```diff
@@ -94,10 +94,9 @@ const HouseholdSummaryTabPanel = memo(
           refetch();
 
           tabs.forEach(({ clientId, enrollmentId }) => {
-            // Make it so the "Assessments" table on the Enrollment page gets refetched next time
+            // Fully refetch the Enrollment
             cache.evict({
               id: `Enrollment:${enrollmentId}`,
```
martha, line 99: @gigxz you were right that the real problem here was that `enrollment.intakeAssessment` was still `null`.

I'm not sure if there really was an issue with `householdSize` or I imagined that/was mistaken. But I removed the addition of `householdSize` that you recommended [here](https://github.com/greenriver/hmis-frontend/pull/745/commits/e00955818faa5b9a44ee93f15594fb1eb999fc21#diff-47969bb3d13a8aabb25165358ccddd0702bba766506a95eadf9c875dd5899507R34), because of build fails that it introduced in other files (see [here](https://github.com/greenriver/hmis-frontend/pull/745/commits/12fa56fb9d5e3532dedf709bd8c15d6a9fd4ba9d)).

With this change, when I tried to reproduce the issue and checked in the debugger for the value of `enrollment.householdSize` [here](https://github.com/greenriver/hmis-frontend/pull/745/commits/e00955818faa5b9a44ee93f15594fb1eb999fc21#diff-0572cdaae28be2c9fec71c4b73468d9810eaa7bfaa99396a13f3e3186f7aeba8R29), it looks like it correctly updates from 2 to 1.

gigxz: Alright interesting, thanks for tracking that down!

gigxz (approved):
👍 😅

PR #746, opened 2024-04-24
Display client alerts for bulk service enrollment

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/5934
This PR adds an alert modal on the Bulk Services page before enrolling or assigning services, if the Client has any Client Alerts.

Assignment table with icons in buttons:
Screenshot 2024-04-24 at 4 34 13 PM

Modal that pops up when you click "Assign" or "Enroll + Assign":
Screenshot 2024-04-24 at 4 35 19 PM

## Type of change
- [x] Bug fix (I guess we could call this a regression bug?)
- [ ] Dependency update

### DIFF
--- a/src/api/operations/bulkService.queries.graphql
+++ b/src/api/operations/bulkService.queries.graphql
@@ -22,6 +22,13 @@ query BulkServicesClientSearch(
       id
       ...ClientName
       ...ClientIdentificationFields
+      alerts {
+        ...ClientAlertFields
+      }
+      access {
+        canViewClientAlerts
+        canManageClientAlerts
+      }
       activeEnrollment(projectId: $projectId, openOnDate: $serviceDate) {
         id
         entryDate
--- a/src/api/operations/bulkService.queries.graphql
+++ b/src/api/operations/bulkService.queries.graphql
@@ -25,10 +25,6 @@ query BulkServicesClientSearch(
       alerts {
         ...ClientAlertFields
       }
-      access {
-        canViewClientAlerts
-        canManageClientAlerts
-      }
       activeEnrollment(projectId: $projectId, openOnDate: $serviceDate) {
         id
         entryDate
--- a/src/modules/services/components/bulk/AssignServiceButton.tsx
+++ b/src/modules/services/components/bulk/AssignServiceButton.tsx
@@ -54,7 +54,7 @@ const AssignServiceButton: React.FC<Props> = ({
     return client.activeEnrollment.services.nodesCount > 0;
   }, [client]);
 
-  const { clientAlerts } = useClientAlerts({ client: client });
+  const { clientAlerts } = useClientAlerts({ client });
   const [clientAlertDialogOpen, setClientAlertDialogOpen] = useState(false);
 
   const onClick = useCallback<NonNullable<ButtonProps['onClick']>>(
--- a/src/modules/services/components/bulk/AssignServiceButton.tsx
+++ b/src/modules/services/components/bulk/AssignServiceButton.tsx
@@ -1,11 +1,20 @@
 import CheckIcon from '@mui/icons-material/Check';
+import WarningAmberRoundedIcon from '@mui/icons-material/WarningAmberRounded';
 import { LoadingButton } from '@mui/lab';
-import { ButtonProps } from '@mui/material';
+import { ButtonProps, DialogActions, DialogContent } from '@mui/material';
+import DialogTitle from '@mui/material/DialogTitle';
 import { useCallback, useEffect, useMemo, useState } from 'react';
 
+import * as React from 'react';
 import { useBulkAssignMutations } from '../../hooks/useBulkAssignMutations';
 import ButtonTooltipContainer from '@/components/elements/ButtonTooltipContainer';
+
+import CommonDialog from '@/components/elements/CommonDialog';
+import ClientAlertStack from '@/modules/client/components/clientAlerts/ClientAlertStack';
+import useClientAlerts from '@/modules/client/hooks/useClientAlerts';
 import ApolloErrorAlert from '@/modules/errors/components/ApolloErrorAlert';
+import FormActions from '@/modules/form/components/FormActions';
+import { FormActionTypes } from '@/modules/form/types';
 import {
   BulkAssignServiceInput,
   BulkServicesClientSearchQuery,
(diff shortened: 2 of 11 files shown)


src/modules/services/components/bulk/AssignServiceButton.tsx
```diff
@@ -77,23 +103,66 @@ const AssignServiceButton: React.FC<Props> = ({
...
+              {
+                action: FormActionTypes.Discard,
+                id: 'discard',
+                label: `Cancel ${serviceTypeName}`,
+                buttonProps: { variant: 'gray' },
+              },
+              {
+                action: FormActionTypes.Submit,
+                id: 'submit',
+                label: `Add ${serviceTypeName}`,
```
martha: Awaiting question about button color in [figma](https://www.figma.com/file/F9EHR2IZYuMAlOJ9aOG2IL?type=design&node-id=3417-24669&mode=design#782287629)

src/modules/services/components/bulk/AssignServiceButton.tsx
martha, line 1: The [ticket](https://github.com/open-path/Green-River/issues/5934) mentions we are not sure whether to make this a Bed Nights-only change, or apply it to all Bulk Services. Is that a decision for us to make, or, whom should we ask?

src/modules/services/components/bulk/AssignServiceButton.tsx
```diff
@@ -77,23 +103,66 @@ const AssignServiceButton: React.FC<Props> = ({
     return 'Enroll + Assign';
   }, [client.activeEnrollment, isAssignedOnDate, localDisabled]);
 
+  const startIcon =
```
martha, line 106: I want to confirm the icon behavior that made the most sense to me:
- Check-mark icon takes priority, if the client is assigned on that date
- Only show the warning icon on alerts if check-mark is not shown

src/api/operations/bulkService.queries.graphql
```diff
@@ -22,6 +22,13 @@ query BulkServicesClientSearch(
       id
       ...ClientName
       ...ClientIdentificationFields
+      alerts {
+        ...ClientAlertFields
+      }
+      access {
+        canViewClientAlerts
+        canManageClientAlerts
```
martha: I thought it was a bit silly to add `canManageClientAlerts` here to keep typescript happy, even though the user is never actually managing client alerts from the bulk services page. I think that the juice is probably not worth the squeeze to avoid it? But let me know if I should do so.

src/modules/services/components/bulk/AssignServiceButton.tsx
martha: Resolved per figma thread - button should be client main color, which it is by default.

src/api/operations/bulkService.queries.graphql
gigxz: I don't think it should be necessary to resolve either of these. The `alerts` array will be empty if the user does not have permission to view alerts, that is already checked [here](https://github.com/greenriver/hmis-warehouse/blob/324fbfe9a2d4e2b748ca0a7ece6e4c402c1dd46c/drivers/hmis/app/graphql/types/hmis_schema/client.rb#L331-L335).

In some cases we check on the frontend because the UX is different based on whether the user has access or not. I don't think that's the case here.

gigxz: ts to make access object optional can be:
`Omit<ClientWithAlertFieldsFragment, 'access'> & { access?: Partial<ClientAccess>; };`
looks like its only used to determine whetehr to show delete button, so we can just hide that if access object is missing

src/modules/services/components/bulk/AssignServiceButton.tsx
gigxz: Let's do it for all of them, and see what feedback we get. A day shelter check-in ("bulk service") would have the same needs as a night shelter check-in.

src/modules/services/components/bulk/AssignServiceButton.tsx
```diff
@@ -43,6 +54,9 @@ const AssignServiceButton: React.FC<Props> = ({
     return client.activeEnrollment.services.nodesCount > 0;
   }, [client]);
 
+  const { clientAlerts } = useClientAlerts({ client: client });
```
gigxz:
```suggestion
  const { clientAlerts } = useClientAlerts({ client });
```

src/modules/services/components/bulk/AssignServiceButton.tsx
```diff
@@ -77,23 +103,66 @@ const AssignServiceButton: React.FC<Props> = ({
...
+                },
+                onCompleted: () => setLocalDisabled(true),
+              });
+              setClientAlertDialogOpen(false);
+            }}
+            config={[
+              {
+                action: FormActionTypes.Discard,
+                id: 'discard',
+                label: `Cancel ${serviceTypeName}`,
```
gigxz: Could use the `FormDialogActionContent` component instead to pass these as props. We have inconsistent usage around modal confirmation buttons, could probably use a refactor to tighten it up at some point.

src/modules/services/components/bulk/AssignServiceButton.tsx
```diff
@@ -77,23 +103,66 @@ const AssignServiceButton: React.FC<Props> = ({
...
           fullWidth
           variant={isAssignedOnDate ? 'contained' : 'gray'}
         >
           {buttonText}
         </LoadingButton>
       </ButtonTooltipContainer>
       {apolloError && <ApolloErrorAlert error={apolloError} />}
+      <CommonDialog fullWidth open={clientAlertDialogOpen}>
+        <DialogTitle>
+          {clientAlerts.length === 1 ? 'Client Alert' : 'Client Alerts'}
```
gigxz: Left feedback in Figma. I feel like the client name should be here, it's not obvious who the alert is for.

src/modules/services/components/bulk/AssignServiceButton.tsx
gigxz: Looks good to me! Thanks

src/api/operations/bulkService.queries.graphql
martha: Ah perfect. Thank you for the suggestion 👍

gigxz (approved):
Looks great!

PR #749, opened 2024-04-25
Refactor Generic Tables to not be responsible for filtering

martha:
## Description

Peripherally related to https://github.com/open-path/Green-River/issues/5918, this PR refactors the `GenericTableWithData` component to take out much of the filtering logic. See comment https://github.com/greenriver/hmis-frontend/pull/744#discussion_r1579557345 for motivation.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/dataFetching/components/GenericTableWithData.tsx
```diff
@@ -61,15 +55,10 @@ export interface Props<
...
-    | TableFilterType<FilterOptionsType>
-    | ((
-        baseFilters: TableFilterType<FilterOptionsType>
-      ) => TableFilterType<FilterOptionsType>);
-  filterPickListArgs?: PickListArgs;
+  filters?: TableFilterType<FilterOptionsType>;
   sortOptions?: SortOptionsType;
   defaultSortOption?: keyof SortOptionsType;
-  defaultFilters?: Partial<FilterOptionsType>;
+  defaultFilterValues?: Partial<FilterOptionsType>;
```
martha, line 61: I renamed this for clarity, the functionality should be unchanged.

src/modules/admin/components/services/ServiceCategoryRuleTable.tsx
martha, line 1: Just flagging that, from what I could tell, this table is not used anywhere in the codebase - but was recently created, so sounds like it should not be removed?

src/modules/admin/components/services/ServiceCategoryTable.tsx
```diff
@@ -50,7 +50,6 @@ const ServiceCategoryTable = () => {
         columns={columns}
         pagePath='serviceCategories'
         noData='No service categories'
-        showFilters
```
martha, line 53: I removed this because `ServiceCategoryFilterOptions` doesn't exist, so no filters were rendering anyway pre-refactor. Is this a bug? If so I'll create a ticket, but I figured it's probably just a copy/paste error.

src/components/clientDashboard/enrollments/ClientAssessments.tsx
```diff
@@ -45,6 +49,10 @@ const ClientAssessments = () => {
     [clientId]
   );
 
+  const filters = useFilters({
+    type: getInputTypeForRecordType('Assessment'),
```
martha: Let me know how you think this looks - I did also consider moving the `getInputTypeForRecordType` inside of `useFilters` so that it's more of a default, but the caller then has to pass in `recordType` anyway. I ended up thinking this was the cleanest approach, but I'd be glad for cleaner suggestion.

src/modules/dataFetching/components/GenericTableWithData.tsx
```diff
@@ -61,15 +55,10 @@ export interface Props<
...
   showFilters?: boolean;
   noSort?: boolean;
   noFilter?: boolean;
```
martha: My big remaining question around this refactor is the `showFilters` / `noFilter` confusion. I think I just don't understand their meaning based on the names -- and there were a couple of components that had both `showFilters` and `noFilter`, which really confused me (ClientScanCards, ClientCaseNotes, and EnrollmentCaseNotes).

If these mean what they sound like, then I think the ideal API for this GenericTable component is that neither one of them exists, and instead the filters are either shown or hidden based on whether `filters` is passed in by the caller.

src/modules/admin/components/services/ServiceCategoryTable.tsx
gigxz: Hm looks like this controls whether the top toolbar shows up at all. Thats not great naming.

With `showFilters`:
Screenshot 2024-04-26 at 12 36 12 PM


Without `showFilters`:
Screenshot 2024-04-26 at 12 36 18 PM

gigxz: It's OK to remove it but it's not a well-named prop. It should probably be called something more descriptive like "show top toolbar". Or it can be removed entirely if we want to decide that the top pagination should only show up if there are sort or filter options. That would be fine imo. Some tables have neither. The pagination still shows up at the bottom regardless.

martha: Ah I see, I think I'd like to keep it and rename it `showTopToolbar`. Looks like it is also used in other places that have no filters (the ones mentioned in my other comment [below](https://github.com/greenriver/hmis-frontend/pull/749#discussion_r1580173837))

src/modules/dataFetching/components/GenericTableWithData.tsx
martha: Thanks for the clarification about `showFilters` on the comment thread above. I'm now not quite sure what the cleanest API would be -- maybe...
- `showFilters` -> renamed to `showTopToolbar`, and either
  - A) the top toolbar is always shown if `filters` are provided, but if not, you can *separately* specify that it should be shown anyway with `showTopToolbar`, OR,
  - B) you must always pass `showTopToolbar` in addition to `filters` if you want the top toolbar to be shown
  - Maybe A is preferable because it avoids the unexpected situation where you're trying to pass in `filters` and confused about why they aren't showing up. ?

- `noFilter` -> I could also go either way
  - A) remove this prop, decide whether or not to show filters based on whether `filters` are passed. OR
  - b) keep it as a way to override, regardless of whether `filters` are passed.
  - Looking at the code, it seems like this was probably invented to give the ability to override when GenericTable always attempted to automatically create filters for the given record type. Now that that's not happening anymore, it seems like `noFilter` prop is not needed, but maybe we should keep it for parity with `noSort` which is still needed and used in a handful of places.

src/components/clientDashboard/enrollments/ClientAssessments.tsx
```diff
@@ -54,7 +62,8 @@ const ClientAssessments = () => {
           GetClientAssessmentsQueryVariables,
           ClientAssessmentType
         >
-          showFilters
+          showTopToolbar
+          filters={filters}
```
gigxz, line 62: This feels a little verbose. Shouldn't the "top tool bar" show up automatically if `filters` are provided? Or are there cases where we want to pass filters but for some reason want to hide them...? I don't think so.

src/components/pages/AllProjects.tsx
```diff
@@ -209,7 +219,7 @@ const AllProjects = () => {
                 rowLinkTo={organizationRowLink}
                 noData='No organizations'
                 pagePath='organizations'
-                showFilters={false}
+                showTopToolbar={false}
```
gigxz: Is this needed? It looks like `showTopToolbar` is false by default.

src/components/clientDashboard/enrollments/ClientAssessments.tsx
gigxz: Hmmmm I see. Honestly I think just specifying `'AssessmentFilterOptions'` feels more straightforward as a pattern. Either way is fine imo.
It would be nice if we had a generated enum for those strings,  but we don't currently.

src/modules/admin/components/services/ServiceCategoryTable.tsx
gigxz: great 👍

src/modules/dataFetching/components/GenericTableWithData.tsx
```diff
@@ -336,15 +283,15 @@ const GenericTableWithData = <
 
   const noDataValue = useMemo(() => {
     if (typeof noData === 'function') return noData(filterValues);
-    if (!showFilters) return noData;
+    if (!showTopToolbar) return noData;
```
gigxz: The intent here is to return if the table doesn't have filtering. If it has filtering, then we show a different message below to make sure the user realizes there is a filter (pretty rudimentary currently..). So, this should probably be `if (!filters) return noData;`.

src/modules/dataFetching/components/GenericTableWithData.tsx
```diff
@@ -308,23 +272,6 @@ const GenericTableWithData = <
     });
   }, [columnDefs, includedOptionalColumns]);
 
-  const filterDefs = useMemo(() => {
```
gigxz, line 311: 👍

src/modules/dataFetching/components/GenericTableWithData.tsx
gigxz: 1)
Ahemmm yes a classic prop dilemma. I wonder if React pros have an opinion about this. I prefer going with option A because it feels more intuitive when using the component. But it might be a personal preference thing.

2)
IMO `noFilter` is no longer necessary, because `filters` are now passed explicitly. I believe this was added to handle scenarios where the table was auto-generating the filter list, and we wanted a prop to let us hide them. Now with this refactor, filters are _always_ passed explicitly, so we shouldn't need this. 
Callers that previously passed `noFilter` should just...not pass any filters :) much nicer.

gigxz (commented):
a few minor comments but overall looks good! feel free to merge into your other branch when you are ready, or re-request if you want another look on it

src/modules/dataFetching/components/GenericTableWithData.tsx
martha: Ok! 1) Agree, choosing 1A. 2) Agree, removing `noFilter`

PR #752, opened 2024-04-29
Show auto-exited in Enrollment Status

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/5741
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4293

This PR updates the Enrollment Status component to say "Auto-Exited" when the enrollment was auto-exited. The component is used in a lot of places, so here are the usages I've QAd:
- [x] `HouseholdMemberTable.tsx` - `EnrollmentStatus` is used in `HOUSEHOLD_MEMBER_COLUMNS`
    - [x] `HouseholdMemberTable` rendered by `EnrollmentOverview`: 
       Screenshot 2024-04-29 at 8 26 59 AM
    - [x] `HouseholdMemberTable` is also used in a couple other places, but without the status column.
- [x] `ProjectClientEnrollmentsTable.tsx` - `EnrollmentStatus` is used in `ENROLLMENT_STATUS_COL`
    - [x] Usage in `ProjectClientEnrollmentsTable`:
       Screenshot 2024-04-29 at 8 36 29 AM
    - [x] `ENROLLMENT_STATUS_COL` is also used in `ClientEnrollmentCard`, which only shows *recent* enrollments ([within 30 days](https://github.com/greenriver/hmis-frontend/blob/c37b83c10d89e6bf897e65d8241ad8c0a3eb1c44/src/modules/hmis/hmisUtil.ts#L332-L341)). Since we require auto-exits to configure a length of absence >= 30 days, I think this usage should never occur in practice? But I did QA it anyway by manually maneuvering the code, and it looks good:
      Screenshot 2024-04-29 at 8 56 38 AM
- [x] `ProjectHouseholdsTable.tsx`
      Screenshot 2024-04-29 at 8 39 27 AM
- [x] `EnrollmentDetails.tsx` - `EnrollmentStatus` component is only used directly here if the enrollment is incomplete.

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/api/operations/customCaseNoteQueries.graphql
+++ b/src/api/operations/customCaseNoteQueries.graphql
@@ -49,6 +49,7 @@ query GetClientCaseNotes(
           projectType
           entryDate
           exitDate
+          autoExited
           inProgress
         }
       }
--- a/src/api/operations/customCaseNoteQueries.graphql
+++ b/src/api/operations/customCaseNoteQueries.graphql
@@ -43,14 +43,7 @@ query GetClientCaseNotes(
       nodes {
         ...CustomCaseNoteFields
         enrollment {
-          id
-          lockVersion
-          projectName
-          projectType
-          entryDate
-          exitDate
-          autoExited
-          inProgress
+          ...ClientEnrollmentFields
         }
       }
     }
--- a/graphql.schema.json
+++ b/graphql.schema.json
@@ -14665,6 +14665,22 @@
             "deprecationReason": null
           },
           {
+            "name": "autoExited",
+            "description": null,
+            "args": [],
+            "type": {
+              "kind": "NON_NULL",
+              "name": null,
+              "ofType": {
+                "kind": "SCALAR",
+                "name": "Boolean",
+                "ofType": null
+              }
+            },
+            "isDeprecated": false,
+            "deprecationReason": null
+          },
+          {
             "name": "ceAssessments",
             "description": null,
             "args": [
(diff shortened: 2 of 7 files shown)


src/api/operations/customCaseNoteQueries.graphql
```diff
@@ -49,6 +49,7 @@ query GetClientCaseNotes(
           projectType
           entryDate
           exitDate
+          autoExited
```
martha: `autoExited` isn't used by custom case notes, but the custom case notes return type is passed to `EnrollmentDateRangeWithStatus` [here](https://github.com/greenriver/hmis-frontend/blob/c37b83c10d89e6bf897e65d8241ad8c0a3eb1c44/src/modules/caseNotes/ClientCaseNotes.tsx#L34).

I thought maybe it's neater to add this here rather than use a partial, since it's just a boolean - but now that I'm thinking about it, maybe it's not good because on the backend we need to join with the Exit table in order to fetch this information.  @gigxz what do you think?

gigxz: It won't cause an extra query because we already join with `exit` to get the exit date here.

But, I think it would make more sense to resolve the enrollment using the  `ClientEnrollmentFields` fragment, since that's what we do everywhere else when we resolve records on client (`GetClientFiles`, `GetClientServices`, `GetClientAssessments`, etc.). I don't think there's any reason Case Notes are different, they probably just got missed when we tightened that up. It will resolve additional fields that we don't _necessarily_ need currently, but I think it's worth it for the consistency.

martha: Ah okay, yeah this makes sense. Thanks for the neater suggestion!

gigxz (approved):

PR #753, opened 2024-04-29
Link enrollment from assessment view

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/5939
This PR adds a link in the Assessment view to get to the relevant Enrollment. A few new props are added to the AssessmentTitle component so that it can display info about the enrollment in the link text.
Screenshot 2024-04-29 at 2 50 50 PM

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/components/elements/RouterLink.tsx
+++ b/src/components/elements/RouterLink.tsx
@@ -24,7 +24,12 @@ const RouterLink = forwardRef<RouterLinkProps, any>(
       target={openInNew ? '_blank' : undefined}
     >
       {openInNew || Icon ? (
-        <Stack direction={'row'} gap={0.75} alignItems='center'>
+        <Stack
+          direction={'row'}
+          gap={0.75}
+          alignItems='center'
+          sx={{ display: 'inline-flex', textDecoration: 'inherit' }}
+        >
           {children}
           {openInNew ? (
             <OpenInNewIcon fontSize='inherit' />
--- a/src/modules/assessments/components/AssessmentTitle.tsx
+++ b/src/modules/assessments/components/AssessmentTitle.tsx
@@ -3,7 +3,10 @@ import { Stack } from '@mui/system';
 import { ReactNode, useState } from 'react';
 import { useRelatedAnnualAssessments } from '../hooks/useRelatedAnnualAssessments';
 import RouterLink from '@/components/elements/RouterLink';
-import { parseAndFormatDate } from '@/modules/hmis/hmisUtil';
+import {
+  parseAndFormatDate,
+  parseAndFormatDateRange,
+} from '@/modules/hmis/hmisUtil';
 import { Routes } from '@/routes/routes';
 import { AssessmentRole } from '@/types/gqlTypes';
 import { generateSafePath } from '@/utils/pathEncoding';
@@ -95,10 +98,8 @@ const AssessmentTitle = ({
             clientId,
           })}
           openInNew
-          sx={{ display: 'inline-flex' }}
         >
-          {projectName} ({parseAndFormatDate(entryDate)} -{' '}
-          {exitDate ? parseAndFormatDate(exitDate) : 'Active'})
+          {projectName} ({parseAndFormatDateRange(entryDate, exitDate)})
         </RouterLink>
       </Typography>
       {subtitle}
--- a/src/modules/assessments/components/AssessmentTitle.tsx
+++ b/src/modules/assessments/components/AssessmentTitle.tsx
@@ -4,15 +4,20 @@ import { ReactNode, useState } from 'react';
 import { useRelatedAnnualAssessments } from '../hooks/useRelatedAnnualAssessments';
 import RouterLink from '@/components/elements/RouterLink';
 import { parseAndFormatDate } from '@/modules/hmis/hmisUtil';
+import { Routes } from '@/routes/routes';
 import { AssessmentRole } from '@/types/gqlTypes';
+import { generateSafePath } from '@/utils/pathEncoding';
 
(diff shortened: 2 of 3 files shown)


src/modules/assessments/components/AssessmentTitle.tsx
```diff
@@ -81,7 +89,17 @@ const AssessmentTitle = ({
...
           {': '}
         </b>
-        {projectName}
+        <RouterLink
+          to={generateSafePath(Routes.ENROLLMENT_DASHBOARD, {
+            enrollmentId,
+            clientId,
+          })}
+          openInNew
+          sx={{ display: 'inline-flex' }}
```
martha: Interesting side note, `RouterLink` [internally](https://github.com/greenriver/hmis-frontend/blob/c37b83c10d89e6bf897e65d8241ad8c0a3eb1c44/src/components/elements/RouterLink.tsx#L27-L34) uses a `Stack` when `openInNew` is passed, with the following result:
- when `openInNew` is passed, by default `RouterLink` renders on its own line
- when `openInNew` is not passed, `RouterLink` renders inline with neighbors

That's why I added `sx={{ display: 'inline-flex' }}` here, but I also considered adding this sx directly on the `Stack` within `RouterLink`, to make that component's behavior more predictable. That would require a quick qa of all instances of `RouterLink` where `openInNew` is true, to make sure it doesn't change the ui in any way that we don't want.

gigxz: @martha thanks for flagging that. I had to look up the difference between `flex` and `inline-flex`. I think it makes sense that the RouterLink should use inline-flex by default when rendering the link next to an icon. Can you move this over to RouterLink? Thank you!

src/modules/assessments/components/AssessmentTitle.tsx
```diff
@@ -81,7 +89,17 @@ const AssessmentTitle = ({
...
+        >
+          {projectName} ({parseAndFormatDate(entryDate)} -{' '}
+          {exitDate ? parseAndFormatDate(exitDate) : 'Active'})
```
gigxz: was there a reason not to use `parseAndFormatDateRange(entryDate, exitDate)` here?

martha: No, no reason 🤦‍♀️

src/modules/assessments/components/AssessmentTitle.tsx
martha: Ok, moved it into RouterLink. And also did a look around the codebase to see if this might mess up any of the existing usages if they expect to be stacked and now appear inlined. The only one I think might be impacted is `ProjectOutgoingReferralPostingsTable`, but I wasn't yet able to successfully QA this -- I went to a `/project/:id/referrals` route and submitted an outgoing referral, but it immediately disappeared. I figure this is something to do with local configuration since I remember referrals right now is a client-specific feature.

appendix:
- [ ] ProjectOutgoingReferralPostingsTable.tsx - maybe impacted since it’s displayed alongside an HmisEnum? How to QA this?
- [x] ClientName.tsx - already inside Stack
- [x] UnitOccupants.tsx - only thing in column definition
- [x] BulkServicesTable.tsx - only thing in column definition
- [x] ClientSearch.tsx - only thing in column definition
- [x] AdminReferralPostingDetails.tsx - only thing in column definition
- [x] ReferralPostingDetails.tsx - only thing in column definition
- [x] FormRuleTable.tsx - used in FormRuleColumns.formDefinition, but this isn’t used anywhere
- [x] MciMatchSelector.tsx - already in Stack
- [x] ServiceTypeDetailPage.tsx - looks good, already in a CommonUnstyledList
- [x] AssessmentTitle.tsx - related annual assessments. Looks good, not impacted since already wrapped inside of a Stack
- [x] NewClientMerge.tsx - only thing in column definition

martha: As a quick aside, I noticed [ExternalLink](https://github.com/greenriver/hmis-frontend/blob/c37b83c10d89e6bf897e65d8241ad8c0a3eb1c44/src/components/elements/ExternalLink.tsx#L5) is a bit similar to RouterLink but it doesn't even use a stack, so it wouldn't have this problem. Should we just use that approach here?

src/components/elements/RouterLink.tsx
```diff
@@ -24,7 +24,12 @@ const RouterLink = forwardRef<RouterLinkProps, any>(
       target={openInNew ? '_blank' : undefined}
     >
       {openInNew || Icon ? (
-        <Stack direction={'row'} gap={0.75} alignItems='center'>
+        <Stack
+          direction={'row'}
+          gap={0.75}
+          alignItems='center'
+          sx={{ display: 'inline-flex', textDecoration: 'inherit' }}
```
martha, line 31: Note of interest, this `textDecoration` is now necessary because text decorations do not propagate to atomic inline-level descendants 😂  https://bugzilla.mozilla.org/show_bug.cgi?id=1344545

gigxz (approved):
looks good!

src/modules/assessments/components/AssessmentTitle.tsx
gigxz: Huh, I'd be happy to look at your local setup for referrals whenever to get that sorted out.

martha: Belated follow-up:
- [x] ProjectOutgoingReferralPostingsTable.tsx - not impacted, because its neighbor `HmisEnum` renders a `<p>` internally which renders on its own line

PR #758, opened 2024-05-01
Update form definition table to accommodate versions

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/5953
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4310

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/admin/components/forms/FormDefinitionTable.tsx
```diff
@@ -49,24 +49,26 @@ const columns: ColumnDef<Row>[] = [
 ];
```
martha, line 49: I haven't made these updates yet, but my understanding is that in-scope for this ticket are all updates to make our forms table match the figma design here: https://www.figma.com/file/F9EHR2IZYuMAlOJ9aOG2IL/OP-%2F-HMIS-%2F-2024-%2F-Q1-%2B-Q2-(Jan-1---June-30)?type=design&node-id=2611-31087&mode=design&t=MS7446xMjNXDLuZX-0
including adding new columns and changing some column names around. Is that correct?

src/modules/admin/components/forms/FormDefinitionDetailPage.tsx
```diff
@@ -38,18 +40,18 @@ const FormDefinitionDetailPage = () => {
       MutationUpdateFormDefinitionArgs
     >({
       formRole: StaticFormRole.FormDefinition,
-      initialValues: formDefinition || {},
-      localConstants: { definitionId: formId },
+      initialValues: formIdentifier?.draft || {},
+      localConstants: { definitionId: formIdentifier?.draft.id },
```
martha: This code is causing typescript build failures since it relies on `formIdentifier` having already resolved, and can't be moved after the formIdentifier's resolution because it is a hook that can't be called conditionally.

In practice, I think this doesn't actually cause problems because the button you click to open this dialog is not visible until the formIdentifier resolves. I think this code will also change/maybe get deleted entirely when we implement https://github.com/open-path/Green-River/issues/6005, so maybe it should be ignored for now?

src/api/operations/form.queries.graphql
```diff
@@ -75,6 +75,53 @@ query GetFormDefinitions(
   }
 }
 
+query GetFormIdentifierForEditor($identifier: String!) {
+  formIdentifier(identifier: $identifier) {
```
gigxz, line 54: This feels like a _lot_ of data to resolve on page load for the page `/admin/forms/:identifier`. Can we re-work this so that the `FormDefinitionFieldsForEditor` fragment is only fetched when you're actually loading the Editor (or the Preview) for a specific form version? Some forms can be really big, there is no need to load all that json just to view the detail page. (here we are loading each json blob 2x or more per version).

src/api/operations/form.queries.graphql
```diff
@@ -75,6 +75,53 @@ query GetFormDefinitions(
...
+    displayVersion {
+      ...FormDefinitionFieldsForEditor
+    }
+    published {
+      ...FormDefinitionFieldsForEditor
+    }
+    draft {
+      ...FormDefinitionFieldsForEditor
+    }
+    allVersions {
```
gigxz: I don't see this being used at all, can we remove it?
Presumably the version history table will use it's own query for a paginated table, using GenericTableWithData to perform the query and pagination logic. I don't see that the version history needs to be part of this query.

src/modules/admin/components/forms/FormDefinitionDetailPage.tsx
```diff
@@ -38,18 +40,18 @@ const FormDefinitionDetailPage = () => {
       MutationUpdateFormDefinitionArgs
     >({
       formRole: StaticFormRole.FormDefinition,
-      initialValues: formDefinition || {},
-      localConstants: { definitionId: formId },
+      initialValues: formIdentifier?.draft || {},
```
gigxz:
```suggestion
      initialValues: formIdentifier?.displayVersion || {},
```

I know it will change later with design, but it seems wrong that it's using `draft` for the `initialValues`, but `displayVersion` to determine which form the "submit" button will actually update.  Am I getting that right?

src/modules/admin/components/forms/UpdateFormDefinitionPage.tsx
```diff
@@ -4,40 +4,40 @@ import FormEditor from './FormEditor';
...
-                input: { definition: JSON.stringify(values) },
-              },
-            });
-          }}
-          saveLoading={loading}
-        />
-      )}
+      <PageTitle title={`Edit Form: ${formIdentifier.displayVersion.title}`} />
+      <FormEditor
+        definition={formIdentifier.displayVersion.rawDefinition}
```
gigxz: Just noting that we'll need to come back to this with https://github.com/open-path/Green-River/issues/5858 to implement the version behavior

src/modules/admin/components/forms/UpdateFormDefinitionPage.tsx
```diff
@@ -4,40 +4,40 @@ import FormEditor from './FormEditor';
 import Loading from '@/components/elements/Loading';
 import PageTitle from '@/components/layout/PageTitle';
 import {
-  useGetFormDefinitionForEditorQuery,
+  useGetFormIdentifierForEditorQuery,
```
gigxz: I'm not convinced on whether this change makes sense. Let's discuss at check-in @martha. In my mind the "FormBuilder" page should still be working off the specific `FormDefinition` version that it's editing.

It may need some more thought before implementing https://github.com/open-path/Green-River/issues/5858
For example, it brings up the question for me:
* When you hit "edit" on a form that doesn't currently have a draft, should it create a draft immediately and _then_ take you to editing that draft? Or, should it take you to a "new draft" page, and only generate the draft record when it auto-saves for the first time, and then going forward future auto-saves will update that draft? (Second option feels more complex imo)

It's kind of out of scope for this PR, which is why I think maybe leaving this page query as-is seems like a good idea.

src/modules/admin/components/forms/FormDefinitionDetailPage.tsx
gigxz: yep I think your solution is fine- this comes up in various places in the code we have the same `|| ''` to pacify ts

src/api/operations/form.queries.graphql
```diff
@@ -75,6 +75,53 @@ query GetFormDefinitions(
...
+      limit
+      nodesCount
+      nodes {
+        ...FormDefinitionFieldsForEditor
+      }
+    }
+  }
+}
+
+query GetFormIdentifiers(
```
gigxz, line 69: Can you remove the `query GetFormDefinitions` since it's not used anymore

src/api/operations/form.queries.graphql
gigxz: From further review, I see that we use the same query `useGetFormIdentifierForEditorQuery ` for both the "form details page" and the "edit form page", and that this issue predates your PR. That is worth changing imo, but you're welcome to do it in a separate PR if this one is feeling big.

An API that would make more sense to me is:
1. `GetFormIdentifierDetails($identifier: String!)` -- a query that loads the details about the form (by `identifier`) but not the form json itself. Used to back the form details page.
2. `GetFormDefinitionFieldsForEditor($id: ID!)` -- a query that loads the `FormDefinitionFieldsForEditor` fragment on a specific form version, by ID. Depending on the situation, this would either be the id of the published version or the draft versions.

For (2) I could see other approaches (mentioned in another comment). But (1) seems pretty standard to me.

src/modules/admin/components/forms/FormDefinitionDetailPage.tsx
martha: Yes this was a mistake - thanks for catching it!

src/api/operations/form.queries.graphql
martha: Agreed, updated to add the two queries you described.

src/api/operations/form.queries.graphql
martha: Removed!

src/modules/admin/components/forms/UpdateFormDefinitionPage.tsx
martha: @gigxz would like to hear your thoughts about the new routing I implemented here, which now uses `forms/:identifier/:formId/edit` for the editor path.

It would be totally possible to just revert back to use `forms/:formId/edit`, but, then the breadcrumbs give a less-nice experience (or perhaps they just need more custom configuration to work nicely?)

gigxz:
@martha retargeting this to release-116 since the backend is going to release-116. I try to create the new release branches around the same time as Elliot but sometimes there is a lag. You're welcome to create it next time you encounter this problem! (I just do: `git checkout release-115; git pull; git checkout -b release-116; git push`

src/api/operations/form.queries.graphql
```diff
@@ -50,55 +50,22 @@ query GetAssessmentFormDefinition(
...
-      ...FormDefinitionFieldsForEditor
-    }
-    allVersions {
-      offset
-      limit
-      nodesCount
-      nodes {
-        ...FormDefinitionFieldsForEditor
-      }
+      ...FormDefinitionMetadata
```
gigxz, line 58: This feels much cleaner, thanks!

src/modules/admin/components/forms/UpdateFormDefinitionPage.tsx
gigxz: That seems fine to me. There is a way to update the breadcrumbs, I think, but it doesn't seem worth the hassle.

gigxz (approved):
looks great!

gigxz:
Feel free to merge https://github.com/greenriver/hmis-warehouse/pull/4310 and then this whenever you're ready. I'll let you handle that since there are related PRs

PR #759, opened 2024-05-06
Add form preview page and adjust permissions

martha:
## Description
GH issue: https://github.com/open-path/Green-River/issues/5981
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4325

This PR adds a new Preview page linked from the form details page. It is visible to everyone who has `canConfigureDataCollection` permissions. Meanwhile, the new permission `canManageForms` now restricts who can see and access the routes related to actually editing forms.

We should also wait to merge this until after the recent form versioning PRs are merged ([backend](https://github.com/greenriver/hmis-warehouse/pull/4310), [frontend](https://github.com/greenriver/hmis-frontend/pull/758)). ~I _think_ the url params for the new Preview route should be updated at that point to use the identifier rather than ID.~ Edit: Decided against this. The url params have been updated to use both `identifier` and `formId`, see comment here https://github.com/greenriver/hmis-frontend/pull/758#discussion_r1592688961

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/graphql.schema.json
+++ b/graphql.schema.json
@@ -37209,6 +37209,22 @@
             "deprecationReason": null
           },
           {
+            "name": "canManageForms",
+            "description": null,
+            "args": [],
+            "type": {
+              "kind": "NON_NULL",
+              "name": null,
+              "ofType": {
+                "kind": "SCALAR",
+                "name": "Boolean",
+                "ofType": null
+              }
+            },
+            "isDeprecated": false,
+            "deprecationReason": null
+          },
+          {
             "name": "canManageIncomingReferrals",
             "description": null,
             "args": [],
--- a/graphql.schema.json
+++ b/graphql.schema.json
@@ -20288,22 +20288,6 @@
             "deprecationReason": null
           },
           {
-            "name": "status",
-            "description": null,
-            "args": [],
-            "type": {
-              "kind": "NON_NULL",
-              "name": null,
-              "ofType": {
-                "kind": "ENUM",
-                "name": "FormStatus",
-                "ofType": null
-              }
-            },
-            "isDeprecated": false,
-            "deprecationReason": null
-          },
-          {
             "name": "system",
             "description": null,
             "args": [],
@@ -22358,35 +22342,6 @@
         "inputFields": null,
         "interfaces": [],
         "enumValues": null,
-        "possibleTypes": null
-      },
-      {
-        "kind": "ENUM",
-        "name": "FormStatus",
-        "description": null,
-        "fields": null,
-        "inputFields": null,
-        "interfaces": null,
-        "enumValues": [
-          {
-            "name": "draft",
-            "description": "3 - Draft",
-            "isDeprecated": false,
-            "deprecationReason": null
-          },
-          {
-            "name": "published",
-            "description": "2 - Published",
-            "isDeprecated": false,
-            "deprecationReason": null
-          },
-          {
-            "name": "retired",
-            "description": "1 - Retired",
-            "isDeprecated": false,
-            "deprecationReason": null
-          }
-        ],
         "possibleTypes": null
       },
       {
--- a/src/modules/admin/components/forms/FormPreview.tsx
+++ b/src/modules/admin/components/forms/FormPreview.tsx
@@ -104,7 +104,7 @@ const FormPreview = () => {
         />
       ))
     );
-  }, [formDefinition, formValues, localConstants, toggleValue]);
+  }, [formDefinition, formValues, initialValues, localConstants, toggleValue]);
 
   if (loading && !formDefinition) return <Loading />;
   if (error) throw error;
--- a/src/modules/admin/components/forms/FormPreview.tsx
+++ b/src/modules/admin/components/forms/FormPreview.tsx
@@ -20,9 +20,23 @@ import {
   AlwaysPresentLocalConstants,
   getInitialValues,
   getItemMap,
-  getLeftNavItems,
+  getFormStepperItems,
 } from '@/modules/form/util/formUtil';
 import { useGetFormDefinitionFieldsForEditorQuery } from '@/types/gqlTypes';
+
+type PreviewMode = 'preview' | 'readOnly';
+const toggleItems: ToggleItem<PreviewMode>[] = [
+  {
+    value: 'preview',
+    label: 'Preview Form',
+    Icon: ListIcon,
+  },
+  {
+    value: 'readOnly',
+    label: 'Read Only View',
+    Icon: AccountBoxIcon,
+  },
+];
 
 const FormPreview = () => {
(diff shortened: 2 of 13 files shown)


src/modules/admin/components/forms/FormPreview.tsx
martha, line 1: @gigxz on our checkin you mentioned this was really slow, but I'm finding the performance to be ok (loading the qa spdat, for example). Could you send an example of a form that loads painfully slowly?

graphql.schema.json
```diff
@@ -20287,6 +20287,22 @@
             "isDeprecated": false,
             "deprecationReason": null
           },
+          {
+            "name": "status",
```
gigxz: Is this PR dependent on https://github.com/greenriver/hmis-warehouse/pull/4326 or did this just slip in here accidentally? 4326 looks like its almost there, so this seems fine either way.

src/modules/admin/components/forms/FormPreview.tsx
gigxz: Hmm yep I can slack it to you. I would try adding CPU throttling in the browser too.

graphql.schema.json
martha: Hm, no this was unintentional - I'll take it out and make a separate PR for the frontend schema changes that go with greenriver/hmis-warehouse#4326.
Edit: Created https://github.com/greenriver/hmis-frontend/pull/763

src/modules/admin/components/forms/FormPreview.tsx
```diff
@@ -0,0 +1,127 @@
...
+      value: 'preview',
+      label: 'Preview Form',
+      Icon: ListIcon,
+    },
+    {
+      value: 'readOnly',
+      label: 'Read Only View',
+      Icon: AccountBoxIcon,
+    },
+  ];
```
gigxz: I don't think this needs to be inside the component (ie being re-defined every render)? I would move it up to line 26

src/modules/form/util/formUtil.ts
```diff
@@ -1476,6 +1476,36 @@ export const parseOccurrencePointFormDefinition = (
...
+  // Remove disabled groups
+  leftNavItems = leftNavItems.filter((item) =>
+    shouldEnableItem({
+      item,
+      values: initialValues,
+      itemMap,
+      localConstants: localConstants || {},
+    })
+  );
+  if (leftNavItems.length < 3) return false;
```
gigxz: Shouldn't this be `minGroupsForLeftNav`?

src/modules/form/util/formUtil.ts
```diff
@@ -1476,6 +1476,36 @@ export const parseOccurrencePointFormDefinition = (
 
   return { displayTitle, isEditable, readOnlyDefinition };
 };
+
+export const getLeftNavItems = (
```
gigxz: Nice refactor. I think we can call it something more descriptive though, like `getFormStepperItems`? The component itself is called `FormStepper`. Usually we use "nav" to talk about the left nav on dashboards, so it could be confusing

src/modules/form/util/formUtil.ts
```diff
@@ -1476,6 +1476,36 @@ export const parseOccurrencePointFormDefinition = (
...
+  localConstants: LocalConstants,
+  minGroupsForLeftNav: number = 3
+) => {
+  if (!formDefinition || !itemMap) return false;
+
+  let leftNavItems = formDefinition.definition.item.filter(
+    (i) => i.type === ItemType.Group && !i.hidden
+  );
+
+  if (leftNavItems.length < minGroupsForLeftNav) return false;
```
gigxz: This check should probably happen after we remove the disabled groups?

gigxz (commented):
Looks great @martha. It still feels really laggy to me on a large assessment. Were you able to repro that?

src/modules/form/util/formUtil.ts
martha: Yes, looks like the [previous code](https://github.com/greenriver/hmis-frontend/blob/9dcf0821612a3d4e6674e0c3709ca4c3f2326d9b/src/modules/form/components/EditRecord.tsx#L99-L110) did the check in both places but I agree it could be cleaned up! 🧹

martha:
@gigxz yup it's definitely laggy for sure. It's especially noticeable with CPU throttling but, like you, I'm noticing it even without.

I poked around using react devtools's Profiler and noticed that the whole DynamicForm is indeed rendering on every change to the form. If I'm interpreting correctly, this is not unique to the Form Preview though; it's also happening elsewhere that the DynamicForm appears, such as the 'Edit Form' page and when filling out an assessment. Are you also noticing lag elsewhere on large forms like uha?

gigxz:
@martha goood question. Yes locally I get a lot of lag typing into the UHA. It is completely resolved with https://github.com/greenriver/hmis-frontend/pull/612, which is great, we just need to find the time for thorough QA so we can get that changed merged in.

Let's go ahead with merging this as-is, then. I'll try to make some forward progress on testing the other PR that addresses form performance.

src/modules/admin/components/forms/FormPreview.tsx
```diff
@@ -0,0 +1,138 @@
...
+          onSaveDraft={(values) => setFormValues(values)}
+          errors={{ errors: [], warnings: [] }}
+          localConstants={localConstants}
+          initialValues={initialValues}
+          FormActionProps={{ config: [] }}
+          ref={formRef}
+        />
+      ))
+    );
+  }, [formDefinition, formValues, localConstants, toggleValue]);
```
gigxz: I think the github annotation here is correct, missing a dependency

gigxz (approved):

src/modules/admin/components/forms/FormPreview.tsx
martha: 🤦‍♀️ thanks!

martha:
re: performance, that sounds fine by me. I'd be glad to share the work of qa-ing the react-hook-forms PR, too, if it's possible to divide it up. I'm not sure (still not really a pro with performance dev tools), but based on a little more poking around, it seems like the reason for the current laggy behavior of Dynamic Form re-rendering on every keypress is due to `values` changing ([here](https://github.com/greenriver/hmis-frontend/blob/9dcf0821612a3d4e6674e0c3709ca4c3f2326d9b/src/modules/form/hooks/useDynamicFields.tsx#L44)) and `disabledLinkIds` changing ([here](https://github.com/greenriver/hmis-frontend/blob/9dcf0821612a3d4e6674e0c3709ca4c3f2326d9b/src/modules/form/hooks/useDynamicFields.tsx#L55C10-L55C25)).
Screenshot 2024-05-08 at 4 01 15 PM
Screenshot 2024-05-08 at 4 01 28 PM

PR #767, opened 2024-05-08
Enable editing supportsBulkAssignment attr of service type

martha:
## Description

GH issue https://github.com/open-path/Green-River/issues/6012
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4343

Changes in this PR:
- Schema changes corresponding to the backend PR, renaming the mutation "renameServiceType" to "updateServiceType"
- That mutation now has the ability to update the `supportsBulkAssignment` boolean attribute of a service type. This PR also adds a "Supports bulk assignment" checkbox to the dialog form for updating service types.
- Factoring out said dialog into its own component. This was mainly because it made it easier to call `useState` with an `initialState` that was the result of a hook from the parent component. But I think this change also has some code legibility benefits too. This also should fix an existing [bug](https://greenriver.slack.com/archives/C061SAW3LFJ/p1715112738599019) in prod (though I believe not one that any users have run into, yet), where the updater throws a 500 error if you open the update modal and hit 'submit' without making any changes.
- The service detail page now notes whether this service does or does not support bulk assignment. Let me know if you agree with this user-facing change -- it was helpful to me during testing, but if there's any reason we don't want to expose it to users, I'm happy to remove

Screenshot 2024-05-08 at 5 24 14 PM

## Type of change
- [x] Bug fix
- [ ] Dependency update


gigxz (approved):
looks good!

PR #770, opened 2024-05-09
Make MUI Tree proof-of-concept for Forms

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6038

This PR adds a proof-of-concept FormTreeView, with a storybook story to check it out (it's not implemented anywhere in the actual codebase yet, per instruction [here](https://github.com/open-path/Green-River/issues/6048#issue-2287996633) that we should leave the existing JSON editor as-is for now.)

This isn't quite there yet either styling-wise or conceptually, but I think it proves that MUI's tree view component is flexible enough to work for us!

See TODOs in the code for open questions and unfinished business.

## Type of change
- [ ] Bug fix
- [x] Dependency update


src/modules/admin/components/forms/FormTreeView.tsx
```diff
@@ -0,0 +1,267 @@
...
+  FormImageIcon,
+  FormIntegerIcon,
+  FormObjectIcon,
+  FormStringIcon,
+  FormTextIcon,
+  FormTimeOfDayIcon,
+} from '@/components/elements/SemanticIcons';
+import theme from '@/config/theme';
+import { FormDefinitionJson, ItemType } from '@/types/gqlTypes';
+
```
gigxz: I'm sure there will end up being lots of components associated with the form builder. I think it would be good to have its own directory for form builder components.

Either
* `src/modules/admin/components/forms/formBuilder/<components>`
or
* `src/modules/formBuilder/components/<components>`

I think I prefer adding an entirely new module just for the form builder, since there will probably be util functions and maybe hooks associated with it. In my eyes it is very distinct from the rest of the application, so encapsulating that inside a module seems right.

src/modules/admin/components/forms/FormTreeView.tsx
```diff
@@ -0,0 +1,267 @@
...
+      return { icon: FormObjectIcon, text: '', ...defaultQuestion };
+  }
+};
+
+interface FormTreeLabelProps {
+  children?: React.ReactNode;
+  question?: FormItemDisplay;
+  required?: boolean;
+}
+const FormTreeLabel: React.FC<FormTreeLabelProps> = ({
```
gigxz: This can probably go in its own file

src/modules/admin/components/forms/FormTreeView.tsx
```diff
@@ -0,0 +1,267 @@
...
+  backgroundColor: string;
+  textColor: string;
+  hoverColor: string;
+}
+
+interface FormTreeItemProps
+  extends Omit<UseTreeItem2Parameters, 'rootRef'>,
+    Omit<React.HTMLAttributes<HTMLLIElement>, 'onFocus'> {}
+
+const FormTreeItem = React.forwardRef(function FormTreeItem(
```
gigxz: This can probably go in its own file

src/modules/admin/components/forms/FormTreeView.tsx
```diff
@@ -0,0 +1,267 @@
...
+        )}
+      </TreeItem2Root>
+    </TreeItem2Provider>
+  );
+});
+
+interface FormTreeViewProps {
+  definition: FormDefinitionJson;
+}
+const FormTreeView: React.FC<FormTreeViewProps> = ({ definition }) => {
```
gigxz: This can probably go in its own file

src/modules/admin/components/forms/FormTreeView.tsx
```diff
@@ -0,0 +1,267 @@
...
+            <Typography variant='caption'>{question.text}</Typography>
+          </Stack>
+        </Badge>
+      )}
+      {children}
+    </TreeItem2Label>
+  );
+};
+
+interface FormItemDisplay {
```
gigxz: This type could go in a types.ts file for the module if its needed across multiple components / util fns

src/modules/admin/components/forms/FormTreeView.tsx
```diff
@@ -0,0 +1,267 @@
...
+  definition: FormDefinitionJson;
+}
+const FormTreeView: React.FC<FormTreeViewProps> = ({ definition }) => {
+  // TODO - let's discuss the best way to support nesting
+  // MUI tree view component expects the 'item' prop to be named 'children', and doesn't provide an overrider function.
+  // OS contribution to react mui that provides an override? (but maybe there's a reason they don't allow)
+  // rewriting our DSL to use 'children' instead of 'item'? (but this feels like big lift + bad coupling)
+  // keep and clean up the below adapter, expanding it to n layers of nesting (n = 5?) ?
+  // We might need something like this adapter anyway to consolidate down some functional groups, like SSN
+  const definitionForTree = useMemo(() => {
```
gigxz: This seems okay to me. We can use the existing `modifyFormDefinition` function instead though, something like the below should add `children` to each item recursively.
```
modifyFormDefinition(definition, (item) => (item.children = item.item)
```

We probably _do_ need to place a limit on recursion, though, that should be incorporated somewhere in the form builder. (e.g. at a certain point, you can no longe add a nested group to an existing group)

gigxz (approved):
This looks awesome!! Nice work! Left some suggestions just around refactoring files.

gigxz:
running `yarn` for me on this branch resulted in changes to `yarn.lock`. do you see the same thing? if so can you commit those changes

src/modules/admin/components/forms/FormTreeView.tsx
martha: Oh this sounds good. Thanks for pointing it out to me! `modifyFormDefinition` doesn't work out-of-the-box exactly, bc typescript is not happy about appending a new property `children` to `item` since that prop doesn't exist on tyep `FormItem`, but something similar will work -- maybe I will adapt the existing recursive function or write a new one.

I agree about the recursion limit. Added a note about that here so it doesn't get lost: https://github.com/open-path/Green-River/issues/6051

gigxz (approved):
looks good! lets merge into 117

PR #772, opened 2024-05-13
Create skeleton layout for FormBuilder

martha:
## Description

GH Issue: https://github.com/open-path/Green-River/issues/6048

This PR creates a basic layout for the FormBuilder, with lots of todos still scattered throughout the code and buttons that aren't hooked up yet, etc. Those are reflected in the tickets for the rest of the FormBuilder, here: https://github.com/open-path/Green-River/labels/form%20builder

## Type of change
- [ ] Bug fix
- [ ] Dependency update


gigxz:
@martha try this for the build issue in the jest config:
`transformIgnorePatterns: ['/node_modules/(?!@mui/x-tree-view|@rails)/'],`

The App test can be a pain because its really the only test that uses jest to transpile the code.  I feel like we can probably get rid of the App test honestly it's not giving us much, and it's overhead to maintain the jest configuration so it will work. But hopefully that should fix it for this case

src/modules/formBuilder/components/FormBuilderHeader.tsx
```diff
@@ -0,0 +1,80 @@
...
+          >
+            <EditIcon fontSize='inherit' />
+          </IconButton>
+        </ButtonTooltipContainer>
+      </Stack>
+      <Stack direction='row' sx={{ alignItems: 'center' }}>
+        <Grid container>
+          <Grid item xs={3}>
+            Form Identifier
+          </Grid>
```
gigxz: We could use the `CommonLabeledTextBlock` we usually use for this type of thing, so we're not introducing a new pattern of key/value display:

```
        <Grid container columnGap={4} rowGap={2}>
          <CommonLabeledTextBlock title='Form Identifier'>
            {formDefinition.identifier}
          </CommonLabeledTextBlock>
          <CommonLabeledTextBlock title='Form Type'>
            {formRole}
          </CommonLabeledTextBlock>
          <CommonLabeledTextBlock title='Status'>
            {formDefinition.status}
          </CommonLabeledTextBlock>
```

src/modules/formBuilder/components/FormBuilderHeader.tsx
```diff
@@ -0,0 +1,80 @@
...
+          <Grid item xs={3}>
+            Form Title (Role)
```
gigxz: I think we landed on "Form Type" as the user-facing language

src/routes/routes.ts
```diff
@@ -149,9 +150,10 @@ export const AdminDashboardRoutes: {
   [k in AdminSubRoutesType]: string;
 } = adminDashboardRoutes;
 
-// Auto-hide left desktop nav for some routes (unused)
+// Auto-hide left desktop nav for some routes
 export const HIDE_NAV_ROUTES: string[] = [
   // EnrollmentDashboardRoutes.NEW_ASSESSMENT
+  AdminDashboardRoutes.EDIT_FORM,
```
gigxz, line 156: good find :)

src/modules/formBuilder/components/FormItemPalette.tsx
```diff
@@ -0,0 +1,89 @@
...
+        {/* TODO - get the Icon from the same source as getItemDisplayAttrs in FormTreeItem.tsx */}
+        <PaletteCard color={lightGreen} icon={FormGroupIcon}>
+          Group
+        </PaletteCard>
+        <PaletteCard color={lightGreen} icon={FormDisplayIcon}>
+          Display Text
+        </PaletteCard>
+        <Typography variant='caption'>Basic Inputs</Typography>
+        <PaletteCard icon={FormStringIcon}>Text</PaletteCard>
+        <PaletteCard icon={FormTextIcon}>Paragraph</PaletteCard>
```
gigxz: It would probably be a good idea to separate out the content from the presentational component. For example having an array of type `{ itemType: ItemType; displayName: string; IconClass: SvgIconComponent }` and using that structure to render the list of PaletteCards.
That way, when we update the _content_ we don't need to touch the component, resulting in hopefully less code churn when we add new item types.


Edit: you are  prob already referring to that with comment `get the Icon from the same source as getItemDisplayAttrs in FormTreeItem.tsx`, I just put that together

src/components/layout/dashboard/DashboardContentContainer.tsx
```diff
@@ -115,16 +118,17 @@ const DashboardContentContainer: React.FC<Props> = ({
...
               pb: 8,
               px: { xs: 1, sm: 3, lg: 4 },
               maxWidth: `${maxPageWidth}px`,
+              ...contentSx,
             }}
           >
-            <Box key='content' component='main'>
-              <SentryErrorBoundary>{children}</SentryErrorBoundary>
-            </Box>
+            <SentryErrorBoundary>{children}</SentryErrorBoundary>
```
gigxz, line 131: This feels like it could be a big change, there might be some styling somewhere that expects two divs. Did you check to see if there is any impact on the dashboards?

gigxz (approved):
looks like a great start!

src/components/layout/dashboard/DashboardContentContainer.tsx
martha: I did check the dashboards, both mobile and desktop, and they look okay. However, of course I'm not super confident that there isn't any styling anywhere that doesn't expect 2 divs here, so I'd be happy to revert this change and leave it be.

The reason to make the change in the first place was only code-stylistic, anyway -- I liked the idea of `DashboardContentContainer` accepting a `contentSx` prop that it applies to the div labelled 'content', whereas with the 2-Box layout it needs to accept an `sx` prop and pass it to the Box one level above 'content,' and I just wasn't sure what to name that sx prop, haha!

Let me know what you think, I could go either way.

gigxz: great! I think it's a nice clean-up, I think we should keep it.

src/modules/formBuilder/components/FormItemPalette.tsx
martha: Yes good point, I had left that as a todo for later, but your comment prompted me to just do it as part of this layout ticket, since it's shared/base functionality!

src/modules/formBuilder/components/formTree/FormTree.tsx
```diff
@@ -1,34 +1,41 @@
...
-      items={definitionForTree}
-      getItemId={(item) => item.linkId}
-      getItemLabel={(item) => item.text || item.helperText || item.linkId}
-      slots={{ item: FormTreeItem }}
-      onItemFocus={(_event, itemId) => {
-        // TODO - add node onclick handler
-        console.log(itemId);
-      }}
-    />
+    <TreeItemContext.Provider value={context}>
```
martha, line 30: (this provider gets around needing to pass props directly to the custom TreeItem element. Took this idea from the example here: https://mui.com/x/react-tree-view/rich-tree-view/customization/#custom-label)

gigxz: Nice. So far we usually use a pattern of creating the context in its own file, like `src/modules/auth/AuthContext.tsx` instead of inside a component file. I find that a bit neater, since the context itself isn't really tied to one component.

gigxz:
Feel free to merge @martha!

gigxz (approved):

PR #777, opened 2024-05-17
Allow scrolling on record picker dialog

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6062

Removing `overflow: hidden` in RecordPickerDialog fixes the usability bug and enables scrolling. However, I think it's still not the originally desired appearance. The record picker table is being told to take up 100% height, but I think that this needs to be a calculated value (100% minus some overhead) in order to not cause the table to always scroll even when its own content is short enough that it shouldn't cause scrolling.

I spent quite a bit of time fiddling with the height CSS, including trying to implement something using flex box; and also trying to change how the `fullHeight` prop is implemented in `GenericTable`. In the end I decided I had spent too much time on it to continue blocking the usability issue. I'd be glad to pair on this though.

The other two changes fix a more minor scrolling issue with the assessment form sidebar. The left hand assessment nav bar should be sticky, but previosuly it would move around a bit when you first scrolled down the page, and the bottom (including the autofill button) would be truncated a bit -- just a visual annoyance, not enough to impact functionality. These problems should both be fixed now.

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/modules/form/components/RecordPickerDialog.tsx
+++ b/src/modules/form/components/RecordPickerDialog.tsx
@@ -70,11 +70,6 @@ const RecordPickerDialog = ({
     return [...assessmentColumns, ...dataColumns];
   }, [item]);
 
-  // Need to set height on the dialog in order for the scrolling to work
-  // Columns are really rows here because this is a vertical table.
-  // The height should be between 550 and 850, with the # of columns determining where it falls within that.
-  const height = `${Math.min(Math.max(columns.length * 60 + 250, 550), 850)}px`;
-
   const hudRoles = [
     AssessmentRole.Intake,
     AssessmentRole.Update,
@@ -91,7 +86,7 @@ const RecordPickerDialog = ({
       fullWidth
       onClose={onCancel}
       sx={{
-        '.MuiDialog-paper': { overflow: 'hidden', height },
+        '.MuiDialog-paper': { overflow: 'hidden' },
       }}
       {...other}
     >
@@ -107,9 +102,8 @@ const RecordPickerDialog = ({
 
       <DialogContent
         sx={{
-          pb: 6,
-          my: 2,
-          height: '100%', // need for scrolling
+          pb: 2,
+          my: 0,
         }}
       >
         {description}
--- a/src/modules/form/components/RecordPickerDialog.tsx
+++ b/src/modules/form/components/RecordPickerDialog.tsx
@@ -1,4 +1,4 @@
-import { Box, Stack } from '@mui/material';
+import { Stack } from '@mui/material';
 import Button from '@mui/material/Button';
 import { DialogProps } from '@mui/material/Dialog';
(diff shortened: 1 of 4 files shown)


src/modules/form/components/RecordPickerDialog.tsx
```diff
@@ -71,6 +71,8 @@ const RecordPickerDialog = ({
   }, [item]);
 
   // Need to set height on the dialog in order for the scrolling to work
+  // Columns are really rows here because this is a vertical table.
+  // The height should be between 550 and 850, with the # of columns determining where it falls within that.
   const height = `${Math.min(Math.max(columns.length * 60 + 250, 550), 850)}px`;
```
gigxz: Ok this REALLY was overly complicated and brittle, that's my bad. I think I did it this way because the designs showed that the table scrolled, but the pagination at the bottom was sticky. 

I think the actual regression came with change behavior for the top toolbar, because that wasn't there before. If you add `noSort` to the table it starts to look sane again.

Aaanyway.. I think we should switch to something more sustainable that doesn't rely on pixels like this. Something like https://github.com/greenriver/hmis-frontend/commit/90f229a46856da8f68f8aae495905d2d23618007 would make it so we just scroll the whole dialog, not just the table. I think that will be more robust to future changes to the table.

martha: Ah gotcha. Thanks for digging up the history. I like the simplification. I guess the downside is that now the pagination buttons jump around a bit when you move from page to page, depending on the line lengths of the records. Maybe that is not a big enough deal to warrant any more complex (brittle) styling.

martha: Ok, https://github.com/greenriver/hmis-frontend/pull/777/commits/974d1332be40a1162f6f0e196003224fb3712855 shows a suggestion for another way - possibly this is still too brittle, but it does have the advantage of keeping the dialog's height consistent as you page through the paginated records!

To be perfectly honest, that `190px` on line 114 was a matter of guess and check 😬

gigxz: Hmm, I still think we should remove the hard-coded pixels because it will be brittle. I'm also seeing 2 scrollbars here, that feels jumpy because when you scroll down it switches from the inner scroll to the outer scroll (and for a moment you see both). I think to keep this simpler for the future, we should just get rid of the inner scroll.

Screenshot 2024-05-21 at 6 51 35 AM
Screenshot 2024-05-21 at 6 52 07 AM

martha: Ok good plan, you're right this was too brittle. I reverted to the version where the height may be inconsistent depending on the height of the rows within the table - maybe a minorly annoying ux but certainly better than 2 scroll bars!

gigxz (approved):
thanks!

PR #778, opened 2024-05-19
Make outgoing referral form configurable

gigxz:
Issue: https://github.com/open-path/Green-River/issues/6037
Backend PR: https://github.com/greenriver/hmis-warehouse/pull/4389 - see description for how to test

* Make outgoing referral form configurable, using DyanmicForm and SubmitForm
* Show custom data elements on the referral detail page
* Link to the referral from the Enrollment details card
* Fix issue with how some statuses were displaying on referral detail page


src/modules/projects/components/ProjectOutgoingReferralDetailsSubForm.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  return (
+    <Box>
+      {formDefinitionLoading && (
+        <LoadingSkeleton width={'100%'} count={1} sx={{ my: 2 }} />
+      )}
+      {formDefinition && (
+        <Box
+          // unset top-level card styling for form sections
+          sx={{ '.HmisForm-card': { px: 0, pt: 1, pb: 0, border: 'unset' } }}
```
ttoomey: This is overriding styles within DynamicForm, right? Do you think it would be worth it to pull this into the component and handle it with a variant prop? \<DynamicForm variant='subform'/\>

ttoomey (approved):
Looks good, thanks!

src/modules/projects/components/ProjectOutgoingReferralDetailsSubForm.tsx
gigxz: Good point, there were actually a few places that we did this. The others are modals. I made the variant `without_top_level_cards` so it's more descriptive, since it applies to some modals and sub-forms.

PR #779, opened 2024-05-20
Create basic form item preview in editor drawer

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6085

The designs for the drawer in general aren't finalized yet, so I hesitated a bit over typography hierarchy here but decided it's probably out of scope for now.

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/modules/form/types.ts
+++ b/src/modules/form/types.ts
@@ -59,7 +59,7 @@ export interface DynamicFieldProps {
 
 export interface DynamicViewFieldProps {
   item: FormItem;
-  nestingLevel: number;
+  nestingLevel?: number;
   value: any;
   horizontal?: boolean;
   pickListArgs?: PickListArgs;
--- a/src/modules/form/types.ts
+++ b/src/modules/form/types.ts
@@ -59,7 +59,6 @@ export interface DynamicFieldProps {
 
 export interface DynamicViewFieldProps {
   item: FormItem;
-  nestingLevel?: number;
   value: any;
   horizontal?: boolean;
   pickListArgs?: PickListArgs;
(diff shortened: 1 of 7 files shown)


src/modules/formBuilder/components/FormEditorItemPreview.tsx
```diff
@@ -0,0 +1,50 @@
...
+}
+
+const FormEditorItemPreview: React.FC<FormEditorItemPreviewProps> = ({
+  selectedItem,
+  value,
+  setValue,
+}) => {
+  const [toggleValue, setToggleValue] = useState<PreviewMode>('input');
+
+  if (selectedItem.type === ItemType.Group) return;
```
gigxz: This feels like it should go in the parent component. We don't want to render the Group based on context, because it doesn't fit in the layout. That's easier to understand than "this component always renders null for groups". Like `selectedItem.type !== ItemType.Group &&  <FormEditorItemPreview />`

src/modules/form/types.ts
```diff
@@ -59,7 +59,6 @@ export interface DynamicFieldProps {
 
 export interface DynamicViewFieldProps {
   item: FormItem;
-  nestingLevel: number;
```
gigxz, line 62: Are we _sure_ this isn't used by any child items in the viewing context? I'm not sure if we should remove this, it might be necessary to know later if we do want the appearance to depend on the level of nesting. It's probably better to make the appearance declarative, as suggested here https://github.com/open-path/Green-River/issues/5759, but we haven't really made that jump yet.

gigxz: `QuestionGroup` looks at `nestingLevel` now so I think this change might be adjusting that behavior. Feels out of scope for a form builder change

gigxz: Oh, nevermind, it looks like I guess we _did_ make that jump, since it looks like nesting level isn't being passed for forms anymore. So this seems okay. I must have forgotten.

gigxz (approved):

src/modules/form/types.ts
martha: Yes, I just removed it to reduce confusion because I could see that it was unused, but I think we removed this entirely on individual items when we were doing updates to improve the mobile experience. It is still used by form groups, I think. Seems like something we could revisit with design?

PR #782, opened 2024-05-22
Make form item basic properties editable in drawer

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6086

This PR adds the most basic form item attributes to the editor drawer; see the linked ticket for which items are added.

Saving is out of scope for this ticket (https://github.com/open-path/Green-River/issues/6103), but for now, the properties are edited and saved in a local-state copy of the form item, so that the preview is updated in real time.

I can already identify lots of changes that we should probably make before releasing this to clients, but that shouldn't (necessarily?) block this PR, partly in the spirit of iterative development and partly because many of them depend on larger design decisions. I think it might be helpful for the design team to see and play with this mapped-exactly-to-the-implementation UI before we decide how to improve upon it. (See [this thread](https://www.figma.com/board/axGI9vwmuBRC4uZ11ZJ3RP?node-id=255-3467#804976250).)

Anyway, maybe we could start tracking these somewhere central, like in a google document? Here's my running list so far:
- Pick list reference could be a dropdown list?
- Two of the enum lists (disabled display and size) have labels in all caps, probably because they don't have human-legible labels defined anywhere so the enum value is just used as a default
- Custom field key mapping should be improved/clarified; maybe we should just add better help text around it for now? (Overall, design/ux team should revisit all of the help text copy that I've composed)
- Some attributes should not be available on all types. For example, “assessment date” should only be checkable on date types. Are there others like this?
- Should we reorder these?
- In general, how can we make this more usable -- e.g. (just brainstorming) combining some attributes into sections, hiding some in an "advanced" panel, language and layout changes, etc.

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/modules/formBuilder/components/FormEditorItemPreview.tsx
+++ b/src/modules/formBuilder/components/FormEditorItemPreview.tsx
@@ -10,13 +10,13 @@ import DynamicViewField from '@/modules/form/components/viewable/DynamicViewFiel
 import { FormItem } from '@/types/gqlTypes';
 
 interface FormEditorItemPreviewProps {
-  selectedItem: FormItem;
+  item: FormItem;
   value?: any;
   setValue: (value: any) => void;
 }
 
 const FormEditorItemPreview: React.FC<FormEditorItemPreviewProps> = ({
-  selectedItem,
+  item,
   value,
   setValue,
 }) => {
@@ -32,12 +32,12 @@ const FormEditorItemPreview: React.FC<FormEditorItemPreviewProps> = ({
         size='small'
         variant='gray'
       />
-      {toggleValue === 'readOnly' ? (
-        <DynamicViewField item={selectedItem} value={value} />
+      {toggleValue === 'readOnly' || item.readOnly ? (
+        <DynamicViewField item={item} value={value} />
       ) : (
         <DynamicField
           value={value}
-          item={selectedItem}
+          item={item}
           itemChanged={({ value }) => setValue(value)}
         />
       )}
--- a/src/modules/formBuilder/components/FormEditorItemPreview.tsx
+++ b/src/modules/formBuilder/components/FormEditorItemPreview.tsx
@@ -7,7 +7,7 @@ import {
 } from '@/modules/admin/components/forms/FormPreview';
 import DynamicField from '@/modules/form/components/DynamicField';
 import DynamicViewField from '@/modules/form/components/viewable/DynamicViewField';
-import { FormItem, ItemType } from '@/types/gqlTypes';
+import { FormItem } from '@/types/gqlTypes';
 
 interface FormEditorItemPreviewProps {
   selectedItem: FormItem;
@@ -21,8 +21,6 @@ const FormEditorItemPreview: React.FC<FormEditorItemPreviewProps> = ({
   setValue,
 }) => {
   const [toggleValue, setToggleValue] = useState<PreviewMode>('input');
-
-  if (selectedItem.type === ItemType.Group) return;
 
   return (
     <>
--- a/src/modules/formBuilder/components/FormEditorItemPreview.tsx
+++ b/src/modules/formBuilder/components/FormEditorItemPreview.tsx
@@ -1,0 +1,50 @@
+import { Typography } from '@mui/material';
+import React, { useState } from 'react';
+import CommonToggle from '@/components/elements/CommonToggle';
+import {
+  PreviewMode,
+  toggleItems,
+} from '@/modules/admin/components/forms/FormPreview';
+import DynamicField from '@/modules/form/components/DynamicField';
+import DynamicViewField from '@/modules/form/components/viewable/DynamicViewField';
+import { FormItem, ItemType } from '@/types/gqlTypes';
+
+interface FormEditorItemPreviewProps {
+  selectedItem: FormItem;
+  value?: any;
+  setValue: (value: any) => void;
+}
+
+const FormEditorItemPreview: React.FC<FormEditorItemPreviewProps> = ({
+  selectedItem,
+  value,
+  setValue,
+}) => {
+  const [toggleValue, setToggleValue] = useState<PreviewMode>('input');
+
+  if (selectedItem.type === ItemType.Group) return;
+
+  return (
+    <>
+      <Typography>Preview</Typography>
+      <CommonToggle
+        value={toggleValue}
+        onChange={setToggleValue}
+        items={toggleItems}
+        size='small'
+        variant='gray'
+      />
+      {toggleValue === 'readOnly' ? (
+        <DynamicViewField item={selectedItem} value={value} />
+      ) : (
+        <DynamicField
+          value={value}
+          item={selectedItem}
+          itemChanged={({ value }) => setValue(value)}
+        />
+      )}
+    </>
+  );
+};
+
+export default FormEditorItemPreview;
--- a/src/modules/formBuilder/components/FormEditorItemPreview.tsx
+++ b/src/modules/formBuilder/components/FormEditorItemPreview.tsx
@@ -1,10 +1,5 @@
-import { Typography } from '@mui/material';
-import React, { useState } from 'react';
-import CommonToggle from '@/components/elements/CommonToggle';
-import {
-  PreviewMode,
-  toggleItems,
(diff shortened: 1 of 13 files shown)


src/modules/formBuilder/components/FormEditorItemProperties.tsx
```diff
@@ -0,0 +1,182 @@
...
+        />
+        <LabeledCheckbox
+          label='Read-only'
+          checked={item.readOnly}
+          onChange={(e) =>
+            onChangeProperty('readOnly', (e.target as HTMLInputElement).checked)
+          }
+        />
+        <LabeledCheckbox
+          label='Hidden'
```
martha, line 131: I was a bit puzzled over what the Preview should display if the Hidden attribute is true. To discuss with design?

gigxz:
@martha great!! I think the easiest way to tackle your list of questions would be to throw a screenshot of the current implementation into Figma and add comments to it. Maybe we can designate a section of the figma board for wip-dev-screenshots.

src/modules/formBuilder/components/FormEditorItemProperties.tsx
```diff
@@ -0,0 +1,182 @@
...
+          options={disabledDisplayPickList}
+          onChange={(_e, value) => {
+            console.log(value);
+            onChangeProperty(
+              'disabledDisplay',
+              isPickListOption(value) ? value.code : undefined
+            );
+          }}
+        />
+        <TextInput
```
gigxz: This can be a single-select for the [pick list types](https://github.com/greenriver/hmis-frontend/blob/2d45a5c5ec38a020d878b1acfa11e1cdf6dc89d0/src/types/gqlEnums.ts#L913). We can do the whole list for now, though eventually it might make sense to slim it down and/or adjust the available options based on the form type (or other contextual factors).

src/modules/formBuilder/components/FormEditorItemPreview.tsx
```diff
@@ -32,12 +32,12 @@ const FormEditorItemPreview: React.FC<FormEditorItemPreviewProps> = ({
         size='small'
         variant='gray'
       />
-      {toggleValue === 'readOnly' ? (
-        <DynamicViewField item={selectedItem} value={value} />
+      {toggleValue === 'readOnly' || item.readOnly ? (
```
gigxz: Looking at this in the context of the editing drawer, I think it would be a bit cleaner to hide the toggle, and just show the read-only view if "read only" is checked. I feel like that gets at the desired functionality, which is to preview what this field will look like when it's rendered within the form. (And the user would still be able to preview that read-only view by temporarily toggling it to read-only, if they wanted to.)

What do you think?
I also had some trouble distinguishing what was the preview vs what was items -- maybe some background color would help, but we can also run that question by design.

Screenshot 2024-05-22 at 2 24 24 PM

src/modules/formBuilder/components/FormEditorItemProperties.tsx
```diff
@@ -0,0 +1,182 @@
...
+      >
+        <TextInput
+          label='Link ID'
+          value={item.linkId}
+          onChange={(e) => {
+            onChangeProperty('linkId', e.target.value);
+          }}
+          helperText='Unique ID for this form item'
+        />
+        <LabeledCheckbox
```
gigxz, line 75: Suggested improvements (in this pr or another)
* maybe `helperText="If checked, this date will be recorded as the Assessment Date on the assessment"`
* only show if type is date, and if form type is one of `AssessmentRole`s. I think there will be other cases where the configurability depends on the `role` of the definition, so we want to thread that data through props

src/modules/formBuilder/components/FormEditorItemProperties.tsx
```diff
@@ -0,0 +1,182 @@
...
+        }}
+        sx={{ maxWidth: MAX_INPUT_AND_LABEL_WIDTH }}
+      >
+        <TextInput
+          label='Link ID'
+          value={item.linkId}
+          onChange={(e) => {
+            onChangeProperty('linkId', e.target.value);
+          }}
+          helperText='Unique ID for this form item'
```
gigxz, line 60: This should be validated on `^[a-zA-Z_$][a-zA-Z0-9_$]*$` (once https://github.com/greenriver/hmis-warehouse/pull/4129 goes in), plus on uniqueness across the form. Those validations will come from the backend too, but with something like the regex of allowed characters, it can be nice to disallow certain things when you type (for example spaces, in this case). Not a huge thing but just pointing it out while I'm here- take or leave, I wouldn't spend a ton of time on it yet.

src/modules/formBuilder/components/FormEditorItemProperties.tsx
```diff
@@ -0,0 +1,182 @@
...
+        <TextInput
+          // TODO(#5776)
+          label='Mapping for custom field key'
+          value={item.mapping?.customFieldKey || ''}
+          onChange={(e) => {
+            onChangeProperty('mapping', { customFieldKey: e.target.value });
+          }}
+        />
+        <LabeledCheckbox
+          label='Warn if empty'
```
gigxz: * Helper text: "If checked, user will see a warning if they don't provide an answer to this question."
* Maybe this and `required` should be grouped together somehow

src/modules/formBuilder/components/FormEditorItemProperties.tsx
```diff
@@ -0,0 +1,182 @@
...
+          checked={item.warnIfEmpty}
+          onChange={(e) =>
+            onChangeProperty(
+              'warnIfEmpty',
+              (e.target as HTMLInputElement).checked
+            )
+          }
+        />
+        <FormSelect
+          label='Data collected about'
```
gigxz: Also only show for `AssessmentRole`s

src/modules/formBuilder/components/FormEditorItemProperties.tsx
```diff
@@ -0,0 +1,182 @@
...
+          onChange={(_e, value) => {
+            console.log(value);
+            onChangeProperty(
+              'size',
+              isPickListOption(value) ? value.code : undefined
+            );
+          }}
+        />
+        <FormSelect
+          label='Disabled Display'
```
gigxz: This should probably only show up if the item has an `enableWhen` clause. If the item is always shown, then this doesn't need to get set.

src/modules/formBuilder/components/FormEditorItemProperties.tsx
```diff
@@ -0,0 +1,182 @@
...
+          onChange={(_e, value) => {
+            console.log(value);
+            onChangeProperty(
+              'disabledDisplay',
+              isPickListOption(value) ? value.code : undefined
+            );
+          }}
+        />
+        <TextInput
+          label='Pick list reference'
```
gigxz: Only if item type is CHOICE or OPEN_CHOICE

gigxz: Let's discuss with design. We haven't mapped out which properties are type-specific, it's just kind of implicit in the code right now. That would be a good thing to do.

gigxz (approved):
Great start!!! Feel free to merge when it's ready to your liking, and I'll deploy it to QA.

src/modules/formBuilder/components/FormEditorItemProperties.tsx
gigxz: In the same vein as https://www.figma.com/board/axGI9vwmuBRC4uZ11ZJ3RP?node-id=368-5046#813135090

src/modules/formBuilder/components/FormEditorItemProperties.tsx
martha: Ok, sounds good. Let me know if you think just passing the whole entire `definition` object is too much data -- I was thinking we will need it later for the sake of conditional logic, for instance, because we need to find out what other link IDs exist in this form.

src/modules/formBuilder/components/FormEditorItemProperties.tsx
martha: I'll put them next to each other for now, but let's discuss more w/ design tomorrow about ordering and grouping these properties.

PR #784, opened 2024-05-24
Patch date display for midnight-datetimes

gigxz:
## Description

Before: `parseAndFormatDate('2024-05-24T00:00:00+00:00') ==  '05/23/2024'`
After:    `parseAndFormatDate('2024-05-24T00:00:00+00:00') ==  '05/24/2024'`

I think this is what we want, because when you call `parseAndFormatDate` on a datetime string you are explicitly requesting the date to be formatted without the time. I think midnight of 5/24 is 5/24.

Also has schema changes from https://github.com/greenriver/hmis-warehouse/pull/4400

## Type of change
- [x] Bug fix

### DIFF
--- a/src/modules/hmis/hmisUtil.ts
+++ b/src/modules/hmis/hmisUtil.ts
@@ -181,6 +181,9 @@ export const parseAndFormatDate = (
   dateString: string | null | undefined
 ): string | null => {
   if (!dateString) return null;
+  // Trim string to date only
+  if (dateString.length > 10) dateString = dateString.slice(0, 10);
+
   const parsed = parseHmisDateString(dateString);
   if (!parsed) return dateString;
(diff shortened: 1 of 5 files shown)


src/modules/hmis/hmisUtil.ts
```diff
@@ -181,6 +181,9 @@ export const parseAndFormatDate = (
   dateString: string | null | undefined
 ): string | null => {
   if (!dateString) return null;
+  // Trim string to date only
+  if (dateString.length > 10) dateString = dateString.slice(0, 10);
```
ttoomey: this is fine, probably don't need the condition though?
```suggestion
  // remove time from ISO8601 date-time string
  dateString = dateString.slice(0, 10);
```

ttoomey (approved):

PR #785, opened 2024-05-28
Implement save logic on form builder

martha:
## Description

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4404

This PR adds frontend support for saving the updates from the Form Builder.

The major open TODO for this PR is to block the user from navigating away when they have unsaved changes.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/formBuilder/components/FormBuilderContents.tsx
martha, line 1: I moved this into its own file `FormBuilderContents` in order to pass `formDefinition` from the parent and then initialize a new piece of state, `workingFormDefinition`, whose initial state depends on `formDefinition`.

I couldn't figure out a simple way to do this while keeping things all in one component, and kept running into a race condition with the initial state being provided by the result of a network call; but maybe that is just due to my limited React knowledge.

I'd also be happy to rename this to something clearer.

src/modules/formBuilder/components/FormBuilder.tsx
```diff
@@ -0,0 +1,174 @@
...
+        <FormBuilderHeader
+          formDefinition={formDefinition}
+          lastUpdatedDate={lastUpdatedDate}
+        />
+        <Box sx={{ p: 4 }}>
+          <FormTree
+            definition={workingDefinition}
+            onEditClick={(item: FormItem) => {
+              // todo @martha - this should NOT auto-save, it should instead prompt to save
+              // onSave(workingDefinition).then((values) => {
```
martha: @gigxz let's discuss the prompt-to-save approach a bit more? Here's what I learned so far:

react-router-dom has some handy built-in functionality for this, `useBlocker`. I played around with it and got something basic running [here](https://github.com/greenriver/hmis-frontend/commit/68385b327771da277b667146bcf25db4ac52420e). But it has some major caveats:
- Most importantly: It doesn’t work if you just plop it inside of a `<BrowserRouter/>`. It needs a Data Router: https://reactrouter.com/en/6.23.1/routers/picking-a-router 
    - This would be a big refactor of our routing, and it definitely feels like scope creep to me.
- It also only works for navigation events within our react app. It does not work for “document events” like hitting reload or navigating to another site directly. https://reactrouter.com/en/main/hooks/use-blocker (It does work on the back button, I think because React Router hijacks session history. It doesn't work on logout, which I wasn't clear about.)
- Finally, they state outright in their docs that it’s an anti-pattern to hijack user navigation, and recommend using session storage instead to persist unsaved values. Given the lift to get this implemented, maybe that would be easier too.

So, I guess, 3 proposed ideas:
- Full steam ahead with upgrading our React Router, which they [recommend doing anyway](https://reactrouter.com/en/6.23.1/routers/picking-a-router#web-projects)
- Accept a non-styled alert component, and attach it directly to the `window.beforeunload` event https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event 
- Just use session storage

gigxz: Thanks Martha. I thought we landed on a lower-lift approach that would just present the save modal when you clicked on certain elements on the page ("preview", "edit item"), and not worry about intercepting browser navigation.

All 3 of the proposed options seem too big for https://github.com/open-path/Green-River/issues/6103. It would be good to capture your research somewhere, though, can you create a ticket for "Implement blocker to stop navigation when user has unsaved changes" so we can address it for Assessments and Form building together? thanks!

martha: Ah, sorry, I must have misunderstood our conclusion on the design session, but the lower lift approach sounds fine to me for now. Yes, I will create a ticket.

edit: created https://github.com/open-path/Green-River/issues/6156

src/modules/formBuilder/components/FormBuilderPage.tsx
```diff
@@ -1,86 +1,116 @@
...
     undefined
   );
+  const [originalLinkId, setOriginalLinkId] = useState<string | undefined>(
+    undefined
+  );
+
+  const closeItemEditor = () => {
+    setSelectedItem(undefined);
+    setOriginalLinkId(undefined);
+  };
```
gigxz: Can wrap in `useCallback` so it's not re declared

src/modules/formBuilder/components/FormBuilderPage.tsx
```diff
@@ -1,86 +1,116 @@
...
+      formDefinition={formDefinition}
+      workingDefinition={workingDefinition}
+      setWorkingDefinition={setWorkingDefinition}
+      errorState={errorState}
+      lastUpdatedDate={lastUpdatedDate || undefined}
+      lastUpdatedBy={lastUpdatedBy}
+      selectedItem={selectedItem}
+      setSelectedItem={setSelectedItem}
+      originalLinkId={originalLinkId}
+      setOriginalLinkId={setOriginalLinkId}
```
gigxz: I think this passing around of "original link id" is a little confusing. If we only allow editing link id for new items -- before they are saved for the first time-- then this wouldn't be needed, correct? I think that would be preferable. Then we also don't have to worry about link id changes causing dependency problems. What do you think about adding that as a constraint? 

It doesn't look like we've yet implemented the behavior whereby when you add an item, the panel opens right away.  That would probably be necessary in order to implement the suggested constraint. (And the panel would have to "know" that it's working on a new item that has not yet been persisted).

src/modules/formBuilder/components/FormBuilder.tsx
```diff
@@ -0,0 +1,224 @@
...
+    formDefinition: FormDefinitionJson
+  ) => Promise<FetchResult<UpdateFormDefinitionMutation>>;
+  saveLoading: boolean;
+  lastUpdatedDate?: string;
+  lastUpdatedBy?: string;
+  selectedItem?: FormItem;
+  setSelectedItem: Dispatch<SetStateAction<FormItem | undefined>>;
+  originalLinkId?: string;
+  setOriginalLinkId: Dispatch<SetStateAction<string | undefined>>;
+  closeItemEditor: () => void;
```
gigxz: `() => void` can be typed as `VoidFunction`

src/modules/formBuilder/components/FormBuilder.tsx
```diff
@@ -0,0 +1,224 @@
...
+  originalLinkId,
+  setOriginalLinkId,
+  closeItemEditor,
+}) => {
+  const dirty = useMemo(() => {
+    return !isEqual(workingDefinition, formDefinition.definition);
+  }, [workingDefinition, formDefinition.definition]);
+
+  const [isBlocked, setIsBlocked] = useState(false);
+  const [unblockCallback, setUnblockCallback] = useState<
```
gigxz: Can you add some comments explaining these states? (I think I get it now, but it took me a bit).

gigxz: It also doesn't seem like we really need 2 pieces of state for this. We always set them together and clear them together, it seems. Maybe it would be simpler to have a single piece of state `blockedActionFunction` that is either opening the panel, navigating to the preview, or undefined.

src/modules/formBuilder/components/FormBuilderPage.tsx
gigxz: Digging through the code some more, another note is that we don't need to be passing the original link id all the way through the tree. The `FormEditorItemProperties` should already know what the "original" link ID is, because it is on the `item` that it receives initially. It could simply store that value:
`const [initialLinkId] = useState<string>(item.linkId);`

src/modules/formBuilder/components/FormBuilder.tsx
```diff
@@ -0,0 +1,224 @@
...
+        loading={saveLoading}
+        confirmText='Save changes'
+        cancelText='Continue editing'
+        title='Unsaved changes'
+        onConfirm={() => {
+          onSave(workingDefinition).then(() => {
+            setIsBlocked(false);
+            if (unblockCallback) unblockCallback(); // todo @Martha typescript
+            setUnblockCallback(undefined);
+          });
```
gigxz: Maybe pull this out to an `const onConfirmSave = useCallback(() => ...`. Same for the `onItemClick ` callback. For me I'm not sure how much of this is a style thing vs actually needing the memoization... it's probably fine as-is, but I find it easier to reason about the component when the logic is not dispersed throughout the jsx.

src/modules/formBuilder/components/FormBuilder.tsx
```diff
@@ -0,0 +1,224 @@
...
+                  setOriginalLinkId(item.linkId);
+                }
+
+                if (dirty) {
+                  setIsBlocked(true);
+                  setUnblockCallback(() => editItem);
+                } else {
+                  editItem();
+                }
+              }}
```
gigxz: Bringing together a couple of my previous comments, this could be streamlined into
```
if (dirty) {
   // form has unsaved changes, so prompt save before proceeding to edit
   setBlockedAction(() => setSelectedItem(item))  //<= no  need for second piece of state
} else {
   setSelectedItem(item); //<== no need for link id state
}
```

src/modules/formBuilder/components/FormBuilderHeader.tsx
```diff
@@ -56,7 +58,11 @@ const FormBuilderHeader: React.FC<FormEditorHeaderProps> = ({
           </CommonLabeledTextBlock>
         </Grid>
         <Stack direction='row' spacing={2}>
-          <Button variant='outlined' startIcon={<VisibilityIcon />}>
+          <Button
+            variant='outlined'
+            startIcon={<VisibilityIcon />}
+            onClick={() => onClickPreview()}
```
gigxz:
```suggestion
            onClick={onClickPreview}
```

src/modules/formBuilder/components/FormBuilderPage.tsx
```diff
@@ -1,86 +1,116 @@
...
+        closeItemEditor();
+      }
+    },
+  });
+
+  const formDefinition = useMemo(() => {
+    if (data && data.updateFormDefinition?.formDefinition)
+      return data.updateFormDefinition.formDefinition;
+    return initialFormDefinition;
+  }, [data, initialFormDefinition]);
```
gigxz: so I could see this causing bugs in the event that the `updateFormDefinition` mutation _failed_, then `data` would be present but `data.updateFormDefinition?.formDefinition` would be nil. I don't think it is correct to use `initialFormDefinition` in that scenario. But I'm not sure what it should be without tinkering with error scenarios.

gigxz (commented):
This is looking good!!

src/modules/formBuilder/components/FormBuilder.tsx
gigxz: thanks!

src/modules/formBuilder/components/FormBuilderPage.tsx
martha: Yes, this is way nicer. Thank you.

I created a new ticket https://github.com/open-path/Green-River/issues/6167 to capture the work for enforcing that constraint. It does seem nice because it lets us avoid some dependency problems. On the other hand, though, it might be frustrating to users (maybe us!) not to be able to update the link IDs of existing form items, especially items that only exist in a draft...

src/modules/formBuilder/components/FormBuilderPage.tsx
martha: That's interesting. I think in that scenario it should be the most recent successfully saved version. Which is `initialFormDefinition` if the user has not already made and saved changes on this page. But if they have, then it's the last result of the update mutation, which gets overridden when the failed mutation returns.

gigxz: ah right, good point. that seems fine to me.

src/modules/formBuilder/components/FormBuilderPage.tsx
gigxz: Awesome thanks :) yeah I could see that, too.

gigxz (approved):
👏  lgtm!

PR #788, opened 2024-05-30
Support editing `enable_when` and `autofill` conditions in Form Builder

gigxz:
## Description

Support specifying conditional rules (EnableWhen) and autofill behavior (AutofillValues) on the form item editor.

https://github.com/open-path/Green-River/issues/6088

## Type of change


src/modules/formBuilder/components/itemEditor/conditionals/EnableWhenCondition.tsx
```diff
@@ -0,0 +1,300 @@
...
+                textInputProps={{
+                  helperText: error?.message,
+                  error: !!error,
+                  inputRef: ref,
+                }}
+              />
+            )}
+          />
+        </Grid>
+        {/* COLUMN 3: Select the comparison value */}
```
gigxz, line 131: This is verbose / brittle. "Column 3" has a bunch of different inputs, depending on the type, but only one of them should show up. Maybe look into an alternative approach where there is only one input element, and it saves to the correct field depending on the type.

gigxz: Requested review but I still want to look at this @martha so no rush

src/modules/formBuilder/formBuilderUtil.ts
```diff
@@ -0,0 +1,24 @@
+import { ItemMap } from '../form/types';
```
martha, line 2: Ah, this is a better place for this, since it's a util not a component. We should move the contents of [components/formBuilderUtil.ts](https://github.com/greenriver/hmis-frontend/blob/release-119/src/modules/formBuilder/components/formBuilderUtil.ts) into here

gigxz: Ah okay, I moved those over

gigxz:
This is ready for review. There are still some TODOs but I think it's ready to get up to QA and play with, and do a design pass.

I re-targeted against https://github.com/greenriver/hmis-frontend/pull/799 which has some refactoring for how we use our inputs with RHF.

src/modules/formBuilder/components/itemEditor/conditionals/AutofillValueCard.tsx
```diff
@@ -0,0 +1,107 @@
...
+          // disabled={!answerInputTypes.includes('answerCode')}
+          label='Value (Formula)'
+          helperText="Formula to calculate the value to fill. Use 'value' to refer to the value of the current item."
+        />
+        <ControlledCheckbox
+          name={`autofillValues.${index}.autofillReadonly`}
+          control={control}
+          label='Autofill in read-only mode'
+        />
+        <ManageEnableWhen
```
martha, line 97: I'm not sure of the right place for this, but I think it could be helpful to add some comments about the overlap between Autofill and EnableWhen. My understanding is that we use the ManageEnableWhen component for Autofill logic because we are trying to do two similar things here -- indicate a set of conditions + when those conditions are applicable -- and so the data shape is the same. And in the [json form schema](https://github.com/greenriver/hmis-warehouse/blob/release-118/drivers/hmis_external_apis/public/schemas/form_definition.json#L181) itself we also reuse EnableWhen for Autofill. Maybe commentary about this already exists somewhere?

src/modules/formBuilder/components/itemEditor/conditionals/EnableWhenCondition.tsx
```diff
@@ -0,0 +1,196 @@
...
+        {/* COLUMN 1: Select the dependent source */}
+        <Grid item xs={4}>
+          <Stack gap={1}>
+            {!advanced.localConstant && (
+              <ControlledSelect
+                name={`${enableWhenPath}.${index}.question`}
+                control={control}
+                label='Dependent Question'
+                options={itemPickList}
+                helperText="Question who's response will determine whether the condition is met"
```
martha:
```suggestion
                helperText="Question whose response will determine whether the condition is met"
```

src/modules/formBuilder/formBuilderUtil.ts
```diff
@@ -1,12 +1,37 @@
...
+// Generate a list of Items to pick from. Used for picking dependent items for conditional logic.
+export const generateItemPickList = (
+  itemMap: ItemMap,
+  exclude: string[] = []
+) => {
+  const pickList: PickListOption[] = [];
+  Object.values(itemMap).forEach((item) => {
+    if (exclude.includes(item.linkId)) return;
+
+    let label = item.briefText || item.text || item.linkId;
```
martha:
```suggestion
    let label = item.briefText || item.text || item.linkId;
    label = label.replace(/<[^>]*>/g, '');
```
?
We could also use a library like html-to-text to do this, but seems like overkill to me.

martha (approved):
Looks awesome. I agree, it would be great to get this on QA and be able to play around with it during the design session tomorrow!

src/modules/formBuilder/formBuilderUtil.ts
gigxz: Good call, added with https://github.com/greenriver/hmis-frontend/pull/788/commits/1f94b4ea62ad098b9b60f06dddd4b9971d761a2b

src/modules/formBuilder/components/itemEditor/conditionals/AutofillValueCard.tsx
gigxz: That's right. I'll add some comments on these components explaining that the "EnableWhen" object type is used in two places in the schema (conditional rules for visibility + conditional rules for autofill).

We used the same graphql type for specifying the logic, because we needed the exact same shape. It would probably make more sense if the shape was called something like "Condition" rather than EnableWhen. https://github.com/greenriver/hmis-warehouse/blob/e4d8e97cbb875889094457e18178cb515692c126/drivers/hmis/app/graphql/types/forms/autofill_value.rb#L22-L24

PR #791, opened 2024-06-03
Enable editing additional form attributes

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6087

This PR adds a few new attributes to the Item Editor drawer, notably the Component dropdown.

Note - There aren't design mocks yet for the multi-pick list option creation, so I've implemented it as a comma-separated text box for now. I think this is v1-only and should be improved upon by the time we actually ship this to any customers.

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/modules/formBuilder/components/FormEditorItemProperties.tsx
+++ b/src/modules/formBuilder/components/FormEditorItemProperties.tsx
@@ -12,7 +12,7 @@ import {
   localResolvePickList,
   MAX_INPUT_AND_LABEL_WIDTH,
 } from '@/modules/form/util/formUtil';
-import { getComponents } from '@/modules/formBuilder/components/formBuilderUtil';
+import { validComponentsForType } from '@/modules/formBuilder/components/formBuilderUtil';
 import {
   AssessmentRole,
   Component,
@@ -52,9 +52,9 @@ const FormEditorItemProperties: React.FC<FormEditorItemPropertiesProps> = ({
   );
   const [initialLinkId] = useState<string>(item.linkId);
 
-  const components = useMemo(
+  const componentOverridePicklist = useMemo(
     () =>
-      getComponents(item.type).map((component) => {
+      validComponentsForType(item.type).map((component) => {
         return {
           code: component,
           label: startCase(component.toLowerCase()),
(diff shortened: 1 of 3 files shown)


src/modules/formBuilder/components/formBuilderUtil.ts
```diff
@@ -0,0 +1,45 @@
+import { ItemType, Component } from '@/types/gqlTypes';
+
+export const getComponents = (type: ItemType) => {
```
gigxz: maybe `validComponentsForType`?

gigxz: It also might be worth adding a comment that these are valid "component" overrides you can select for a given item type, but that each item type also has a default display (Except for `Object`?) that is used for it, which is specified in code in the DynamicField/DynamicViewField component. From a user-facing perspective this is really a "component override". We could come back to that later if it's confusing, but for now I think it's OK since they can see the default component in the preview.

src/modules/formBuilder/components/FormEditorItemProperties.tsx
```diff
@@ -49,6 +52,17 @@ const FormEditorItemProperties: React.FC<FormEditorItemPropertiesProps> = ({
   );
   const [initialLinkId] = useState<string>(item.linkId);
 
+  const components = useMemo(
```
gigxz: Naming suggestion: `componentOverridePicklist`?

src/modules/formBuilder/components/FormEditorItemProperties.tsx
```diff
@@ -201,23 +228,51 @@ const FormEditorItemProperties: React.FC<FormEditorItemPropertiesProps> = ({
...
-                (o) => o.code === item.pickListReference
-              ) || undefined
-            }
-            options={pickListTypesPickList}
-            onChange={(_e, value) => {
+        {([ItemType.Choice, ItemType.OpenChoice].includes(item.type) ||
+          (item.type === ItemType.Object &&
+            item.component === Component.Address)) && (
+          <LabeledCheckbox
+            label='Repeats'
```
gigxz: Maybe label can be `Allow multiple responses`?

src/modules/formBuilder/components/FormEditorItemProperties.tsx
```diff
@@ -60,6 +74,19 @@ const FormEditorItemProperties: React.FC<FormEditorItemPropertiesProps> = ({
         }}
         sx={{ maxWidth: MAX_INPUT_AND_LABEL_WIDTH }}
       >
+        {components.length > 0 && (
+          <FormSelect
+            label='Component'
```
gigxz:
```suggestion
            label='Component Override'
```

gigxz (commented):
looks good! just a few small comments

gigxz (approved):

PR #792, opened 2024-06-03
Add CLS tab to Projects dash

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6075
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4430

This PR adds the Current Living Situation tab to the Projects dashboard.

Pending confirmation on custom columns such as Notes, see: https://github.com/open-path/Green-River/issues/6075#issuecomment-2146017018

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/modules/projects/components/ProjectCurrentLivingSituations.tsx
+++ b/src/modules/projects/components/ProjectCurrentLivingSituations.tsx
@@ -6,6 +6,7 @@ import useSafeParams from '@/hooks/useSafeParams';
 import ClientName from '@/modules/client/components/ClientName';
 import GenericTableWithData from '@/modules/dataFetching/components/GenericTableWithData';
 import { baseColumns } from '@/modules/enrollment/components/dashboardPages/EnrollmentCurrentLivingSituationsPage';
+import EnrollmentDateRangeWithStatus from '@/modules/hmis/components/EnrollmentDateRangeWithStatus';
 import { EnrollmentDashboardRoutes } from '@/routes/routes';
 import {
   GetProjectCurrentLivingSituationsDocument,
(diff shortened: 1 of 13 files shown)


src/modules/projects/components/ProjectCurrentLivingSituations.tsx
```diff
@@ -0,0 +1,81 @@
...
+          GetProjectCurrentLivingSituationsQueryVariables,
+          ProjectCurrentLivingSituationFieldsFragment
+        >
+          queryVariables={{ id: projectId }}
+          queryDocument={GetProjectCurrentLivingSituationsDocument}
+          columns={columns}
+          pagePath='project.currentLivingSituations'
+          noData='No current living situations'
+          recordType='CurrentLivingSituation'
+          rowLinkTo={rowLinkTo}
```
gigxz, line 79: I think we should aim for consistency with the Project Services table here.
We still need to do more of a [table rework](https://www.figma.com/design/F9EHR2IZYuMAlOJ9aOG2IL/OP-%2F-HMIS-%2F-2024-%2F-Q1-%2B-Q2-(Jan-1---June-30)?node-id=3543-15008) but I think we at least agreed that the row should be clickable if there is only one action, like you have here, but we don't have on Project Services.

Suggestions to streamline the two to match:
1. Update Project Services to use a rowLinkTo that takes you to the services tab of the enrollment dash (matching your CLS impl)
2. Add `Enrollment Period` as the last column of the CLS table
  * I noticed that the Project Services enrollment period doesn't use the new component. Can you update that while you're at it so they both use `EnrollmentDateRangeWithStatus` to display the enrollment range? (That may involve resolving additional data, and updating the input type  for the EnrollmentDateRangeWithStatus component to be more general, like `Pick<Enrollment, 'entryDate' | 'exitDate' | 'inProgress'>`)
3. Remove Location Details col from the CLS table since they won't both fit


The table interactions and styles definitely more of a revisit, but this will at least get us consistency across the Assessments, Services, and CLS tabs.

martha: Cool! Thanks for this suggestion. I implemented it using a new fragment `EnrollmentRangeFields`, let me know if it looks okay -- I'm happy to do it the other way with a `Pick` type too.

gigxz: That looks good to me!

gigxz (approved):

PR #793, opened 2024-06-04
Update Form Palette

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6083

This PR:
- Moves the form palette over to the right side of the screen. It re-institutes the left-hand side Admin nav, per design.
- Updates the ordering, icons, and styling for the form palette items. They are now proper buttons with hover and tooltip effects.

There are still two changes I'd consider in-scope for this ticket, but I'd like to hold off pending more discussion and pairing:
- Generating link IDs for new form items - let's discuss, see inline comment
- Correctly updating the form structure when a new item is added - should wait for React Hook Forms before putting more time into this.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/formBuilder/components/FormBuilder.tsx
```diff
@@ -107,11 +108,96 @@ const FormBuilder: React.FC<FormBuilderProps> = ({
           proceeding?
         </Typography>
       </ConfirmationDialog>
-      <Box sx={{ display: 'flex' }}>
+      {selectedItem && (
```
martha, line 110: GH is making this change look bigger than it is -- it's just a rearrangement of these components due to moving the Form Palette over to the right-hand side.

src/modules/formBuilder/components/FormBuilder.tsx
```diff
@@ -107,11 +108,96 @@ const FormBuilder: React.FC<FormBuilderProps> = ({
...
+                </Stack>
+              </Paper>
+            )}
+          </Box>
+        </Box>
         <FormBuilderPalette
           onItemClick={(itemType) => {
             const newItem: FormItem = {
-              linkId: crypto.randomUUID(), // TODO(#6083) - this is a placeholder, but right now will only work in localhost
+              linkId: crypto.randomUUID(), // TODO(#6083) - this is a placeholder
```
martha: I'd like to discuss this. I was wrong in my comment that it'd only work in localhost -- seems it works as long as the context is secure (https). However, these randomly generated UUIDs aren't compliant with our new link ID requirements because they have hyphens.

In general, do we want to take the approach of randomly generating link IDs? Or should we try to programmatically guess at a more human-readable link ID?

src/modules/formBuilder/components/FormBuilder.tsx
```diff
@@ -122,100 +211,9 @@ const FormBuilder: React.FC<FormBuilderProps> = ({
               enableBehavior: EnableBehavior.Any,
             };
 
-            const newDefinition: FormDefinitionJson = {
-              ...workingDefinition,
-              item: [...workingDefinition.item, newItem],
-            };
-
-            setWorkingDefinition(newDefinition);
+            setSelectedItem(newItem);
```
martha, line 213: Okay, one more thread to pick up here. Should the user be able to create a new item and immediately save it, without changing any of the item's attributes? Right now this isn't possible, because the new item's form state is clean. 

It might be a confusing UX if you can't find the save button. But on the other hand, maybe it's a good thing to require changing those defaults before save.

gigxz: Hmm, I think that's fine to require them to change some values before submitting.  If that become an issue we can revisit

src/modules/formBuilder/components/FormBuilder.tsx
gigxz: All good questions! I think generating a random link id initially is appropriate, since we have nothing else to go off. Lets use the uuid library instead of crypto though:
```
import { v4 } from 'uuid';
v4() // remove dashes, or just take the first part `v4().split('-')[0]`
```

It would be even nicer if we replaced it with a readable LinkID once a user typed in a label. The advantage of readable Link IDs is that when they are referenced in other conditionals,  you can tell at a glance what they're talking about (when looking at the JSON). That might make it worth implementing.


I don't think it would be _too_ tricky to implement, now that we integrated RHF. We could use an onBlur on the label element to update the link id value. It probably shouldn't re-set itself if it was manually updated, which we should be able to tell from the dirtyFields state. The Link ID should probably be hidden in some "Advanced" section, too, since the average user shouldn't need to see it.

I fiddled around with a simple slugifier below. We'll likely want something similar/same for generating custom data element definition `key`s too.
```
linkId = kebabCase(item.label)
    .toLowerCase()
    .replace(/\-/g, '_')
    .slice(0, 40)
// if generated linkId not unique in the form, append unique uuid to it
```

What do you think of that approach?
If it's feeling too scope-creeping for this ticket, we can just go with uuid to start.

src/modules/admin/components/AdminDashboard.tsx
```diff
@@ -116,7 +116,10 @@ const AdminDashboard: React.FC = () => {
   const formEditorContentSx = {
     px: 0,
     py: 0,
+    maxWidth: '100%',
   };
+  const isFormEditor =
+    dashboardState.currentPath === '/admin/forms/:identifier/:formId/edit';
```
gigxz: We have a handy helper for this:
```
const currentPath = useCurrentPath();

currentPath === AdminDashboardRoutes.EDIT_FORM ? ...
```

src/modules/formBuilder/components/FormBuilder.tsx
```diff
@@ -105,12 +108,98 @@ const FormBuilder: React.FC<FormBuilderProps> = ({
...
+            )}
+          </Box>
+        </Box>
         <FormBuilderPalette
           onItemClick={(itemType) => {
             const newItem: FormItem = {
-              linkId: crypto.randomUUID(), // TODO(#6083) - this is a placeholder, but right now will only work in localhost
-              text: itemType.toString(),
+              linkId: crypto.randomUUID(), // TODO(#6083) - this is a placeholder
+              text: FORM_ITEM_PALETTE[itemType].displayName,
```
gigxz: I think this looks a little confusing. In the preview it's almost like "oh, is that just an extra label they put in to tell me what kind of question this is"? Idk. It might be best to just leave this blank, or put something generic like "label" or "untitled question"

gigxz (commented):
Overall this looks great @martha! 📈 
Just a few comments. This seems to fit in seamlessly with RHF saving, I'm not seeing any issues with it.

src/modules/form/util/formUtil.ts
```diff
@@ -295,10 +295,15 @@ export const updateFormItem = (
...
     }
 
-    throw new Error(`Link ID  ${newItem.linkId} not found`);
+    return false;
+  }
+
+  // Recurse through the item tree and replace the item
+  if (!recursiveReplace(copy.item)) {
+    // If recursiveReplace returns false, we didn't find the link ID, so this is a new element
+    copy.item = [...copy.item, newItem];
```
gigxz: I think maybe this `updateFormItem` function should go in a util function inside the `formBuilder` module. We already have soo many util functions for rendering the forms, it might be wise to keep the formBuilder utils separate.

src/modules/form/util/formUtil.ts
```diff
@@ -295,10 +295,15 @@ export const updateFormItem = (
       }
     }
 
-    throw new Error(`Link ID  ${newItem.linkId} not found`);
+    return false;
```
gigxz: I could see this potentially leading to a bug where if `initialLinkId` is wrong somehow, then the item gets duplicated rather than an error being thrown. I think we plan to disallow changing the LinkID after initial save, though, so at that point we can get rid of `initialLinkId` altogether. So I think probably a  non-issue :)

martha: Good point, I think we haven't 100% decided we are sure to make that change? Here's the ticket for it: https://github.com/open-path/Green-River/issues/6167

src/modules/formBuilder/components/FormBuilder.tsx
martha: Thanks for these suggestions! I don't think it's too scope-creeping, I'll take a stab at it.

src/modules/form/util/formUtil.ts
gigxz: I think we should. If for some reason we need to change Link IDs we can use the JSON editor. I can't think of a compelling reason why we would want to change them, either.

src/modules/formBuilder/components/itemEditor/FormEditorItemProperties.tsx
```diff
@@ -77,6 +83,29 @@ const FormEditorItemProperties: React.FC<FormEditorItemPropertiesProps> = ({
...
+    () => getItemMap(definition.definition),
+    [definition.definition]
+  );
+
+  const onLabelBlur = useCallback(
+    (event: React.ChangeEvent<HTMLInputElement | HTMLTextAreaElement>) => {
+      // When the user edits the label, update the link ID to be a human-readable 'slugified' version of the label.
+      // But, don't make an update to the link ID if it has been manually set.
+      if (!dirtyFields.linkId) {
+        let newLinkId = slugifyItemLabel(event.target.value);
```
gigxz, line 96: @martha If I clear an existing label, it clears the Link ID. We probably don't want that. Maybe only if `!dirtyFields.linkId && event.target.value`

gigxz (approved):
One small change to link id generation, looks awesome though!

src/modules/formBuilder/components/itemEditor/FormEditorItemProperties.tsx
martha: Oh true, thanks!

PR #794, opened 2024-06-04
Use react-hook-forms for form item properties

gigxz:
## Description

https://github.com/open-path/Green-River/issues/6179

* Use React Hook Form for item properties form
* Use SaveSlide at the bottom
* Allow a background click to close the panel if the form is not dirty

## Type of change


gigxz:
@martha this still needs to be cleaned up, but wanted to get it over to you asap in case you want to work off this branch for https://github.com/open-path/Green-River/issues/6083. I'd be happy to pair on incorporating that / walking through the RHF changes.

relevant docs:
* https://react-hook-form.com/docs/useform
* https://react-hook-form.com/docs/usecontroller/controller for wrapping MUI inputs
* https://react-hook-form.com/docs/usewatch watch state changes for Preview component

src/components/elements/input/TextInput.tsx
```diff
@@ -66,6 +67,7 @@ const TextInput = ({
         !props.multiline && e.key === 'Enter' && e.preventDefault()
       }
       autoComplete={formAutoCompleteOff}
+      value={value || ''} // always used as controlled input, so don't pass null or undefined
```
gigxz, line 70: this could have a wide impact, revisit

src/modules/formBuilder/components/itemEditor/FormEditorItemProperties.tsx
```diff
@@ -0,0 +1,380 @@
+import { LoadingButton } from '@mui/lab';
```
gigxz, line 1: I moved this file into a new dir, but git is showing it as a new file.

src/modules/formBuilder/components/itemEditor/FormItemEditor.tsx
```diff
@@ -0,0 +1,79 @@
...
+  item,
+  definition,
+  onSave,
+  onDiscard,
+  onClose,
+  saveLoading,
+  errorState,
+}) => {
+  // Top-level form state lives here so that we know if it's dirty when attempting to close the drawer
+  const handlers = useForm<DeepPartial<FormItem>>({
```
gigxz, line 32: I used `DeepPartial` as the form item state, so that it can support empty values on required fields, for example an empty link ID while you're changing it. This results in a few places where I needed to cast the state into a FormItem, e.g. before submitting it. This means that we need to make sure the Form state has the validations necessary for required fields (I added an example for Link ID), or else we could end up attempting to submit an invalid form, which would cause issues.

I'm curious to hear if anyone has ideas about other patterns to approach this.

src/modules/formBuilder/components/itemEditor/FormEditorItemProperties.tsx
```diff
@@ -0,0 +1,380 @@
...
+                <SelectOption
+                  label='Component Override'
+                  options={componentOverridePicklist}
+                  {...field}
+                  textInputProps={{
+                    helperText: error?.message,
+                    error: !!error,
+                    inputRef: ref,
+                  }}
+                />
```
gigxz, line 128: There's a fair amount of repeated code here. I tried to pull out the common parts (passing ref and errors to `textInputProps`) but it was finicky to get the function types right. I can try again though. Not sure if this is a common issue with RHF maybe there are patterns we can use.

src/modules/formBuilder/components/itemEditor/FormEditorItemProperties.tsx
```diff
@@ -0,0 +1,380 @@
...
+            />
+          )}
+          <Controller
+            name='linkId'
+            control={control}
+            rules={{
+              required: 'Link ID is required',
+              // TODO: add pattern validation once it is enforced
+              // https://github.com/greenriver/hmis-warehouse/pull/4424
+              // pattern: /^[a-zA-Z_$][a-zA-Z0-9_$]*$/
```
gigxz, line 139: @martha you raised a good point about this validation, I turned it off for now, and we can re-enable it once https://github.com/greenriver/hmis-warehouse/pull/4424 is merged.

src/modules/formBuilder/components/FormEditorItemPreview.tsx
```diff
@@ -1,24 +1,36 @@
...
 
-interface FormEditorItemPreviewProps {
-  item: FormItem;
-  value?: any;
-  setValue: (value: any) => void;
+interface Props {
+  control: Control<DeepPartial<FormItem>>;
 }
+export const FormEditorItemPreview: React.FC<Props> = ({ control }) => {
+  // watch whole form. probably doesn't need to watch everything (like enablewhen, etc). can pass `name` to specify fields to watch
```
martha, line 13: I think watching the whole form seems fine. If we do decide to pass `name` and restrict the list of fields it watches, then maybe it would be good to add a comment here explaining why (my understanding is that we only need to watch the fields that could impact this individual form item's display, since the preview only shows one item so it can't be used for previewing conditional logic like enableWhen)

src/modules/formBuilder/components/itemEditor/FormItemEditor.tsx
```diff
@@ -0,0 +1,79 @@
...
+          // p: 2,
+          width: '50vw',
+        },
+      }}
+    >
+      <Typography variant='cardTitle' sx={{ p: 2 }}>
+        Edit Form Item: {item.linkId}
+      </Typography>
+      <Divider />
+      {/** maybe just inline this component?  it gets basically the same props */}
```
martha, line 65: Yep I think this would be fine. I just had it in a separate component because it's so large, but I agree passing so many props around gets a bit unwieldy too

src/modules/formBuilder/components/itemEditor/FormItemEditor.tsx
martha: I think the way you have done it seems pretty good to me. There aren't that many required fields in this form, so it doesn't seem to add a burdensome amount of repetitive code to have each required field need that `rules` prop. I guess the risk is that we may forget to add such rules when adding a new required field to the form, but the risk of that seems low - the form json structure seems pretty stable to me, it seems unlikely we will add a lot more required fields?

martha (approved):
Cool, it looks like overall this went really smoothly 🥳

src/modules/formBuilder/components/FormEditorItemPreview.tsx
gigxz: Yeah, agree that watching the whole item is okay. If you pass `name` then it returns an array of the values in the same order, so that will just get super wordy since there are so many attributes. I'll leave as is and update the comment!

src/modules/formBuilder/components/itemEditor/FormItemEditor.tsx
gigxz: Okay good point. I agree it seems OK for us to be responsible for the validation there.

src/modules/formBuilder/components/itemEditor/FormItemEditor.tsx
gigxz: Okay- I won't do this now, thought, because it will cause more merge conflicts with  https://github.com/greenriver/hmis-frontend/pull/793

PR #795, opened 2024-06-05
Fix bug with LabeledCheckbox and RHF

gigxz:
## Description

React-hook-form controller passes `value` to the checkbox, so update LabeledCheckbox to consider itself checked if a value is passed. This should not affect dynamic form usages, which [already passed checked](https://github.com/greenriver/hmis-frontend/blob/b8ed4197791dcf831a168d0ba93cd39ecf18b987/src/modules/form/components/DynamicField.tsx#L187).

Repro bug:
* Open a form item in the form builder that has "warn if empty: true". The UI will show that the checkbox is not checked, even though its true.

## Type of change
- [x] Bug fix

### DIFF
--- a/src/components/elements/input/LabeledCheckbox.tsx
+++ b/src/components/elements/input/LabeledCheckbox.tsx
@@ -6,7 +6,13 @@ import {
   FormGroup,
   FormHelperText,
 } from '@mui/material';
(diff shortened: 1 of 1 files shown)


martha (approved):
Good catch, thanks Gig!

PR #797, opened 2024-06-06
Implement basic movement of form items

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6094

This PR is a first pass at updating the form structure, per Bobby's [mock](https://www.figma.com/proto/F9EHR2IZYuMAlOJ9aOG2IL/OP-%2F-HMIS-%2F-2024-%2F-Q1-%2B-Q2-(Jan-1---June-30)?page-id=783%3A4565&node-id=4065-10495&viewport=279%2C-8295%2C0.53&t=BBiohoIiafaBNHIa-1&scaling=min-zoom&starting-point-node-id=4065%3A10495).

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/modules/formBuilder/components/formBuilderUtil.ts
+++ b/src/modules/formBuilder/components/formBuilderUtil.ts
@@ -1,9 +1,9 @@
 import { cloneDeep, kebabCase } from 'lodash-es';
 import {
+  ItemType,
   Component,
   FormDefinitionJson,
   FormItem,
-  ItemType,
   Maybe,
 } from '@/types/gqlTypes';
 
--- a/src/modules/formBuilder/components/formBuilderUtil.ts
+++ b/src/modules/formBuilder/components/formBuilderUtil.ts
@@ -1,9 +1,9 @@
 import { cloneDeep, kebabCase } from 'lodash-es';
 import {
-  ItemType,
   Component,
   FormDefinitionJson,
   FormItem,
+  ItemType,
   Maybe,
 } from '@/types/gqlTypes';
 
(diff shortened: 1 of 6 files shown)


src/modules/formBuilder/components/formTree/FormTreeItem.tsx
```diff
@@ -79,10 +83,28 @@ const FormTreeItem = React.forwardRef(function FormTreeItem(
               sx={{
                 display: 'flex',
                 flexDirection: 'column',
-                mr: 4,
+                mr: 10,
```
martha, line 86: This looks pretty silly right now, but I didn't put any time into updating the styles, figuring that we will tackle all styling updates with https://github.com/open-path/Green-River/issues/6093

src/modules/formBuilder/components/formBuilderUtil.ts
```diff
@@ -90,6 +90,137 @@ export const updateFormItem = (
...
+
+export const reorderFormItems = (
+  formDefinition: FormDefinitionJson,
+  itemToMove: FormItem,
+  direction: 'up' | 'down'
+) => {
+  const copy = cloneDeep(formDefinition);
+  const toExpand: string[] = [];
+
+  function recurseFindAndMove(
```
martha, line 134: Overall I think the code I've added in this file is pretty messy and I kept wondering if there were a more graceful way to do it. We could think of this conceptually as a depth-first search to unwrap the whole tree into a list, swapping two consecutive items, and then re-wrapping it into a tree... but I wasn't convinced that taking that approach would actually be performant, so I ended up with this messy morass of special case if-statements. I tried to add lots of comments for clarity.

gigxz: Hmm, I wonder if it would be more graceful with React Hook Forms. I'm looking at  [useFieldArray](https://react-hook-form.com/docs/usefieldarray) (and maybe [setValue](https://react-hook-form.com/docs/useform/setvalue)).

RHF syntax would have the following style of "name"s for each item
```
item.0
item.0.item.0
item.0.item.1.item.0
item.0.item.1.item.1
item.0.item.2
item.1
item.2
item.2.item.0
```


useFieldArray provides functions to prepend/append/remove/swap within a given array. I used useFieldArray to add/remove items from the EnableWhen condition array in https://github.com/greenriver/hmis-frontend/pull/788..

In this context, each Group Item would have it's own FieldArray
```
  // In the context of the component that is rendering the Group that is at location `item.0.item.1`:
  const { prepend, append, remove, swap } = useFieldArray({ control, name: "item.0.item.1.item" });

  // can easily use "swap" to in onClick handlers to move items up and down within this array

  // moving the last item "down" or first item "up" feels trickier
   const movefirstItemUp = (item) => {
     insertAboveGroup(item) // insert above this group. this prop would need to be passed into this component from the parent group, I think. This would encapsulate logic for which nesting level it should land at (?) based on whether it's a Group or a question
     remove(item); // remove from this group
   }
   const moveLastItemDown = (item) => {
     // same but other direction
   }
```


This is a super rough idea, but if it works it could cut down on code complexity. Maybe we can take a stab at it in a pairing session?

src/modules/formBuilder/components/formBuilderUtil.ts
```diff
@@ -90,6 +90,137 @@ export const updateFormItem = (
...
+    if (currentIndex >= 0) {
+      // Found the item in this layer
+      if (direction === 'up') {
+        if (currentIndex > 0) {
+          // This item isn't the first item in the list
+          const prevItem = thisLayer[currentIndex - 1];
+
+          if (prevItem.type === ItemType.Group && nestingDepth < 5) {
+            toExpand.push(prevItem.linkId);
+            // If the previous item is a group, and we haven't reached max nesting depth, then move it into the group
```
gigxz, line 155: Phew I think I need an example to fully grasp this!

In the below example, assume all items are "questions" except the ones that have nested items, which are groups.
```
{
  "item": {
    { "link_id": 1 },
    { "link_id": 2 },
    { "link_id": 3,
      "item": [
        { "link_id": 4 },
        { "link_id": 5 },
        { "link_id": 6 },
      ]
    },
    { "link_id": 7,
      "item": [
        { "link_id": 8 },
        { "link_id": 9 },
        { "link_id": 10 },
      ]
    },
    { "link_id": 11}
  }
}
```
Are my below assumptions correct? This is not an order of events, each is a separate operation on its own.

* Moving item 3 "up" will swap items 2 and 3
* Moving item 4 "up" will insert item 4 after item 2
* Moving item 7 "up" will insert item 7 after item 6 (in other words, "append" to item 3)
* Moving item 8 "up" will insert item 8 after item 3 ("pops out" of group, moves down a nesting level)
* Moving item 11 "up" will insert item 11 after item 10 (in other words, "append" to item 7) (moves up a nesting level)


I think that lets you move an item to any other position, but it is laborious to move something up far. E.g. if you want to move item 11 to the top, you need to step through all the nested items in all the groups above it.

No changes requested, this is purely for validating my understanding. Thanks!

martha: Yes, I think what you've described here is what I've tried to implement, basically trying to match the prototype Bobby created here: (see link, but I'm linking his comment for background) https://github.com/open-path/Green-River/issues/5793#issuecomment-2112462489

I agree it is super laborious for items to move far. Keyboard shortcuts would be a big improvement, drag/drop would be even better.

gigxz: Okay gotcha, thanks 👍

src/modules/formBuilder/components/formBuilderUtil.ts
martha: Yeah this looks interesting. I didn't even think of using rhf. I would be glad to take a stab at it together! I want to take a closer look at your PR #788 first, maybe it will help me get a better sense of `useFieldArray`.

PR #799, opened 2024-06-07
Use RHF-controlled components in Form Item Properties Editor

gigxz:
## Description

Slight refactor to pull out shared logic into special wrapped input components that are used just for React Hook Forms, rather than wrapping them in `Controller` every time. The controlled components use the `useController` hook instead.

## Type of change


src/modules/form/components/rhf/ControlledSelect.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  // The 'value' thats stored in the form state is a string, but the value that gets
+  // passed to GenericSelect is a PickListOption
+  const valueOption = useMemo(
+    () => options.find(({ code }) => code === field.value) || null,
+    [field.value, options]
+  );
+
+  const getOptionLabel = useCallback(
+    (option: PickListOption) => getOptionLabelFromOptions(option, options),
```
martha: Nice, I wonder if `getOptionLabelFromOptions` should be moved into the `formUtil`? Although I know that file is very bulky already 🤔

src/modules/form/components/rhf/ControlledTextInput.tsx
```diff
@@ -0,0 +1,59 @@
...
+    name,
+    control,
+    shouldUnregister: true,
+    rules: {
+      required: props.required ? 'This field is required' : false,
+      ...rules,
+    },
+  });
+
+  const wrappedOnBlur = useCallback(
```
martha, line 37: Am I understanding this correctly?:
- the parent can pass in a custom `onBlur` callback to this component
- the RHF also needs to know when an `onBlur` event happens, so we call its onBlur callback on line 40
- a similar `wrappedOnBlur` isn't needed in the two other controlled components added in this PR, Select and Checkbox, because we don't currently have any custom `onBlur`s on those elements in our form logic (since, right now, this is used only for the label/linkId). A `wrappedOnBlur` could be added to them in the future if we did need such functionality

martha (approved):
Lovely 🥳

src/modules/form/components/rhf/ControlledSelect.tsx
gigxz: Definitely, updated with https://github.com/greenriver/hmis-frontend/pull/799/commits/2ab12ed7db65d9aa33073b8dec842c4c22c82701

src/modules/form/components/rhf/ControlledTextInput.tsx
gigxz: Yes that's correct.  I made a few changes with https://github.com/greenriver/hmis-frontend/pull/799/commits/ed1da8ea820269e2696298e6c586734cae32f464 to clarify this:
1)`ControlledCheckbox` and `ControlledSelect` should not _accept_ onBlur props, because they are ignored (at least right now)
2) RHF onBlur wasn't being passed to ControlledSelect, I updated that

PR #800, opened 2024-06-08
6064-sentry-grouping

ttoomey:
## Description

[GH issue](https://github.com/open-path/Green-River/issues/6064)

* upgrade sentry 7->8, add reactRouter integration
* add apollo-link-sentry to improve sentry instrumentation for graphql errors 
* adjust apollo error link handling to improve grouping and reporting
* upgrade @sentry/vite-plugin, configure to include component names
* code organization: pull sentry config out of index.tsx, extract errorLink from apollo link

### Testing
for testing, I set `PUBLIC_SENTRY_DSN` to the development sentry project

### Testing case auth failure
1) login
1) log your user out on the server from the console `Hmis::User.where(id:1).update_all(unique_session_id:'')`
1) on api request: error should not be reported to sentry, user prompted to sign in
### Testing server error
1) login and visit /projects
1) adjust graphqlController#execute to return `return render status: 500, json: { errors: [ { message: 'test' }, ], data: {}, }`
1) click on a project
1) 1 "graphql error" should be logged to sentry. Message should contain informative breadcrumbs

## Type of change
- [x] Dependency update


vite.config.js
```diff
@@ -55,15 +55,22 @@ export default defineConfig(({ command, mode }) => {
...
               authToken: env.SENTRY_AUTH_TOKEN,
-              release: env.PUBLIC_GIT_COMMIT_HASH,
-              setCommits: {
-                repo: 'greenriver/hmis-frontend',
-                commit: env.FULL_GITHASH,
+              release: {
+                uploadLegacySourcemaps: {
+                  paths: ['dist/assets/'],
+                  urlPrefix: '~/assets/',
+                },
```
ttoomey, line 65: tried to preserve this config from the previous version of the vite sentry plugin (0.6)

src/providers/apolloErrorLink.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  try {
+    result = JSON.stringify(obj);
+  } catch (error) {
+    result = undefined;
+  }
+  return result;
+}
+
+const apolloErrorLink = onError(
```
ttoomey, line 26: extracted from apolloClient

src/sentryInit.ts
```diff
@@ -0,0 +1,46 @@
...
+} from 'react-router-dom';
+
+const dsn = import.meta.env.PUBLIC_SENTRY_DSN;
+const hostname = window.location.hostname;
+const environment = /\b(qa|training|staging)\b/.test(hostname)
+  ? 'staging'
+  : 'production';
+
+if (dsn) {
+  Sentry.init({
```
ttoomey, line 19: extracted from index.ts

src/providers/apolloClient.tsx
```diff
@@ -7,18 +7,13 @@ import {
 } from '@apollo/client';
 import { BatchHttpLink } from '@apollo/client/link/batch-http';
 import { setContext } from '@apollo/client/link/context';
-import { onError } from '@apollo/client/link/error';
-import * as Sentry from '@sentry/react';
+import { SentryLink } from 'apollo-link-sentry';
```
gigxz, line 10: @ttoomey this branch is up on QA right now, as you can see there is an issue with the syntax (require is not defined).

This package has [both cjs and esm exports](https://github.com/DiederikvandenB/apollo-link-sentry/blob/bcc93a12b2d3b63140f01e8aef7fb94100132a0e/package.json#L8-L11) and it seems like vite is pulling the cjs one (probably because its "main"). Not sure if there is a neater way to specify this, but it seems to work to replace this (and the other usage) with importing from `'apollo-link-sentry/lib-esm';`.

gigxz: I'm not sure if that will fix the issue, but might be worth a shot. I'm deploying a different commit to qa so its accessible again, but feel free to deploy this branch again if you need!

gigxz: That didn't fix it. It's weirdly trying to use "require" to load the sentryInit file. I wonder if this has something to do with the fact that the order of imports changed now that we call sentryInit first thing in the index file.

It's really weird that I can't reproduce this with `yarn build` -- when I run that locally, the index.js file does not use required. But the deployment leads me to believe its generating differently [when building in the container](https://github.com/greenriver/hmis-frontend/blob/a4a6045eac20f818bade8c3dcd717314439f42c1/docker/Dockerfile#L32)(?)

Screenshot 2024-06-10 at 2 20 03 PM

ttoomey:
paper trail on `require not defined` problem:
The issue seems to be caused by enabling `reactComponentAnnotation` in the vite sentry plugin. Not sure what's going on with that

gigxz:
Good find @ttoomey! Ok, lets just leave it out then. Component name annotation doesn't seem too important, we can get by with the html selectors.

gigxz (approved):
💯

PR #801, opened 2024-06-10
Disallow changing linkid on persisted item, add constraints

gigxz:
## Description

* Disallow changing LinkID on item once it has been persisted
* Add regex constraints to LinkID when adding new item
* Present as "Question ID" to end user

https://github.com/open-path/Green-River/issues/6167

## Type of change

### DIFF
--- a/src/modules/formBuilder/components/itemEditor/FormItemEditor.tsx
+++ b/src/modules/formBuilder/components/itemEditor/FormItemEditor.tsx
@@ -1,5 +1,6 @@
 import { Divider, Drawer, Typography } from '@mui/material';
-import React from 'react';
+import { startCase } from 'lodash-es';
+import React, { useMemo } from 'react';
 import { DeepPartial, useForm } from 'react-hook-form';
 import { ErrorState } from '@/modules/errors/util';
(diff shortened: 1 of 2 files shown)


src/modules/formBuilder/components/itemEditor/FormItemEditor.tsx
```diff
@@ -39,6 +40,9 @@ const FormItemEditor: React.FC<FormItemEditorProps> = ({
     // errors: errorState?.errors,
   });
 
+  // If typename is not present, then this is a new item that hasn't been saved yet.
+  const isNewItem = useMemo(() => !item.__typename, [item]);
```
martha, line 44: We could also make this a piece of state in `FormBuilder` that's set when a form builder palette button is pressed... might be a bit neater, but I could go either way

martha (approved):
Wonderful :)

src/modules/formBuilder/components/itemEditor/FormItemEditor.tsx
gigxz: I could go either way too. I opted for managing fewer pieces of state for simplicity, even though it feels a bit un-neat to rely on the presence of the typename, it does signify that the record was resolved from the backend so it should be persisted. We don't resolve the record on validation.

PR #802, opened 2024-06-11
Form property editor styling and sections

gigxz:
## Description

* Move properties into sections, use header tags
* Move Required/WarnifEmpty into custom radio
* Use overline typography
* Scroll up to errors, fix issue with errors persisting too long

## Type of change


src/modules/formBuilder/components/itemEditor/RequiredOptionalRadio.tsx
```diff
@@ -0,0 +1,113 @@
...
+    [readOnlyField, requiredField, warnIfEmptyField]
+  );
+
+  return (
+    <RadioGroupInput
+      options={options}
+      label='Response Validation'
+      value={options.find((o) => o.code === radioValue)}
+      onChange={handleChange}
+      labelSx={{ typography: 'body2', color: 'text.secondary', pt: 1 }}
```
gigxz, line 95: We should adjust default styles for this component, it looks different from others. Want to run by design first

src/modules/formBuilder/components/FormBuilderPage.tsx
```diff
@@ -44,6 +44,9 @@ const FormBuilderPage = () => {
 
   const [errorState, setErrorState] = useState<ErrorState>(emptyErrorState);
 
+  useEffect(() => {
+    setErrorState(emptyErrorState);
+  }, [selectedItem, setErrorState]);
```
gigxz, line 49: Fixes a bug where if there was an error in the item editor and you close the editor, the error still appears on the main page.

martha: Thanks 🙏  this has been bugging me too!

src/modules/admin/components/forms/FormDefinitionDetailPage.tsx
```diff
@@ -30,21 +30,13 @@ const FormDefinitionDetailPage = () => {
...
-          Selected Form
-        </Typography>
-        <Stack direction='row' gap={1}>
-          <Typography variant='h3'>
-            {formIdentifier.displayVersion.title}
-          </Typography>
-        </Stack>
-      </Stack>
       <Stack gap={2}>
+        <Typography variant='h3'>
```
martha: just noting here, for interest, though I don't think it should be fixed on this PR, that one of the recommendations coming out of the a11y audit for this app is that we should divorce heading hierarchy from heading appearance. For good screen reader nav, ideally, the first heading on a page is a h1 -- even if it's styled in a way that appears like an h3 in other parts of the app.

It's a "best practice" as opposed to a WCAG standard, though, so perhaps that makes it lower priority to fix!

src/modules/formBuilder/components/itemEditor/RequiredOptionalRadio.tsx
martha: 👍 also I wonder if it would be nicer to display the helper text for each option below that option, rather than one big block below the whole item?

martha (approved):
Great :)

src/modules/formBuilder/components/itemEditor/RequiredOptionalRadio.tsx
gigxz: Maybe, I pulled this approach from the designs. I think it would be possible to have extended labels on each radio element as another alternative. (But each long text would become the label for each element, not helper text).

src/modules/admin/components/forms/FormDefinitionDetailPage.tsx
gigxz: Well thats an easy fix, updated with https://github.com/greenriver/hmis-frontend/pull/802/commits/af570354984ab4da66eb62ef9214663930be4683. I'm not sure if it's correct to have the overline text span inside the h1 tag.

PR #804, opened 2024-06-11
Add draft, published, retired workflow for forms

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/5858
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4449

This PR:
- updates the schema corresponding to the warehouse updates
- updates the FormDefinitionDetailPage's two Edit buttons -- the Json editor and the FormBuilder editor -- to create a new Draft if none exists. Non-Draft forms should not be editable
- updates the FormPreview page to include a Publish button
- updates the FormBuilder's header publish button to work, though more changes are to come in https://github.com/open-path/Green-River/issues/6091

I have a few open questions which I'll post inline -- I can also spend more time researching/figuring them out, but wanted to put up a review at the same time to confirm the overall approach.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/admin/components/forms/FormPreview.tsx
```diff
@@ -102,23 +109,58 @@ const FormPreview = () => {
...
+      },
+      onCompleted: () => {
+        cache.evict({ id: `FormIdentifier:${formDefinition?.identifier}` });
+      },
+    });
+  }, [publishForm, formDefinition]);
+
   if (loading && !formDefinition) return <Loading />;
   if (error) throw error;
+  if (publishError) throw publishError;
```
martha: I'd like to confirm my judgment call that there's no need to use `ErrorState` and `apolloError` here, because any errors here would be bugs (as opposed to user input validation), does that sound right?

src/modules/admin/components/forms/EditFormButton.tsx
```diff
@@ -0,0 +1,82 @@
...
+        generateSafePath(route, {
+          identifier: formIdentifier.identifier,
+          formId: draftId,
+        })
+      );
+    },
+    [formIdentifier, editorRoute, navigate]
+  );
+
+  const [createDraftForm, { loading, error }] = useCreateDraftFormMutation({
```
martha: I think we may need to move this up into the parent component so that each button can be blocked while the other is loading. We don't want the user to be able to click "Edit Form" and then "Json Editor" in rapid succession, leading to an error message since two drafts can't both exist at the same time.

However, moving it up into the parent proved a little difficult because of the `goToEditor` callback, so I put it aside for now; will return with fresh eyes tomorrow

src/modules/admin/components/forms/EditFormButton.tsx
```diff
@@ -0,0 +1,82 @@
...
+    variables: { identifier: formIdentifier.identifier },
+    onCompleted: (data) => {
+      if (data.createDraftForm?.formIdentifier?.draftVersion?.id) {
+        cache.evict({ id: `FormIdentifier:${formIdentifier.identifier}` });
```
martha: Also I'd like to hack a bit more on trying to use `cache.modify` instead of `cache.evict` here in order to avoid an extra round-trip for data we already have.

src/modules/admin/components/forms/EditFormButton.tsx
```diff
@@ -0,0 +1,82 @@
...
+  }, [formIdentifier.draftVersion, goToEditor, createDraftForm]);
+
+  if (error) throw error;
+
+  return (
+    <LoadingButton
+      onClick={() => onClick()}
+      startIcon={Icon ? <Icon /> : <DashboardCustomizeIcon />}
+      loading={loading}
+      role='link'
```
martha, line 78: I'd like to test this for accessibility - I'm giving it a role of `link` since it is linking out to another page, but I'm not sure if this works well in screen readers without a `to` attribute

src/modules/admin/components/forms/FormPreview.tsx
martha: resolved in discussion on backend PR

src/modules/admin/components/forms/EditFormButton.tsx
martha: The new [change from the backend PR](https://github.com/greenriver/hmis-warehouse/pull/4449#discussion_r1636773496) where we just return the existing draft instead of throwing an error kinda solves this problem, I think. Also, since the JSON editor is only accessible to staff level admins, I'd propose to not put any more time into it now.

src/modules/admin/components/forms/EditFormButton.tsx
martha: edit: this seems to work OK; from voiceover, the link looks the same as other links on the page

gigxz: nice 👍

src/modules/admin/components/forms/EditFormButton.tsx
```diff
@@ -0,0 +1,85 @@
...
+}
+
+interface EditFormButtonProps {
+  formIdentifier: FormIdentifierDetailsFragment;
+  text?: string;
+  Icon?: React.ComponentType;
+  editorRoute?: FormEditorType;
+}
+
+const EditFormButton: React.FC<EditFormButtonProps> = ({
```
gigxz, line 24: Nice refactor, looks good!

gigxz (approved):
👍

PR #805, opened 2024-06-12
Prevent error on empty item.initial list

martha:
## Description

https://greenriver.slack.com/archives/C061SAW3LFJ/p1718196001472859

The bug appears when `item.initial` list is present but empty.

I didn't find any forms in QA where this was actually the case, barring the one I created copied from a [local definition](https://qa-hmis.openpath.host/admin/forms/martha_test) that somehow got into this state. But, I think it's cleaner for our code to handle this case given that I think it's legal per the [json schema](https://github.com/greenriver/hmis-warehouse/blob/6702d55a7a29cba68b29bbf0342cadd83de7754a/drivers/hmis_external_apis/public/schemas/form_definition.json#L251).

I also updated the initial `enableBehavior` on newly added form items from the Palette to be `Any` instead of `All` - that initial `Any` was causing the newly added items to not appear in the preview, since none of the zero conditions were true 😭 . (The reason to include it at all here is just to make typescript happy, but perhaps it should instead be made optional and removed here?)

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/modules/formBuilder/components/FormBuilder.tsx
+++ b/src/modules/formBuilder/components/FormBuilder.tsx
@@ -207,7 +207,7 @@ const FormBuilder: React.FC<FormBuilderProps> = ({
               repeats: false,
               prefill: false,
               disabledDisplay: DisabledDisplay.Hidden,
-              enableBehavior: EnableBehavior.Any,
+              enableBehavior: EnableBehavior.All,
             };
 
             setSelectedItem(newItem);
--- a/src/modules/form/util/formUtil.ts
+++ b/src/modules/form/util/formUtil.ts
@@ -777,7 +777,7 @@ export const getInitialValues = (
       if (Array.isArray(item.item)) {
         recursiveFillInValues(item.item, values);
       }
-      if (!item.initial || item.initial.length === 0) {
+      if (!item.initial) {
         // Make sure all linked fields are present in values to begin
         if (item.mapping && behavior !== InitialBehavior.Overwrite)
           values[item.linkId] = null;
--- a/src/modules/form/util/formUtil.ts
+++ b/src/modules/form/util/formUtil.ts
@@ -777,7 +777,7 @@ export const getInitialValues = (
       if (Array.isArray(item.item)) {
         recursiveFillInValues(item.item, values);
       }
-      if (!item.initial) {
+      if (!item.initial || item.initial.length === 0) {
         // Make sure all linked fields are present in values to begin
         if (item.mapping && behavior !== InitialBehavior.Overwrite)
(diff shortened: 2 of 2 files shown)


src/modules/formBuilder/components/FormBuilder.tsx
```diff
@@ -207,7 +207,7 @@ const FormBuilder: React.FC<FormBuilderProps> = ({
               repeats: false,
               prefill: false,
               disabledDisplay: DisabledDisplay.Hidden,
-              enableBehavior: EnableBehavior.Any,
+              enableBehavior: EnableBehavior.All,
```
gigxz, line 210: I think it would be less brittle to update the evaluation logic such that `enableBehavior: 'ANY', enableWhen: []` results in the item always being enabled.

That can be achieved by changing the below line to `if (!item.enableWhen || item.enableWhen.length === 0) return true;`
https://github.com/greenriver/hmis-frontend/blob/5b49a16f1a2fe729030858dd113ca1305c476426/src/modules/form/util/formUtil.ts#L433



There are some other places where we use `!tem.enableWhen` to mean that there are no rules, like [here](https://github.com/greenriver/hmis-frontend/blob/c5dcb90dc0e2b53f85cce42de0700b4b039cdcf8/src/modules/form/util/formUtil.ts#L139), so that concerns me... rather than  (or in addition to) updating these to always check `|| item.enableWhen.length === 0`, can we update the save function such that it strips out empty arrays? I think we discussed that was needed for the form "updateability" ticket, so maybe its already in the works. Since we have some evaluation code expecting empty arrays not to exist, it feels safest to strip them out.

martha: Good point! I was just working on this: https://github.com/greenriver/hmis-warehouse/pull/4450

martha: I'd propose to
- keep this change on FormBuilder.tsx:210; it doesn't really matter which one is the case for the initial default, I think I picked one at random?
- not make the change on formUtil.ts:433, for consistency, since it sounds like that will have to be made in several places and we're not sure of all of them, so it's brittle?
- remove the change to formUtil.ts:780 (above), also for consistency?

what do you think?

gigxz: That sounds good to me. Thanks! There is a risk now that our local/QA envirnoments might end up with some forms containing empty arrays, and we will need to re-submit those in the form builder interface to clean them up. I think that's OK (eg no need for a cleanup routine) since we haven't used the form builder in other environments yet.

PR #807, opened 2024-06-12
Standardize use body2 for labels

gigxz:
## Description

Fix style discrepancies for certain elements (radio, checkbox notably) when used outside of DynamicForm context.

QA:
* form item property form looks consistent now
* dynamic assessments look the same

## Type of change
- [x] Bug fix


src/config/theme.ts
```diff
@@ -245,6 +245,21 @@ const createThemeOptions = (theme: Theme) => ({
         },
       },
     },
+    MuiFormLabel: {
+      styleOverrides: {
+        root: theme.unstable_sx({
+          typography: 'body2',
+          fontWeight: 600,
```
gigxz, line 252: Fields rendered within a DynamicForm get special labels ([RequiredLabel] component(https://github.com/greenriver/hmis-frontend/blob/201900cd61d2e961a957c4a494c8110f950bcd03/src/modules/form/components/RequiredLabel.tsx#L17)). Here I update the theme to match that style (body2 bold), so that inputs rendered _outside_ of dynamic form match the same styles as dynamic forms.

martha (approved):
Looks great. I still think the radio button overall styling could use another pass from design, maybe? The option labels are really far away from each other, as you noted, and also the label for the whole input is really far away from those options (compared to e.g. the dropdown components).

gigxz:
Agreed! I fixed that up with the latest commit:

Before:
Screenshot 2024-06-13 at 3 24 56 PM



After:

Screenshot 2024-06-13 at 3 33 10 PM

PR #821, opened 2024-06-24
Enable form item deletion

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6165

The solution I've landed on is to enable the `useReorder` hook (here renamed to `useUpdateFormStructure`) to return meaningless values when an item has been deleted, instead of throwing up an error. 

I still haven't figured out why the Mui `RichTree` keeps rendering `FormTreeLabel`s for an item for 1 more render after the item has been deleted from the form definition structure. After pairing with @ttoomey, our thought was that it probably isn't worth sinking more time into. @gigxz happy to defer to you on that though.

Possibly related, I'll record here for posterity that it's still the case that if you set a breakpoint in the browser debugger at line 42 of `useUpdateFormStructure`, you can see the following warning (not visible without breakpoint):

> Warning: Cannot update a component (`ForwardRef(RichTreeView2)`) while rendering a different component (`FormTreeLabel`). To locate the bad setState() call inside `FormTreeLabel`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/modules/formBuilder/components/formTree/FormTree.tsx
+++ b/src/modules/formBuilder/components/formTree/FormTree.tsx
@@ -1,7 +1,11 @@
 import { RichTreeView } from '@mui/x-tree-view/RichTreeView';
 import React, { useMemo } from 'react';
 import { useFormContext, useWatch } from 'react-hook-form';
-import { displayLabelForItem, getRhfPathMap } from '../../formBuilderUtil';
+import {
+  displayLabelForItem,
+  getAncestorLinkIdMap,
+  getRhfPathMap,
+} from '../../formBuilderUtil';
 import Loading from '@/components/elements/Loading';
 import { getItemMap } from '@/modules/form/util/formUtil';
 import { FormTreeContext } from '@/modules/formBuilder/components/formTree/FormTreeContext';
@@ -28,6 +32,11 @@ const FormTree: React.FC<FormTreeProps> = ({ onEditClick }) => {
 
   const rhfPathMap = useMemo(() => getRhfPathMap(values.item), [values]);
 
+  const ancestorLinkIdMap = useMemo(
+    () => getAncestorLinkIdMap(values.item),
+    [values]
+  );
+
   const [expandedItems, setExpandedItems] = React.useState<string[]>([]);
 
   const handleExpandedItemsChange = (
@@ -46,14 +55,15 @@ const FormTree: React.FC<FormTreeProps> = ({ onEditClick }) => {
         setExpandedItems((prev) => [...prev, itemId]),
       collapseItem: (itemId: string) =>
         setExpandedItems((prev) => prev.filter((id) => id !== itemId)),
-      // `itemMap` and `rhfPathMap` are both regenerated on every change to the form, due to the `useWatch`.
+      // These maps are regenerated on every change to the form, due to the `useWatch`.
       // This shouldn't cause a perf issue right now, because this context is only consumed by the FormTreeLabel,
       // which is already calling useWatch({ control }) anyway (inside the useUpdateFormStructure hook).
       // But it could cause performance issues if other components consume this context.
       itemMap,
       rhfPathMap,
+      ancestorLinkIdMap,
     }),
-    [onEditClick, itemMap, rhfPathMap]
+    [onEditClick, itemMap, rhfPathMap, ancestorLinkIdMap]
   );
 
   if (!values.item) return <Loading />;
--- a/src/modules/formBuilder/components/formTree/FormTree.tsx
+++ b/src/modules/formBuilder/components/formTree/FormTree.tsx
@@ -44,6 +44,10 @@ const FormTree: React.FC<FormTreeProps> = ({ onEditClick }) => {
         setExpandedItems((prev) => [...prev, itemId]),
       collapseItem: (itemId: string) =>
         setExpandedItems((prev) => prev.filter((id) => id !== itemId)),
+      // `itemMap` and `rhfPathMap` are both regenerated on every change to the form, due to the `useWatch`.
+      // This shouldn't cause a perf issue right now, because this context is only consumed by the FormTreeLabel,
+      // which is already calling useWatch({ control }) anyway (inside the useUpdateFormStructure hook).
+      // But it could cause performance issues if other components consume this context.
       itemMap,
       rhfPathMap,
     }),
--- a/src/modules/formBuilder/components/formTree/FormTree.tsx
+++ b/src/modules/formBuilder/components/formTree/FormTree.tsx
@@ -37,6 +37,8 @@ const FormTree: React.FC<FormTreeProps> = ({ onEditClick }) => {
     setExpandedItems(itemIds);
   };
 
+  // Passing these values via FormTreeContext here enables us to easily pass props down
+  // to the FormTreeItem and FormTreeLabel components, avoiding slotProps and its typescript issues
   const context = React.useMemo(
     () => ({
       openFormItemEditor: (item: FormItem) => onEditClick(item),
@@ -51,7 +53,7 @@ const FormTree: React.FC<FormTreeProps> = ({ onEditClick }) => {
       itemMap,
       rhfPathMap,
     }),
-    [onEditClick, itemMap]
+    [onEditClick, itemMap, rhfPathMap]
   );
 
   if (!values.item) return <Loading />;
--- a/src/modules/formBuilder/components/formTree/FormTree.tsx
+++ b/src/modules/formBuilder/components/formTree/FormTree.tsx
@@ -1,7 +1,7 @@
 import { RichTreeView } from '@mui/x-tree-view/RichTreeView';
 import React, { useMemo } from 'react';
 import { useFormContext, useWatch } from 'react-hook-form';
-import { displayLabelForItem } from '../../formBuilderUtil';
+import { displayLabelForItem, getRhfPathMap } from '../../formBuilderUtil';
 import Loading from '@/components/elements/Loading';
 import { getItemMap } from '@/modules/form/util/formUtil';
 import { FormTreeContext } from '@/modules/formBuilder/components/formTree/FormTreeContext';
@@ -26,6 +26,8 @@ const FormTree: React.FC<FormTreeProps> = ({ onEditClick }) => {
     [values]
   );
 
+  const rhfPathMap = useMemo(() => getRhfPathMap(values.item), [values]);
+
   const [expandedItems, setExpandedItems] = React.useState<string[]>([]);
 
   const handleExpandedItemsChange = (
@@ -43,6 +45,7 @@ const FormTree: React.FC<FormTreeProps> = ({ onEditClick }) => {
       collapseItem: (itemId: string) =>
         setExpandedItems((prev) => prev.filter((id) => id !== itemId)),
       itemMap,
+      rhfPathMap,
     }),
     [onEditClick, itemMap]
   );
--- a/src/modules/formBuilder/components/formTree/FormTree.tsx
+++ b/src/modules/formBuilder/components/formTree/FormTree.tsx
@@ -13,9 +13,12 @@ interface FormTreeProps {
 }
 const FormTree: React.FC<FormTreeProps> = ({ onEditClick }) => {
   const { control } = useFormContext();
-  const items = useWatch({ name: 'item', control });
+  const values = useWatch({ control });
 
-  const definitionForTree = useMemo(() => getItemsForTree(items), [items]);
+  const definitionForTree = useMemo(
+    () => getItemsForTree(values.item),
+    [values.item]
+  );
 
   const [expandedItems, setExpandedItems] = React.useState<string[]>([]);
 
@@ -37,7 +40,7 @@ const FormTree: React.FC<FormTreeProps> = ({ onEditClick }) => {
     [onEditClick]
   );
 
-  if (!items) return <Loading />;
+  if (!values.item) return <Loading />;
 
   return (
     <FormTreeContext.Provider value={context}>
--- a/src/modules/formBuilder/components/formTree/FormTree.tsx
+++ b/src/modules/formBuilder/components/formTree/FormTree.tsx
@@ -3,10 +3,11 @@ import React, { useMemo } from 'react';
 import { useFormContext, useWatch } from 'react-hook-form';
 import { displayLabelForItem } from '../../formBuilderUtil';
 import Loading from '@/components/elements/Loading';
+import { getItemMap } from '@/modules/form/util/formUtil';
 import { FormTreeContext } from '@/modules/formBuilder/components/formTree/FormTreeContext';
 import FormTreeItem from '@/modules/formBuilder/components/formTree/FormTreeItem';
 import { getItemsForTree } from '@/modules/formBuilder/components/formTree/formTreeUtil';
-import { FormItem } from '@/types/gqlTypes';
+import { FormDefinitionJson, FormItem } from '@/types/gqlTypes';
 
 interface FormTreeProps {
   onEditClick: (item: FormItem) => void;
@@ -18,6 +19,11 @@ const FormTree: React.FC<FormTreeProps> = ({ onEditClick }) => {
   const definitionForTree = useMemo(
     () => getItemsForTree(values.item),
     [values.item]
+  );
+
+  const itemMap = useMemo(
+    () => getItemMap(values as FormDefinitionJson),
+    [values]
   );
 
   const [expandedItems, setExpandedItems] = React.useState<string[]>([]);
@@ -36,8 +42,9 @@ const FormTree: React.FC<FormTreeProps> = ({ onEditClick }) => {
         setExpandedItems((prev) => [...prev, itemId]),
       collapseItem: (itemId: string) =>
         setExpandedItems((prev) => prev.filter((id) => id !== itemId)),
+      itemMap,
     }),
-    [onEditClick]
+    [onEditClick, itemMap]
   );
 
   if (!values.item) return <Loading />;
--- a/src/modules/formBuilder/components/formTree/FormTreeLabel.tsx
+++ b/src/modules/formBuilder/components/formTree/FormTreeLabel.tsx
@@ -50,7 +50,7 @@ const FormTreeLabel: React.FC<FormTreeLabelProps> = ({
     disabled,
   });
 
-  const { itemMap } = useContext(FormTreeContext);
+  const { itemMap, rhfPathMap } = useContext(FormTreeContext);
   const item = useMemo(() => itemMap[itemId], [itemMap, itemId]);
 
   const displayAttrs = useMemo(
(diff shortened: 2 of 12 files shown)


src/modules/formBuilder/components/formTree/FormTreeLabel.tsx
```diff
@@ -38,32 +39,35 @@ const FormTreeLabel: React.FC<FormTreeLabelProps> = ({
...
     children,
     label,
     disabled,
   });
 
-  // we could get this from the form instead of from the tree api
-  const treeItem = publicAPI.getItem(itemId);
-  const item = useMemo(() => getItemFromTree(treeItem), [treeItem]);
+  const itemMap = useMemo(
+    () => getItemMap(values as FormDefinitionJson),
```
martha: Should provide itemMap in `FormTreeContext`?

src/modules/formBuilder/components/formTree/FormTreeLabel.tsx
```diff
@@ -91,6 +91,24 @@ const FormTreeLabel: React.FC<FormTreeLabelProps> = ({
...
     >
+      <ConfirmationDialog
+        open={!!errors}
+        title='Cannot delete'
+        onConfirm={() => setErrors(undefined)}
+        onCancel={() => setErrors(undefined)}
+        loading={false}
+      >
+        Cannot delete "{labelProps.children}," because one or more other
+        questions depend on it.
```
gigxz: Did this come out of Figma or is it off the cuff?

I would add some text like: "Please remove all conditional dependencies from dependent items and try again. "

src/modules/formBuilder/formBuilderUtil.ts
```diff
@@ -214,3 +214,79 @@ export const insertItemToDefinition = ({
...
+
+/**
+ * Validates that the attempted deletion action is allowed. It is allowed if the item
+ * the user is trying to delete doesn't have any other items that refer to it.
+ * If dependents exist, then the operation is not allowed and the function returns a user-facing error string.
+ *
+ * @param toDelete the linkId of the item the user is trying to delete
+ * @param definition the current form definition
+ */
+export const validateDeletion = ({
```
gigxz: UI suggestion for modal:
```
Items with autofill condition(s) that depend on this item:
* foo
* bar

Items with visibility condition(s) that depends on this item:
* foo

Items with min/max bound(s) that depends on this item:
* foo
```

as opposed to:


Screenshot 2024-06-26 at 2 03 55 PM


---

Thoughts on that?

That could be accomplished by making this function return the actual dependencies (or their labels) rather than strings to be rendered, so we have some more flexibility in adjusting the alert interface (i.e. adding links to those items later). It also could be helpful if we ever want to add links to dependent items into the property editor drawer.

```
const { autofillDependents, enableWhenDependents, boundDependents } = getDependentItems(itemMap, linkId)
```

src/modules/formBuilder/formBuilderUtil.ts
```diff
@@ -214,3 +214,79 @@ export const insertItemToDefinition = ({
...
+  definition: FormDefinitionJson;
+}): { success: boolean; errors?: Set<string> } => {
+  // Mapping of linkId of the dependent question => an array of user-facing strings
+  // representing where the problematic dependency is located on that item, e.g. autofill or enableWhen
+  const errors = new Set<string>();
+
+  function checkDependence(
+    toDelete: string,
+    toCheck: FormItem,
+    parentPath: string
```
gigxz: Tiny thing, but I would rename this. (I know I have a separate comment that might make this irrelevant). We use "parentPath"/"itemPath" elsewhere in the code to describe paths like `item.0.item.5`, which I think is pretty intuitive. Here it's describing `"label > label > label"`, I think, which feels less intuitive. Maybe call this `parentLabelPrefix` or `parentLabels` (as an array) or simply `labelPrefix`.

In general imo this helps with code readability if we have consistent meanings for words like "path" and "linkId" etc. Not always possible to be 100% consistent but we can try :)

src/modules/formBuilder/formBuilderUtil.ts
```diff
@@ -214,3 +214,79 @@ export const insertItemToDefinition = ({
...
+    toCheck.enableWhen?.forEach((enableWhen) => {
+      if (
+        enableWhen.question === toDelete ||
+        enableWhen.compareQuestion === toDelete
+      ) {
+        errors.add(`${prefix}: <b>conditional visibility</b>`);
+      }
+    });
+
+    toCheck.autofillValues?.forEach((autofillValue) => {
```
gigxz: We're duplicating logic with `buildAutofillDependencyMap` helper here -- sorry I didn't flag that in the spec, I didn't think about that! `buildEnabledDependencyMap` exists as well. 

`buildAutofillDependencyMap` also checks for dependencies in the `formula`, which this doesn't. It might be best to use those helper methods, so we don't have to worry about them falling out of sync.

Hopefully that is an easy change, the only issue I see is that it doesn't provide this "parent path prefix". We would probably need some an additional helper to find the label prefix for a given link id. I wonder if that can be written in a way that is useful if we need it elsewhere, for example:
```
parentLinkIds = findParentLinkIds(itemMap, linkId)   // <== api that would be usable for other needs
prefixedLabel = parentLinkIds.map { |x| itemMap[x].text }.join(" > ") + itemMap[linkId].text
```

src/modules/formBuilder/components/formTree/useUpdateFormStructure.tsx
```diff
@@ -32,24 +39,20 @@ export default function useReorderItem(
...
   const parentItemId = parentArrayPath.replace(/\.item$/, '');
   const { parentPath: grandParentArrayPath, index: parentIndex } =
     getPathContext(parentItemId);
 
   // RHF swap is used for swapping items within an array
-  const { swap } = useFieldArray({ control, name: parentArrayPath });
+  const { swap, remove } = useFieldArray({ control, name: parentArrayPath });
 
   const thisLayer = get(values, parentArrayPath);
-  if (!thisLayer)
```
gigxz, line 46: Can you add a comment indicating that `thisLayer` could be undefined if an item was just deleted

src/modules/formBuilder/components/formTree/FormTree.tsx
```diff
@@ -33,11 +42,12 @@ const FormTree: React.FC<FormTreeProps> = ({ onEditClick }) => {
         setExpandedItems((prev) => [...prev, itemId]),
       collapseItem: (itemId: string) =>
         setExpandedItems((prev) => prev.filter((id) => id !== itemId)),
+      itemMap,
```
gigxz, line 62: This feels like a change that could have a negative performance impact, since we're putting something in the context that is being updated on every change to the form (due to the watch).

Looking at where the context is consumed, though, it's only consumed in the FormTreeLabel which is already calling `useWatch({ control })` anyway (inside the useUpdateFormStructure hook). So I don't _think_ this has any perf impact _now_, but it does make it so that this context very often causes re-renders, which could cause issues if we ever have other components consuming from it.

Just wanted to flag this as a code smell but i'm happy to leave it as-is and come back to it if there are perf problems.

src/modules/formBuilder/components/formTree/FormTreeLabel.tsx
```diff
@@ -38,32 +39,31 @@ const FormTreeLabel: React.FC<FormTreeLabelProps> = ({
...
     itemId,
     children,
     label,
     disabled,
   });
 
-  // we could get this from the form instead of from the tree api
-  const treeItem = publicAPI.getItem(itemId);
-  const item = useMemo(() => getItemFromTree(treeItem), [treeItem]);
+  const { itemMap } = useContext(FormTreeContext);
```
gigxz: This component already consumes this context on line 41

src/modules/formBuilder/components/formTree/useUpdateFormStructure.tsx
```diff
@@ -12,10 +18,11 @@ import {
   getPathContext,
   insertItemToDefinition,
   removeItemFromDefinition,
+  validateDeletion,
 } from '@/modules/formBuilder/formBuilderUtil';
 import { FormDefinitionJson, FormItem, ItemType } from '@/types/gqlTypes';
 
-export default function useReorderItem(
+export default function useUpdateFormStructure(
```
gigxz, line 19: trying to comment on line 34 `const itemIdMap = useMemo(() => getItemIdMap(values.item), [values]);` -- does this need to be re-calculated here? It looks like you added it to the context, so you could consume it from that, or better yet just pass it as an argument to this hook: `useUpdateFormStructure(control, itemMap, itemId)`

gigxz (commented):
looking good, let me know if you want to discuss any comments. lets retarget to 122. thanks!

src/modules/formBuilder/formBuilderUtil.ts
bobbyw-ng: What about:
```
Cannot delete item

[Item Title] cannot be deleted because it is referenced elsewhere:
* Items with autofill condition(s):
  * [Other Item Title]
  * [Other Item Title]
* Items with visibility condition(s):
  * [Other Item Title]
* Items with min/max bound(s):
  * [Other Item Title]

All references to this item must be removed before it can be deleted
```

The first and last sentence may be redundant, but since it's instructions on how to resolve the issue it might benefit from some repetition. Also this doesn't account for questions with the same titles.

The formatting of item titles should be consistent as well, so we should either use just "quotes" or **bold** but not both

src/modules/formBuilder/components/formTree/useUpdateFormStructure.tsx
martha: Hm, this is confusing, let me think about this a bit more. The problem is that there are two different maps with too-similar names, `itemMap` and `itemIdMap`. I think we should rename them for clarity, and probably also improve where they are generated and how they are passed around for consistency and performance.
- `itemMap`:
  - Maps `linkId => FormItem`
  - Predates the form builder, is used many places across form related code
  - For the Form Builder, it's currently created inside of `FormTree` and passed via `FormTreeContext` so that it doesn't get re-generated by every `FormTreeLabel`, which is what was happening before [this commit](https://github.com/greenriver/hmis-frontend/pull/821/commits/3bef69c92f20a396eee618282e30cd5cb51545eb).
  - It's used by the `FormTreeLabel` in order to grab the item from the form definition instead of from RichTreeView's "publicAPI", which is what was happening before [this commit](https://github.com/greenriver/hmis-frontend/pull/821/commits/063f06aa47f006a364da18c901b1596b4b7e2f7e#diff-0531236b7fea21dc33f861077b3ca9822708adeb428d830d25d7c72ee92dd46eL51-L52).
- `itemIdMap`
  - Maps `linkId => path` where path is e.g. `item.0.item.1`
  - New with the RHF/form builder
  - Only used here in the `useUpdateFormStructure` hook in order to get the RHF paths for manipulating the location of items within the tree.

Whew thanks for the opportunity to parse out that confusion in writing! I think my proposed changes are:
- [x] rename `itemIdMap`, maybe `itemPathMap`? `itemRhfPathMap`? `rhfPathMap`?
- [x] move `itemIdMap`'s generation into `FormTree` and pass to `FormTreeContext` alongside `itemMap`, that way it isn't regenerated per `FormTreeLabel`.

martha: Ok, I've attempted to clarify a bit in this commit: https://github.com/greenriver/hmis-frontend/pull/821/commits/78c2e729d494aba64034ed775c23c422f3f5a07b

src/modules/formBuilder/components/formTree/FormTree.tsx
martha: Thanks for this! I added a comment to this effect.

While I was composing that comment, I started to wonder, regarding our custom hook `upseUpdateFormStructure,` what do you think are the best practices around passing something via context vs. passing as arguments to the hook?
- should we move the `useWatch` call inside of `useUpdateFormStructure` up into the context? `FormTree` already has this `useWatch` call so it would be easy to pass in, but maybe that would be us duplicating the purpose of the rhf context?
- `useUpdateFormStructure` also gets `expandItem` from `FormTreeContext`, but it now gets `rhfPathMap` passed as an arg, in response to your "better yet" comment [here](https://github.com/greenriver/hmis-frontend/pull/821#discussion_r1655365400). Should `expandItem` also be passed as an arg?

gigxz: The alternative to `context` would be to pass those values as props down to FormTreeItem and FormTreeLabel respectively (using `slotProps`). So I don't think we have a super _strong_ case for using a custom Context in our form tree, since we're just basically passing data down to one component through 2 levels. The main benefit imo is to avoid the annoying slotProp thing and it's ts issues. I recommend reviewing https://react.dev/learn/passing-data-deeply-with-context#before-you-use-context. 

> should we move the useWatch call inside of useUpdateFormStructure up into the context? FormTree already has this useWatch call so it would be easy to pass in, but maybe that would be us duplicating the purpose of the rhf context?

Do you mean adding the watched `values` into the context? I think that is unnecessary. RHF already provides its own context and a way to consume it, so lets just use that where needed

> useUpdateFormStructure also gets expandItem from FormTreeContext, but it now gets rhfPathMap passed as an arg, in response to your "better yet" comment https://github.com/greenriver/hmis-frontend/pull/821#discussion_r1655365400. Should expandItem also be passed as an arg?

The point of `useUpdateFormStructure` hook is just to pull out some logic from FormTreeLabel so it's not quite so bloated with reordering logic. It's not a hook that we plan to _reuse_ so I'm not so worried about the api, I guess. But here is someone that asked the same question and got a logical response: https://stackoverflow.com/questions/55147479/is-it-better-to-pass-context-to-a-custom-hook-or-call-usecontext-inside-the-cust

src/modules/formBuilder/components/formTree/useUpdateFormStructure.tsx
gigxz: Oh sorry, totally missed that getItemIdMap and getItemMap were different! Sounds good

src/modules/formBuilder/components/formTree/FormTree.tsx
martha: Wonderful, thanks for all this. I added a comment indicating why we wanted a `FormTreeContext` in the first place, and moved `expandItem` up so it's explicitly passed down to the hook, as the stack overflow post recommended. This way it's especially clear that the `FormTreeContext` is only used in one place, and it's for the specific purpose of getting around the slotProps annoyance.

src/modules/formBuilder/formBuilderUtil.ts
martha: Great, I think these are both big improvements.

I think often there will be only one dependent item -- should the display change when that's the case? such as removing all the nesting and just saying,

> [Item Title] cannot be deleted because [Other Item Title]'s autofill condition(s) depend on it.

Agree about bold vs. quotes. Which do you prefer? I uploaded 2 quick screenshots to [figjam](https://www.figma.com/board/axGI9vwmuBRC4uZ11ZJ3RP/Form-Builder-Workflow?node-id=900-7646&t=LPqSP0K6FaR93jiY-0) so we can compare. (Sorry for the lack of bullet points, I still need to add them)

~I'm not 100% sure about "it is referenced" -- the grammarian in me (lol) wants to say that the verb is "to refer" rather than "to reference" but I guess "to reference" is also common usage nowadays. Maybe "because it is referred to elsewhere" or "because other item(s) refer to it"?~ edit: after a spirited dinner table conversation, my conclusion is that "reference" as a verb is fine 😂

src/modules/formBuilder/components/formTree/FormTreeLabel.tsx
martha: Per Bobby's designs, adding similar: "All references to this item must be removed before it can be deleted."

src/modules/formBuilder/formBuilderUtil.ts
martha: Yes good point, thanks for this! (didn't end up making any changes here because, as you said, your other comment meant this code got deleted)

src/modules/formBuilder/components/formTree/FormTree.tsx
gigxz: Great, thanks!

src/modules/formBuilder/components/formTree/FormTreeLabel.tsx
```diff
@@ -91,6 +112,47 @@ const FormTreeLabel: React.FC<FormTreeLabelProps> = ({
         height: '48px',
       }}
     >
+      {!!itemDependents && (
+        <ConfirmationDialog
```
gigxz: Might be nice to pull this out into a separate file, since it's pretty far from "Form Tree Label". This can be simplified to `<DeleteItemConfirmationDialog dependentItems={dependentItems} />`

src/modules/formBuilder/components/formTree/FormTreeLabel.tsx
```diff
@@ -91,6 +112,47 @@ const FormTreeLabel: React.FC<FormTreeLabelProps> = ({
...
+          <Alert
+            color='error'
+            icon={false}
+            sx={{ my: 2, '& .MuiAlert-message': { width: '100%' } }}
+          >
+            {Object.entries(itemDependents).map(([key, val]) =>
+              val?.length > 0 ? (
+                <Box key={key}>
+                  {dependentLabelMap[key]}:
+                  <List sx={{ listStyleType: 'disc', py: 0 }}>
```
gigxz: This would probably be a good opportunity to add a `CommonUnorderedList` component, that can be used like so:
```
<CommonUnorderedList>
  <li> item 1 </li>
  <li> item 2 </li>
</CommonUnorderedList>
```


That matches the pattern we use for `CommonUnstyledList` and `CommonOrderedList` components. They make use of simple ol/ul tags instead of MUI which is a bit heavy-weight for this use case. This will also help ensure that future bulleted lists we place in the app will have the same spacing. (Assuming the `CommonUnorderedList` component is discoverable).

See usage here: https://github.com/greenriver/hmis-frontend/blob/5a274046ef33057613c29a37aa16053790bf8197/src/modules/assessments/components/household/HouseholdSummaryIntakeHelpCard.tsx#L44-L51

src/modules/formBuilder/components/formTree/FormTreeLabel.tsx
```diff
@@ -28,42 +32,56 @@ export interface FormTreeLabelProps
...
+      itemId,
+      item,
+      itemMap,
+      rhfPathMap,
+      expandItem
+    );
+
+  const [itemDependents, setItemDependents] = useState<
+    ItemDependents | undefined
+  >(undefined);
```
gigxz: Suggest adding a comment to explain that this piece of state relates specifically to items that are blocking a deleting that is currently being attempted. It could be mistaken for just holding the items that are dependent on this item, at all times.

gigxz (approved):
Looks great! I left few more suggestions on code style, but go ahead and merge when those are addressed to your liking.

PR #825, opened 2024-06-27
Stricter behavior for form property editor

gigxz:
## Description

Update property editor to be match stricter schema. Sorry this is a big diff here. It includes:
* Only show the "relevant" fields in EnableWhen/AUtofill/Initial sections
* Some bug fixes on formUtil
* Some UI improvements

Relates to (but does not depend on): https://github.com/greenriver/hmis-warehouse/pull/4498 which enforces the stricter schema

**How to test**:
* pull down the related backend branch `stricter-hmis-form-json-schema`
* you will probalby need to delete any local forms that you have updated using the form builder to date, because it injected values that are now considered invalid
* use the form builder to create new forms. they should be valid (Eg not a bunch of JSON Schema errors when you save)

## Type of change


gigxz:
Still need to update the schema to match some of the changes made here, particularly around the "oneOf" needs https://github.com/greenriver/hmis-warehouse/pull/4498#discussion_r1661489534

src/modules/formBuilder/components/itemEditor/conditionals/EnableWhenCondition.tsx
```diff
@@ -27,11 +31,68 @@ interface EnableWhenConditionProps {
   enableWhenPath?: 'enableWhen' | `autofillValues.${number}.autofillWhen`; // path to enableWhen in form
 }
 
+const validOperatorsForType = (type: ItemType, repeats: boolean) => {
```
gigxz, line 34: should we try to capture this logic in the JSON Schema or validation?

there are many instances of this. It will be difficult (impossible?) for me to express this in JSON schema but I can validate it with code in the validator.

src/modules/formBuilder/components/itemEditor/conditionals/EnableWhenCondition.tsx
```diff
@@ -27,11 +31,68 @@ interface EnableWhenConditionProps {
...
+    operators.push(EnableOperator.GreaterThanEqual);
+    operators.push(EnableOperator.LessThanEqual);
+  }
+
+  // any item type supports enableWhen, so we can always evaluate the condition of whether the item is enabled or not
+  operators.push(EnableOperator.Enabled);
+  return operators;
+};
+
+export const determineEnableWhenComparisonField = (
```
gigxz, line 65: should we try to capture this logic in the json schema

src/modules/formBuilder/formBuilderUtil.ts
```diff
@@ -82,6 +82,39 @@ export const validComponentsForType = (type: ItemType) => {
       return [];
   }
 };
+export const determineAutofillField = (itemType: ItemType) => {
```
martha, line 85: Could these be combined into 1 function?

src/modules/formBuilder/components/itemEditor/conditionals/EnableWhenCondition.tsx
martha: Nice, I think this is a much better ux than before!

As for validating in the json schema or in the backend validator code, I think it is a good idea, but now that these restrictions are in place in the frontend interface, I also think it would be fine to leave as a todo/cleanup for later.

src/modules/formBuilder/components/itemEditor/conditionals/InitialValue.tsx
```diff
@@ -74,54 +94,80 @@ const InitialValue: React.FC<Props> = ({ itemType, control }) => {
...
-                options={localConstantsPickList}
-                helperText='Set initial value based on a local constant'
-              />
+              <Box sx={{ mt: 2 }}>
+                <Typography typography='body2' fontWeight={600}>
+                  Advanced Options
+                </Typography>
+                <Stack>
+                  <LabeledCheckbox
+                    label='Initialize using a Local Constant'
```
martha: I wonder if "initialize" might be a bit jargon-y for our users, though maybe this is ok because it's an advanced setting anyway. Could this be 'Set initial value based on a local constant', like the helper text above?

src/modules/formBuilder/components/itemEditor/FormEditorItemProperties.tsx
```diff
@@ -108,6 +136,15 @@ const FormEditorItemProperties: React.FC<FormEditorItemPropertiesProps> = ({
     [definition.definition]
   );
 
+  // If a field is marked as "always hidden", clear the enable_when conditions
```
martha, line 138: I could imagine this causing ux issues if someone taps the "always hidden" button accidentally (just clicking around) on a form item that already has a complicated enableWhen setup. I'm sure it's fine since they can just cancel their changes instead of saving. But maybe we could consider also adding helper text such as "Removes any existing conditions for visibility"? We even could set that text conditionally depending on `hasEnableWhen`.

src/modules/formBuilder/components/itemEditor/FormEditorItemProperties.tsx
```diff
@@ -198,41 +238,59 @@ const FormEditorItemProperties: React.FC<FormEditorItemPropertiesProps> = ({
...
+          {isQuestionItem && (
+            <ControlledTextInput
+              control={control}
+              name='briefText'
+              label='Brief label'
+              helperText="Label to display when the item is referenced briefly, such as in an Autofill dialog box. If not specified, the item's normal label text is shown."
+            />
+          )}
+          {!isGroupItem && (
+            // Read-only text use cases:
```
martha, line 266: nice, thanks for the use case comment 😎

src/modules/formBuilder/components/itemEditor/FormEditorItemProperties.tsx
```diff
@@ -266,31 +324,112 @@ const FormEditorItemProperties: React.FC<FormEditorItemPropertiesProps> = ({
...
+                      <b>Hidden</b>: When this item is disabled, it will be
+                      completely hidden from view.
+                      <br />
+                      {/* TODO: do we actually use  'protected' at all? can we remove it? */}
+                      <b>Protected</b>: When this item is disabled, it will
+                      still show up but not be interactable.
+                      <br />
+                      <b>Protected with Value</b>: When this item is disabled,
+                      it will still appear and its value will be visible but not
+                      interactible. It's value will be submitted.
```
martha:
```suggestion
                      interactible. Its value will be submitted.
```

src/modules/formBuilder/components/itemEditor/FormEditorItemProperties.tsx
```diff
@@ -266,31 +324,112 @@ const FormEditorItemProperties: React.FC<FormEditorItemPropertiesProps> = ({
...
+                  placeholder='Select disabled display'
+                  options={disabledDisplayPickList}
+                  helperText={
+                    <>
+                      <b>Hidden</b>: When this item is disabled, it will be
+                      completely hidden from view.
+                      <br />
+                      {/* TODO: do we actually use  'protected' at all? can we remove it? */}
+                      <b>Protected</b>: When this item is disabled, it will
+                      still show up but not be interactable.
```
martha:
```suggestion
                      still appear but not be interactable.
```

martha (approved):
Wonderful, I think this is a huge ux improvement! Thanks for spending so much time in the logical weeds 🌱

src/modules/formBuilder/components/itemEditor/FormEditorItemProperties.tsx
gigxz: Hmm good point, I considered that too. I just reduced the click target on the "always hidden" checkbox so it only covers the label, not the full width, which might reduce accidental clicks. I will leave as-is for now since user can discard changes if anything was lost.

martha (approved):

src/modules/formBuilder/formBuilderUtil.ts
gigxz: They could -- they are currently equivalent -- but they won't remain that way. They have som fields that differ but aren't support yet. I didn't want to give the wrong impression that they were equivalent.

src/modules/formBuilder/components/itemEditor/conditionals/EnableWhenCondition.tsx
gigxz: Some validations and TODOs added to https://github.com/greenriver/hmis-warehouse/pull/4498

PR #826, opened 2024-06-27
Implement "Form Applicability" design for viewing form rules

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6108
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4494

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/admin/components/formRules/FormRuleCard.tsx
martha, line 1: I renamed this to `FormRulesCard`, plural, to better differentiate it from singular `FormRule`

src/modules/admin/components/formRules/FormRuleTable.tsx
```diff
@@ -127,75 +46,39 @@ const FormRuleTable: React.FC<Props> = ({
     getVariables: (values) => ({
       input: { input: values as FormRuleInput, id: selectedRule?.id || '' },
     }),
-    onCompleted: () => {},
+    onCompleted: () => {
+      evictQuery('formDefinition', { id: formId });
```
martha, line 50: Maybe could use more precise cache manipulation to _just_ invalidate/refetch the `formDefinition`'s `projectMatches`, and not the entire record.

src/modules/admin/components/formRules/FormRule.tsx
```diff
@@ -0,0 +1,121 @@
...
+        <IconButton
+          aria-label='edit form rule'
+          disabled={rule.system}
+          onClick={() => {
+            setSelectedRule(rule);
+            openFormDialog();
+          }}
+          size='small'
+        >
+          <EditIcon fontSize='inherit' />
```
martha, line 114: This edit button and will go away soon (https://github.com/open-path/Green-River/issues/6109) but I left it for now in order to easily test the result of changing rules on the project applicability.

src/modules/dataFetching/components/GenericTableWithData.tsx
```diff
@@ -61,6 +61,7 @@ export interface Props<
   defaultFilterValues?: Partial<FilterOptionsType>;
   showTopToolbar?: boolean;
   noSort?: boolean;
+  noPaginate?: boolean;
```
martha: I added `noPaginate` for now just to get everything hooked up. But I would like to discuss what to do in terms of pagination, sorting, and filtering on the `FormProjectMatchTable`.

The underlying record type, `InstanceProjectMatch` is just a ruby object, not a rails model. My understanding is that from the backend's perspective, there's not much benefit to paginating this list, because we aren’t querying the db.

Is it worthwhile to enable pagination for this table just for the ux benefit? Is there anywhere else in the codebase already where we paginate, filter, or sort on a non-ActiveRecord?

gigxz: If it's not too hard, I think it would be worthwhile to resolve the array as paginated in the GraphQL API, so that the frontend can use the same patterns for pagination.

We have almost had this need before (paginating an array) and I'm sure we'll have it again. I started an implementation one time that never made it into the codebase. Here are some not-fully-fleshed thoughts about that:
* `Types::Forms::ProjectMatch.page_type`
* Update [this function](https://github.com/greenriver/hmis-warehouse/blob/771099a1842631e424086ceadd617fec33ded06c/drivers/hmis/app/graphql/types/paginated_scope.rb#L18) to have conditional logic for how to apply `limit` or `offset` based on whether `@all_nodes` is an AR scope or an array. 
  * Actually, maybe the more appropriate thing is to create a `PaginatedArray` that is similar to `PaginatedScope`, and choose the appropriate one [here](https://github.com/greenriver/hmis-warehouse/blob/771099a1842631e424086ceadd617fec33ded06c/drivers/hmis/app/graphql/types/base_field.rb#L58-L59) based on whether the object is an array or a scope.
* There is a downside which is we could accidentally resolve an array when really we should use a scope, and it would be slow. For that reason, it might be better to have something explicit (eg specifying `Types::Forms::ProjectMatch.array_page_type` but that might be a little tricker to work in

martha: Yes sounds good! See https://github.com/greenriver/hmis-warehouse/pull/4494/commits/a67d6179b9971b5f6cb635c81378734437183380

src/modules/admin/components/formRules/FormRuleTable.tsx
martha: Since this is related to editing, I'd like to address it in https://github.com/open-path/Green-River/issues/6109 which I'll pick up next!

src/components/elements/TitleCard.tsx
```diff
@@ -60,6 +62,7 @@ const TitleCard: React.FC<Props> = ({
         }}
       >
         <Typography
+          component={headerComponent || 'span'}
```
gigxz: Defaulting to `span` might not be right. Can we just leave it empty if `headerComponent` was not passed?
It will set the component based on the default `variantMapping` https://mui.com/material-ui/api/typography/ so maybe we should leave that behavior as-is until we address the header tag issue across the board

src/modules/admin/components/formRules/FormRulesCard.tsx
```diff
@@ -0,0 +1,101 @@
...
+      evictQuery('formDefinition', { id: formId });
+      evictQuery('formRules');
+    },
+  });
+
+  // Fetch here in order to display the total number of project matches outside of the table.
+  // FormProjectMatchTable also requests this, so it will fetch from the apollo cache.
+  const { data: fetchFormData } = useGetFormProjectMatchesQuery({
+    variables: { id: formId },
+  });
```
gigxz, line 47: I am seeing this make 2 queries on page load
Screenshot 2024-07-01 at 11 40 26 AM

But I see why it's needed, since we don't have an easy way to get that number "out" of the GenericTableWithData component. There are definitely improvements that could be made to that component api, since we have this need elsewhere too.

This is OK to leave as-is but maybe we can set `limit: 0` explicitly since we're not doing anything with the returned nodes.

src/modules/admin/components/forms/FormDefinitionDetailPage.tsx
```diff
@@ -91,7 +91,13 @@ const FormDefinitionDetailPage = () => {
...
         <Grid container spacing={2}>
           <Grid item xs={12} md={8}>
-            <CommonCard title='Details'>
+            <CommonCard
+              title={
+                <Typography variant='h4' component='h2' sx={{ mb: 2 }}>
+                  Details
+                </Typography>
+              }
+            >
```
gigxz: This doesn't seem ideal.. the point of CommonCard is to hide these style details. Is there a reason we have `h5` CommonCards some places but this page in particular needs `h4` cards? We should discuss this w design

gigxz (commented):
Looks great!! just a few minor comments

src/modules/admin/components/formRules/FormRulesCard.tsx
martha: Ah thanks for checking this. I updated the query to use `limit: 0` and updated the comment as well to be more honest.

src/modules/admin/components/forms/FormDefinitionDetailPage.tsx
martha: Hm, you're right -- I was just trying to figure out a way to update the Form Definition page to match the designs, without messing with CommonCards across the app -- but we should definitely take a broader view.

I started a thread in the figjam here: https://www.figma.com/board/axGI9vwmuBRC4uZ11ZJ3RP?node-id=773-10818#858654148

My guess is that Bobby made the change from `<h5>` to `<h4>` in the card titles on this page so that the cards could have subtitles at the `<h5>` level, such as Form Applicability - Rules and Projects. So as not to block this PR on that conversation, I'll revert the change here to use `<h5>` and then make those sub-headings `<h6>`, if that sounds good to you.

martha: Here's a preview of how that approach looks:

Screenshot 2024-07-01 at 12 30 29 PM

gigxz: okay sounds good, thanks for following up there

gigxz (approved):
👍

PR #829, opened 2024-07-01
Implement new form rule dialog

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6109

Here's a screenshot of the current pass:
Screenshot 2024-07-02 at 8 43 33 AM

It's not an exact match for the [designs](https://www.figma.com/board/axGI9vwmuBRC4uZ11ZJ3RP/Form-Builder-Workflow?node-id=0-1&t=O9bCOwTTmM3E1JTX-0), since some assumptions have changed since those were created:
- for v1, we are returning to the model where you can only edit one rule at once (for simplicity)
- we are removing the Project Applicability info from the dialog, since it is redundant
- some visual changes make this component more consistent with other usages elsewhere, like using an X instead of a trash icon for RemovableCard, and the modal width itself.

To discuss:
- [ ] Correct approach for restoring lost functionality for Service Type and Service Category and other variances based on Form Role: https://github.com/greenriver/hmis-frontend/pull/826/files#diff-263242b358a931201dbc10b01b439b9a544c1ec5bbc1cd611d00dfce7a30534dL134-L149
- [x] Should this reuse the existing `ProjectApplicabilitySummary` chip, which I just discovered 🤦‍♀️ ? Edit: Added a todo for a future cleanup
- [x] Ordering of the conditions in the group should ideally depend on the order in which clients add remove them, but it's complicated to implement; is this worth it? Edit: Could be fixed as part of Gig's suggestion here https://github.com/greenriver/hmis-frontend/pull/829#discussion_r1662983538, but after a [timebox](https://github.com/greenriver/hmis-frontend/commit/348fd1cedde753497ba2b7b6af30a1c5e2652b6d) we aren't continuing with this.
- [x] Enable turning rules to inactive (in lieu of deletion) - https://github.com/open-path/Green-River/issues/6289 is ~not pulled into the current sprint, but without it, there is no UI way to deactivate rules now that I've removed the old edit modal.~ pulled into this PR

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/formBuilder/components/itemEditor/conditionals/CardGroup.tsx
```diff
@@ -21,8 +23,8 @@ export const RemovableCard: React.FC<RemovableCardProps> = ({
         border: '1px solid white',
         borderRadius: 1,
         borderColor: 'borders.light',
-        mb: 1,
```
martha, line 24: This change does result in a visual change to the `CardGroup`s in the Form Item editor drawer -- they have less spacing between them now. The reason to make this change is so that the margin between the last card and the "Add Condition" button at the bottom is independent of the margin set on the cards themselves. I think it's also better for the spacing between these cards to be controlled by one prop, the `spacing` on the `Stack`, rather than 2 (the stack's spacing + the margin on the cards).

It is possible to set the stack's spacing to 3, which reverts the visual change, and then we can make that a passable prop from the parent in order to achieve the desired spacing on the new Form Rule dialog, but I wasn't sure this was worth it. Let me know what you think!

src/components/elements/input/GenericSelect.tsx
```diff
@@ -48,6 +49,7 @@ const GenericSelect = <
     <Autocomplete
       options={options}
       value={value}
+      disableClearable={disableClearable}
```
gigxz: This shouldn't be needed because `...rest` is already spread to the `Autocomplete` component. So it should be receiving the `disableClearable` prop already. Was it not working without this?

src/modules/admin/components/formRules/NewFormRuleDialog.tsx
```diff
@@ -0,0 +1,304 @@
...
+          />
+        </Stack>
+        <CardGroup
+          onAddItem={() => {
+            if (conditionsAvailable.length > 0) {
+              const conditionType = conditionsAvailable[0];
+              const defaultValue = pickListMap[conditionType][0].code;
+              // We display all conditions with a non-null value, so setting this condition's value
+              // to some default value for its type (e.g. the first Project in the list) causes it to appear.
+              setRule({ ...rule, [conditionType]: defaultValue });
```
gigxz: It seems like using `FormRuleInput` as the internal component state caused some UX limitations:
* the "condition types" are in a fixed order
* requiring us to auto-fill an initial value for the field to appear

The first bullet doesn't seem like a big deal, but I think the second one is worth addressing. I could imagine someone thinking that the selected project is intentional (i.e. "am I editing an existing rule now? did someone already set this to project X? Why does it say Coordinated Entry? etc..) and that causing some support questions/confusion.

The better way to do this might be to refactor so that the component state mimics the interface (e.g. an array of conditions), and it only gets converted into a FormRuleInput on submit. Though that feels like too big of a change to make at this stage, for this component, so I would suggest a different workaround instead:
- set value to an empty string `''` to indicate that it's present. switch to filtering out like `.filter((condition) => typeof condition.value !== 'undefined')`.

src/modules/admin/components/formRules/NewFormRuleDialog.tsx
```diff
@@ -0,0 +1,304 @@
...
+              }}
+            >
+              <Grid container spacing={1}>
+                <Grid item xs={4}>
+                  <FormRuleLabelTypography>
+                    And funding source is
+                  </FormRuleLabelTypography>
+                </Grid>
+                <Grid item xs={8}>
+                  <TextInput
```
gigxz, line 333: It would be great for this to be a picklist of distinct values from the `OtherFunder` column for all HMIS Projects. The string needs to match exactly, so it's more error-prone to leave it up to the user. Would you mind creating a follow-up ticket for that? Or, you're welcome to bring it into this work instead if you'd like. It will require another backend change to add a new remote pick list (perhaps called `LOCAL_OR_OTHER_FUNDERS`).

src/modules/admin/components/formRules/NewFormRuleDialog.tsx
```diff
@@ -0,0 +1,340 @@
...
+                        sx={isTiny ? {} : { textAlign: 'center' }}
+                      >
+                        is
+                      </FormRuleLabelTypography>
+                    </Grid>
+                    <Grid item xs={10} sm={5}>
+                      <FormSelect
+                        // disable clearing the input because we expect users to remove the whole condition
+                        disableClearable={true}
+                        value={value ? { code: value } : undefined}
```
gigxz: - If `value` is allowed to be an empty string (re: other suggestion) the initial value should be set to null, not undefined. undefined indicates that its an uncontrolled component so you get an error.
- Suggest adding placeholder text
```suggestion
                        disableClearable
                        value={value ? { code: value } : null}
                        placeholder={`Select ${startCase(
                          conditionType.replace(/Id$/, '')
                        )}`}
```

src/modules/admin/components/formRules/FormRule.tsx
```diff
@@ -51,69 +59,81 @@ const FormRule: React.FC<Props> = ({
...
-            openFormDialog();
-          }}
-          size='small'
-        >
-          <EditIcon fontSize='inherit' />
-        </IconButton>
-      </ButtonTooltipContainer>
+      <ActiveChip
+        active={active || false}
+        // TODO(#6289) - Deletion should deactivate. Inactive rules should not be shown/returned
```
gigxz: On this branch, I'm not seeing a way to edit/deactive/delete existing rules. That is a functionality regression, so we should either add the deletion button on this branch, or keep the old "edit" functionality until it's ready.

gigxz (commented):
looks good! let me know if there's anything you want to discuss from my comments or otherwise

src/modules/admin/components/formRules/FormRule.tsx
```diff
@@ -26,11 +18,25 @@ const ActiveChip = ({ active }: { active: boolean }) => (
...
+];
+
+const nonProjectFormRoles = [FormRole.Organization];
+
+interface Props {
+  rule: FormRuleFieldsFragment;
+  formRole: FormRole;
+}
+
+const FormRule: React.FC<Props> = ({ rule, formRole }) => {
```
gigxz: > Should this reuse the existing ProjectApplicabilitySummary chip, which I just discovered

That component was just a quick implementation for a way to display project applicability rules. It would be a plus if we could use the same underlying component to render rules in `ProjectConfigTable`. The project config settings are more limited, which is why that component [accepted a union type and used type guards to access the non-overlapping fields](https://github.com/greenriver/hmis-frontend/blob/90336b447c650dad7cc758e0e5ce49fca0c1d796/src/modules/admin/components/formRules/ProjectApplicabilitySummary.tsx#L23-L25). The same approach could be taken with this FormRule component (Eg have the incoming rule prop be `FormRuleFieldsFragment | ProjectConfigFieldsFragment`, and `formRole` be optional). 

But that's purely a nice-to-have code cleanup. We can always do it later.

src/components/elements/input/GenericSelect.tsx
martha: Ah good point, thanks. I'll remove.

src/modules/admin/components/formRules/NewFormRuleDialog.tsx
martha: Thanks -- yes, I had some version of this in the past and would be happy to revert to it.

I'm also happy to spend a few minutes trying to implement your first suggestion. I think it would probably be more readable, and it's probably also a useful exercise for me. I'll timebox my effort to half an hour or so.

gigxz: Sounds good. I'm totally fine with as-is. This might be helpful for your timebox though https://react.dev/learn/updating-arrays-in-state

martha: Ok, in a short time box I didn't come up with something that was as functional as / more readable than the code as-is, so I'll leave it!

src/modules/admin/components/formRules/NewFormRuleDialog.tsx
martha: Sounds good, I created https://github.com/open-path/Green-River/issues/6317

src/modules/admin/components/formRules/FormRule.tsx
```diff
@@ -130,10 +140,14 @@ const FormRule: React.FC<Props> = ({ rule, formRole }) => {
...
       </Typography>
-      <ActiveChip
-        active={active || false}
-        // TODO(#6289) - Deletion should deactivate. Inactive rules should not be shown/returned
-      />
+      <IconButton
+        onClick={() => deactivate()}
+        size='small'
+        sx={{ height: '28px' }}
+        disabled={loading}
```
gigxz: There's a moment after the deletion completes that the button becomes enabled again, when it's no longer loading.
You could disable this on `called` instead so it remains disabled https://www.apollographql.com/docs/react/data/mutations/#mutationresult-interface-called

src/modules/admin/components/formRules/FormRule.tsx
```diff
@@ -34,17 +30,29 @@ const nonProjectFormRoles = [FormRole.Organization];
 interface Props {
   rule: FormRuleFieldsFragment;
   formRole: FormRole;
+  formId: string;
 }
 
-const FormRule: React.FC<Props> = ({ rule, formRole }) => {
+const FormRule: React.FC<Props> = ({ rule, formRole, formId }) => {
+  // TODO - This duplicates some functionality from ProjectApplicabilitySummary used in the Project config table.
```
gigxz, line 31: To that point, it will be easier to refactor if the `useDeactivateFormRuleMutation` doesn't live here. (Since we don't need that for  Project config context).

Previously `FormRule` served as more of a presentational component, but we make it a data-fetching component by adding this mutation. I would suggest moving the mutation out into `FormRuleTable`, which is already doing data fetching. The `FormRule` component could accept an optional `actionButtons`(?) prop that we can use to pass the `<IconButton`. That makes it a little more flexible for use elsewhere, and keeps it "presentational".

src/modules/admin/components/formRules/NewFormRuleDialog.tsx
gigxz: Thanks!

gigxz (approved):
Looks great! Left a few suggestions on the delete functionality.

src/modules/admin/components/formRules/FormRule.tsx
martha: I think this should be fixed now with my recent updates to cache eviction after the query completes.

src/modules/admin/components/formRules/FormRule.tsx
```diff
@@ -51,69 +70,87 @@ const FormRule: React.FC<Props> = ({
...
-    conditions.push(
-      <FormRuleChip label={`Funding Source is ${otherFunder}`} />
-    );
-  } else if (funder) {
-    conditions.push(
-      <FormRuleChip
-        label={`Funding Source is ${HmisEnums.FundingSource[funder]}`}
-      />
-    );
+  const conditions: Record<string, ReactNode> = {};
```
martha, line 44: @gigxz I'd especially appreciate your review on the changes here. I've tried to add back functionality that I lost in implementing the new Form Rule table (in the [previous PR](https://github.com/greenriver/hmis-frontend/pull/826/files#diff-263242b358a931201dbc10b01b439b9a544c1ec5bbc1cd611d00dfce7a30534dL42-L97)).

One specific thing I'm worried about is ending up with a form rule that's displayed as a blank row in the table. I've tried to protect against that [here](https://github.com/greenriver/hmis-frontend/pull/829/files#diff-9ea65162911b6860febdc61e834e3f30615253867cc120610e61b9c20cad815eR126-R132).

src/modules/admin/components/formRules/FormRule.tsx
martha: Makes sense to me. Thanks! This change causes all the deletion buttons to be greyed out during a mutation (instead of just the button pertaining to the rule that's getting deleted by this mutation.) I'm not sure if there was a way around this, I was checking the [useMutation docs](https://www.apollographql.com/docs/react/data/mutations/#usemutation-api) to see if there is an `onStarted` or similar callback that we could use to 'listen' to which ID is currently being deleted and then separately manage a piece of state for which one the `loading` applies to, but I don't think such a callback exists. It doesn't seem like a huge deal to me though, I'm fine with leaving it as-is.

src/modules/admin/components/formRules/FormRuleTable.tsx
```diff
@@ -73,17 +57,22 @@ const FormRuleTable: React.FC<Props> = ({ formId, formRole }) => {
...
+                actionButtons={
+                  <IconButton
+                    onClick={() => deactivate({ variables: { id: rule.id } })}
+                    size='small'
+                    sx={{ height: '28px' }}
+                    disabled={loading}
+                  >
+                    <DeleteIcon fontSize='small' />
+                  </IconButton>
+                }
```
gigxz: Removal should be disabled for form rules that are marked as "system". They should raise: https://github.com/greenriver/hmis-warehouse/blob/26ef8c27b51dba80b065dfa9a3034722f165a15e/drivers/hmis/app/graphql/mutations/update_form_rule.rb#L19

"system" rules exist to ensure that forms that are required for application functioning -- like Project form -- do not get removed
```
actionButtons={
  <ButtonTooltipContainer
    title={rule.system ? 'System rule cannot be removed' : ''}
  >
    <IconButton
      onClick={() => deactivate({ variables: { id: rule.id } })}
      size='small'
      sx={{ height: '28px' }}
      disabled={rule.system || loading}
    >
      <DeleteIcon fontSize='small' />
    </IconButton>
  </ButtonTooltipContainer>
}
```

src/modules/admin/components/formRules/FormRule.tsx
gigxz: Thanks for the call-out. I played around with it a little and have a snippet suggestion below. It includes:

1) More logical display for service type field. The service fields indicate which service type(s) this form collects, so it doesn't really make sense to display it as a "condition". If you look at the default HUD service form you can see lots of examples (screencapped below).
Screenshot 2024-07-03 at 4 59 25 PM

2) Show as "Applies to all clients in all projects" if there are no project conditions selected.:
Screenshot 2024-07-03 at 5 07 47 PM


```
  // remove service stuff from "conditions". 

  const serviceCollected =
    formRole === FormRole.Service && (serviceType || serviceCategory) ? (
      <>
        Collects{' '}
        <FormRuleChip
          label={serviceType?.name || serviceCategory?.name || 'Service'}
        />
      </>
    ) : null;


  return (
    <Stack direction='row' gap={2}>
      <Typography variant='body2' sx={{ flexGrow: 1 }}>
        {serviceCollected}
        {!nonClient && (
          <>
            {serviceCollected ? ' for' : 'Applies to'} {appliesTo}
          </>
        )}
        {conditionCount === 0 &&
          (nonClient ? (
            <>
              Applies to <FormRuleChip label='All Projects' />
            </>
          ) : (
            <>
              {' '}
              in <FormRuleChip label='All Projects' />
            </>
          ))}

        {conditionCount > 0 && (
          //... etc.
```

src/modules/admin/components/formRules/FormRuleTable.tsx
martha: Thanks for this, sorry for the regression!

src/modules/admin/components/formRules/FormRule.tsx
martha: Thanks Gig this is great. I refactored it a little for clarity/simplicity, but I think I've checked each case and kept the functionality from your snippet.

gigxz (approved):
looks good

PR #831, opened 2024-07-02
Make UX improvements to form tree reordering

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6274

- Enforces maximum depth of the form tree at 5
  - I implemented this strictly, by disabling moving a Group item beyond depth 4. Since you can't move an item into a non-Group item, this also has the effect that no item can go beyond depth 5.
- After form reset, re-focuses the most recent up/down button selected so that users can more easily use the keyboard to move an item up or down through a long form.
  - The focus behavior is still a bit unpredictable when you use the keyboard to move an item all the way to the end.

## Type of change
[//]: # 'remove options that are not relevant'
- [x] Bug fix
- [x] New feature (adds functionality)
- [ ] Code clean-up / housekeeping
- [ ] Dependency update

## Checklist before requesting review
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [x] My code includes comments and/or descriptive variable names to help other engineers understand the intent (or not applicable)
- [ ] My code follows the style guidelines of this project (eslint)
- [ ] I have updated the documentation (or not applicable)
- [ ] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

PR #833, opened 2024-07-03
Make exhaustive-deps a lint error, not warning

martha:
## Description

I'm proposing to add a flag to the eslint step that runs on pre-commit, failing the commit if there are warnings (not just errors) from eslint.

Follow-up from https://github.com/greenriver/hmis-frontend/pull/831#discussion_r1664261298, moved to its own PR so as not to block unrelated work from being merged.

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   },
   "lint-staged": {
     "*.{js,ts,jsx,tsx}": [
-      "eslint --fix --ignore-path .gitignore --max-warnings=0"
+      "eslint --fix --ignore-path .gitignore"
     ],
     "*": [
       "prettier --write"
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   },
   "lint-staged": {
     "*.{js,ts,jsx,tsx}": [
-      "eslint --fix --ignore-path .gitignore"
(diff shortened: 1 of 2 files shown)


gigxz (approved):
Thanks! I think we'll need to improve those console.logs if they cause a failure?

package.json
```diff
@@ -24,7 +24,7 @@
   },
   "lint-staged": {
     "*.{js,ts,jsx,tsx}": [
-      "eslint --fix --ignore-path .gitignore"
+      "eslint --fix --ignore-path .gitignore --max-warnings=0"
```
gigxz: Although, I'm not sure if it makes sense for these to fail on staged but not on `yarn lint` (eg not fail on CI). I feel like those should match. This will have no change in terms of enforcing no warnings in the codebase, since precommit hooks are an optional part of the dev setup. What do you think?

If there are specific warnings that we want to be errors (eg missing hook dependencies) then we can mark them as errors in eslintrc.

martha: Ok yes, that makes sense to me. Warning should stay warnings, but anything that is always going to be a blocking comment on a PR (unless it has been explicitly ignored for a specific reason) should be elevated to an error.

gigxz (approved):

PR #835, opened 2024-07-08
Filter form rules by active status

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6109

This PR is meant to update the Forms table so that under the Applicability Rules table, it only returns the number of Active rules, now that we are using Inactive as a deletion flag.

Depends on: https://github.com/greenriver/hmis-warehouse/pull/4520

Should target 124?

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/api/operations/form.queries.graphql
+++ b/src/api/operations/form.queries.graphql
@@ -94,7 +94,7 @@ query GetFormIdentifiers(
       displayVersion {
         ...FormDefinitionMetadata
         system
-        formRules(limit: 1) {
+        formRules(limit: 1, filters: { activeStatus: ACTIVE }) {
           nodesCount
         }
       }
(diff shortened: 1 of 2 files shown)


gigxz:
Ah I see, we'd need to call the resolve function explicitly instead of using the instances method. Like this: https://github.com/greenriver/hmis-warehouse/blob/c261d989ac3919ba875a6d29c09f9149af159023/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb#L426-L431 but with `resolve_form_rules(object.instances, **args)`

Looks like this an N+1 on the table right now. Probalby worth addressing at some point but we have mocks to introduce other columns (eg # Projects where the form applies) so lets not go down that rabbit hole... adding the filter so at least the # reflects the number of active rules seems like a decent fix.

gigxz (approved):

PR #836, opened 2024-07-09
5727-hmis-graphql-retry

ttoomey:
## Description
[GH Issue](https://github.com/open-path/Green-River/issues/5727)

Retry graphql queries under certain circumstances. We are only retrying "Get*" queries here but it may also be safe to retry 'update' or 'delete' queries too.

## Type of change

### DIFF
--- a/src/providers/apolloClient.tsx
+++ b/src/providers/apolloClient.tsx
@@ -127,8 +127,8 @@ export const cache = new InMemoryCache({
 
 const retryLink = new RetryLink({
   delay: {
-    initial: 800,
-    max: 2400,
+    initial: 800, // Wait 800ms before the first retry after a failed request
+    max: 1200, // Maximum wait time between retries
     jitter: true,
   },
   attempts: {
--- a/src/providers/apolloClient.tsx
+++ b/src/providers/apolloClient.tsx
@@ -128,7 +128,6 @@ export const cache = new InMemoryCache({
 const retryLink = new RetryLink({
   delay: {
     initial: 800, // Wait 800ms before the first retry after a failed request
-    max: 1200, // Maximum wait time between retries
     jitter: true,
   },
   attempts: {
--- a/src/providers/apolloClient.tsx
+++ b/src/providers/apolloClient.tsx
@@ -7,6 +7,7 @@ import {
 } from '@apollo/client';
 import { BatchHttpLink } from '@apollo/client/link/batch-http';
 import { setContext } from '@apollo/client/link/context';
+import { RetryLink } from '@apollo/client/link/retry';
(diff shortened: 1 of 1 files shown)


src/providers/apolloClient.tsx
```diff
@@ -124,10 +125,37 @@ export const cache = new InMemoryCache({
...
 });
 
+const retryLink = new RetryLink({
+  delay: {
+    initial: 800, // Wait 800ms before the first retry after a failed request
+    max: 1200, // Maximum wait time between retries
+    jitter: true,
+  },
+  attempts: {
+    max: 3,
```
gigxz: Docs:
> Starting with initialDelay, the delay of each subsequent retry is increased exponentially, meaning it's multiplied by 2 each time. For example, if initialDelay is 100, additional retries will occur after delays of 200, 400, 800, etc.
> With the jitter option enabled, delays are randomized anywhere between 0ms (instant), and 2x the configured delay. This way you get the same result on average, but with random delays.


I am just wondering if there is a particular logic behind setting the `delay.max` so low?  It's less than 2x the initial, so we don't really get the exponential backoff behavior, if I understand correctly. (And the default is `Infinity`)

src/providers/apolloClient.tsx
```diff
@@ -124,10 +125,37 @@ export const cache = new InMemoryCache({
...
+      return !!error;
+    },
+  },
+});
+
 const apolloClient = new ApolloClient({
   link: from([
     apolloErrorLink, // handle errors
     sentryLink, // instrument graphql requests for sentry
+    retryLink,
```
gigxz, line 157: with this placement, if a request fails 2 times and succeeds on the 3rd try, we will not get a Sentry error at all, right?

ttoomey: Yes, that should be the case. It should cut down on the chatter in sentry but maybe we want those alerts?

src/providers/apolloClient.tsx
ttoomey: Good question. We could go with the default exponential backoff here. I think the idea with the max value is it lets us keep a higher initial value but not wait quite so long to get to an error message in the worst case. I could believe it's better to have a longer back off though; it's kinder to our servers in case the failures are due to performance problems and not short-lived network blips

I'll change it and we can always revisit if it seems like a problem

src/providers/apolloClient.tsx
gigxz: Alright, I think that is good 👍

src/providers/apolloClient.tsx
gigxz: Okay, sounds good. Thanks. Being kind to our servers seems like a good place to start

PR #837, opened 2024-07-09
Warn on referral when client is already enrolled

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4526

This PR
- Adds a query for checking for enrollment conflict in the project receiving the referral
- Adds an alert when that query returns information about a conflict (step 1 of the linked ticket, see screenshot)
- Makes schema updates corresponding to the linked warehouse change
- Moves some referral-related queries from `src/api/operations/referral.operations.graphql` to a new file `src/api/operations/referral.queries.graphql`, which is consistent with our pattern elsewhere

Screenshot 2024-07-09 at 2 52 16 PM

Note that there's no frontend change needed for step 2 of the linked ticket; that is included in the existing error handling of the referral creation form:

Screenshot 2024-07-09 at 2 51 42 PM

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/modules/projects/components/ProjectOutgoingReferralDetailsSubForm.tsx
+++ b/src/modules/projects/components/ProjectOutgoingReferralDetailsSubForm.tsx
@@ -1,3 +1,4 @@
+import { Alert } from '@mui/material';
 import { Box } from '@mui/system';
 import { useMemo } from 'react';
 
@@ -7,7 +8,10 @@ import { useDynamicFormHandlersForRecord } from '@/modules/form/hooks/useDynamic
 import useFormDefinition from '@/modules/form/hooks/useFormDefinition';
 import { SubmitFormAllowedTypes } from '@/modules/form/types';
 import { AlwaysPresentLocalConstants } from '@/modules/form/util/formUtil';
-import { RecordFormRole } from '@/types/gqlTypes';
+import {
+  RecordFormRole,
+  useGetCanProjectAcceptReferralQuery,
+} from '@/types/gqlTypes';
 
 interface Props {
   enrollmentId: string; // HoH enrollment that is being referred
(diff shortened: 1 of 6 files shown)


src/modules/projects/components/ProjectOutgoingReferralDetailsSubForm.tsx
```diff
@@ -57,11 +61,24 @@ const ProjectOutgoingReferralDetailsSubForm: React.FC<Props> = ({
     );
   }
 
+  const { data: receivingProject } = useGetCanProjectAcceptReferralQuery({
```
gigxz: feels a little funny to call this `receivingProject` when it's just a data object with a boolean. I think it would be clearer to just leave it as `data` or destructure the boolean field directly

src/modules/projects/components/ProjectOutgoingReferralDetailsSubForm.tsx
```diff
@@ -57,11 +61,24 @@ const ProjectOutgoingReferralDetailsSubForm: React.FC<Props> = ({
...
+    },
+  });
+
   return (
     <Box>
       {formDefinitionLoading && (
         <LoadingSkeleton width={'100%'} count={1} sx={{ my: 2 }} />
       )}
+      {receivingProject && !receivingProject.canProjectAcceptReferral && (
+        <Alert color='warning' sx={{ mb: 2 }}>
```
gigxz: I noticed this still has a checkmark for the warning. I think if you use severity instead of warning it will adjust the icon as well: https://mui.com/material-ui/api/alert/#alert-prop-severity

gigxz (approved):
Looks good once the backend piece is updated!Can you re target to 124?

src/modules/projects/components/ProjectOutgoingReferralDetailsSubForm.tsx
martha: I'll leave it as `data`, that way we can also check whether data has been returned yet or not, so the alert doesn't show up when the result hasn't been returned yet.

PR #847, opened 2024-07-14
6239 adjust vite proxy for insufficient rate limiting

ttoomey:
## Notes:
* this is a development only fix. The production configuration in `docker/templates/proxy_headers.conf.template` appears to already be sending all the XFF headers. So with this change the vite proxy more closely mirrors production behavior
* it's likely that the value for DataSource.hmis will need to be updated in the local database because rails will use the X-Forwarded-Host value to set request.origin

## Description
* Adjust vite (http-proxy) to send x-forward-* headers
* Don't set an origin. To pass csrf validation, the request.origin and base_url must match. Due to sending X-Forwarded-Host and X-Forwarded-Proto these seem to be mismatched with the manual origin header. In theory, we shouldn't need to set it since the forwarded headers should match with the origin header send by the browser. This was quite hard to track down as it surfaces as generic CSRF issue

Depends on hmis-warehouse PR: [4514](https://github.com/greenriver/hmis-warehouse/pull/4514) 

## Type of change

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -35,10 +35,6 @@ export default defineConfig(({ command, mode }) => {
     target: env.HMIS_SERVER_URL || DEFAULT_WAREHOUSE_SERVER,
     xfwd: true,
     changeOrigin: true, // sets Host header
-    headers: {
-      Origin: env.HMIS_SERVER_URL || DEFAULT_WAREHOUSE_SERVER,
-      'X-Proxy-Secret': env.HMIS_PROXY_SECRET,
-    },
     secure: false,
   };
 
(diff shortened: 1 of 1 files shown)


vite.config.js
```diff
@@ -33,10 +33,8 @@ export default defineConfig(({ command, mode }) => {
 
   const warehouseProxyServer = {
     target: env.HMIS_SERVER_URL || DEFAULT_WAREHOUSE_SERVER,
+    xfwd: true,
     changeOrigin: true, // sets Host header
-    headers: {
-      Origin: env.HMIS_SERVER_URL || DEFAULT_WAREHOUSE_SERVER,
-    },
```
gigxz, line 39: see https://github.com/greenriver/hmis-frontend/blob/release-125/docker/nginx.conf and template files in the same dir

gigxz:
Can we close this?

ttoomey:
> Can we close this?

Yes, thanks for the nudge

PR #850, opened 2024-07-17
Add pick list option manager to form builder

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6278

This PR adds the ability to manage pick list options from the form builder. I'll add a few comments/questions inline.

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/components/elements/input/RadioGroupInputOption.tsx
+++ b/src/components/elements/input/RadioGroupInputOption.tsx
@@ -1,9 +1,4 @@
-import {
-  Checkbox,
-  FormControlLabel,
-  FormHelperText,
-  Radio,
-} from '@mui/material';
+import { Checkbox, FormControlLabel, Radio } from '@mui/material';
 import { KeyboardEventHandler, useCallback, useId } from 'react';
 
 import CommonHtmlContent from '@/components/elements/CommonHtmlContent';
@@ -15,7 +10,6 @@ interface Props {
   onChange: (option: PickListOption) => void;
   variant: 'checkbox' | 'radio';
   value: string | undefined;
-  row?: boolean;
 }
 
 const RadioGroupInputOption: React.FC<Props> = ({
@@ -24,7 +18,6 @@ const RadioGroupInputOption: React.FC<Props> = ({
   onChange,
   variant,
   value,
-  row,
 }) => {
   const descriptionId = useId();
   const labelId = useId();
@@ -87,22 +80,9 @@ const RadioGroupInputOption: React.FC<Props> = ({
         }}
       />
       {helperText && (
-        <FormHelperText
-          component={'div'}
-          sx={{
-            mt: 0,
-            display: 'flex',
-            alignItems: 'center',
-          }}
-        >
-          <CommonHtmlContent
-            sx={row ? {} : { ml: 4 }}
-            variant='body2'
-            id={descriptionId}
-          >
-            {helperText}
-          </CommonHtmlContent>
-        </FormHelperText>
+        <CommonHtmlContent variant='body2' sx={{ ml: 4 }} id={descriptionId}>
+          {helperText}
+        </CommonHtmlContent>
       )}
     </>
   );
--- a/src/components/elements/input/RadioGroupInputOption.tsx
+++ b/src/components/elements/input/RadioGroupInputOption.tsx
@@ -1,4 +1,9 @@
-import { Checkbox, FormControlLabel, Radio } from '@mui/material';
+import {
+  Checkbox,
+  FormControlLabel,
+  FormHelperText,
+  Radio,
+} from '@mui/material';
 import { KeyboardEventHandler, useCallback, useId } from 'react';
 
 import CommonHtmlContent from '@/components/elements/CommonHtmlContent';
@@ -10,6 +15,7 @@ interface Props {
   onChange: (option: PickListOption) => void;
   variant: 'checkbox' | 'radio';
   value: string | undefined;
+  row?: boolean;
 }
 
 const RadioGroupInputOption: React.FC<Props> = ({
@@ -18,6 +24,7 @@ const RadioGroupInputOption: React.FC<Props> = ({
   onChange,
   variant,
   value,
+  row,
 }) => {
   const descriptionId = useId();
   const labelId = useId();
@@ -80,9 +87,22 @@ const RadioGroupInputOption: React.FC<Props> = ({
         }}
       />
       {helperText && (
-        <CommonHtmlContent variant='body2' sx={{ ml: 4 }} id={descriptionId}>
-          {helperText}
-        </CommonHtmlContent>
+        <FormHelperText
+          component={'div'}
+          sx={{
+            mt: 0,
+            display: 'flex',
+            alignItems: 'center',
+          }}
+        >
+          <CommonHtmlContent
+            sx={row ? {} : { ml: 4 }}
+            variant='body2'
+            id={descriptionId}
+          >
+            {helperText}
+          </CommonHtmlContent>
+        </FormHelperText>
       )}
     </>
   );
--- a/src/modules/formBuilder/components/itemEditor/FormEditorItemProperties.tsx
+++ b/src/modules/formBuilder/components/itemEditor/FormEditorItemProperties.tsx
@@ -24,7 +24,6 @@ import {
   MAX_INPUT_AND_LABEL_WIDTH,
   getItemMap,
   localResolvePickList,
-  chooseSelectComponentType,
 } from '@/modules/form/util/formUtil';
 import ManagePickListOptions from '@/modules/formBuilder/components/itemEditor/pickLists/ManagePickListOptions';
 import {
@@ -101,40 +100,11 @@ const FormEditorItemProperties: React.FC<FormEditorItemPropertiesProps> = ({
     formState: { isDirty, dirtyFields },
   } = handlers;
 
-  // console.log(errors);
+  // Monitor changes to FormItem fields
   const itemTypeValue = watch('type');
-
-  // Monitor changes to the FormItem.component field
   const itemComponentValue = watch('component');
-  const repeatsValue = watch('repeats');
   const hiddenValue = watch('hidden');
-  const pickListOptionsValue = watch('pickListOptions');
-  const pickListReferenceValue = watch('pickListReference');
   const hasEnableWhen = !!watch('enableWhen.0');
-
-  // Whereas `itemComponentValue` above contains the user-provided component override (null if not specified),
-  // `realItemComponent` is the actual component for the item. It's used specifically for Choice/Open Choice types,
-  // which have their own rules for whether to show a dropdown or radio if the user hasn't specified anything.
-  // If needed, we could support other components here, but right now this will be undefined for non-Choice types.
-  const realItemComponent = useMemo(() => {
-    if (
-      itemTypeValue &&
-      [ItemType.Choice, ItemType.OpenChoice].includes(itemTypeValue)
-    ) {
-      return chooseSelectComponentType(
-        itemComponentValue,
-        repeatsValue,
-        pickListOptionsValue?.length,
-        !pickListReferenceValue
-      );
-    }
-  }, [
-    itemTypeValue,
-    itemComponentValue,
-    repeatsValue,
-    pickListOptionsValue,
-    pickListReferenceValue,
-  ]);
 
   const itemCategory = useMemo<ItemCategory>(() => {
     if (!itemTypeValue) return 'question';
@@ -328,15 +298,11 @@ const FormEditorItemProperties: React.FC<FormEditorItemPropertiesProps> = ({
             />
           )}
         </Section>
-        {[ItemType.Choice, ItemType.OpenChoice].includes(itemTypeValue) &&
-          realItemComponent && (
-            <Section title='Choices'>
-              <ManagePickListOptions
-                control={control}
-                formItemComponent={realItemComponent}
-              />
-            </Section>
-          )}
+        {[ItemType.Choice, ItemType.OpenChoice].includes(itemTypeValue) && (
+          <Section title='Choices'>
+            <ManagePickListOptions control={control} />
+          </Section>
+        )}
         <Section title='Visibility'>
           {isAssessment && (
             <ControlledSelect
--- a/src/modules/formBuilder/components/itemEditor/FormEditorItemProperties.tsx
+++ b/src/modules/formBuilder/components/itemEditor/FormEditorItemProperties.tsx
@@ -24,7 +24,9 @@ import {
   MAX_INPUT_AND_LABEL_WIDTH,
   getItemMap,
   localResolvePickList,
+  chooseSelectComponentType,
 } from '@/modules/form/util/formUtil';
+import ManagePickListOptions from '@/modules/formBuilder/components/itemEditor/pickLists/ManagePickListOptions';
 import {
   COMPARABLE_ITEM_TYPES,
   ItemCategory,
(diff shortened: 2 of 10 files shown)


src/modules/formBuilder/components/itemEditor/pickLists/PickListOption.tsx
```diff
@@ -0,0 +1,111 @@
...
+  index,
+  formItemComponent,
+  isCodeUnique,
+  setPickListInitialSelected,
+}) => {
+  const onChangeInitialSelected = useCallback(
+    (value: boolean) => {
+      // Even though it's a controlled RHF checkbox, we still provide an onChange callback
+      // in order to set all the other pick list options' `initialSelected` values to false when this one is set to true.
+      if (value && code) setPickListInitialSelected(code);
```
martha: I think there is an existing bug with initial selected, but I'm considering it out of the scope of this ticket; I've created a bug ticket that links to a form that reproduces the bug on QA. https://github.com/open-path/Green-River/issues/6368

src/components/elements/input/RadioGroupInputOption.tsx
```diff
@@ -80,9 +87,22 @@ const RadioGroupInputOption: React.FC<Props> = ({
         }}
       />
       {helperText && (
-        <CommonHtmlContent variant='body2' sx={{ ml: 4 }} id={descriptionId}>
-          {helperText}
-        </CommonHtmlContent>
+        <FormHelperText
```
martha: Let me know what you think about the change to add a <FormHelperText> wrapper around the helper text of radio buttons. I think it's better for helper text to be consistent, but if we do merge this, we may want to re-open the SPDAT and specify in the custom html there that those helper texts should continue to use the color black, instead of the default helper text color of grey.

src/modules/formBuilder/components/itemEditor/pickLists/PickListOption.tsx
```diff
@@ -0,0 +1,111 @@
...
+          <Stack direction='row' gap={1}>
+            <ControlledTextInput
+              control={control}
+              name={`pickListOptions.${index}.groupLabel`}
+              label='Group Label'
+              helperText='Label for group that this choice belongs to, if grouped'
+            />
+            <ControlledTextInput
+              control={control}
+              name={`pickListOptions.${index}.groupCode`}
```
martha: I think that it would be nice for v2 to provide a better UI for managing choice groups in the UI, maybe with some help from design first. For v1, I think this is ok.

It looks like we use groups for 2 purposes:
- in dropdowns, we visually arrange choices under groups
- we can use the groupCode in conditional logic like enableWhen, but it looks like that's not currently turned on in the form builder because it has only one use case? https://github.com/greenriver/hmis-frontend/blob/c7482bfacf0ec245c06ebcd1a8dd4f55cb1294fd/src/modules/formBuilder/components/itemEditor/conditionals/EnableWhenCondition.tsx#L77 Does this mean we should disable groupCode here too?

src/modules/formBuilder/components/itemEditor/pickLists/ManagePickListOptions.tsx
```diff
@@ -0,0 +1,104 @@
...
+      if (!fieldsWatch) return;
+      const newFields = fieldsWatch.map((field) => {
+        return { ...field, initialSelected: field?.code === code };
+      });
+      replace(newFields);
+    },
+    [replace, fieldsWatch]
+  );
+
+  const isCodeUnique = useCallback(
```
martha: Just noting this uniqueness check is not validated in the backend schema

src/modules/formBuilder/components/itemEditor/pickLists/PickListOption.tsx
```diff
@@ -0,0 +1,111 @@
...
+        name={`pickListOptions.${index}.code`}
+        label='Code'
+        helperText='Must be unique'
+        required={true}
+        rules={{
+          validate: (input) => isCodeUnique(input),
+        }}
+      />
+      <ControlledTextInput
+        control={control}
```
martha: Originally I had wanted to hide a lot of these options under an "advanced" slideout/collapse panel. It's a bit complicated though, because using `replace` (which we do in the onChangeInitialSelected callback) causes the whole thing to re-render, so we would need to either remove that logic, or keep track in a separate piece of state of which advanced panels were open.

I'm ok either way for v1, let me know your preference.

src/modules/formBuilder/components/itemEditor/pickLists/ManagePickListOptions.tsx
```diff
@@ -0,0 +1,104 @@
...
+      <ControlledSelect
+        name='pickListReference'
+        control={control}
+        label='Or, use a reference list for choices'
+        placeholder='Select list'
+        options={pickListTypesPickList}
+        onChange={(value) => {
+          if (value) replace([]);
+        }}
+      />
```
gigxz: Passing on onChange to the controlled RHF component feels like a bit of an anti pattern.
Another way to do this would be to `watch` the `pickListReference` value, and call `replace([])` if it has a value. That's the pattern [we use elsewhere](https://github.com/greenriver/hmis-frontend/blob/9fe5e55a56903ac4f6e89f15c63be030c924d315/src/modules/formBuilder/components/itemEditor/FormEditorItemProperties.tsx#L139-L145), so maybe we should stick to it for consistency.

src/modules/formBuilder/components/itemEditor/pickLists/PickListOption.tsx
gigxz: Thanks for making a ticket!

src/components/elements/input/RadioGroupInputOption.tsx
gigxz: There is certainly complexity in maintaining both of these orientations. Moving the conversation [to figma](https://www.figma.com/design/F9EHR2IZYuMAlOJ9aOG2IL?node-id=2867-9260#878133179) to propose simplifying to a single orientation (vertical) to reduce maintenance burden. Maybe you can pull this part out to a different branch while we sort that out?

src/modules/formBuilder/components/itemEditor/pickLists/ManagePickListOptions.tsx
martha: Okay! This works for removing the pick list options when a pick list reference is chosen. It does not quite work for nulling out the `initialSelected` values when one option's `initialSelected` is checked. I think this is because calling `replace` inside the `useEffect` with a non-empty list causes a rerender, which triggers `useEffect` again, and it creates an infinite loop.

One workaround I thought of is to use a custom validator instead of calling `replace`, similar to the custom validator `isCodeUnique`. The downside of this is that it does less work for the user and instead forces them to fix the error themselves, but maybe this is better anyway? Edit: I tried implementing this workaround and it's not quite right, either. The validation gets stale because it is only run when that field changes, so it doesn't invalidate or re-validate when other options' `initialSelected` value changes.

I was shying away from `useEffect` because of https://react.dev/learn/you-might-not-need-an-effect. Since this is ultimately a response to a user-generated event, I thought it would be more of an antipattern to use useEffect, and passing an additional callback would be preferable. Maybe the name of the callback could be changed from `onChange` to something else? Let me know what you think!

Edit: One more idea is to implement your suggestion, but try using `setValue` on each individual item's `initialSelected` in the whole list of options, instead of `replace` on the whole list. I think this will avoid the infinite loop.

src/modules/formBuilder/components/itemEditor/pickLists/PickListOption.tsx
```diff
@@ -0,0 +1,111 @@
...
+        label='Label'
+        helperText='If Label is not provided, Code will be displayed'
+      />
+
+      <ControlledCheckbox
+        name={`pickListOptions.${index}.initialSelected`}
+        control={control}
+        label='Initially selected'
+        helperText='Whether this choice is selected by default'
+        onChange={onChangeInitialSelected}
```
gigxz: Same here, could watch `pickListOptions.${index}.initialSelected` and call `onChangeInitialSelected` in a useEffect. I think it is a little more explanatory, because it really is happening as a _side effect_ of the change that occurred. Supplying an `onChange` prop for these controlled components is confusing imo since they actually don't support being controlled from anything other than RHF.

src/modules/formBuilder/components/itemEditor/pickLists/PickListOption.tsx
```diff
@@ -0,0 +1,111 @@
...
+      />
+
+      {formItemComponent !== Component.Dropdown && (
+        // Helper text is only supported for radio/checkbox, not dropdown options
+        <ControlledTextInput
+          control={control}
+          name={`pickListOptions.${index}.helperText`}
+          label='Helper text'
+          helperText='Helper text, including html'
```
gigxz: Maybe:
```suggestion
          helperText='Helper text (may contain HTML)'
```

src/modules/formBuilder/components/itemEditor/pickLists/PickListOption.tsx
gigxz: Yeah, I think this is good for v1! There is only 1 specific use case for conditional based on the group code (prior living situation). However, I could imagine plenty of situations where you would want to visually group the choices to separate them into groups. So, yes, I think we should support it here.

src/modules/formBuilder/components/itemEditor/pickLists/PickListOption.tsx
```diff
@@ -0,0 +1,111 @@
...
+      )}
+
+      <ControlledTextInput
+        control={control}
+        name={`pickListOptions.${index}.numericValue`}
+        label='Numeric value'
+        helperText='Numeric value, such as a score, used for comparison in conditional logic'
+        rules={{
+          validate: (input) => !input || /^-?\d+$/.test(input),
+        }}
```
gigxz: In other places we use `type='number'` for `ControlledTextInput`, which restricts it using the html type. Debatable whether that's the best way to do numeric inputs, but we should probably be consistent.

src/modules/formBuilder/components/itemEditor/pickLists/PickListOption.tsx
martha: Sounds good, yes I definitely think `groupLabel` should be supported for v1. I was asking here specifically about `groupCode`, which isn't currently supported in the form builder's EnableWhen manager. (it seems to me that `groupCode` and `groupLabel` are operating independently of one another.) However, I'm happy to leave both if you prefer.

src/modules/formBuilder/components/itemEditor/FormEditorItemProperties.tsx
```diff
@@ -105,9 +106,36 @@ const FormEditorItemProperties: React.FC<FormEditorItemPropertiesProps> = ({
...
+        !pickListReferenceValue
+      );
+    }
+  }, [
+    itemTypeValue,
+    itemComponentValue,
+    repeatsValue,
+    pickListOptionsValue,
+    pickListReferenceValue,
+  ]);
```
gigxz: I would probably move this into the `ManagePickListOptions` component, since it's specific to pick-lists and choice options.

src/modules/formBuilder/components/itemEditor/FormItemEditor.tsx
```diff
@@ -32,6 +32,7 @@ const FormItemEditor: React.FC<FormItemEditorProps> = ({
     // TODO: can transform server errors into RHF FieldErrors so they get inlined into the form
     // https://react-hook-form.com/ts#FieldErrors
     // errors: errorState?.errors,
+    mode: 'onChange', // allows custom RHF validator functions to run when fields change
```
gigxz: Seems fine to me, do you notice a performance impact?
> Warning: this often comes with a significant impact on performance.

src/components/elements/input/RadioGroupInputOption.tsx
martha: Sure sounds good. https://github.com/greenriver/hmis-frontend/commit/f907310a5b0279cb7ff50cfa86e3ba72775e00d2

src/modules/formBuilder/components/itemEditor/pickLists/PickListOption.tsx
gigxz: Oh, I see. leaving both seems fine for this first pass. Thanks!

src/modules/formBuilder/components/itemEditor/pickLists/ManagePickListOptions.tsx
gigxz: Thanks for flagging. I think it is sufficient to validate on the frontend.

src/modules/formBuilder/components/itemEditor/pickLists/PickListOption.tsx
gigxz: yeah fair enough!! sorry the "initial selected" caused some complexity and it's not even working :\ I can't say confidently whether it's used anywhere or not.

src/modules/formBuilder/components/itemEditor/FormItemEditor.tsx
martha: I didn't notice a performance impact, I think it still seems okay

src/modules/formBuilder/components/itemEditor/pickLists/ManagePickListOptions.tsx
gigxz: Hmm, ok. I think I was wrong, passing an onChange handler is fine. It doesn't accept a `value`, so in that sense it's pretty clear that it can't act as a controlled component. (Ugh, maybe naming them ControlledSelect etc was confusing...since they really masquerade as "uncontrolled" from the perspective of the caller...)

Using validate also seems like a good solution to me, even if the validation only happens on save.

martha: Ok, I've pushed a change implementing this as an `afterChange` callback.

But, since this is so much effort and introducing a new pattern all for the sake of a little checkbox that isn't even working correctly and we aren't sure if there is a use case for it... Maybe we should:
- Remove this `afterChange` callback from the controlled checkbox
- Change the [validation mode](https://github.com/greenriver/hmis-frontend/pull/850#discussion_r1681404064) back to onSubmit instead of onChange
- Remove any onChange behavior on the initialSelected checkbox
- Validate on submit that only one option has initialSelected checked
- And as a bonus, this would eliminate the barrier [here](https://github.com/greenriver/hmis-frontend/pull/850#discussion_r1681105701) to implementing an 'advanced' collapse panel. It's all connected lol

Update: I've made these updates and I'm much happier with it! Wish I'd come around to this way sooner.

src/modules/formBuilder/components/itemEditor/pickLists/PickListOption.tsx
martha: Ok update, after [these changes](https://github.com/greenriver/hmis-frontend/pull/850#discussion_r1681533408) implementing the collapsible advanced panel turned out to be easy. Let me know if you think it looks okay -- maybe as currently implemented (with an icon button) it's too unobtrusive? I've also tried implementing it as a SimpleAccordion, (commit on separate branch https://github.com/greenriver/hmis-frontend/commit/eca9e231ee36a751fd348313fa4689ec414a45a9). It's nice to reuse another pattern we have elsewhere, but it's pretty bulky, which loses part of the ux benefit.

src/modules/formBuilder/components/itemEditor/pickLists/ManagePickListOptions.tsx
gigxz: Alright this looks great, thanks @martha!!!

gigxz (approved):
nice 👍 💯

src/components/elements/input/RadioGroupInputOption.tsx
gigxz: https://github.com/open-path/Green-River/issues/6378

PR #860, opened 2024-07-30
6369 hmis service category type rules

ttoomey:
## Description

[GH issue
](https://github.com/open-path/Green-River/issues/6369)

## Type of change


src/modules/admin/components/formRules/NewFormRuleDialog.tsx
```diff
@@ -219,9 +244,28 @@ const NewFormRuleDialog: React.FC<Props> = ({
         New rule for: <b>{formTitle}</b>
       </DialogTitle>
       <DialogContent>
-        <Stack gap={2} direction='row' sx={{ mt: 2, mb: 1, display: 'flex' }}>
-          <FormRuleLabelTypography>Applies to</FormRuleLabelTypography>
```
ttoomey, line 223: this didn't seem accessible since the label isn't associated with the input. Changed to use FormRuleSelectField

src/modules/admin/components/formRules/NewFormRuleDialog.tsx
```diff
@@ -128,6 +130,25 @@ const NewFormRuleDialog: React.FC<Props> = ({
     },
   });
 
+  // This would be better handled with real form validation
```
ttoomey, line 133: I feel like it would be better to use rhf for forms like this that aren't "dynamic" but that would require some refactoring.

src/modules/admin/components/formRules/NewFormRuleDialog.tsx
```diff
@@ -128,6 +130,25 @@ const NewFormRuleDialog: React.FC<Props> = ({
...
+  const [validationError, setValidationError] = useState<string>();
+  const handleSubmit = () => {
+    let error: string | undefined = undefined;
+    if (formRole === FormRole.Service) {
+      if (rule.serviceCategoryId && rule.serviceTypeId) {
```
ttoomey, line 138: I'm reading the "type OR category" as "XOR" but perhaps we have some case where someone would set both?

src/modules/admin/components/formRules/NewFormRuleDialog.tsx
martha: Hm that's a good point about accessibility. Your concern also applies to the other places in this modal where a pick list's label changes dynamically with another picklist (the other places where we use FormRuleLabelTypography). Would it be possible to update FormRuleSelectField to keep this design, but make it more accessible perhaps by using a dynamic aria label? Or is this pattern just inaccessible by default and there's no saving it? If so, I think we should go back to the design team and ask for revisions.

Designs here https://www.figma.com/board/axGI9vwmuBRC4uZ11ZJ3RP/OP-%2F-HMIS-%2F-5792-%3A-Form-Builder-Workflow?node-id=239-7978&t=f3pwwMGHLq7moNRr-0

src/modules/admin/components/formRules/NewFormRuleDialog.tsx
martha: Yes I agree with you that it is xor.

When we wrote the ticket I think we were envisioning that this would be another horizontal pair of dropdowns, similar to the Conditions you can add below. In the first dropdown you would select Service Type or Service Category, and in the second dropdown you would pick from a dynamic pick list whose contents change depending on the value of the first dropdown. 

That would eliminate the need for an error validation step since the interface would inherently prevent the user from picking both service type and service category.

I think it would be nice to use this design consistently, but I also agree with your concern about the accessibility of this pattern, so let's talk about it with design if we can't figure out a more accessible implementation.

src/modules/admin/components/formRules/NewFormRuleDialog.tsx
martha: I played around with it a little and came up with a demo of this idea of the dynamic aria label: https://github.com/greenriver/hmis-frontend/commit/ec0353e4c790e028e6f66a5e85561870167ea600

Sorry it's quite messy and not really functional at the moment, I don't want to put more time into cleaning it up unless you agree that this could be an okay approach!

martha: We talked about this briefly at the design session today, design team seemed to lean towards using dynamic aria labels and keeping the design.

ttoomey: Hi @martha - thanks for working up a demo! Sorry I missed the last part of the design call. I put a little time on the calendar tomorrow to check in on this if that's okay.

martha: We discussed on a call, and decided that the current implementation is fine for an MVP of this task, and we will create a follow-up ticket to improve both the accessibility and design for this form.

martha (approved):

PR #861, opened 2024-07-30
Add Staff Assignment column to Project Enrollments and Households tables

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6336
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4582

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/projects/components/tables/ProjectClientEnrollmentsTable.tsx
```diff
@@ -146,6 +148,15 @@ export const ENROLLMENT_COLUMNS: {
       return null;
     },
   },
+  assignedStaff: {
+    ...ASSIGNED_STAFF_COL,
+    render: (e) => {
+      if ('household' in e && isRenderFunction(ASSIGNED_STAFF_COL.render)) {
```
ttoomey: So it looks like isRenderFunction casts ASSIGNED_STAFF_COL.render to be `(e:any)=> React.ReactNode`. This seems like an indirect way of just casting e to `any`.

Suggest we either cast e in a more obvious way, or use a guard function. So:

```ts
    render: (e:any) => {
      if ('household' in e){
        return ASSIGNED_STAFF_COL.render(e.household);
      }
      return null;
    },
```

or


```ts
//guard function
const isHouseholdWithStaff = (e: any): e is {household: HouseholdWithStaffAssignmentsFragment} => {
  return 'household' in e && 'staffAssignments' in e.household
}
// ....

    render: (e) => {
      return isHouseholdWithStaff(e) ? ASSIGNED_STAFF_COL.render(e.household) : null
    }
```

src/modules/projects/components/tables/ProjectClientEnrollmentsTable.tsx
```diff
@@ -178,20 +189,33 @@ const ProjectClientEnrollmentsTable = ({
...
   );
 
+  const {
+    project: { staffAssignmentsEnabled },
+  } = useProjectDashboardContext();
+
   const defaultColumns: ColumnDef<ProjectEnrollmentQueryEnrollmentFieldsFragment>[] =
     useMemo(() => {
-      return [
+      const c = [
```
ttoomey: I know this is a small function but suggest we use more descriptive variable names, such as `cols` or just `results`

src/modules/projects/components/tables/ProjectHouseholdsTable.tsx
```diff
@@ -32,6 +34,27 @@ const TableCellContainer = ({ children }: { children: ReactNode }) => (
...
 
+export const ASSIGNED_STAFF_COL = {
+  header: 'Assigned Staff',
+  optional: true,
+  defaultHidden: true,
+  render: (hh: HouseholdWithStaffAssignmentsFragment) => {
+    if (!hh?.staffAssignments || hh.staffAssignments.nodes.length === 0) return;
+
+    const first = hh.staffAssignments.nodes[0].user.name;
+    const rest = hh.staffAssignments?.nodesCount - 1;
```
ttoomey: micro optimization, I think this is a little more concise
```suggestion
    if (!hh.staffAssignments?.nodes.length) return;

    const first = hh.staffAssignments.nodes[0].user.name;
    const rest = hh.staffAssignments.nodesCount - 1;
```

ttoomey (approved):
looks good! I left a few optional suggestions

src/modules/projects/components/tables/ProjectClientEnrollmentsTable.tsx
martha: I like the idea of a guard function. Thanks for the suggestion @ttoomey !

PR #863, opened 2024-08-01
Make first pass at toolbar mobile usability

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6433
See latest design thinking: https://greenriver.slack.com/archives/C033YNSN0Q1/p1722449528561929

I have several worries about this approach, but it is more functional that the current mobile toolbar especially with the recent addition of 'My Dashboard' as a fourth tab.

Worries/things to improve:
- Is there a better alternative to the way I've injected the dashboard navigational elements into the mobile menu? For example, would it be better if we have just 1 place where the MobileMenu component is rendered, and make it learn about what children nav elements it should render using path parameters directly?
- Can we make a more consistent treatment between the different types of NavItems? (user nav, toolbar nav, and dashboard nav)
- Can we improve the mobile experience for search? (mui's Autocomplete component seems to be very geared towards popover behavior; would it be better to remove reliance on Autocomplete entirely on mobile?)
- There are some weird bugs with scroll behavior that we should investigate more

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/components/layout/MainLayout.tsx
+++ b/src/components/layout/MainLayout.tsx
@@ -1,15 +1,7 @@
-import {
-  AppBar,
-  Box,
-  CssBaseline,
-  SxProps,
-  Theme,
-  Toolbar,
-} from '@mui/material';
+import MenuIcon from '@mui/icons-material/Menu';
+import { AppBar, Box, CssBaseline, IconButton, Toolbar } from '@mui/material';
+import { Stack } from '@mui/system';
 import * as React from 'react';
-
-import { useLocation } from 'react-router-dom';
-import ButtonLink from '../elements/ButtonLink';
 import RouterLink from '../elements/RouterLink';
 
 import {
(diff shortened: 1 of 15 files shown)


ttoomey (commented):
This is looking like a big improvement, thanks Martha!

I fiddled about with this branch and made some adjustments / improvements. It seemed easer to just make a PR:
https://github.com/greenriver/hmis-frontend/pull/866
( feel free to ignore or cherry pick out anything useful)

* I think injecting the sidebar components into the mobilenav seems fine for now, it's nice that it re-uses the current sidebars. But I agree we might want to explore other approaches, perhaps calling our customHooks from within the mobileMenu(?)
* Probably should ensure that the nave follows a11y standards (use a <nav> element?)
* I agree the current autocomplete isn't ideal in mobile, maybe worth writing a mobile version, particularly for search (aside, I think it would feel better to have the search results render in the sidebar, rather than in a popover)

PR #866, opened 2024-08-03
Tt/6433 toolbar

ttoomey:
Some thoughts on 6433

* there are two variants, expanded with a scrolling region, and collapsed where just the top-level items are shown
* fix max-width on drawer
* fix scrolling behavior in drawer.
* roll up user menu
* move search above nav items (maybe that's better?)

Screenshot 2024-08-03 at 01 38 12
Screenshot 2024-08-03 at 01 38 37

### DIFF
--- a/src/components/layout/MainLayout.tsx
+++ b/src/components/layout/MainLayout.tsx
@@ -129,7 +129,7 @@ const MainLayout: React.FC<Props> = ({ mobileMenuContext, children }) => {
                 // Dashboards render their own Mobile Menus with additional nav elements inside.
                 <MobileMenu
                   mobileNavIsOpen={mobileNavIsOpen}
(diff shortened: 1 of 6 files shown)


martha (approved):
Thanks @ttoomey  this looks much better to me. I would be fine with merging this into my branch and then merging that to release soon, on the premise that now that we're adding 'My Dashboard,' some kind of mobile usability improvement is needed sooner rather than later.

I think what we have now is still pretty different from the latest design thinking around mobile nav (https://greenriver.slack.com/archives/C033YNSN0Q1/p1722457826488179?thread_ts=1722449528.561929&cid=C033YNSN0Q1), but since I think those designs warrant more thought and discussion, maybe it would be better to split that work into a separate, lower-priority ticket. What do you think?

ttoomey:
Good to merge, thanks!

PR #867, opened 2024-08-05
Geolocation hook and wip component

gigxz:
## Description

Proof-of-concept component for requesting geolocation using [Geolocation API](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation_API).

I wired this up for testing it out, figured we might as well include it in the codebase. Currently it's only used in the Story.

Related issues: https://github.com/open-path/Green-River/issues/5726, https://github.com/open-path/Green-River/issues/5896
## Type of change

### DIFF
--- a/src/components/elements/input/GeolocationInput.tsx
+++ b/src/components/elements/input/GeolocationInput.tsx
@@ -1,5 +1,5 @@
 import { Typography } from '@mui/material';
-import React from 'react';
+import React, { ReactNode } from 'react';
 import { useGeolocation } from '../../../hooks/useGeolocation';
 import LoadingButton from '../LoadingButton';
 
@@ -8,13 +8,18 @@ import LoadingButton from '../LoadingButton';
  * - Needs to be wired up with Props to be used as an Input component.
  * - Likely needs to use a mapping API to display the coordinates on a map.
  */
-const GeolocationInput = () => {
+
+interface GeolocationInputProps {
+  children?: ReactNode;
+}
+
+const GeolocationInput: React.FC<GeolocationInputProps> = ({ children }) => {
   const { coordinates, requestCoordinates, loading, error } = useGeolocation();
 
   return (
     <>
       <LoadingButton onClick={requestCoordinates} loading={loading}>
-        Request Location
+        {children || 'Request Location'}
       </LoadingButton>
       {error && <Typography color='error'>{error}</Typography>}
       <div>
--- a/src/components/elements/input/GeolocationInput.tsx
+++ b/src/components/elements/input/GeolocationInput.tsx
@@ -1,0 +1,39 @@
+import { Typography } from '@mui/material';
+import React from 'react';
+import { useGeolocation } from '../../../hooks/useGeolocation';
+import LoadingButton from '../LoadingButton';
+
+/**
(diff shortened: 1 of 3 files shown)


martha (approved):
Cool!
May I ask how you tested this locally? Even with VPN turned off, I'm not finding the location component gets my coordinates accurately, but, neither does Google Maps.

src/hooks/useGeolocation.tsx
```diff
@@ -0,0 +1,51 @@
...
+          break;
+        default:
+          setError('An unknown error occurred.');
+          break;
+      }
+    };
+
+    if (navigator.geolocation) {
+      setLoading(true);
+      navigator.geolocation.getCurrentPosition(handleSuccess, handleError);
```
ttoomey, line 38: I'm not familiar with this api but [looks like there's a third options param](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation/getCurrentPosition). For example:
```
const options = {
  timeout: 30 * 1000, // timeout after 30 seconds
  maximumAge: 5 * 60 * 1000, // use a cached result from 5 minutes ago
  // enableHighAccuracy: true
};
navigator.geolocation.getCurrentPosition(handleSuccess, handleError, options);

```

src/components/elements/input/GeolocationInput.tsx
```diff
@@ -0,0 +1,44 @@
...
+          <img
+            alt='Map Holder'
+            src={`https://maps.googleapis.com/maps/api/staticmap?center=${location.latitude + ',' + location.longitude}&zoom=16&size=400x300&output=embed`}
+          />
+        )} */}
+        {coordinates && (
+          <Typography>
+            Latitude: {coordinates.latitude}
+            <br />
+            Longitude: {coordinates.longitude}
```
ttoomey, line 36: could show accuracy too? Maybe a threshold for usable accuracy (like 500 meters)
```
Accuracy: {coordinates.accuracy}
```

ttoomey (approved):
Looks good! 

I guess the plan is to use this in combo with a map, so if your geolocation isn't working, or you're back filling data from another location you could drop a pin or enter an address?

gigxz:
> May I ask how you tested this locally? Even with VPN turned off, I'm not finding the location component gets my coordinates accurately, but, neither does Google Maps.

I tested by running storybook and clicking "Request location" on the GelocationInput component. It requests location from my browser, and then shows the latitude and longitude on the page. That's interesting, I'm not sure why you might be getting inaccurate coordinates-- but its definitely the case that this api is more accurate on mobile, so maybe that has something to do with it.


> I guess the plan is to use this in combo with a map, so if your geolocation isn't working, or you're back filling data from another location you could drop a pin or enter an address?

Yes on no, may be as a phase 2 for so. [some more context here](https://docs.google.com/document/d/1mlh91wDHOZ81w5RFYRERYP8izXvPRMqWq9cf9XZwkS8/edit#heading=h.83tgztf3bii9)

src/components/elements/input/GeolocationInput.tsx
gigxz: We're probably going to want to show a map here, not the coordinates, this is just a placeholder!

PR #871, opened 2024-08-05
Show overall household status, not HoH's status

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6337

## Type of change
- [x] Bug fix
- [ ] Dependency update


graphql.schema.json
```diff
@@ -12,6 +12,7 @@
         "kind": "ENUM",
         "name": "ActiveStatus",
         "description": null,
+        "isOneOf": null,
```
gigxz: opened https://github.com/greenriver/hmis-frontend/pull/872 for this unrelated change

src/modules/hmis/components/HouseholdStatus.tsx
```diff
@@ -0,0 +1,30 @@
+import { memo } from 'react';
+import EnrollmentStatus from '@/modules/hmis/components/EnrollmentStatus';
+import { HouseholdWithEnrollmentFieldsFragment } from '@/types/gqlTypes';
+
+interface Props {
+  household: HouseholdWithEnrollmentFieldsFragment;
+}
+
+const HouseholdStatus: React.FC<Props> = ({ household }) => {
+  // When showing the status for the household, show the 'most important' status - Incomplete, Open, or Exited.
```
gigxz: This duplicates logic that we already have in the [household view](https://github.com/greenriver/hmis-warehouse/blob/f4a3f8ca1712bd50f53506c178ebb68a2ec59444/db/views/hmis_households_v07.sql#L19). Sorry I did not mention that we already have it! 

I think it would be best to re-use that logic by having the household type resolve an `in_progress` bool (or `any_in_progress`?) that is resolved as `object.any_wip?`. It wouldn't hurt to resolve the other existing fields as well, in case we need them later:
```
household {
   anyInProgress // any_wip
   earliestEntryDate // earliest_entry
   latestExitDate // latest_exit
}
```

graphql.schema.json
martha: Thanks! 👍

src/modules/hmis/components/EnrollmentStatus.tsx
```diff
@@ -2,75 +2,88 @@ import ErrorOutlineIcon from '@mui/icons-material/ErrorOutline';
...
-        {!hideIcon && <Icon fontSize='small' />}
-        {text}
+        <statusProps.Icon fontSize='small' />
+        {statusProps.text}
       </Stack>
     </Typography>
   );
 };
 
+const EnrollmentStatus = ({
```
martha, line 57: @gigxz this refactor removes some customizability that we weren't using, as far as I could tell:
- hideIcon
- withActiveRange
- activeColor
- closedColor

all were unused props. I can add back support for them if you prefer, let me know.

src/modules/hmis/components/HouseholdStatus.tsx
martha: Good plan, thanks for this suggestion!!

src/modules/hmis/components/EnrollmentStatus.tsx
gigxz: great, thanks! no need to keep them around

src/modules/hmis/components/EnrollmentStatus.tsx
```diff
@@ -2,75 +2,88 @@ import ErrorOutlineIcon from '@mui/icons-material/ErrorOutline';
...
+
+export const HouseholdStatus = ({
+  household,
+}: {
+  household: HouseholdWithEnrollmentFieldsFragment;
+}) => {
+  if (household.anyInProgress) return <CommonStatus variant='inProgress' />;
+  if (!household.latestExitDate) return <CommonStatus variant='open' />;
+  return <CommonStatus variant='exited' />;
+};
```
gigxz, line 76: Nice improvement

src/api/operations/staffAssignment.fragments.graphql
```diff
@@ -40,11 +40,9 @@ fragment StaffAssignmentWithClients on StaffAssignment {
   assignedAt
   unassignedAt
   household {
-    id
+    ...HouseholdWithEnrollmentFields
```
gigxz: This looks like it resolves ssn and dob (and access object) for all household members which we don't need for staff assignment. I think we should avoid that since it's not necessary in this case and it is highly sensitive.

Maybe just inline the fields needed for staff assignment?

src/modules/hmis/components/EnrollmentStatus.tsx
```diff
@@ -2,75 +2,88 @@ import ErrorOutlineIcon from '@mui/icons-material/ErrorOutline';
...
+  if (enrollment.inProgress) return <CommonStatus variant='inProgress' />;
+  if (enrollment.autoExited) return <CommonStatus variant='autoExited' />;
+  if (!enrollment.exitDate) return <CommonStatus variant='open' />;
+  return <CommonStatus variant='exited' />;
+};
+
+export const HouseholdStatus = ({
+  household,
+}: {
+  household: HouseholdWithEnrollmentFieldsFragment;
```
gigxz: Might be better to loosen these constraints a bit, since only 2 fields are needed. Then we can use this component in other places and not worry about type issues.

```suggestion
  household: Pick<Household, 'anyInProgress' | 'latestExitDate'>;
```
The same could be done for `EnrollmentStatus` component above as well. I think that approach would be more maintainable than jamming in all the fragments we happen to use. (I know, we do that in a lot of places)

graphql.schema.json
```diff
@@ -12,7 +12,6 @@
         "kind": "ENUM",
         "name": "ActiveStatus",
         "description": null,
-        "isOneOf": null,
```
gigxz: Can you try running `yarn` and then codegen again? I think these should be present.

gigxz (approved):
LGTM after fixing codegen changes!

PR #876, opened 2024-08-07
Remove support for horizontal checkboxes and radio groups

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6378
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4605

This PR
- makes codegen updates in accordance w/ the linked backend PR to deprecate RADIO_BUTTONS_VERTICAL
- removes support for the `row` prop in `DynamicField`, so all checkboxes and radio groups will now be rendered vertically
- removes `RadioButtonsVertical` from the list of valid component overrides, so it doesn't show in the dropdown

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/components/elements/input/RadioGroupInput.tsx
+++ b/src/components/elements/input/RadioGroupInput.tsx
@@ -64,7 +64,6 @@ const RadioGroupInput = ({
   value,
   error,
   warnIfEmptyTreatment,
-  row, // Not supported by dynamic forms anymore, but used internally by YesNoRadio. Support is partial (e.g. helper text is not supported)
   sx,
   clearable,
   helperText,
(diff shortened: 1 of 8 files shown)


src/components/elements/input/RadioGroupInput.tsx
```diff
@@ -64,7 +64,7 @@ const RadioGroupInput = ({
   value,
   error,
   warnIfEmptyTreatment,
-  row,
+  row, // Not supported by dynamic forms anymore, but used internally by YesNoRadio. Support is partial (e.g. helper text is not supported)
```
martha: @gigxz let me know your thoughts on leaving this in place.. I would also be happy to fully remove all support for the `row` prop so that YesNoRadio, TableFilterItemCheckboxes etc. also render vertically.

gigxz: I say lets go ahead and remove it entirely.

martha: Sounds good, done. I think this looks fine on the table filters, but are you at all concerned about the impact in the form builder? We use this in EnableWhen in a design that I think assumed the horizontal checkbox layout:

Screenshot 2024-08-07 at 3 42 54 PM

gigxz: Hmm good point, I think it could be resolved by switching to a dropdown there. Lets just table it for now pedning the next design session around conditionals, which I think is on the docket for next week

gigxz (approved):

src/components/elements/input/RadioGroupInput.tsx
martha: Sounds good, I'll go ahead and merge, and we can discuss next week.

PR #877, opened 2024-08-07
Clear autofilled value when autofill condition turns false

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/5954
This PR enacts the behavior described in this comment: https://github.com/open-path/Green-River/issues/5954#issuecomment-2273975433

To test, I made a form with relevant autofill conditions in QA: https://qa-hmis.openpath.host/admin/forms/marthas_custom_assessment/165/preview-draft

You can copy that form's JSON to a local form, and play around with the behavior on this branch locally.

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/modules/form/util/formUtil.ts
+++ b/src/modules/form/util/formUtil.ts
@@ -528,7 +528,7 @@ export const autofillValues = ({
   if (!item.autofillValues) return false;
 
   // use `some` to stop iterating when true is returned
-  return item.autofillValues.some((av) => {
+  const autofilled = item.autofillValues.some((av) => {
     // Skip autofills that should not run on read-only views
     if (viewOnly && !av.autofillReadonly) return false;
 
@@ -567,6 +567,17 @@ export const autofillValues = ({
     // Stop iterating through autofill values
     return true;
   });
+
+  // For readonly items, if no autofill condition matched, clear out the item's value.
+  // Don't make any changes if the whole form is in viewOnly mode, though, because
+  // in that case this is being called on form load (not on change of an autofill condition),
+  // so we don't want to clear any form values.
+  if (!autofilled && item.readOnly && !viewOnly) {
+    values[item.linkId] = undefined;
+    return true;
+  }
+
+  return autofilled;
 };
 
 const getBoundValue = (
--- a/src/modules/form/util/formUtil.ts
+++ b/src/modules/form/util/formUtil.ts
@@ -568,11 +568,14 @@ export const autofillValues = ({
     return true;
   });
 
-  // For readonly items, if no autofill condition matched, clear out the item's value.
-  // Don't make any changes if the whole form is in viewOnly mode, though, because
-  // in that case this is being called on form load (not on change of an autofill condition),
-  // so we don't want to clear any form values.
-  if (!autofilled && item.readOnly && !viewOnly) {
+  // For read-only items that are displayed on editable forms, the autofill value should nullify when its conditions are no longer met.
+  // For example, a read-only field showing the sum of 2 input fields should be cleared if the inputs are cleared.
+  // (In that example, we assume the autofill rule for summing has an autofill_when condition requiring the inputs to be present)
+  if (
+    !autofilled &&
+    (item.readOnly || item.type === ItemType.Display) &&
+    !viewOnly
+  ) {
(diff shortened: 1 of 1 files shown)


src/modules/form/util/formUtil.ts
```diff
@@ -567,6 +567,17 @@ export const autofillValues = ({
...
   });
+
+  // For readonly items, if no autofill condition matched, clear out the item's value.
+  // Don't make any changes if the whole form is in viewOnly mode, though, because
+  // in that case this is being called on form load (not on change of an autofill condition),
+  // so we don't want to clear any form values.
+  if (!autofilled && item.readOnly && !viewOnly) {
+    values[item.linkId] = undefined;
+    return true;
+  }
```
gigxz: I think the "read only items" needs to be defined as `item.readOnly || item.type == 'display', unfortunately.. We have items like the below-- can you add something like this to the test form to ensure it works as expected?

```
                {
                  "text": "<b>Duration</b>: ${FORMAT_DURATION(value)}",
                  "type": "DISPLAY",
                  "link_id": "ServiceCode_1_duration",
                  "autofill_values": [
                    {
                      "formula": "ABS(ServiceCode_1_start - ServiceCode_1_end)",
                      "autofill_when": [
                        {
                          "operator": "EXISTS",
                          "question": "ServiceCode_1_start",
                          "answer_boolean": true
                        },
                        {
                          "operator": "EXISTS",
                          "question": "ServiceCode_1_end",
                          "answer_boolean": true
                        }
                      ],
                      "autofill_behavior": "ALL",
                      "autofill_readonly": true
                    }
                  ]
                }
```

src/modules/form/util/formUtil.ts
```diff
@@ -567,6 +567,17 @@ export const autofillValues = ({
     // Stop iterating through autofill values
     return true;
   });
+
+  // For readonly items, if no autofill condition matched, clear out the item's value.
+  // Don't make any changes if the whole form is in viewOnly mode, though, because
+  // in that case this is being called on form load (not on change of an autofill condition),
+  // so we don't want to clear any form values.
```
gigxz: Thanks, I think this is right. I tried to word it in another way: `For read-only items that are displayed on editable forms, the autofill value should nullify when it's conditions are no longer met. For example, a read-only field showing the sum of 2 input fields should be cleared if the inputs are cleared. (In that example, we assume the autofill rule for summing has an autofill_when condition requiring the inputs to be present)`. 
We're saying the same thing, right? Might be nice to our future selves to include the example in the code comment.

martha: Yes we are saying the same thing, and I think your comment is much clearer so I replaced mine 😂 thank you!

src/modules/form/util/formUtil.ts
martha: Yes, I added this, and with your suggested change it appears to be behaving as we expect on editable forms. I think this might be exposing an unrelated bug ~with the duration component~ with many components(?) on view-only forms but I have to dig a little more to confirm.

martha: Ok, I haven't tracked down the reason for the bug but to reproduce:
- Visit https://qa-hmis.openpath.host/admin/forms/marthas_custom_assessment/165/preview-draft
- Type something into the "Autofill when" field (Group 1)
- Type something into the "just a simple text box" field (Group 3)
- Switch to Read Only View
- Why does "Autofill when" retain its value, but "just a simple text box" does not??

gigxz: Oh, that's a good call out. The items that don't have `mapping`s (yet) don't retain their values when switching between edit/view modes, because the `FormPreview` component does the transformation process that changes the form state from `keyed by link id => keyed by field name => keyed by link id` when switching modes. We could probably get it working without that extra transformation, which would resolve this issue. I'll make a ticket. Thanks for finding!

gigxz: https://github.com/open-path/Green-River/issues/6487

gigxz (approved):

src/modules/form/util/formUtil.ts
martha: Ahaa, that makes sense, thanks for the sleuthing!

PR #879, opened 2024-08-07
Clean up mobile toolbar

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6433

This PR
- Renders the toolbar menu on mobile (rather than hand-mapping the toolbar nav elements) so that we can take advantage of its permission filtering
- In the toolbar menu, uses the new `MobileMenuItem` for consistent sx
- Cleans up comments

I think, while we should still consider mobile nav (or just nav in general - see my comment [here](https://github.com/open-path/Green-River/issues/6475#issuecomment-2273634035)) to be an area of tech/design debt, I'd propose considering this PR as closing out the linked ticket, and moving further work based on TBD designs to a future ticket.

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/components/layout/nav/ToolbarMenu.tsx
+++ b/src/components/layout/nav/ToolbarMenu.tsx
@@ -1,9 +1,9 @@
-import { alpha, lighten, MenuItem, SxProps, Theme } from '@mui/material';
-import React, { useCallback } from 'react';
+import { alpha, Theme } from '@mui/material';
+import React from 'react';
 import { useLocation } from 'react-router-dom';
 import ButtonLink from '@/components/elements/ButtonLink';
-import RouterLink from '@/components/elements/RouterLink';
 import { NavItem } from '@/components/layout/dashboard/sideNav/types';
+import MobileMenuItem from '@/components/layout/nav/MobileMenuItem';
 import { useIsMobile } from '@/hooks/useIsMobile';
 import { PERMISSIONS_GRANTING_ADMIN_DASHBOARD_ACCESS } from '@/modules/admin/components/AdminDashboard';
 import { RootPermissionsFilter } from '@/modules/permissions/PermissionsFilters';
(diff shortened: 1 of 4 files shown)


src/components/layout/nav/ToolbarMenu.tsx
```diff
@@ -65,60 +65,32 @@ export const useActiveNaveItem = () => {
...
         key={item.id}
-      >
-        {item.title}
-      </MenuItem>
+        title={item.title as string}
+        selected={activeItem === item.activeItemPathIncludes}
+        path={item.path}
+      />
     ) : (
       <ButtonLink
```
ttoomey, line 81: so since this is rendered within a `List` component, it looks like you end up with `<ul><a>item</a</ul>` - which isn't valid. We could take out the ul as a quick fix. Although ideally the markup would be more like `<nav><ul><li><a>item</a</li></ul></nav>` for accessibility

src/components/layout/nav/ToolbarMenu.tsx
```diff
@@ -65,60 +65,32 @@ export const useActiveNaveItem = () => {
...
         variant='text'
         to={item.path}
         data-testid={item.id}
-        sx={navItemSx(activeItem === item.activeItemPathIncludes)}
+        sx={{
+          fontWeight: 600,
+          fontSize: 14,
```
ttoomey: When we have custom typography, maybe we should start calling out with a FIXME comment so we can clean it up later. Ideally we shouldn't doing this

ttoomey (approved):
Looks great, thanks! I have one additional suggestion: I think we should change the name of the 'User' item in the menu to 'My Account' to be more clear

src/components/layout/nav/ToolbarMenu.tsx
martha: Good point, do you think it's ok to add <li> as a quick fix as I've done here? I agree we should add `nav` too, but when I briefly tried it messed with your styling. We do have an a11y ticket for that somewhere so maybe better to revisit and fix when we have a clearer idea of the long term vision for this nav.

ttoomey: Looks good, thanks for the quick fix!

PR #881, opened 2024-08-08
Make form rules dialog more accessible and consistent

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6474

This PR
- Modifies the `FormRule` component so it exposes a couple of distinct interfaces for different contexts where it might be used.
- Extracts a new `FormRuleCondition` component, basically comprising 2 dropdowns for Condition Type (e.g. project, organization) and Condition Value (e.g. project ID, organization ID). This component is now used by both regular conditions and service-related conditions. I added labels to both dropdowns to improve accessibility
- Modifies the `NewFormRuleDialog` to store the form state separately and then use an effect to update the form rule. It also now reflects the current state of the form rule so the user can hopefully better understand the changes they are making.

~I'm stuck on a typescript bug which is marked with a todo.~ Otherwise, I'd be very grateful for feedback about both the implementation and changes to design, please do let me know what you think.

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/modules/admin/components/formRules/FormRule.tsx
+++ b/src/modules/admin/components/formRules/FormRule.tsx
@@ -25,54 +25,43 @@ const nonClientFormRoles = [
 
 const nonProjectFormRoles = [FormRole.Organization];
 
-interface Props {
-  rule: FormRuleFieldsFragment | ProjectConfigFieldsFragment;
+interface BaseFormRuleProps {
+  projectType?: string;
+  project?: string;
+  organization?: string;
+  dataCollectedAbout?: string;
+  funder?: string;
+  otherFunder?: string | null;
+  serviceCategory?: string;
+  serviceType?: string;
   formRole?: FormRole;
   actionButtons?: ReactNode;
 }
 
-function isFormRuleFragment(
-  obj: FormRuleFieldsFragment | ProjectConfigFieldsFragment
-): obj is FormRuleFieldsFragment {
-  return obj && obj.hasOwnProperty('funder');
-}
-
-const FormRule: React.FC<Props> = ({ rule, formRole, actionButtons }) => {
-  // Fields that are shared across Project Configs and Form Rules
-  const { projectType, project, organization } = rule;
-
-  // Fields that are specific to Form Rules
-  let dataCollectedAbout;
-  let funder;
-  let otherFunder;
-  let serviceCategory;
-  let serviceType;
-  if (isFormRuleFragment(rule)) {
-    dataCollectedAbout = rule.dataCollectedAbout;
-    funder = rule.funder;
-    otherFunder = rule.otherFunder;
-    serviceCategory = rule.serviceCategory;
-    serviceType = rule.serviceType;
-  }
-
+export const BaseFormRule: React.FC<BaseFormRuleProps> = ({
+  projectType,
+  project,
+  organization,
+  dataCollectedAbout = HmisEnums.DataCollectedAbout.ALL_CLIENTS,
+  funder,
+  otherFunder,
+  serviceCategory,
+  serviceType,
+  formRole,
+  actionButtons,
+}) => {
   const conditions: Record<string, ReactNode> = {};
 
   if (!formRole || !nonProjectFormRoles.includes(formRole)) {
     if (projectType)
       conditions.projectType = (
-        <FormRuleChip
-          label={`Project Type is ${HmisEnums.ProjectType[projectType]}`}
-        />
+        <FormRuleChip label={`Project Type is ${projectType}`} />
       );
     if (project)
-      conditions.project = (
-        <FormRuleChip label={`Project is ${project.projectName}`} />
-      );
+      conditions.project = <FormRuleChip label={`Project is ${project}`} />;
     if (organization)
       conditions.organization = (
-        <FormRuleChip
-          label={`Organization is ${organization.organizationName}`}
-        />
+        <FormRuleChip label={`Organization is ${organization}`} />
       );
     if (otherFunder) {
       conditions.otherFunder = (
@@ -80,9 +69,7 @@ const FormRule: React.FC<Props> = ({ rule, formRole, actionButtons }) => {
       );
     } else if (funder) {
       conditions.funder = (
-        <FormRuleChip
-          label={`Funding Source is ${HmisEnums.FundingSource[funder]}`}
-        />
+        <FormRuleChip label={`Funding Source is ${funder}`} />
       );
     }
   }
@@ -100,9 +87,7 @@ const FormRule: React.FC<Props> = ({ rule, formRole, actionButtons }) => {
         {isServiceForm ? (
           <>
             Collects{' '}
-            <FormRuleChip
-              label={serviceType?.name || serviceCategory?.name || 'Service'}
-            />{' '}
+            <FormRuleChip label={serviceType || serviceCategory || 'Service'} />{' '}
             for{' '}
           </>
         ) : (
@@ -110,13 +95,7 @@ const FormRule: React.FC<Props> = ({ rule, formRole, actionButtons }) => {
         )}
         {isClientForm && (
           <>
-            <FormRuleChip
-              label={
-                dataCollectedAbout
-                  ? HmisEnums.DataCollectedAbout[dataCollectedAbout]
-                  : HmisEnums.DataCollectedAbout.ALL_CLIENTS
-              }
-            />{' '}
+            <FormRuleChip label={dataCollectedAbout} />{' '}
           </>
         )}
         {conditionCount === 0 && (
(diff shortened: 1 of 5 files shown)


src/modules/admin/components/formRules/FormRule.tsx
```diff
@@ -143,4 +122,57 @@ const FormRule: React.FC<Props> = ({ rule, formRole, actionButtons }) => {
...
+export const FormRule: React.FC<FormRuleProps> = ({
+  rule,
+  formRole,
+  actionButtons,
+}) => {
+  return (
+    <BaseFormRule
+      project={rule.project?.projectName}
+      projectType={
+        rule.projectType ? HmisEnums.ProjectType[rule.projectType] : ''
```
gigxz: What's the utility of pulling out the display logic into the parent components (FormRule/ProjectConfigFormRule), as opposed to having BaseFormRule be responsible for displaying values?

If it was a TS concern, an option is:
```
interface BaseFormRuleProps {
  projectType?: ProjectType | null;
}
// in BaseFormRule
label={`Project Type is ${HmisEnums.ProjectType[projectType]}`}

// passing the prop:
projectType={rule.projectType}
```

If `BaseFormRuleProps` continues to accept all strings, it might be worth renaming them for clarity. Eg `projectType=>projectTypeDisplay` `project=>projectName` etc. to clarify.

src/modules/admin/components/formRules/FormRule.tsx
```diff
@@ -143,4 +122,57 @@ const FormRule: React.FC<Props> = ({ rule, formRole, actionButtons }) => {
...
+}
+export const FormRule: React.FC<FormRuleProps> = ({
+  rule,
+  formRole,
+  actionButtons,
+}) => {
+  return (
+    <BaseFormRule
+      project={rule.project?.projectName}
+      projectType={
```
gigxz: I'm seeing "project type is undefined" after selecting a project type:

Screenshot 2024-08-09 at 10 33 25 AM

src/modules/admin/components/formRules/FormRuleCondition.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { PickListOption } from '@/types/gqlTypes';
+
+const FormRuleLabelTypography = ({
+  sx,
+  children,
+}: {
+  sx?: SxProps;
+  children: ReactNode;
+}) => (
+  <Typography variant='body2' sx={{ pt: 4.25, ...sx }}>
```
gigxz: Looks like we lost the `fontWeight: 600` here.

The 4.25 thing is fine, but just a learning-oriented comment: another way it could be done is `py: 1` with  `alignItems='flex-end'` on the Grid container. I think that feels slightly less janky to me but they look pretty much identical.

src/modules/admin/components/formRules/NewFormRuleDialog.tsx
```diff
@@ -245,114 +268,151 @@ const NewFormRuleDialog: React.FC<Props> = ({
...
-                onChange={(option) => onChangeRule('serviceCategoryId', option)}
-                options={pickListMap.serviceCategoryId}
-              />
-              <FormRuleSelectField
-                rule={rule}
-                label='Applies to Service Type'
-                name='serviceTypeId'
-                onChange={(option) => onChangeRule('serviceTypeId', option)}
-                options={pickListMap.serviceTypeId}
+              <FormRuleCondition
```
gigxz, line 313: A design suggestion:
- wrap this in a Box and move it below the data collected about, so it looks more cohesive with the rest of the rules:
Screenshot 2024-08-09 at 10 56 30 AM


```
              <Box
                sx={{
                  backgroundColor: theme.palette.grey[100],
                  borderRadius: 1,
                  py: 1,
                  px: 2,
                }}
              >
```

src/modules/admin/components/formRules/NewFormRuleDialog.tsx
```diff
@@ -130,52 +160,33 @@ const NewFormRuleDialog: React.FC<Props> = ({
...
-  // This would be better handled with real form validation
-  const [validationError, setValidationError] = useState<string>();
-  const handleSubmit = () => {
-    let error: string | undefined = undefined;
-    if (formRole === FormRole.Service) {
-      if (rule.serviceCategoryId && rule.serviceTypeId) {
-        error =
-          'You cannot choose both Service Category and Service Type. Please only choose one';
-      } else if (!(rule.serviceCategoryId || rule.serviceTypeId)) {
-        error = 'One of either Service Category or Service Type are required';
```
gigxz, line 142: It looks like we lost this validation error, which we should keep. I'm able to create a service rule without specifying any service type/category

src/modules/admin/components/formRules/FormRule.tsx
```diff
@@ -25,64 +25,51 @@ const nonClientFormRoles = [
...
-    otherFunder = rule.otherFunder;
-    serviceCategory = rule.serviceCategory;
-    serviceType = rule.serviceType;
-  }
-
+export const BaseFormRule: React.FC<BaseFormRuleProps> = ({
+  projectType,
+  project,
+  organization,
+  dataCollectedAbout = HmisEnums.DataCollectedAbout.ALL_CLIENTS,
```
gigxz: If we have this default, can we remove the ternary down on line 145?

src/modules/admin/components/formRules/FormRuleCondition.tsx
martha: Oh, thanks for this, I agree that is much less janky, ha!

Regarding `fontWeight`, yes, I removed this on purpose to visually distinguish the "logical flow" label text from the form element label text, but it was a pretty arbitrary decision. I'll loop in @bobbyw-ng for a preference between these 2 screenshots:

1. With all labels bold

Screenshot 2024-08-09 at 12 01 09 PM


3. With only Form Control labels bold, and "logical flow" labels in normal weight

Screenshot 2024-08-09 at 12 00 53 PM

src/modules/admin/components/formRules/NewFormRuleDialog.tsx
martha: I like the idea of adding a box around it. One argument for keeping it above the "Applies to client type" dropdown is that this ordering matches the order of prose when we display the completed rule, both in the top box here on this modal, and on the Rules table outside of this modal. What do you think about this change - adding the box but keeping the order the same?

Screenshot 2024-08-09 at 12 06 55 PM

src/modules/admin/components/formRules/FormRule.tsx
martha: Hm, yes I agree, this refactor is better. It can just accept a typed `ProjectType`, `DataCollectedAbout`, and `Funder`. For organization name, project name, and service type/category name, I think it must accept strings since those are the values that the `NewFormRuleDialog` has access to, not the full project/org object. I will rename those params for clarity as you suggested.

src/modules/admin/components/formRules/FormRuleCondition.tsx
gigxz: Oh alright no preference, just pointing it out in case it was accidentally missed. I think it looks fine as you have it without the bold!

src/modules/admin/components/formRules/NewFormRuleDialog.tsx
gigxz: That looks good to me

src/modules/admin/components/formRules/FormRuleCondition.tsx
bobbyw-ng: I agree I think it's fine without the bold; it would be more consistent with the rest of the site as we typically use bold for labels.

This is very minor but I wonder if these should be all caps, since that's typically how logical operators are formatted (IF instead of If, AND instead of And).

src/modules/admin/components/formRules/FormRule.tsx
martha: @gigxz  I wasn't able to reproduce this locally. Are you still seeing it?

src/modules/admin/components/formRules/FormRuleCondition.tsx
martha: I have a preference for keeping the normal case since it matches the "display sentence" of the form rule, shown at the top. I guess we could change it there too, like this:
Screenshot 2024-08-13 at 11 20 34 AM

But I think there are downsides to doing that, because there is some non-logical-operator text included there too.

src/modules/admin/components/formRules/NewFormRuleDialog.tsx
martha: Thanks, good point, I added this back. The 'cannot choose both' error is no longer needed, but we still need 'one of them is required'

Is there a better way to do this? I was curious about @ttoomey's comment that this would be better handled using real form validation. Should we be trying to standardize around using RHF for forms like this?

gigxz: Yes probably, but we are only using RHF inside the form builder right now and we haven't really hardened off our conventions around validation there either. I think that woudl be ideal, though, that we use RHF for most form inputs including the integrated client side validation

gigxz (approved):

PR #882, opened 2024-08-08
Use most recent version of form for editing Assessments

gigxz:
## Description

https://github.com/open-path/Green-River/issues/6112
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4607

### New behavior for editing old assessments:
* [unchanged] When you open an “old” Assessment for viewing, it is re-opened using the form that was most recently used to submit it (record.form_processor.definition_id). If no such reference exists, use the “most relevant” form.
* [new] When you open an “old” Assessment for editing, it is re-opened using the most recent version of the form that was last used to submit it. **This form is resolved as `upgradedDefinitionForEditing`**

### How to test
**Basic behavior for individual assessment**:
* Create a new custom assessment form definition, publish it
* Perform the assessment for a client (save or submit)
* Edit the assessment definition. Make sure to change the title and add a new question so you can tell them apart.
* Go back to the assessments you submitted and/or saved. Submitted one should use old form for viewing and switch to new form when you hit "unlock". Saved one should continue to use old form for editing.

**Other things to QA**:
* Test same thing as above but for a household intake. (System test tests scenario where some household members used "old" forms and some members used "new" forms).
* Test item rule and data collected about behavior, since that was refactored. (Needs system test https://github.com/open-path/Green-River/issues/6506, I tested manually)

## Type of change


src/modules/assessments/components/AssessmentForm.tsx
```diff
@@ -61,10 +64,8 @@ import {
 interface Props {
   enrollment: EnrollmentFieldsFragment;
   client: AssessedClientFieldsFragment;
-  formRole?: FormRole;
```
gigxz, line 64: This was not needed, as it's always the same as the `definition.role`.

src/modules/assessments/components/AssessmentForm.tsx
```diff
@@ -61,10 +64,8 @@ import {
...
   definition: FormDefinitionFieldsFragment;
   assessment?: FullAssessmentFragment;
-  assessmentTitle?: ReactNode;
```
gigxz, line 67: Instead of passing the title in, we generate it in this component. That's because it depends on the form definition we end up using–– which may be the `definition` or the `upgradedDefinitionForEditing`.

src/modules/assessments/components/AssessmentForm.tsx
```diff
@@ -119,9 +117,30 @@ const AssessmentForm: React.FC<Props> = ({
...
+    if (
+      !locked &&
+      !assessment?.inProgress &&
+      assessment?.upgradedDefinitionForEditing
+    ) {
+      fd = assessment.upgradedDefinitionForEditing;
+    }
+    // Apply "data collected about" rules to filter down the definition to relevant items
+    const relationshipToHoH = enrollment.relationshipToHoH;
+    return applyDefinitionRulesForClient(fd, client, relationshipToHoH);
```
gigxz, line 129: Calling `applyDefinitionRulesForClient` was previously peppered throughout the codebase, at the locations where the definition is fetched. Now, it is only called in one place, right here, closer to where the form rendering occurs.

I think this is easier to reason about. We don't have system test coverage for this, but I opened https://github.com/open-path/Green-River/issues/6506 -- I started working on a system test for this locally, but decided to de-scope it. QA will be needed on this front.

src/modules/assessments/components/AssessmentFormSideBar.tsx
```diff
@@ -136,11 +130,6 @@ const AssessmentFormSideBar: React.FC<Props> = ({
                 onSuccess={navigateToEnrollment}
               />
             )}
-            {showAssessmentId && (
-              <Typography variant='body2'>
-                <b>Assessment ID:</b> {assessment.id}
```
gigxz, line 141: Remove this, it is only shown in development (`import.meta.env.MODE === 'development'`) and not useful, since the id is in the URL bar

src/modules/assessments/components/AssessmentForm.tsx
```diff
@@ -61,13 +64,10 @@ import {
 interface Props {
   enrollment: EnrollmentFieldsFragment;
   client: AssessedClientFieldsFragment;
-  formRole?: FormRole;
   definition: FormDefinitionFieldsFragment;
   assessment?: FullAssessmentFragment;
-  assessmentTitle?: ReactNode;
   alerts?: ReactNode;
   top?: number;
-  navigationTitle: ReactNode;
```
gigxz, line 70: This is no longer needed due to simplifying the household stepper header

current

=>
proposed 1

src/modules/assessments/components/AssessmentForm.tsx
```diff
@@ -117,11 +114,24 @@ const AssessmentForm: React.FC<Props> = ({
...
     if (assessment && !assessment.inProgress) setLocked(true);
-  }, [assessment, canEdit]);
+  }, [assessment, canEdit]); // re-runs after assessment is refetched, which leads to assessment re-locking after submit
+
+  // Choose the FormDefiniton to use for rendering, and filter it down based on client attributes (Data Collected About rules).
+  const definition = useMemo(() => {
+    let fd = definitionProp;
+    if (!locked && assessment && !assessment.inProgress) {
+      // When editing a non-WIP assessment, we want to use the most recent version of the Form Definition.
+      // If the most recent version differs from the version that was most recently used to updated it, then 'upgradedDefinitionForEditing' is present.
```
martha, line 124: I remember some back-and-forth in design sessions about whether it would be confusing to users that in-progress forms are still opened using the old form version. We discussed adding a button to prompt the user to upgrade to the latest version, and decided against it since it seems like too much effort for an edge case.

I wonder, though, if just adding a brief sentence to the top of the form, something like, "A new version of this assessment has been published since the draft was started. To upgrade, delete this draft and begin again" could be a low effort way to help clarify things.

Or we can just push what you have and see if users actually get confused 🤷‍♀️

martha (approved):
Looks awesome @gigxz . Thank you for this hard work 🙏 🚀 I tested as described in the PR description and everything looked good to me.

One fun fact that I learned while QA-ing this, completely unrelated to your changes, is that we don't restrict submitting an assessment for a definition that's still in draft. You can't get there from any buttons/links within our app, but you can easily get there by replacing the form definition ID in the URL with the ID of the latest draft (e.g. https://qa-hmis.openpath.host/client/==ZDg0MXE4My93d29xaXhKbWZ1ZHY0QT09/enrollments/==U3dTREVOdk11NzBxaXhKbWZ1ZHY0QT09/assessments/new/174). And then you can fill out the assessment and submit it as normal.

Is this a bug? Should I create a ticket? Seems pretty low priority to fix because it's unlikely to ever come up, but it just feels icky, ha.

src/modules/assessments/components/AssessmentForm.tsx
gigxz: We landed on not showing anything, I think because of the underlying goal to make Provider-facing interfaces as simple as possible. If 90% of the time if the change is to the wording of a question, it doesn't really matter––they might be alarmed or confused by seeing a new message, when it is (most likely?) not relevant (or actionable) to them as a Provider.

gigxz:
> One fun fact that I learned while QA-ing this, completely unrelated to your changes, is that we don't restrict submitting an assessment for a definition that's still in draft. You can't get there from any buttons/links within our app, but you can easily get there by replacing the form definition ID in the URL with the ID of the latest draft (e.g. https://qa-hmis.openpath.host/client/==ZDg0MXE4My93d29xaXhKbWZ1ZHY0QT09/enrollments/==U3dTREVOdk11NzBxaXhKbWZ1ZHY0QT09/assessments/new/174). And then you can fill out the assessment and submit it as normal.

Good find! A ticket for that would be awesome, thanks. I think it should be implemented in both places:
1) frontend should show a NotFound page if the definition it fetches turns out to be a `draft`
2) SubmitForm/SubmitAssessment/SaveAssessment should all have a check to ensure the definition if `published` or `retired` (aka, not `draft`).

martha:
Ok sounds great, thank you for the implementation notes! Created https://github.com/open-path/Green-River/issues/6520

PR #883, opened 2024-08-08
Validate that either pick list options or reference exists

martha:
## Description
GH issue: https://github.com/open-path/Green-River/issues/6405

How to test:
- Make a custom form with a choice element that doesn't have any options or a link to a pick list reference
- Try to hit save, you should see a (more) user-friendly error

A few things that are not ideal about this solution
- it doesn't clear out the error after the error condition is no longer true, only on save
- it only selects the dropdown as having an error state, not the 'add condition' button, since 'add condition' is not an RHF controlled component
- on error, it auto-focuses the dropdown of pick list references, which in turn opens the dropdown and covers the actual error message that we want to show

but, given our conversations in the estimation session yesterday, I didn't want to spend a ton more time on it. (Edit: spent about 1 hour troubleshooting the above so far and could not come up with an improvement.) Maybe this is clear enough?  Or @gigxz maybe you can easily see a way to improve this that escaped me?

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/modules/formBuilder/components/itemEditor/pickLists/ManagePickListOptions.tsx
+++ b/src/modules/formBuilder/components/itemEditor/pickLists/ManagePickListOptions.tsx
@@ -90,6 +90,11 @@ const ManagePickListOptions: React.FC<ManagePickListOptionsProps> = ({
     [fieldsWatch]
   );
 
+  const eitherOptionsOrReferenceProvided = useMemo(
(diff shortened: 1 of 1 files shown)


gigxz (approved):
Thanks for the thorough description @martha. I think this meets our needs, agree that we shouldn't spend more time here. Thanks!

PR #884, opened 2024-08-12
Style adjustments to mobile toolbar, and decrease `md` breakpoint size

gigxz:
## Description

Style changes from https://github.com/open-path/Green-River/issues/6433#issuecomment-2286332151 and reduce `md` breakpoint (see comment)

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/components/layout/MainLayout.tsx
+++ b/src/components/layout/MainLayout.tsx
@@ -115,7 +115,11 @@ const MainLayout: React.FC<Props> = ({ mobileMenuContext, children }) => {
           </RouterLink>
           <Box display='flex' sx={{ flexGrow: 1 }}></Box>
           {!isTiny && (
-            <Stack direction='row' spacing={{ md: 0.5, lg: 2 }}>
+            <Stack
+              direction='row'
+              alignItems='center'
+              spacing={{ md: 0.5, lg: 2 }}
+            >
               <ToolbarMenu />
               <OmniSearch />
               <UserMenu hideName={isMobile} />
--- a/src/components/layout/MainLayout.tsx
+++ b/src/components/layout/MainLayout.tsx
@@ -40,8 +40,7 @@ const MainLayout: React.FC<Props> = ({ mobileMenuContext, children }) => {
     error,
   } = useGetRootPermissionsQuery();
 
-  const isMobile = useIsMobile('md');
-  const isTiny = useIsMobile('sm');
+  const isMobile = useIsMobile();
 
   const { mobileNavIsOpen, handleOpenMobileMenu, handleCloseMobileMenu } =
     mobileMenuContext;
@@ -114,7 +113,7 @@ const MainLayout: React.FC<Props> = ({ mobileMenuContext, children }) => {
             {appName || 'Open Path HMIS'}
           </RouterLink>
           <Box display='flex' sx={{ flexGrow: 1 }}></Box>
-          {!isTiny && (
+          {!isMobile && (
             <Stack
               direction='row'
               alignItems='center'
@@ -125,7 +124,7 @@ const MainLayout: React.FC<Props> = ({ mobileMenuContext, children }) => {
               <UserMenu hideName={isMobile} />
             </Stack>
           )}
-          {isTiny && (
+          {isMobile && (
             <>
               <IconButton
                 aria-label='Navigation'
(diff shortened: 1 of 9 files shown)


src/components/layout/MainLayout.tsx
```diff
@@ -109,14 +114,14 @@ const MainLayout: React.FC<Props> = ({ mobileMenuContext, children }) => {
             {appName || 'Open Path HMIS'}
           </RouterLink>
           <Box display='flex' sx={{ flexGrow: 1 }}></Box>
-          {!isMobile && (
+          {!isTiny && (
             <Stack direction='row' spacing={{ md: 0.5, lg: 2 }}>
```
martha: Can add `sx={{alignItems: 'center'}}` to the `Stack` in order to keep the button padding unchanged from what's currently on prod

src/components/layout/nav/UserMenu.tsx
```diff
@@ -23,18 +24,21 @@ import { useIsMobile } from '@/hooks/useIsMobile';
...
+interface UserMenuProps {
+  hideName?: boolean;
+}
+const UserMenu: React.FC<UserMenuProps> = ({ hideName }) => {
   const popupState = usePopupState({ variant: 'popover', popupId: 'userMenu' });
   const { user, logoutUser } = useAuth();
   const { manageAccountUrl, warehouseUrl, warehouseName } =
     useHmisAppSettings();
-  const isMobile = useIsMobile();
+  const isTiny = useIsMobile('sm');
```
martha: This leads to the desktop-style buttons getting displayed on `md` size screens, which I think is not intentional, right?

Screenshot 2024-08-13 at 10 53 43 AM

martha: Also, I think this is not a problem with your current changes, but just wanted to flag that this is another place in the code where I made a breakpoint change that relies on other breakpoint changes in a kind of code-smelly way. https://github.com/greenriver/hmis-frontend/blob/release-128/src/modules/admin/components/AdminDashboard.tsx#L145-L147

Maybe the simplest way to clean up this code smell would be to have some breakpoints, ones that are used sort of systemically like this, should simply be set somewhere as constants?

src/components/layout/MainLayout.tsx
gigxz: Thanks, added

src/components/layout/nav/UserMenu.tsx
gigxz: I ended up rolling back the change because there were a lot of downstream effects of treating `sm` as mobile here but `md` as mobile elsewhere. The MUI default for `md` is 900px which is quite large, even larger than the norm for tablet size which appears to be 768px.

So, to simplify this, I rolled back the change so that we are still using `md`, but I reduced the `md` breakpoint to 769 in the theme. This negatively affects the Enrollment Dashboard when viewed in the 800-900px width page _with the nav menu open_, which should be fixed, but I think is an OK tradeoff. (The benefit is that desktop users on 800-900 are not forced to use the mobile navigation menu).

Thoughts on that @martha @ttoomey?

src/components/layout/dashboard/contextHeader/ContextHeader.tsx
```diff
@@ -115,7 +117,7 @@ const ContextHeader: React.FC<Props> = ({
         </Box>
       ) : (
         <Box display='flex' alignItems='stretch' width='100%' flex={1}>
-          {!isOpen && (
+          {!isOpen && !isMobile && (
```
gigxz, line 120: Prevent both menu items from  appearing. To reproduce, make sure menu is closed when you reduce screen width
Screenshot 2024-08-13 at 4 54 00 PM

src/components/layout/nav/UserMenu.tsx
gigxz: Oops, the issue @martha pointed out is still occurring, hold on.

Edit: addressed in latest commit, the issue only popped up when using a "mobile" screen size with the nav open, and then expanding the width until it was not "mobile" anymore. Fixed it by passing mobile as a prop to the toolbar instead of it reading it from the current size.

martha (approved):
@gigxz  all the changes you made I think are great. 🚀  Let's merge this and revisit mobile nav with fresh eyes and more design thought sometime soon!

PR #857, opened 2024-07-27
6349 upgrade dependencies

ttoomey:
## Description

[GH issue](https://github.com/open-path/Green-River/issues/6349)

prominent upgrades
* most recent version of MUI 5x
* typescript
* apollo GraphQL client
* react 18.2 -> 18.3
* update @rails/activestorage to match rails server version
* prettier 2-> 3

Also:
* fix whitespace and formatting issues
* update versions on workflow files

Note I also attempted to update eslint and react-pdf to new major versions but backed out those updates since they weren't straight forward

## Type of change
[//]: # 'remove options that are not relevant'
- [x] Dependency update

## Checklist before requesting review
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [x] My code includes comments and/or descriptive variable names to help other engineers understand the intent (or not applicable)
- [x] My code follows the style guidelines of this project (eslint)
- [x] I have updated the documentation (or not applicable)
- [x] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

PR #885, opened 2024-08-14
Fix bugs with Form Builder conditional logic

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6403

This PR fixes the higher priority bugs outlined in the linked ticket.

Sample form for testing various types of conditional logic:
[conditions_qa.json](https://github.com/user-attachments/files/16616150/conditions_qa.json)

## Type of change
- [x] Bug fix
- [ ] Dependency update


src/modules/form/components/rhf/ControlledSelect.tsx
```diff
@@ -35,6 +36,7 @@ const ControlledSelect: React.FC<Props> = ({
   loading,
   placeholder,
   helperText,
+  afterChange,
```
martha: For reference here's a link to our previous convo about implementing an `afterChange` callback in one of these `Controlled` form components, which we ended up scrapping at the time. https://github.com/greenriver/hmis-frontend/pull/850#discussion_r1681533408

TLDR `afterChange` could alternatively be implemented as an effect, but overusing effects when they are not needed can cause extra renders. (https://react.dev/learn/you-might-not-need-an-effect)

src/modules/form/components/rhf/ControlledSelect.tsx
```diff
@@ -22,6 +22,7 @@ export type ControlledSelectProps = Omit<
 
 interface Props extends ControlledSelectProps {
   placeholder?: string;
```
martha, line 22: I was curious about this change to add a new interface `Props` that extends `ControlledSelectProps` -- what is the reason for the separation? Seems to come from Typescript https://github.com/greenriver/hmis-frontend/pull/857/files#diff-f060165369ed70f2b53e95c90ca4ef0d6aa9d8a906b7dcc926fac7b569bc7211

src/modules/formBuilder/components/itemEditor/FormEditorItemProperties.tsx
```diff
@@ -142,6 +142,12 @@ const FormEditorItemProperties: React.FC<FormEditorItemPropertiesProps> = ({
     }
   }, [hiddenValue, setValue]);
 
+  useEffect(() => {
```
martha: And, well, to contradict myself, here I switched to using an effect instead of the previous callback (see `ManageEnableWhen` below) which wasn't working, though I haven't deeply investigated why not, and would be happy to put some more time into it if we think it's worthwhile.

Possibly it might relate to us mix and matching  rhf's `setValue` with `onChange`, which the docs explain we should avoid. https://www.react-hook-form.com/api/usecontroller/controller/

src/modules/formBuilder/components/itemEditor/conditionals/EnableWhenCondition.tsx
```diff
@@ -217,34 +244,40 @@ const EnableWhenCondition: React.FC<EnableWhenConditionProps> = ({
...
                 render={({
                   field: { ref, ...field },
                   fieldState: { error },
                 }) => (
                   <YesNoRadio
                     label='Value'
                     error={!!error}
-                    helperText={error?.message}
+                    helperText={
+                      error?.type === 'required' ? 'Required' : undefined
```
martha: I don't really love this pattern for adding the "Required" error message, especially because it's repeated twice, here and below in `InitialValue.tsx`. I'd be glad for a suggestion on how to improve this - maybe a new ControlledRadio component is in order, for example...?

gigxz: Yeah - I think since this is vertical checkboxes now it looks pretty funky- I would suggest making it a ControlledSelect so that all the comparison "values" have a consistent look. I wanted to test this out before suggestion, so, here is a working approach: https://github.com/greenriver/hmis-frontend/pull/888/files

src/modules/formBuilder/components/itemEditor/conditionals/InitialValue.tsx
```diff
@@ -84,12 +84,20 @@ const InitialValue: React.FC<Props> = ({ itemType, control }) => {
                 control={control}
                 defaultValue={InitialBehavior.IfEmpty}
                 shouldUnregister
-                render={({ field: { ref, value, onChange, ...field } }) => (
+                rules={{ required: true }}
+                render={({
+                  field: { ref, value, onChange, ...field },
+                  fieldState: { error },
+                }) => (
```
gigxz: Could DRY up this and the RadioGroupInput in `ManageEnableWhen` by refactoring out to a ControlledRadioGroupInput like we have for some others input types. There are some differences in implementation-- like the other sets `helperText={error?.message}` which is more flexible for supporting any kind of error.

src/modules/formBuilder/components/itemEditor/conditionals/ManageEnableWhen.tsx
```diff
@@ -89,11 +89,7 @@ const ManageEnableWhen: React.FC<ManageEnableWhenProps> = ({
       {fields.map((condition, index) => (
         <RemovableCard
           key={JSON.stringify(condition)} // fixme could be non unique
-          onRemove={() => {
-            const isLastCondition = fields.length === 1;
-            remove(index);
-            if (isLastCondition) onChangeEnableBehavior(null);
-          }}
+          onRemove={() => remove(index)}
```
gigxz, line 76: Nice - it seems like this is no longer needed because the "Conditional Behavior" input gets unmounted when the last index is removed, is that right?

src/modules/formBuilder/components/itemEditor/conditionals/EnableWhenCondition.tsx
```diff
@@ -101,18 +102,42 @@ const EnableWhenCondition: React.FC<EnableWhenConditionProps> = ({
...
+  const { pickList: remotePickList, loading: remotePickListLoading } =
+    usePickList({
+      item: {
+        linkId: 'fake',
+        type: ItemType.Choice,
+        pickListReference: dependentItem?.pickListReference,
+      },
+      fetchOptions: {
+        skip:
+          !dependentItem || !dependentItem.pickListReference || !!localPickList,
```
gigxz: I would have thought that we don't need the `!!localPickList` check, nor the `localResolvePickList` call above. The usePickList hook should [detect](https://github.com/greenriver/hmis-frontend/blob/14715dbffd123391bf11d8f77803dd26f6488a0a/src/modules/form/hooks/usePickList.tsx#L80) whether it's a locally resolvable pick list, and if so returns that (And skips the remote call).

src/modules/formBuilder/components/itemEditor/conditionals/EnableWhenCondition.tsx
```diff
@@ -101,18 +102,42 @@ const EnableWhenCondition: React.FC<EnableWhenConditionProps> = ({
...
+  } = useController({
+    control,
+    name: `${enableWhenPath}.${index}.operator`,
+  });
+
   const dependentItem = useMemo(
     () => (state?.question ? itemMap[state?.question] : undefined),
     [itemMap, state?.question]
   );
+
```
gigxz, line 110: Found an unrelated bug while testing this, but it looks like its on release-129 too. Added a separate issue: https://github.com/open-path/Green-River/issues/6512

src/modules/form/components/rhf/ControlledSelect.tsx
gigxz: Sounds good. I think it might be worth making this more general, e.g. `onChange?: (option: PickListOption | null) => void`, so that it's possible to use it for side effects that might depend on the value. As I belatedly come to in the linked conversation, that is a common(ish?) pattern even for uncontrolled components. I think "onChange" probably clearer than "afterChange" because they are triggered in the same render.

src/modules/formBuilder/components/itemEditor/conditionals/ManageEnableWhen.tsx
martha: Hm, no, I removed this because it was not working -- it seems to work instead with a `useEffect` which I mentioned [here](https://github.com/greenriver/hmis-frontend/pull/885#discussion_r1717121606). But, unmounting is also a good idea, let me try it out...

src/modules/formBuilder/components/itemEditor/conditionals/EnableWhenCondition.tsx
martha: Oh perfect, looks great. Before seeing your comment I put a few minutes into a new `ControlledYesNo` component, but I like your solution much better.

It just needed some tweaking to validation, related to this RHF issue: https://github.com/react-hook-form/react-hook-form/issues/2323

I added that here https://github.com/greenriver/hmis-frontend/pull/885/files#diff-2368e4c27ae82cd3c902453efcfd59df5a1e85dab95f44b903dec4569284ce4eR254-R258

src/modules/formBuilder/components/itemEditor/conditionals/ManageEnableWhen.tsx
martha: Hrm! It has to do with the `useController` call to get the `onChangeEnableBehavior` handler in this file. For some reason, this is setting the enable behavior back to 'ALL' _after_ the component gets unmounted and sets it to `undefined`. Curiously, this happens even when the `onChangeEnableBehavior` call on line 61 is commented out.

Anyway, I _think_ removing the `useController` call and then adding a default value to the radio group input should achieve the desired behavior without interfering with the component's ability to unregister when it unmounts. Investigating...

martha: Ok, I see now that this behavior must've been written the way it was originally because RHF's controller's `defaultValue` doesn't quite work for this use case. Maybe related to its strict requirements around not passing `undefined`: https://react-hook-form.com/docs/usecontroller/controller

Anyway, passing `setValue` from `FormEditorItemProperties` does seem to work, if you think it's not too messy.

martha: Alternatively, we could remove this and leave it as a required field that doesn't get a default value.

src/modules/formBuilder/components/itemEditor/conditionals/InitialValue.tsx
martha: Good idea to dry up into a controlled radio component, I implemented that.

I was seeing that the `error` object coming from RHF did not actually have a message on it, just a type of 'required.' I can investigate a bit more to double check that behavior and see if I can find where it's coming from.

martha: Ah it was a pretty simple fix. https://github.com/greenriver/hmis-frontend/pull/885/commits/26082bd076389dcaef16a4d83246dc7e86fc1a64

src/modules/formBuilder/components/itemEditor/conditionals/EnableWhenCondition.tsx
martha: You are right. This can be simplified a lot! Thank you!

martha: I'm not sure if `|| { linkId: 'fake', type: ItemType.Choice }` on line 118 is really the best way to keep typescript happy; maybe `usePickList` should accept item as optional and return early if it's not provided, but I was more hesitant to change a broadly used interface. (Passing `skip` option doesn't work because it skips the network call but not the whole hook, and it also overwrites the `skip` logic internal to that hook, which turns out to be important)

src/modules/form/components/rhf/ControlledSelect.tsx
gigxz: Weird, I don't know why it's like that. Typescript seems perfectly happy with the below.
```
export type ControlledSelectProps = Omit<
  GenericSelectProps<PickListOption, false, false>,
  'value' | 'onChange' | 'onBlur'
> & {
  name: string;
  control?: Control; // Optional when using FormProvider
  rules?: RhfRules;
  required?: boolean;
  helperText?: ReactNode;
  placeholder?: string;
  onChange?: (option: PickListOption | null) => void;
  setValueAs?: (option: PickListOption | null) => any; // allow transform PickListOption to desired value (to support boolean)
};

// React-Hook-Form wrapper around GenericSelect.
// This component stores a string as the field value, but passes a PickListOption to the GenericSelect. (Logic that is redundant with TableFilterItemSelect, among others)
const ControlledSelect: React.FC<ControlledSelectProps> = ({
```

src/modules/formBuilder/components/itemEditor/conditionals/ManageEnableWhen.tsx
gigxz: > For some reason, this is setting the enable behavior back to 'ALL' after the component gets unmounted and sets it to undefined.

Hmm, I can't reproduce this. To me the behavior looks correct. (When reverting to the useController approach). Maybe we can pair?

src/modules/formBuilder/components/itemEditor/conditionals/ManageEnableWhen.tsx
```diff
@@ -56,44 +53,24 @@ const ManageEnableWhen: React.FC<ManageEnableWhenProps> = ({
   return (
     <CardGroup
       onAddItem={() => {
-        if (fields.length === 0) {
-          // when adding first condition, set default enable behavior
-          onChangeEnableBehavior('ALL');
-        }
+        setValue(enableBehaviorPath, EnableBehavior.All);
```
gigxz: I think you still need this `if` condition or else its randomly setting it back to `ALL` each time you add a rule, even if you already specified you want `ANY`

src/modules/formBuilder/components/itemEditor/conditionals/ManageEnableWhen.tsx
gigxz: oookay, summary:
- bug was reproducible by working from an item that was persisted with a condition (since that gets set as the default value)
- `useController`'s `onChange` is meant to only be used with onChange to an input component, it shoudl not be invoked manually (per docs) -- you are correct to use `setValue` for these instances
- there were 2 other places where we used useController hook in this way that may be prone to the same bug and should be replaced with setValue

src/modules/formBuilder/components/itemEditor/FormEditorItemProperties.tsx
```diff
@@ -172,6 +172,7 @@ const FormEditorItemProperties: React.FC<FormEditorItemPropertiesProps> = ({
 
   if (!itemTypeValue) throw Error('Item type must be defined');
 
+  // todo @martha - new text item condition type is dropdown
```
martha: Flagging that I still have to fix this 🚶‍♀️

src/modules/formBuilder/components/itemEditor/conditionals/EnableWhenCondition.tsx
gigxz: Yes I agree, the interface for usePickList could certainly be improved!

gigxz (approved):

PR #886, opened 2024-08-14
Hide FormBuilder Image, File, Initial, Autofill, and some Bounds features

martha:
## Description

GH issues:
- https://github.com/open-path/Green-River/issues/6482
- https://github.com/open-path/Green-River/issues/6484

Depends on warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4623

This PR is a grab bag of hiding Form Builder features that will require more work to fully debug/support.
- Hide Image and File upload types entirely for now (even from super-admins)
- Hide Initial and Autofill cards for non-super-admin users
- For Bounds the story was a little more complicated.
  - Desired behavior is to hide "Advanced Bounds" features behind super-admin permission filter. This includes local constant, dependent question, and offset.
  - For dates, while working on this, I discovered that setting a simple bound (aka not based on a local constant or dependent question) on a date is not supported by the [schema](https://github.com/greenriver/hmis-warehouse/blob/84255c42328e4e63885887a53735c91aabadb7c4/drivers/hmis_external_apis/public/schemas/form_definition.json#L45-L52). I can't think of a use case for this anyway, so I just disabled it. This also meant disabling date bounds entirely for non-Super Admin users.
  - For numbers, this still left open my [question](https://github.com/open-path/Green-River/issues/6484#issuecomment-2289296161) about showing/hiding/disabling advanced features if they are already filled in. I've implemented a suggestion here, which is to show, but disable, advanced features that have values filled in. I'm very open to alternative proposals though, this just seemed like the simplest thing to me in the moment (?) - so that we don't have to deal with filtering which value bound cards the user can see, and any ramifications that might have on dealing with indexes in the Bounds list within the actual structure of the form ...

## Type of change
- [x] Bug fix
- [ ] Dependency update


src/modules/formBuilder/components/FormBuilderPalette.tsx
```diff
@@ -135,8 +135,13 @@ const FormBuilderPalette: React.FC<FormBuilderPaletteType> = ({
     >
       <Stack gap={1}>
         {Object.entries(FORM_ITEM_PALETTE)
-          // TODO - Object is a special case, remove it from the list for now until we support it
-          .filter(([key]) => key !== ItemType.Object)
+          // TODO - Image, File, and Object are special cases. Remove from the list for now until we add support
```
gigxz: Don't think this needs to be marked as a TODO since there is no immediate action, maybe add a reference to issue 6401 instead? https://github.com/open-path/Green-River/issues/6401

src/modules/formBuilder/components/itemEditor/FormEditorItemProperties.tsx
```diff
@@ -349,31 +355,37 @@ const FormEditorItemProperties: React.FC<FormEditorItemPropertiesProps> = ({
...
         <Section
           title='Min/Max Bounds'
           hidden={
-            // bounds are supported by numbers and dates only
-            !COMPARABLE_ITEM_TYPES.includes(itemTypeValue)
+            // "simple" bounds are supported for numbers only.
+            // For super admins only, "advanced" bounds on dates are also supported.
+            !COMPARABLE_ITEM_TYPES.includes(itemTypeValue) ||
+            (itemTypeValue === ItemType.Date && !canAdministrateConfig)
           }
```
gigxz: I'm not sure about the need for hiding the "Date" type in particular. We had talked about the fact that it's less useful to set as a fixed constant, but I don't see why it _shouldn't_ be possible to set a fixed date bound.  (Vague example could be: certain funding stream started at a certain fixed date, so make sure people don't record xyz field before that,?)

```
        {
          "text": "Bounded date",
          "type": "DATE",
          "link_id": "bounded_date",
          "required": false,
          "read_only": false,
          "warn_if_empty": false,
          "disabled_display": "HIDDEN",
          "bounds": [
            {
              "id": "foo",
              "type": "MIN",
              "severity": "error",
              "value_date": "2024-01-01"
            }
          ]
        }
```

The GraphQL schema and frontend implementation do support the above date bound. The only change needed is to allow the field [here in the JSON schema](https://github.com/greenriver/hmis-warehouse/blob/40df1f6881d9212cf3e20002591b90b50d949473/drivers/hmis_external_apis/public/schemas/form_definition.json#L127-L129) by adding the below
```
"value_date": {
  "type": "string"
},
```
I feel like that's worth doing it only to reduce the number of special cases. 

---

Separately, and this is getting away a little bit from the task at hand, but bounds are supported for strings, which set min/max char counts. We do have several custom fields on assessments that use that functionality.

Would you mind defining a separate constant for which types support bounds, just to get that squared away? 
```
export const BOUNDABLE_ITEM_TYPES = [
  ItemType.String, // bound sets min/max chars
  ItemType.Text,    // bound sets min/max chars
  ItemType.Integer,
  ItemType.Currency,
  ItemType.Date,
];
```

gigxz (approved):
Looks good! I think it would be nice to fix the issues with the bounds (allow strings and dates).

I tested manually to ensure that as a non-super-admin, when I edit items it doesn't blow away fields that I can't access (like autofill) 👍

src/modules/formBuilder/components/itemEditor/FormEditorItemProperties.tsx
martha: Ah, makes sense and sounds good.

For text bounds, I'm assuming it doesn't make sense to pin these to a local constant or dependent question, right -- in other words, text field bounds can only be simple numeric bounds?

gigxz: Yes that's right

gigxz (approved):
new changes look good, re-targeted to 130 since thats where https://github.com/greenriver/hmis-warehouse/pull/4623 went!

PR #895, opened 2024-08-19
Address misc Editor Drawer grumps

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6406

This PR includes a couple of minor changes and a refactor that looks big but hopefully isn't too disruptive.

Minor changes:
- Hides Disabled Display for groups, since its behavior is unpredictable. (See [figma](https://www.figma.com/board/axGI9vwmuBRC4uZ11ZJ3RP/OP-HMIS-%2F-5792-%3A-Form-Builder-Workflow?node-id=1066-7835&t=Dcm0pN2tZnHUo0we-0) for bug description and decision)
- Sets Link ID rules as part of the helper text so that they always appear, rather than only showing when the user tries to submit something that doesn't match the rules.

The refactor was motivated by a low priority item on this ticket, making the drawer open/close transition smooth. Since the previous 2 fixes were so simple, I put the remaining time into making this update too. It involved:
- Updating the `FormItemEditor` component, here renamed `FormItemDrawer`, to make the option prop optional, and removing the conditional around displaying this component based on whether or not the item exists. This allows the transition to be smooth
- Moving the `useForm` call and other logic that requires a non-optional item prop into the `FormEditorItemProperties` component, here renamed `FormItemEditor`. I think this naming is clearer -- the ItemDrawer is in charge of the open/closeable drawer, and when the item is non-null, it renders the ItemEditor which contains the actual form. Unfortunately, the renaming adds confusion to this PR because `FormItemEditor` is what the overarching Drawer component used to be called, and git doesn't understand that when displaying the diff. Here is the commit to look at to understand the diff before the rename: https://github.com/greenriver/hmis-frontend/commit/174f3a5bcfc69eaed0e421ec31a0e58abbab2336

I'm willing to walk any part of this back, or change names to reduce confusion, if you don't think it is worth while!

## Type of change
- [x] Bug fix
- [ ] Dependency update


gigxz (approved):
A little hard to follow the diff but, I think it looks look a good cleanup and much better naming. thanks!

PR #896, opened 2024-08-19
Fix bug evaluating date comparisons in Enable When conditions

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6403

This PR:
- Fixes the existing bug with the form conditional logic (not related to the form builder) where it fails to correctly compare dates in Enable When conditions
- In the form builder, allows setting a date using a datepicker in an enable when condition
- Adds .storybook to .gitignore to avoid
Screenshot 2024-08-19 at 9 56 05 AM

To test, create a custom assessment where one field is shown conditionally based on the result of another field that is a date. Example json:
[conditional_date.json](https://github.com/user-attachments/files/16661298/conditional_date.json)

## Type of change
- [x] Bug fix
- [ ] Dependency update


.gitignore
```diff
@@ -29,3 +29,4 @@ yarn-debug.log*
 yarn-error.log*
 
 docker-compose.override.yml
+.storybook
```
gigxz, line 32: We don't want to ignore this directory, because that's where all the storybook configuration lives. 
What command were you running that triggered creation of the js files in your screenshot? I haven't seen that before

src/modules/form/util/formUtil.ts
```diff
@@ -323,11 +324,17 @@ const evaluateEnableWhen = ({
       en.compareQuestion ? values[en.compareQuestion] : undefined,
     ].filter((e) => !isNil(e))[0];
   }
-
+  if (isDate(currentValue)) {
+    comparisonValue = parseHmisDateString(comparisonValue);
```
gigxz, line 328: I'm not sure about changing the pattern here. In other places, like `ValueBound`, if we're comparing with a date then we use the date-typed `value_date` field.

With this change, we would be using the `answer_code` (presumably?) field to store the comparison date on the `EnableWhen`  object, rather than adding a typed `answer_date` field to the schema. Do we have a reason for that inconsistency?

.gitignore
martha: Hm, I see. I'm not sure, I don't remember running anything other than `yarn storybook`, but I probably did. Anyway, I can just delete the files and pay attention if they get generated again.

src/modules/form/util/formUtil.ts
martha: Good point, no reason, we can update the schema instead.

This was sort of a sidebar bugfix that doesn't have its own ticket, but I can create a ticket now or just proceed until this is ready to merge, whatever you prefer

gigxz: Ok, Lets add a separate ticket since its new functionality

PR #897, opened 2024-08-20
Add aria labels for ease of system testing

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6340

Depends on hmis-warehouse PR: Unusually, in this case I think the warehouse PR depends upon this one and not the other way around.

This PR
- Adds aria labels to item label, ~action menu (edit/delete)~, and up/down buttons
- Adds aria labels to form palette buttons
- Fixes a display inconsistency in the way item types are shown

## Type of change
- [x] Bug fix
- [ ] Dependency update


src/modules/formBuilder/components/itemEditor/FormItemEditor.tsx
```diff
@@ -79,7 +79,8 @@ const FormItemEditor: React.FC<FormItemEditorProps> = ({
           {isNewItem ? 'Add New Form item' : 'Edit Form Item'}
         </Typography>
         {isNewItem
-          ? `${startCase(item.type.toLowerCase())} Item`
+          ? // todo @Martha - any change here will cause merge conflicts
```
martha: Will conflict with https://github.com/greenriver/hmis-frontend/pull/895

src/components/elements/CommonMenuButton.tsx
```diff
@@ -98,28 +99,30 @@ const CommonMenuButton = ({
...
-            </MenuItem>
-          )
+        {items.map(
+          ({ key, to, title, divider, onClick, disabled, ariaLabel }) =>
+            divider ? (
+              <Divider key={key} />
+            ) : to ? (
+              <MenuItem key={key} component={RouterLink} to={to}>
+                {title}
+              </MenuItem>
```
gigxz: aria label should be applied if its a link, too, right?

src/components/elements/CommonMenuButton.tsx
```diff
@@ -98,28 +99,30 @@ const CommonMenuButton = ({
...
+                key={key}
+                onClick={() => {
+                  if (onClick) {
+                    // close menu before triggering onClick
+                    setAnchorEl(null);
+                    onClick();
+                  }
+                }}
+                disabled={disabled}
+                aria-label={ariaLabel}
```
gigxz: > Every menuitem must have an accessible name. This name comes from the element's contents by default. If the contents don't provide for a useful accessible name, [aria-labelledby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby) can be used to reference visible label. If no visible content is available to provide the accessible name, an accessible name can be provided with [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label).

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menuitem_role

Based on that description, I'm not sure this is the best solution. In this case the menu items are "Edit" and "Delete", right? Were you having trouble selecting those just based on text (with capybara)?

martha: Hm, yes I was, but I can give it another try.

martha: Ok, yes, it is possible like this: https://github.com/greenriver/hmis-warehouse/pull/4627/commits/b1ccbc2d98eaa1448d694e52e430323ea0cac389

gigxz: Hmm ok, I think we want to avoid relying on MUI classes in the system tests as much as possible, since those can change, and are related to styling not function. Can you use `li[role="menuitem"]` instead?

gigxz: If not, I guess aria label is better than using MUI class

martha: It seems to work with `find('li', exact_text: 'Edit')`. (Not `find("li[exact_text='Edit']")` 😞 )

gigxz (approved):
👍

PR #899, opened 2024-08-21
Fix misc Form Builder bugs

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6512

This PR fixes a couple of straightforward bugs ~and shows some thoughts around the more complicated bug having to do with `ControlledSelect`, default values, and unregistering.~ Edit: See https://github.com/greenriver/hmis-frontend/pull/906 for this one

Simple bugfixes:
- The required/optional radio previously didn't set the form to dirty, so you couldn't save changes to that field unless you also made changes to another field. This is fixed with passing the `shouldDirty` option to `setValue`
- The boolean `ControlledSelect` requires custom validation, otherwise it is considered invalid when the value is set to `false`, which is actually a valid value.

~I'll add a comment thread inline to discuss the other bug 🧵 🐞 ~

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/modules/formBuilder/components/itemEditor/conditionals/EnableWhenCondition.tsx
+++ b/src/modules/formBuilder/components/itemEditor/conditionals/EnableWhenCondition.tsx
@@ -229,7 +229,13 @@ const EnableWhenCondition: React.FC<EnableWhenConditionProps> = ({
                   return null;
                 }}
                 rules={{
-                  required: 'This field is required',
+                  validate: (value) => {
+                    if (value === null) {
+                      // Requires custom validation to accommodate the valid value `false`
+                      return 'This field is required';
+                    }
+                    return true;
+                  },
                 }}
               />
             )}
--- a/src/modules/formBuilder/components/itemEditor/conditionals/EnableWhenCondition.tsx
+++ b/src/modules/formBuilder/components/itemEditor/conditionals/EnableWhenCondition.tsx
@@ -1,6 +1,6 @@
 import { Box, Grid, Stack, Typography } from '@mui/material';
 import { startCase } from 'lodash-es';
-import { useCallback, useMemo, useState } from 'react';
+import { useMemo, useState } from 'react';
 import { UseFormSetValue, useWatch } from 'react-hook-form';
 import { FormItemControl, FormItemState } from '../types';
 import { useLocalConstantsPickList } from '../useLocalConstantsPickList';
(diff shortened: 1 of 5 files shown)


src/modules/formBuilder/components/itemEditor/conditionals/EnableWhenCondition.tsx
```diff
@@ -163,6 +163,22 @@ const EnableWhenCondition: React.FC<EnableWhenConditionProps> = ({
...
+  const clearAnswer = useCallback(() => {
+    setValue(`${enableWhenPath}.${index}.answerCode`, null);
+    setValue(`${enableWhenPath}.${index}.answerCodes`, null);
+    setValue(`${enableWhenPath}.${index}.answerBoolean`, null);
+    setValue(`${enableWhenPath}.${index}.answerNumber`, null);
+    setValue(`${enableWhenPath}.${index}.answerGroupCode`, null);
+  }, [setValue, enableWhenPath, index]);
+
+  const onChangeDependentQuestion = useCallback(() => {
+    // todo @Martha - fix typescript's unhappiness
```
martha: Ok so: What I found when playing around with this is that:
- `shouldUnregister` appears to be the culprit for the bug that's video'd in the original ticket. I wonder if it is because `shouldUnregister` causes a reset on the form, which resets to `defaultValues`?
- `shouldUnregister` also doesn't appear to be working -- at least, in the "answer value" dropdowns. Using the debug button, I could see that even when the dropdown or text box component unmounted, the `answerCode` or `answerBoolean` property of the `enableWhen` would still be present.
- The code as it is here appears to me to be working, but typescript is unhappy with passing `null` to the enableWhen operator. I've added a couple ideas for how to move forward, @gigxz I'd like to get your thoughts and/or pair :)

gigxz: I did some digging here to try to understand _why_ the components were being unregistered when changing/clearing the value.  I pasted the below code in various components in this hierarchy to see which ones were being mounted and unmounted
```
  useEffect(() => {
    console.log('parent component mounted');
    return () => {
      console.log('parent component unmounted');
    };
  }, []);
```

That tracked me back to several keys [like this](https://github.com/greenriver/hmis-frontend/blob/3298215d9ffb94ba22357409d87df5fb9e0d1934/src/modules/formBuilder/components/itemEditor/conditionals/ManageEnableWhen.tsx#L91) (I'm sorry 🙈 ). Those keys were causing the whole component tree below them to mount unmount when values changed. The unmounting caused the inputs to unregister, and the remounting caused them to reregister with their default values.

I opened https://github.com/greenriver/hmis-frontend/pull/906, can you try that out and see if it addresses the bug from the ticket? I think it should, but I only did some basic testing.

Also, I had no idea that useFieldArray generated `id`s.. I found that out from seeing that you used it on the PickListOption array. So thanks!

gigxz (approved):

PR #901, opened 2024-08-21
Fix bug on submitting upgraded assessments

gigxz:
## Description

Requires backend PR: https://github.com/greenriver/hmis-warehouse/pull/4649

* Patch bug: https://github.com/open-path/Green-River/issues/6112#issuecomment-2302737952 Fix:
  * Always pass the Form Definition ID of the "editingDefinition" to the `useAssessmentHandlers` hook, so that when an assessment is submitted it is always submitted as using the "editing" definition
  * Raise an error if you try to submit form state that has keys that are not present as link_ids in the underlying definition. (We already do this check on the backend in the Form Processor, but this was dropping them silently during the transformation).
 
* Patch another bug (also introduced in 129) whereby the submitted values for assessments would have some extra hidden fields, because the `createHudValuesForSubmit` transformation was being called with a definition that had NOT been yet transformed  (e.g. it still had adult-only questions even if form was for a child). I don't have a regression test for this yet. The impact could be that some empty records are unnecessarily created, I think. Fix:
  * Move values transformation (`createHudValuesForSubmit`) into the `useDynamicFields` hook so that the DynamicForm is responsible for transforming the form state into a "submittable" shape, using the same FormDefinition it used for editing it. I think this is a cleaner API, and makes sense since the DynamicForm already knows about the definition.

#### How to Test
* Follow script in https://github.com/open-path/Green-River/issues/6112#issuecomment-2302737952
* and "how to QA" section of https://github.com/open-path/Green-River/issues/6112#issuecomment-2302737952

## Type of change
- [x] Bug fix

### DIFF
--- a/src/modules/assessments/components/AssessmentForm.tsx
+++ b/src/modules/assessments/components/AssessmentForm.tsx
@@ -64,13 +64,16 @@ import {
 interface Props {
   enrollment: EnrollmentFieldsFragment;
   client: AssessedClientFieldsFragment;
-  definition: FormDefinitionFieldsFragment;
+  // FormDefiniton to use for rendering the assessment in read-only mode
+  viewingDefinition: FormDefinitionFieldsFragment;
+  // FormDefiniton to use for rendering the assessment for editing
+  editingDefinition: FormDefinitionFieldsFragment;
   assessment?: FullAssessmentFragment;
   alerts?: ReactNode;
   top?: number;
   embeddedInWorkflow?: boolean;
   FormActionProps?: DynamicFormProps['FormActionProps'];
-  onSubmit: (formDefinitionId: string) => DynamicFormProps['onSubmit'];
+  onSubmit: DynamicFormProps['onSubmit'];
   onSaveDraft?: DynamicFormProps['onSaveDraft'];
   errors: ErrorState;
   onCancelValidations?: VoidFunction;
@@ -87,7 +90,8 @@ const AssessmentForm: React.FC<Props> = ({
   assessment,
   client,
   alerts,
-  definition: definitionProp,
+  viewingDefinition,
+  editingDefinition,
   enrollment,
   embeddedInWorkflow,
   FormActionProps,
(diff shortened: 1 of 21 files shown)


gigxz:
This bug appears in release-129

gigxz:
Also the system test should be updated

src/modules/assessments/components/AssessmentForm.tsx
```diff
@@ -118,16 +122,17 @@ const AssessmentForm: React.FC<Props> = ({
 
   // Choose the FormDefiniton to use for rendering, and filter it down based on client attributes (Data Collected About rules).
   const definition = useMemo(() => {
-    let fd = definitionProp;
-    if (!locked && assessment && !assessment.inProgress) {
-      // When editing a non-WIP assessment, we want to use the most recent version of the Form Definition.
-      // If the most recent version differs from the version that was most recently used to updated it, then 'upgradedDefinitionForEditing' is present.
-      fd = assessment.upgradedDefinitionForEditing || fd;
-    }
+    const fd = locked ? viewingDefinition : editingDefinition;
```
gigxz, line 125: This change requires https://github.com/greenriver/hmis-warehouse/pull/4649#discussion_r1733379534

src/modules/form/components/DynamicForm.tsx
```diff
@@ -115,42 +126,43 @@ const DynamicForm = forwardRef(
       }
     }, []);
 
-    const { renderFields, getCleanedValues } = useDynamicFields({
-      definition,
-      initialValues,
-      localConstants,
-      onFieldChange,
-    });
+    const { renderFields, getCleanedValues, getValuesForSubmit } =
```
martha: Maybe, add a comment here explaining why `getCleanedValues` is returned separately from the hook even though it's also called by `getValuesForSubmit`? My understanding is that `getCleanedValues` is needed by the `MciClearance` component and that's why it's passed to the context here. It seems like `MciClearance` also calls `createHudValuesForSubmit` internally too, so maybe it would be desirable to clean this up further by just making `MciClearance` receive `getValuesForSubmit` from the context instead of `getCleanedValues`?

It was confusing to me that we refer to the result of `getCleanedValues` as `rawValues`, which is part of why I'm wondering if it's possible to further clean up -- maybe if `getCleanedValues` were only internal to the `useDynamicFields` hook it would be easier to follow. 

The comments you've added showing how these objects look are really helpful, thanks 🙏

martha (approved):
I think this is a great refactor, thanks!

martha:
I noticed a caching thing that I'm not sure it is worthwhile to fix, but I'll note it here:
- Save assessment in progress with Form V1
- Publish Form V2
- Go back to assessment and submit using V1
- Go back to assessment again and unlock for editing. Should now see V2, but you have to reload the page first to refresh the cache.

src/modules/form/components/DynamicForm.tsx
gigxz: Thanks, good point. I renamed `getCleanedValues` to `getValues` since the values are not really "cleaned", they just have the hidden fields dropped. I don't want to mess too much with the MCI Clearance component, so I left that function in the context. I also added a comment here explaining the usage of getValues vs getValuesForSubmit.

gigxz:
Thanks good spot. I think the cache issue is OK. Publishing is rare, I think it's OK that someone only gets the latest form once their cache refreshes (session timeout / log out)

PR #903, opened 2024-08-22
Fix form builder bugs and grumps

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6400

This PR fixes the following bugs/grumps noted in the ticket:
- Deleting a draft, when there is no published version "underneath", leads to a "page not found."
- Update action button text
  - I also added Ash's suggested mini-text with info about last publication/edit date and user
  - I followed Ash's suggestion to move the JSON Editor button into the Form Builder, though this is extremely low impact since that button won't be visible to non-GR users
- Items with extra long labels cause horizontal scroll
- “Save Draft” slide should stick to bottom of screen, not float in the middle
- Add version numbers to history
- The central form tree panel should be able to scroll independently of the left-hand nav and the right-hand palette

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/components/layout/dashboard/DashboardContentContainer.tsx
+++ b/src/components/layout/dashboard/DashboardContentContainer.tsx
@@ -9,6 +9,7 @@ import ContextHeader from './contextHeader/ContextHeader';
 import DashboardContentNav from './DashboardContentNav';
 
 import { useIsMobile } from '@/hooks/useIsMobile';
+import useMaxPageWidth from '@/hooks/useMaxPageWidth';
 import SentryErrorBoundary from '@/modules/errors/components/SentryErrorBoundary';
 
 interface Props {
@@ -45,7 +46,7 @@ const DashboardContentContainer: React.FC<Props> = ({
   contentSx,
 }) => {
   const theme = useTheme();
-  const maxPageWidth = theme.breakpoints.values.xl;
+  const maxPageWidth = useMaxPageWidth();
 
   const desktopContainerWidth = desktopNavIsOpen
     ? // FIXME using vw causes horizontal scrollbar when vertical scroll is present
--- a/src/modules/admin/components/forms/FormDefinitionDetailPage.tsx
+++ b/src/modules/admin/components/forms/FormDefinitionDetailPage.tsx
@@ -15,6 +15,7 @@ import NotFound from '@/components/pages/NotFound';
 import useSafeParams from '@/hooks/useSafeParams';
 import EditFormButton from '@/modules/admin/components/forms/EditFormButton';
 import HmisEnum from '@/modules/hmis/components/HmisEnum';
+import { parseAndFormatDate } from '@/modules/hmis/hmisUtil';
 import { RootPermissionsFilter } from '@/modules/permissions/PermissionsFilters';
 import { AdminDashboardRoutes } from '@/routes/routes';
 import { HmisEnums } from '@/types/gqlEnums';
@@ -25,11 +26,12 @@ import {
 } from '@/types/gqlTypes';
 
 const FormStatusText: React.FC<{
-  identifer: FormIdentifierDetailsFragment;
-}> = ({ identifer }) => {
-  const isPublished = identifer.displayVersion.status === FormStatus.Published;
-  const hasDraft = !!identifer.draftVersion;
-  const isRetired = identifer.displayVersion.status === FormStatus.Retired;
+  identifier: FormIdentifierDetailsFragment;
+}> = ({ identifier }) => {
+  console.log(identifier);
+  const isPublished = identifier.displayVersion.status === FormStatus.Published;
+  const hasDraft = !!identifier.draftVersion;
+  const isRetired = identifier.displayVersion.status === FormStatus.Retired;
 
   if (isPublished && hasDraft) {
     // Form is currently published, but there is also a Draft
@@ -79,7 +81,20 @@ const FormDefinitionDetailPage = () => {
 
   const isPublished =
     formIdentifier.displayVersion.status === FormStatus.Published;
+  const publishedBy = isPublished
+    ? formIdentifier.displayVersion.updatedBy
+    : undefined;
+  const publishedOn = isPublished
+    ? formIdentifier.displayVersion.dateUpdated
+    : undefined;
+
   const hasDraft = !!formIdentifier.draftVersion;
+  const draftUpdatedBy = hasDraft
+    ? formIdentifier.draftVersion?.updatedBy
+    : undefined;
+  const draftUpdatedOn = hasDraft
+    ? formIdentifier.draftVersion?.dateUpdated
+    : undefined;
 
   return (
     <>
@@ -108,7 +123,7 @@ const FormDefinitionDetailPage = () => {
                   />
                 </CommonLabeledTextBlock>
                 <CommonLabeledTextBlock title='Status'>
-                  <FormStatusText identifer={formIdentifier} />
+                  <FormStatusText identifier={formIdentifier} />
                 </CommonLabeledTextBlock>
               </Stack>
             </CommonCard>
@@ -127,6 +142,12 @@ const FormDefinitionDetailPage = () => {
                 >
                   View Published Form
                 </ButtonLink>
+                {isPublished && (
+                  <Typography variant='body2'>
+                    Published on {parseAndFormatDate(publishedOn)}{' '}
+                    {publishedBy && `by ${publishedBy.name}`}
+                  </Typography>
+                )}
                 <RootPermissionsFilter permissions='canManageForms'>
                   <Divider />
                   <EditFormButton
@@ -135,19 +156,25 @@ const FormDefinitionDetailPage = () => {
                     variant='outlined'
                   />
                   {hasDraft && (
-                    <ButtonLink
-                      to={generatePath(
-                        AdminDashboardRoutes.PREVIEW_FORM_DRAFT,
-                        {
-                          identifier: formIdentifier.identifier,
-                          formId: formIdentifier.draftVersion?.id || '',
-                        }
-                      )}
-                      variant='contained'
-                      fullWidth
-                    >
-                      Preview / Publish Draft
-                    </ButtonLink>
+                    <>
+                      <ButtonLink
+                        to={generatePath(
+                          AdminDashboardRoutes.PREVIEW_FORM_DRAFT,
+                          {
+                            identifier: formIdentifier.identifier,
+                            formId: formIdentifier.draftVersion?.id || '',
+                          }
+                        )}
+                        variant='contained'
+                        fullWidth
+                      >
+                        Preview / Publish Draft
+                      </ButtonLink>
+                      <Typography variant='body2'>
+                        Last edited on {parseAndFormatDate(draftUpdatedOn)}{' '}
+                        {draftUpdatedBy && `by ${draftUpdatedBy.name}`}
+                      </Typography>
+                    </>
                   )}
                 </RootPermissionsFilter>
               </Stack>
--- a/src/modules/admin/components/forms/FormDefinitionDetailPage.tsx
+++ b/src/modules/admin/components/forms/FormDefinitionDetailPage.tsx
@@ -15,7 +15,10 @@ import NotFound from '@/components/pages/NotFound';
 import useSafeParams from '@/hooks/useSafeParams';
 import EditFormButton from '@/modules/admin/components/forms/EditFormButton';
 import HmisEnum from '@/modules/hmis/components/HmisEnum';
-import { parseAndFormatDate } from '@/modules/hmis/hmisUtil';
+import {
+  formatRelativeDateTime,
+  parseHmisDateString,
+} from '@/modules/hmis/hmisUtil';
 import { RootPermissionsFilter } from '@/modules/permissions/PermissionsFilters';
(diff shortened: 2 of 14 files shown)


src/components/layout/dashboard/DashboardContentContainer.tsx
```diff
@@ -45,7 +46,7 @@ const DashboardContentContainer: React.FC<Props> = ({
   contentSx,
 }) => {
   const theme = useTheme();
-  const maxPageWidth = theme.breakpoints.values.xl;
+  const maxPageWidth = useMaxPageWidth();
```
martha, line 49: `FormBuilder.tsx` was already defining its own `maxPageWidth` that was supposed to match the styles used by `DashboardContentContainer.tsx` (see the [comment](https://github.com/greenriver/hmis-frontend/pull/903/files#diff-d278cce8d05aa45f7d08250bb57d807cc320462c5e0ef165263b095bf4561a7dR140-R141)). But for some reason it was set to `lg` instead of `xl`. I think they should be kept consistent, so I factored it out into a common hook.

src/modules/admin/components/forms/FormDefinitionDetailPage.tsx
```diff
@@ -109,52 +123,60 @@ const FormDefinitionDetailPage = () => {
...
-                  >
-                    Preview / Publish Draft
-                  </ButtonLink>
-                  <EditFormButton
-                    formIdentifier={formIdentifier}
-                    text={hasDraft ? 'Edit Draft' : 'New Draft'}
-                    editorType={FormEditorType.FormBuilder}
-                    variant='outlined'
-                  />
-                  <RootPermissionsFilter permissions='canAdministrateConfig'>
```
martha, line 138: Removing the JSON editor button from this panel and tucking it into the Form Builder was a [suggestion from Ash](https://www.figma.com/board/axGI9vwmuBRC4uZ11ZJ3RP/OP-HMIS-%2F-5792-%3A-Form-Builder-Workflow?node-id=1125-8863&t=AeD7zuvyJ9FM0tPz-0). It doesn't matter much since it won't be client-facing, but I went for it because I think it has the added advantage of removing some complexity from the `EditFormButton` above.

src/modules/admin/components/forms/FormDefinitionDetailPage.tsx
```diff
@@ -109,52 +123,60 @@ const FormDefinitionDetailPage = () => {
...
-                    />
-                  </RootPermissionsFilter>
-                </RootPermissionsFilter>
                 <ButtonLink
                   to={generatePath(AdminDashboardRoutes.PREVIEW_FORM, {
                     identifier: formIdentifier.identifier,
                     formId: formIdentifier.displayVersion.id,
                   })}
-                  variant='outlined'
+                  variant={hasDraft ? 'outlined' : 'contained'}
```
martha, line 141: Suggestion from me: When there is a draft, the primary action is "Preview / Publish Draft". When there's no draft, the primary action is "View Published Form".
Background: [Figma](https://www.figma.com/board/axGI9vwmuBRC4uZ11ZJ3RP?node-id=1125-8806#921966539), related [Slack](https://greenriver.slack.com/archives/C033YNSN0Q1/p1724174821969259)

src/modules/admin/components/forms/FormDefinitionDetailPage.tsx
```diff
@@ -12,25 +13,25 @@ import PageTitle from '@/components/layout/PageTitle';
...
-  FormIdentifierDetailsFragment,
 } from '@/types/gqlTypes';
 
 const FormStatusText: React.FC<{
-  identifer: FormIdentifierDetailsFragment;
-}> = ({ identifer }) => {
-  const isPublished = identifer.displayVersion.status === FormStatus.Published;
-  const hasDraft = !!identifer.draftVersion;
-  const isRetired = identifer.displayVersion.status === FormStatus.Retired;
+  identifier: FormIdentifierDetailsFragment;
```
martha, line 32: No change here, just a typo fix

src/modules/admin/components/forms/FormDefinitionDetailPage.tsx
```diff
@@ -80,7 +81,20 @@ const FormDefinitionDetailPage = () => {
 
   const isPublished =
     formIdentifier.displayVersion.status === FormStatus.Published;
+  const publishedBy = isPublished
```
martha, line 86: This might be contentious - I'm not 100% sure about the assumption that the last person to update a published form was indeed the person to publish it. (We get the person from the [audit history](https://github.com/greenriver/hmis-warehouse/blob/11be216286e6b81d4e1367813d8ea53bb3a69b59/drivers/hmis/app/graphql/types/forms/form_definition.rb#L117))

I feel like that assumption _should_ be true because you can't edit a published form... but maybe making the assumption is brittle?

src/modules/formBuilder/components/FormBuilder.tsx
```diff
@@ -124,15 +126,24 @@ const FormBuilder: React.FC<FormBuilderProps> = ({
         component='form'
         onSubmit={rhfMethods.handleSubmit(updateForm)}
       >
-        <Box sx={{ flexGrow: 1 }}>
+        <Box
+          sx={{
+            flexGrow: 1,
+            // Can't exceed 100% width of the container minus width of the Palette
+            maxWidth: 'calc(100% - 52px)',
```
martha, line 133: This is to prevent the long labels causing horizontal scroll. Normal pages that use `DashboardContentContainer` don't need this because of the max width set [here](https://github.com/greenriver/hmis-frontend/blob/ced6ce01206afe925d8005af6251af9d3f89bfaf/src/components/layout/dashboard/DashboardContentContainer.tsx#L75), I think. But we circumvented that in the FormBuilder in order to pin the Palette on the right, as described below.

Let's discuss: it's a bit of a code smell to have `52px` hard coded here -- it comes from the current width of the palette items + their padding, which could change.

src/modules/formBuilder/components/FormBuilder.tsx
```diff
@@ -124,15 +126,24 @@ const FormBuilder: React.FC<FormBuilderProps> = ({
...
           <Box
             sx={{
               // Matches the styles usually applied in DashboardContentContainer.
               // (Moved in here because of the Palette drawer)
-              maxWidth: `${theme.breakpoints.values.lg}px`,
+              maxWidth: `${maxPageWidth}px`,
               pt: 2,
               pb: 8,
               px: { xs: 1, sm: 3, lg: 4 },
+              flexGrow: 1,
```
martha, line 146: This is to prevent the SaveSlide from creeping up the page, it should stick to the bottom

src/modules/formBuilder/components/FormBuilder.tsx
```diff
@@ -171,7 +182,12 @@ const FormBuilder: React.FC<FormBuilderProps> = ({
               </Box>
             </Paper>
           </Box>
-          <SaveSlide in={isDirty} direction='up' loading={saveLoading}>
+          <SaveSlide
+            in={isDirty}
+            direction='up'
+            loading={saveLoading}
+            unmountOnExit
```
martha, line 189: `unmountOnExit` prevents the SaveSlide from adding extra padding to the bottom of the page when it isn't active. I wasn't sure if this prop should be added to the SaveSlide component itself, but decided not to mess with it since it's used in many places and would require more qa.

src/modules/formBuilder/components/FormBuilderHeader.tsx
```diff
@@ -50,15 +52,23 @@ const FormBuilderHeader: React.FC<FormEditorHeaderProps> = ({
       }),
     });
 
+  const isFirstDraft = formDefinition.version === '0';
```
martha, line 55: Curious to hear your thoughts on deriving this from the form version -- maybe it's a bit brittle. We could also add something to the schema coming from the backend if that feels more certain, but I wasn't sure it was worth the extra API complexity.

src/modules/formBuilder/components/FormBuilderHeader.tsx
```diff
@@ -98,6 +119,22 @@ const FormBuilderHeader: React.FC<FormEditorHeaderProps> = ({
...
+              <Box>
+                <Typography variant='body1'>
+                  Are you sure you want to delete this draft?
+                </Typography>
+                <Typography variant='body1'>
+                  This action cannot be undone.
+                </Typography>
+                {isFirstDraft && (
+                  <Typography variant='body1' sx={{ mt: 2 }}>
+                    This form has no previously published versions, so
```
martha: New suggested text from me, seeking design feedback: https://www.figma.com/board/axGI9vwmuBRC4uZ11ZJ3RP?node-id=1092-9943#898549210

src/modules/formBuilder/components/FormBuilderPalette.tsx
```diff
@@ -119,22 +119,21 @@ interface FormBuilderPaletteType {
 const FormBuilderPalette: React.FC<FormBuilderPaletteType> = ({
   onItemClick,
 }) => {
+  const top = STICKY_BAR_HEIGHT + CONTEXT_HEADER_HEIGHT;
+
   return (
-    <Drawer
-      variant='persistent'
-      open={true}
+    <Paper
```
martha, line 125: This didn't actually need to be a Drawer -- I simplified it to just Paper in the course of troubleshooting css, and ended up liking the simplification, so I'm proposing to keep it.

src/modules/formBuilder/components/FormBuilderPalette.tsx
```diff
@@ -119,22 +119,21 @@ interface FormBuilderPaletteType {
...
-          }px)`,
-          position: 'static',
-          borderTop: 'none',
-          boxSizing: 'border-box',
-          p: 1,
-          pt: 2,
-        },
+        height: `calc(100vh - ${top}px)`,
+        position: 'sticky',
+        left: 'auto',
```
martha, line 130: For some reason setting `right: 0` leaves the palette hanging out in the middle of the screen when the left-hand admin dashboard is toggled closed. `left: auto` seemed to do the trick though, I haven't observed any similar bugs with it.

src/modules/formBuilder/components/FormBuilderHeader.tsx
gigxz: Yeah this seems fine

src/modules/admin/components/forms/FormDefinitionDetailPage.tsx
```diff
@@ -12,25 +13,25 @@ import PageTitle from '@/components/layout/PageTitle';
...
 
 const FormStatusText: React.FC<{
-  identifer: FormIdentifierDetailsFragment;
-}> = ({ identifer }) => {
-  const isPublished = identifer.displayVersion.status === FormStatus.Published;
-  const hasDraft = !!identifer.draftVersion;
-  const isRetired = identifer.displayVersion.status === FormStatus.Retired;
+  identifier: FormIdentifierDetailsFragment;
+}> = ({ identifier }) => {
+  console.log(identifier);
```
gigxz: Remove console log?

src/modules/admin/components/forms/FormDefinitionDetailPage.tsx
gigxz: Yep thats how we are using those fields  in the version history. There is no other mechanism for updating forms. I agree there could be drawbacks (for example, if we need to update forms in a migration and accidentally set updated_at in the process). Such a change could be rolled back though since we use paper trail. I don't think it makes sense to change this pattern at this stage

src/modules/admin/components/forms/FormDefinitionDetailPage.tsx
```diff
@@ -109,52 +123,60 @@ const FormDefinitionDetailPage = () => {
...
                 >
-                  Preview Published
+                  View Published Form
                 </ButtonLink>
+                {isPublished && (
+                  <Typography variant='caption'>
+                    Published on {parseAndFormatDate(publishedOn)}{' '}
+                    {publishedBy && `by ${publishedBy.name}`}
+                  </Typography>
+                )}
```
gigxz: We have the _exact_ dates and times down in the version history table, so it might be more useful to show a relative date here `formatRelativeDateTime(parseHmisDateString(publishedOn))`. Since it seems like it's more for understanding, at a glance, how old the form is.

src/modules/formBuilder/components/FormBuilder.tsx
gigxz: Yeah good point. If we need to have a fixed width like this, I think its better if we pull it out into a constant `PALETTE_WIDTH`, and use that constant to fix the actual width on the palette, too. That way, we (hopefully) won't run into the scenario where the palette width inadvertently changes, and this 52px gets out of sync.

gigxz (approved):
Looks good! left some minor suggestions but would be fine with merging as is

src/modules/admin/components/forms/FormDefinitionDetailPage.tsx
martha: Sounds good.

I noticed while clicking around dev that this displays a publication date for forms like Enrollment and Client. It looks like it displays the date of the last time the form was updated, for example using seed_definitions. Is this okay?

gigxz: Eh, not ideal, but it's OK for now

PR #906, opened 2024-08-27
Fix bad keys in form builder

gigxz:
## Description

Fix incorrect keys that were leading to buggy behavior https://github.com/open-path/Green-River/issues/6512
The `id` used in all cases is a unique id generated by useFieldArray: https://react-hook-form.com/docs/usefieldarray

## Type of change
- [x] Bug fix

### DIFF
--- a/src/modules/formBuilder/components/itemEditor/conditionals/AutofillProperties.tsx
+++ b/src/modules/formBuilder/components/itemEditor/conditionals/AutofillProperties.tsx
@@ -29,7 +29,7 @@ const AutofillProperties: React.FC<AutofillPropertiesProps> = ({
       {fields.map((value, index) => (
         <RemovableCard
           onRemove={() => remove(index)}
-          key={JSON.stringify(value)}
+          key={value.id}
           removeTooltip='Remove Autofill'
         >
           <AutofillValueCard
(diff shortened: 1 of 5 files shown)


martha:
@gigxz this looks great and is a change we should for sure merge. I think it doesn't 100% fix the issue, I'm investigating a couple of odd bugs:
- Add a condition to an item and save. Reopen the item and clear the "Dependent Question" field. "Operator" is not cleared, but if you debug the question, you can see that it _is_ set to undefined in the form state. When you select a new "Dependent Question," you have to make some change to "Operator" in order to get the answer input to appear.
  - I'm not sure of the priority of this fix. It is an edge case and there is a workaround, but it seems not-unlikely to come up in regular use.
  - using `setValue` to `null` instead of `undefined`, which is similar to what I tried [here](https://github.com/greenriver/hmis-frontend/pull/899/files#diff-2368e4c27ae82cd3c902453efcfd59df5a1e85dab95f44b903dec4569284ce4eR178), appears to make it behave normally.
- Add a condition to an item and save. Reopen the item and clear "Dependent Question" and "Operator." Hit "Remove condition" and then "Add Condition." The dependent question and operator former values reappear.
  - My first thought was: this could be happening because the component unmounts (`shouldUnregister` now works, great!) and then when it re-mounts it reuses `defaultValues`. However, the answer input's value _doesn't_ reappear when re-adding the condition, which I think it probably would if this were a default values issue..
  - This seems low priority to me to fix, it's an edge case and there is a workaround.

gigxz:
I thank that is because we call `setValue(undefined)` for the operator, and it's not allowed to pass undefined to setValue https://react-hook-form.com/docs/useform/setvalue
I think using null would work?

gigxz:
Thanks @martha , I pushed the change for setting to null isntead of undefined which works as you describe. Agree the other issue is low priority

martha:
Thanks @gigxz `setValue` to `null` works if you are fine with the cast to `any`

(`resetField` resets to the default value unless you explicitly supply a new default value, but this is also not allowed to be `undefined`.)

martha (approved):

PR #909, opened 2024-09-03
View External Form Submissions by values, not CDEs

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6546
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4657

This PR
- updates the External Submissions review ui to be compatible with the new processing mechanism, which doesn't create CDEs until after the submission is reviewed

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/form/hooks/useStaticFormDialog.tsx
```diff
@@ -86,8 +87,8 @@ export function useStaticFormDialog<
         >
           <DialogTitle>{title}</DialogTitle>
           <DialogContent>
+            <Box mt={2}>{beforeFormComponent}</Box>
```
martha, line 90: fixes visual bug: 
Screenshot 2024-09-03 at 10 48 42 AM

src/modules/projects/components/tables/ProjectExternalSubmissionsTable.tsx
```diff
@@ -87,26 +79,31 @@ const ProjectExternalSubmissionsTable = ({
...
   }, []);
 
   const [selected, setSelected] =
     useState<ExternalFormSubmissionFieldsFragment | null>(null);
 
+  const { data: { externalFormDefinition: definition } = {}, error } =
+    useGetExternalFormDefinitionQuery({
+      variables: {
+        formDefinitionIdentifier: formDefinitionIdentifier,
+        id: selected?.definition?.id,
```
gigxz: Let's discuss whether it makes sense to use an approach more similar to assessments, where the definition to use for viewing the assessment is loaded on the assessment object itself:
```
externalFormSubmission(id: 3) {
    definition {... } // form_submission.form_processer.definition_id reference
    values {... }
}
```

src/modules/projects/components/tables/ProjectExternalSubmissionsTable.tsx
```diff
@@ -84,29 +69,37 @@ const ProjectExternalSubmissionsTable = ({
...
     Object.keys(itemMap).forEach((key) => {
       const item = itemMap[key];
       const customFieldKey = item.mapping?.customFieldKey;
-      if (!customFieldKey) return;
 
-      const value = customDataElementValueForKey(
-        customFieldKey,
-        selected.customDataElements
-      );
+      // TODO - this doesn't support reading the record type from the parent's record type. see transformSubmitValues
```
martha: Let's discuss this -- I think the impact is that with what I have currently, external forms don't support setting a record type on a Group and then having the children of that group use that record type without needing to individually specify. It looks(?) like regular forms do support that. Do we use it anywhere? Do we plan to on the PIT?

gigxz: @martha I think that was previous behavior (before we introduced the `mapping` object) that is no longer needed. I just verified in the [schema](https://github.com/greenriver/hmis-warehouse/blob/release-132/drivers/hmis_external_apis/public/schemas/form_definition.json) that it's not valid to have a mapping on a group, so we shouldn't need to worry about that and can probably pull the `parentRecordType` code out of `transformSubmitValues`.

martha: Sounds good thanks! I removed the todo. Should I pull that out of `transformSubmitValues` on this PR or separately?

gigxz: I think a separate PR would be wise. THanks!

src/modules/form/components/viewable/DynamicViewField.tsx
```diff
@@ -115,8 +115,12 @@ const DynamicViewField: React.FC<DynamicViewFieldProps> = ({
         <TextContent
           {...commonProps}
           value={
-            [TRUE_OPT, FALSE_OPT].find((o) => o.code === String(value))
-              ?.label || <NotCollectedText variant='body2' />
+            [TRUE_OPT, FALSE_OPT].find((o) => {
+              return (
+                o.code === String(value) ||
+                String(o.numericCode) === String(value)
```
gigxz: I'm not sure about this. "numeric code" was introduced as a way to attach a different "score" to each pick list option. These are simple yes/no's, they don't need a score associated with it.

This is a downstream effect of the fact that External Forms collect boolean values as `1/0`, while HMIS forms collect them as `true/false`. I think it would be better to normalize that behavior instead. (Either adjusting External Forms to collect as booleans, or normalize values before storing them in form processor?).

martha: Discussed on a call. Decision: Stop using `numericCode` (it's unrelated), but just special-case the 0/1 display

martha: Use YesNoDisplay

src/modules/projects/components/tables/ProjectExternalSubmissionsTable.tsx
```diff
@@ -132,6 +144,7 @@ const ProjectExternalSubmissionsTable = ({
     getVariables: (values) => {
       return {
         id: selected?.id || '', // selected should never be undefined when dialog is open
```
gigxz:
```suggestion
        id: selectedId || '', // selected should never be undefined when dialog is open
```
`selected` is null while useGetExternalFormSubmissionQuery is fetching. So it's possible to hit submit while the view is loading, which results in an error

gigxz (approved):
Looks good!

PR #914, opened 2024-09-06
Switch to vitest for testing, remove jest

gigxz:
## Description

* Switch from jest to vitest for unit tests
* I didn't enable globals, which is why the explicit imports are added. It's possible, though, if others have a preference towards using globals
* Remove App.test.ts which isn't testing anything beyond coverage we already have in our system tests

https://github.com/open-path/Green-River/issues/6529

## Type of change
- [x] Dependency update


martha (approved):
Cool! It seems significantly faster too, which is great.

Is it intentional that in dev this is running tests in [watch mode](https://vitest.dev/guide/features#watch-mode)? It's fine by me, but a behavior change from before. I'm not sure if I will really use this for developing, but maybe, if our test suite becomes more robust.

Also, running `yarn` locally on this branch generates a diff for me. Do you see that too?

gigxz:
Just pushed the yarn lock change, thanks!

Good point, watching is the default behavior for vitest. There is some discussion on that here https://github.com/vitest-dev/vitest/discussions/540.

I updated it to match the previous behavior by adding a setting to vite.config.js.
```sh
yarn test # run test suite
yarn test -w # run and watch test suite
```

PR #922, opened 2024-09-13
Adjust error logic for Form Builder bounds

martha:
## Description

Fixes the following 2 related bugs:
1. Log in as a Super Admin. In the Form Builder, create a new Bound on an item. Save without specifying the bound value. It is possible to save and hit the backend error message, since the fields are not marked as required.
- Log in as a non Super Admin. In the Form Builder, create a new Bound on a Date item. Save without specifying the bound value. Same here, you see the unintuitive backend error message

Follows the pattern we created here in ManagePickListOptions: https://github.com/greenriver/hmis-frontend/blob/c07299eb9db6f99018c574e78890b69f1bd2e70c/src/modules/formBuilder/components/itemEditor/pickLists/ManagePickListOptions.tsx#L125-L127

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/modules/formBuilder/components/itemEditor/conditionals/ValueBoundCard.tsx
+++ b/src/modules/formBuilder/components/itemEditor/conditionals/ValueBoundCard.tsx
@@ -1,6 +1,6 @@
 import { Typography } from '@mui/material';
 import { Stack } from '@mui/system';
-import React from 'react';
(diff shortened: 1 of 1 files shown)


gigxz (approved):
thanks!

PR #925, opened 2024-09-13
Use implicit paths in storybook

gigxz:
## Description

1. switch to implicit paths https://github.com/open-path/Green-River/issues/6689
2. clean up some import warnings by updating all storybook deps to the latest version (minor)

This makes the Storybook pretty messy, so we need some follow-up work to move some code files around. I would like to have a team discussion about code organization before that. 

## Type of change


.storybook/main.ts
```diff
@@ -27,7 +27,13 @@ const config = {
       },
     });
   },
-  stories: ['../src/**/*.stories.mdx', '../src/**/*.stories.@(js|jsx|ts|tsx)'],
+  stories: [
+    // Use implicit paths for all stories
+    '../src/**/*.stories.@(js|jsx|ts|tsx)',
+    // For a better structure, need to reorganize some things. Should first confirm approach that "src" has comment element components and "modules" have components for specific pages or modules, which is the vague pattern right now.
+    // - Move src/components/clientDashboard components into module(s)
```
martha, line 34: Sorry for the disorganization! In a way this definition is a bit unintuitive to me, since "module" maybe implies some standard part that can be reused in different contexts, as in "modular"? But upon researching a little it seems like it's the convention, right? So I am fine with it 👍 

Under that convention I agree with moving `pages/` into `modules/`.

Other misc questions about our app structure/conventions, now that I'm looking at it -- these aren't related to storybook though: 
- What goes in `scripts` vs. in util directories like `src/utils` or `src/modules/form/util`, or even other inlined util files like hmisUtil.ts?
- Should we move  `api/operations` -> `api/` ? (since it's not just operations in there)

.storybook/main.ts
```diff
@@ -27,7 +27,13 @@ const config = {
       },
     });
   },
-  stories: ['../src/**/*.stories.mdx', '../src/**/*.stories.@(js|jsx|ts|tsx)'],
+  stories: [
```
martha, line 30: This isn't related to this PR but I think comes from some recent storybook-related change. I keep getting this file showing up in my diff when touching PRs related to storybook: `.storybook/main.js`. It looks generated to me, but I am not sure what's generating it. Do you get this too and do you know why it happens?

martha (approved):
Great change thank you 👍

.storybook/main.ts
ttoomey: +1 for conventions. Did you find documentation or other reference for a convention?

ttoomey (approved):
looks good, maybe needs a yarn dedup?

.storybook/main.ts
gigxz: Yeah we can discuss! This was more of a placeholder note.

The `modules` structure was based off another GR application. Imo it should be similar to `features/` in [bulletproof react](https://github.com/alan2207/bulletproof-react/blob/master/docs/project-structure.md) structure., but we don't follow that super consistently. I can come up with some proposed structure documentation  that we can review together

.storybook/main.ts
gigxz: Nope I haven't gotten that. Do you know what command generates the file for you? I would try: yarn, yarn storybook, yarn build-storybook

gigxz:
it's already been dedup'ed https://github.com/greenriver/hmis-frontend/pull/925/commits/51a922941072f81fac5848e4cf18b97f05d9751b, merging

PR #933, opened 2024-09-18
Create UX for bulk external form submission review

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6598
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4713/files
- Add schema updates and frontend chrome for the bulk review mutation
- Adjust the page title when a form has been autoselected
- Hide the "Mark reviewed" checkbox if it's already been reviewed
- Update the "Date Submitted" column to show a relative date
- Add an Action column to open the modal, rather than having the rows be clickable to open the modal. This is to avoid having rows that are both clickable and checkbox-selectable.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/projects/components/tables/ProjectExternalSubmissionsTable.tsx
```diff
@@ -212,20 +245,36 @@ const ProjectExternalSubmissionsTable = ({
...
         noData='No external form submissions'
         pagePath='project.externalFormSubmissions'
         recordType='ExternalFormSubmission'
         paginationItemName='submission'
         filters={filters}
-        handleRowClick={(row) => {
-          setSelectedId(row.id);
-          openFormDialog();
+        EnhancedTableToolbarProps={{
+          renderBulkAction: (selectedIds, selectedRows) => (
```
martha, line 150: I haven't passed a title here, so the blank space is a bit awkward when no rows are selected:
Screenshot 2024-09-18 at 8 59 04 AM

src/modules/projects/components/tables/ProjectExternalSubmissionsTable.tsx
```diff
@@ -170,6 +137,71 @@ const ProjectExternalSubmissionsTable = ({
...
+                  variant='outlined'
+                  sx={{ ml: 1, color: theme.palette.error.dark }}
+                />
+              )}
+            </>
+          );
+        },
+      },
+      {
+        header: 'Date Submitted',
```
martha: Should we update the "Date Submitted" column title now that it's a relative time?

src/modules/projects/components/tables/ProjectExternalSubmissionsTable.tsx
gigxz: In other places, like https://hmis.dev.test:5173/admin/client-merge-history/candidates, we just add a secondary title there. I agree it's weird though, I added a note for the table design session. For now can you just add a title like "Form Submissions"?

src/modules/projects/components/tables/ProjectExternalSubmissionsTable.tsx
gigxz: We could use this existing component which displays both time and relative time, and keep the label the same
```
<RelativeDateTableCellContents
            dateTimeString={s.submittedAt}
            horizontal
          />
```

src/modules/projects/components/tables/ProjectExternalSubmissionsTable.tsx
```diff
@@ -212,20 +245,36 @@ const ProjectExternalSubmissionsTable = ({
...
+              onClick={() => {
+                bulkUpdate({
+                  variables: {
+                    ids: selectedIds as string[],
+                  },
+                });
+              }}
+              loading={bulkLoading}
+            >
+              Convert {selectedRows.length} to Client
```
gigxz: this language isn't quite right, since some form types don't generate clients. Maybe `Bulk Review ({selectedRows.length}) Submissions`? (parens to match bulk services style)

src/modules/projects/components/tables/ProjectExternalSubmissionsTable.tsx
```diff
@@ -170,6 +137,71 @@ const ProjectExternalSubmissionsTable = ({
...
+        render: (s: ExternalFormSubmissionSummaryFragment) => {
+          const parsedDate = parseHmisDateString(s.submittedAt);
+          if (parsedDate) return formatRelativeDateTime(parsedDate);
+          return '';
+        },
+      },
+      {
+        header: 'Action',
+        render: (s: ExternalFormSubmissionSummaryFragment) => (
+          <Button
```
gigxz: I think it's probably best to disable this button while the bulk mutation is loading. Then we don't need to worry about the situation where a user initiates a bulk action, and while that's loading, they also perform an individual action.

src/modules/projects/components/tables/ProjectExternalSubmissionsTable.tsx
martha: great!

src/modules/projects/components/tables/ProjectExternalSubmissionsTable.tsx
martha: Yes much better, good point that it's not generic enough for all external forms. I also like this better because it's more consistent with the language we use inside the dialog when you are doing the review action for 1 individual submission.

src/modules/projects/components/ExternalSubmissionsViewModal.tsx
```diff
@@ -134,17 +134,19 @@ const ExternalSubmissionsModalContent = ({
...
+                    checked
+                      ? ExternalFormSubmissionStatus.Reviewed
+                      : ExternalFormSubmissionStatus.New
+                  );
+                }}
+              />
+            )}
             <LabeledCheckbox
               value={spam}
               label='Mark as Spam'
```
martha: I wonder if we should also disable 'Mark as Spam' when the submission has already been reviewed. It seems meaningless to mark a submission as spam at that point, if the cdes/client/enrollment have already been created and we aren't taking any action to delete them. All it does is hide the submission from the default view of this table.

gigxz: yeah I agree, we could hide it

gigxz (approved):

src/components/elements/RouterLink.tsx
```diff
@@ -30,6 +30,7 @@ const RouterLink = forwardRef<RouterLinkProps, any>(
           gap={0.75}
           alignItems='center'
           sx={{ display: 'inline-flex', textDecoration: 'inherit' }}
+          component='span'
```
martha, line 33: @gigxz this is to avoid `Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>.`

We can also fix it in the newly added alert component, but I think `RouterLink` should be allowed as a child of `<p>` in general, so I've added it here. It doesn't seem to impact the visual styling.

gigxz: Okay yeah this seems relatively low risk since it's already inside the `a` tag and it's only affecting links with icons

gigxz: Thanks!

PR #935, opened 2024-09-18
Add CI step for storybook test runner

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6145

Summary of changes:
- Add a CI step for the Storybook test runner
- Add a simple test that checks, for each story, whether it has any a11y violations.
  - Example build that shows the passing tests: https://github.com/greenriver/hmis-frontend/actions/runs/13317749398/job/37195788692?pr=935
- Disable certain violations, particularly contrast check, which we should roll back after our theme updates are introduced.*
  - To see how it looks when tests fail, locally, you can comment out [these lines](https://github.com/greenriver/hmis-frontend/pull/935/files#diff-0bf5908655f9e530c64c03d6084091a38f86e59e75f9266134bdf27513e2e7e8R16-R20), then run `yarn test-storybook`
- Add a note to the README about how to run a11y tests locally
- Fix a few minor issues that were causing a11y tests to fail

*Note that these violations are _not_ disabled on the Storybook Stories themselves, only in testing. So you can still see them when you go to the story:
Screenshot 2025-02-13 at 4 55 10 PM
It is possible to disable that too, but I figured it would be useful for development to still be able to see them.

## Type of change
- [ ] Bug fix
- [ ] Dependency update
- [x] CI enhancement ✨ 


.github/workflows/test.yml
```diff
@@ -52,4 +52,11 @@ jobs:
         projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
         zip: true
         exitZeroOnChanges: true
-        onlyChanged: true
\ No newline at end of file
+        onlyChanged: true
+    - name: Storybook test
```
gigxz, line 55: @martha I am wondering about the choice to user storybook test runner on CI as opposed to running tests via Chromatic on CI. The storybook gives the [example](https://storybook.js.org/docs/writing-tests/test-runner#whats-the-difference-between-chromatic-and-test-runner) that you might want to "Use [test runner] locally and Chromatic on your CI". I would guess that a benefit of Chromatic would be if the tests fail, maybe you can see some more detailed output on chromatic as opposed to just test logs. (Judging from [chromatic docs about PR workflow](https://www.chromatic.com/docs/in-pull-request/) it seems pretty advanced).

Linking https://github.com/open-path/Green-River/issues/6687 for reference too

.storybook/test-runner.ts
```diff
@@ -0,0 +1,26 @@
+// COMMENTED OUT for now until we want to actually turn on accessibility tests on CI
```
gigxz: Do the tests pass for you locally? test-storybook passed for me (which I was surprised by, if it included accessibility tests?) but I'm not 100% confident that I resolved the merge conflicts correctly

.github/workflows/test.yml
gigxz: Sorry, I didn't read all the way through your description 🙈 

> I believe, alternatively, it's possible to run tests in CI against a Chromatic instance (see [this other recipe](https://storybook.js.org/docs/writing-tests/test-runner#run-against-deployed-storybooks-via-github-actions-deployment)), but based on the existing "Run Chromatic" build step, it looks like we only want to run chromatic on push to a release branch and not on pull request.

That was set up mostly for the purpose of publishing the "release version" of the storybook. I don't have anything against using chromatic to run testing on PRs. (And I guess publishing, maybe that's the norm for chromatic usage).

.storybook/test-runner.ts
martha: We can see the accessibility failures both locally and on CI by uncommenting the contents of this file. See [this build](https://github.com/greenriver/hmis-frontend/actions/runs/11236420220/job/31236495367?pr=935), for example.

Do you want to actually turn this on now and make all our CI tests fail until we fix the a11y issues? I had assumed not yet... 

But maybe it would be a good idea to turn it on and factor it out into a separate build step, that way when it fails we can immediately recognize the failure is expected, instead of clicking into the gh action log to make sure it's failing on the expected step?

.github/workflows/test.yml
martha: Sounds good, I updated the PR to run the tests against the Chromatic instance of Storybook.

We do still need the test runner to run the accessibility tests. Chromatic supports running interaction tests, but [doesn't support a11y tests](https://github.com/storybookjs/storybook/discussions/17966) as far as I can tell. You can see this in the example build [here](https://github.com/greenriver/hmis-frontend/actions/runs/11223801489/job/31199218726?pr=935). The test runner step shows a11y failures, but the Chromatic step succeeds. In the corresponding [Chromatic build](https://www.chromatic.com/build?appId=66e877f06b978106d75d04bc&number=34) for that action, there are 51 "unreviewed changes" (bc I had temporarily turned on the "visual tests" setting) but there are no a11y-related failures or warnings.

.storybook/test-runner.ts
gigxz: Huh okay I wasn't seeing them locally, but I'll try again.  No I agree, lets not intentionally have CI fail.

src/components/elements/Loading.tsx
```diff
@@ -11,7 +11,7 @@ const Loading = () => (
     aria-live='polite'
     aria-busy='true'
   >
-    <CircularProgress size={50} />
+    <CircularProgress aria-label='Loading' size={50} />
```
martha, line 14: Is there a reason we don't just use the usual `<Loading/>` here?

src/modules/client/providers/ClientSsnDobVisibility.tsx
```diff
@@ -71,12 +70,9 @@ export const ContextualSsnToggleButton: React.FC<ButtonProps> = (props) => (
           {...props}
           on={show}
           onToggle={() => setShow((prev) => (isNil(prev) ? true : !prev))}
-          aria-hidden
         >
           <strong>SSN</strong>
         </VisibilityToggleButton>
```
martha, line 75: This reverts a change introduced in an a11y fix released last year: https://github.com/greenriver/hmis-frontend/pull/941#discussion_r1771936182

We waffled about it at the time, so given that it's causing automated tests to now fail, I would be in favor of reverting it. We should have tried running the axe tools in browser at the time 🤦‍♀️  (and hopefully automated tests will help us catch this kind of thing in the future!)

src/components/elements/Loading.tsx
gigxz: This _is_ the `Loading` component 🙂

.storybook/test-runner.ts
```diff
@@ -0,0 +1,33 @@
...
+    await configureAxe(page, {
+      // Skip some rules for now as we work on addressing them
+      rules: [
+        {
+          // TODO(#7108, #a11y-162) - turn this back on
+          id: 'color-contrast',
+          enabled: false,
+        },
+        {
+          // TODO(#a11y-1)
```
gigxz: What is this referencing?

martha: Ah, sorry, it is referencing https://github.com/open-path/hmis-accessibility/issues/1. I will add the link in the code.

src/components/elements/Loading.tsx
martha: Ack sorry hahaha you are right, I was rushing, but meant to put this comment [here](https://github.com/greenriver/hmis-frontend/pull/935/files#diff-12793900bb3fac98fe746720c86e1a33888ef8fd9b188976a0ed7c9026ab6e1bR54) below on the `ClientEnrollmentCard`.

It looks like there are a few places in our app where we use MUI's `CircularProgress` outside of our own `Loading` component; looks like often these are in the 'disabled' color as opposed to the usual primary color.

.storybook/test-runner.ts
gigxz: ah okay, thanks!

src/components/elements/Loading.tsx
gigxz: I see, I assume that was just before we added the Loading component. My IDE is showing that ClientEnrollmentCard CircularProgress was  last updated 2 years ago, so that's probably it.

gigxz (approved):

src/components/elements/Loading.tsx
martha: Ok sounds good. I will add a note to maybe consolidate these when we update the code to match the Design Library.

PR #936, opened 2024-09-18
Display summary columns on external forms review table

gigxz:
## Description

Issue: https://github.com/open-path/Green-River/issues/6622
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4720

1. Add summary field columns
2. Add link to enrollment on the modal (it felt too busy to add it to the table, and I thought this might be sufficient? but want feedback on that!)

Screenshot 2024-09-18 at 3 50 44 PM

Screenshot 2024-09-18 at 3 50 52 PM

## Type of change
- [x] New feature (adds functionality)

## Checklist before requesting review
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [x] My code includes comments and/or descriptive variable names to help other engineers understand the intent (or not applicable)
- [x] My code follows the style guidelines of this project (eslint)
- [x] I have updated the documentation (or not applicable)
- [ ] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

PR #937, opened 2024-09-18
Change external submissions from using static form to React form

martha:
## Description

GH issue: https://github.com/greenriver/hmis-warehouse/pull/4713#discussion_r1765474000
This PR moves the External Forms review UI from a static form to a directly configured React form, since the static form doesn't seem to be making development any easier.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/projects/ExternalSubmissionsViewModal.tsx
```diff
@@ -0,0 +1,165 @@
+import {
```
gigxz: can we move this up into the "components" dir?

src/modules/projects/ExternalSubmissionsViewModal.tsx
```diff
@@ -0,0 +1,165 @@
...
+              value={status === ExternalFormSubmissionStatus.Reviewed}
+              label='Mark as Reviewed'
+              onChange={(_e, checked) => {
+                setStatus(
+                  checked
+                    ? ExternalFormSubmissionStatus.Reviewed
+                    : ExternalFormSubmissionStatus.New
+                );
+              }}
+            />
```
gigxz: Assuming that a future PR on the PIT branch will hide this checkbox for forms that have been reviewed.

If we go with my suggestion from https://github.com/greenriver/hmis-frontend/pull/936, we could replace it with something like `This form has been reviewed [and here is the Enrollment it generated, if applicable]`

gigxz (approved):
This feels much nicer, thanks!

src/modules/projects/ExternalSubmissionsViewModal.tsx
martha: Yes, sorry!

src/modules/projects/ExternalSubmissionsViewModal.tsx
martha: Yes -- good plan.

martha:
There was no separate ticket for this work outside of https://github.com/open-path/Green-River/issues/6598, the bulk of which is on `pit-external-forms` branch, but just adding a note that I qa'd this portion which was deployed with release-134 and it looks good.
- Can navigate to an external form submission
- Can submit review with a note and new status/spam value

PR #941, opened 2024-09-23
Fix a11y for ClickToShow

martha:
## Description

GH issue: https://github.com/open-path/hmis-accessibility/issues/55
This PR makes a couple of accessibility related fixes to the `ClickToShow` component and its callers. It also adds a Storybook interaction test to check the behavior. 

## Type of change
- [x] Bug fix
- [ ] Dependency update


src/components/elements/ClickToShow.stories.tsx
```diff
@@ -10,3 +11,22 @@ export default {
 type Story = StoryObj<typeof ClickToShow>;
 
 export const Default: Story = { args: { children: 'Hidden Value' } };
+
+export const Clicked: Story = {
```
martha, line 14: When discussing this [ticket](https://github.com/open-path/hmis-accessibility/issues/55) we mentioned adding an a11y test in Storybook to ensure the correct behavior.

It looks like Storybook's actual [a11y testing](https://storybook.js.org/docs/writing-tests/accessibility-testing) is limited to Axe, i.e. we are limited to configuring which auto-checks from Axe get run. We can separately add interaction tests, however, like this one, but let's discuss the approach. Here I've done it by adding a separate story to check the behavior, but I remember at our design meeting about storybook structuring we wanted to err on the side of fewer stories.

src/modules/client/providers/ClientSsnDobVisibility.tsx
```diff
@@ -66,6 +66,7 @@ export const ContextualSsnToggleButton: React.FC<ButtonProps> = (props) => (
         {...props}
         on={show}
         onToggle={() => setShow((prev) => (isNil(prev) ? true : !prev))}
+        aria-label={show ? 'Hide SSN' : 'Show SSN'}
```
martha: This one is a bit of a judgment call, I think:
- If we leave off the aria-label property, then a11y tools will just read the text on the button, "DOB" or "SSN", without giving any indication about state (is it hidden or shown?) or functionality (what does clicking this button do?)
- On the other hand, if we add this aria-label property, the a11y tool will read the whole thing, "Hide SSN" or "Show SSN" etc., every time the user navigates into this column, which is a bit annoying.

Maybe there is a better way, should these be moved out of the table entirely? Or is there a way to get the accessible text on the component to be different when it has focus vs. when it's being read out to establish context elsewhere in a table? (I wasn't able to find anything like this)

@bobbyw-ng I'm curious to get your thoughts. Here's the component: 
https://qa-hmis.openpath.host/?textSearch=test -> SSN and DOB column headers

martha: Ok, what I've proposed here after checking out the examples in this thread https://greenriver.slack.com/archives/C033YNSN0Q1/p1727203783067949 is to hide the full-column visibility toggle in screen reader contexts (where it doesn't make much sense anyway), and add a visually hidden plaintext column header.

bobbyw-ng: @martha I like the idea of hiding the full-column visibility toggle. I also think it's acceptable to repeat "Hide SSN" / "Show SSN" inside each of those cells since it's not _that_ long. At the end of the day the user still needs to understand what the toggle does across a range of contexts (i.e. within a table, or standalone elsewhere).

> Maybe there is a better way, should these be moved out of the table entirely?

I think the alternative here is to just do a hide/show for an entire column only (instead of by cell) but that leads to some trickiness of discoverability. i.e. if a user somehow jumps straight to a cell with hidden content, how will they figure out where they need to go to unhide it? Putting additional text there to inform the user lands us back at square 1 (text too long), without the benefit of being able to hide/show on the spot.

> Or is there a way to get the accessible text on the component to be different when it has focus vs. when it's being read out to establish context elsewhere in a table? (I wasn't able to find anything like this)

I did some digging and couldn't find anything either.

martha: Now that we have removed this column almost everywhere, except Bulk Services, we might be able to get away with removing the column-level toggle and only allowing users to toggle one value at a time.

gigxz (approved):

src/modules/client/providers/ClientSsnDobVisibility.tsx
martha: @gigxz thanks for merging this! I had kinda let it fall by the wayside because I figured we wanted to revisit in general once the new table patterns were implemented. Now that it is only used in Bulk Services for dob, I think I would advocate to remove the column-level toggle and only allow toggling one at a time. (The user can still differentiate clients by age, which is not toggled.) What do you think about that change? It seems nice to be able to rip out some code that is now barely used.

gigxz: Yeah I agree that would be nice! It's hard to know if it would negatively impact anyones workflow. I tend to agree, though, it shouldn't be necessary for a typical workflow. The only thing I can think of is if someone is entering data from an external bed register or something that has DOBs on it.

PR #942, opened 2024-09-25
Make small fixes and code cleanup on PIT branch

martha:
## Description

This PR makes a couple of suggested changes to the PIT external forms branch (https://github.com/greenriver/hmis-frontend/pull/916/files). See inline comments for reasoning.

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/modules/projects/components/ExternalSubmissionsViewModal.tsx
+++ b/src/modules/projects/components/ExternalSubmissionsViewModal.tsx
@@ -41,7 +41,6 @@ const ExternalSubmissionsModalContent = ({
   onClose: VoidFunction;
 }) => {
   const submissionValues = useMemo(() => {
-    if (!submission) return {};
     const itemMap = getItemMap(submission.definition.definition, true);
     const submissionValues: FormValues = {};
     Object.entries(itemMap).forEach(([key, item]) => {
@@ -96,7 +95,7 @@ const ExternalSubmissionsModalContent = ({
 
   const deleteButton = useMemo(
     () =>
-      submission && (
+      submission.status === ExternalFormSubmissionStatus.New && (
         <DeleteMutationButton<
           DeleteExternalFormSubmissionMutation,
           DeleteExternalFormSubmissionMutationVariables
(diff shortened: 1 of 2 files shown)


src/modules/projects/components/ExternalSubmissionsViewModal.tsx
```diff
@@ -143,12 +142,10 @@ const ExternalSubmissionsModalContent = ({
             </Alert>
           )}
           <Card sx={{ p: 2 }}>
-            {submission && (
-              <DynamicView
-                values={submissionValues}
-                definition={submission.definition.definition}
-              />
-            )}
+            <DynamicView
```
martha, line 145: Remove guards in this component against falsy `submission`, since with the code restructure it will only be rendered when `submission` is present.

src/modules/projects/components/ExternalSubmissionsViewModal.tsx
```diff
@@ -96,7 +95,7 @@ const ExternalSubmissionsModalContent = ({
 
   const deleteButton = useMemo(
     () =>
-      submission && (
+      submission.status === ExternalFormSubmissionStatus.New && (
```
martha, line 98: Suggestion: Hide the delete button if the submission has already been reviewed and converted to a client/enrollment.

Question: If we do this at all, should we do it for all submissions, or only ones that correspond to an enrollment?

src/modules/projects/components/tables/ProjectExternalSubmissionsTable.tsx
```diff
@@ -25,7 +25,7 @@ const ProjectExternalSubmissionsTable = ({
   projectId: string;
   formDefinitionIdentifier: string;
 }) => {
-  const [selectedId, setSelectedId] = useState<string | null>(null);
+  const [modalOpenId, setModalOpenId] = useState<string | null>(null);
```
martha, line 28: Rename to avoid confusion over what "selected" means in this table. I'm also open to alternative names, maybe `viewingId`, `viewingSubmissionId`, etc.

src/modules/projects/components/ExternalSubmissionsViewModal.tsx
gigxz: This seems ok for all submissions. It looks like we don't have acts_as_paranoid on the form submissions, meaning they can't be recovered if deleted. So probably makes sense to restrict it as you suggest

gigxz (approved):

PR #948, opened 2024-09-27
Add aspirational file structure guidelines to README

gigxz:
Add documentation for desired file structure. This won't be "true" until we do the reorginazation happens https://github.com/open-path/Green-River/issues/6848 which I am working on

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,79 +183,22 @@ src
 |
 +-- api               # graphql queries/mutations/fragments
 |
-+-- app               # application layer containing:
-|   +-- App.tsx              # main application component
-|   +-- AppProvider.tsx      # application provider that wraps the entire application
-|   +-- protectedRoutes.ts   # define routes for the authorized applciation
-+-- assets            # assets folder can contain all the static files such as images, fonts, etc.
-|
 +-- components        # shared components used across the entire application
 |
-+-- config            # global configurations, exported env variables etc.
-|
-+-- modules           # feature based modules
++-- modules           # feature-based modules
 |
 +-- hooks             # shared hooks used across the entire application
 |
-+-- providers         # shared contexts and providers? (added)
-|
 +-- test              # test utilities and mocks
 |
-+-- types             # shared types used across the application
++-- types             # shared types used across the application (auto-generated from the graphql schema)
 |
 +-- utils             # shared utility functions
 ```
 
-Important modules
-```
-app
-components             # no components, just a dir
-components/elements    # common components
-components/input       # base input components that are used across the app
-components/layout      # layout components that are used across the app
-components/table       # table components that are used across the app (could be a module or just in elements/ ?)
-components/pages       # ***keep or move to modules or src/app/pages?? these are not re-used
 
-modules:
+Inner structure of a module:
 
-modules/admin                  # everything related to top-level Admin page (Should be further split out)
-modules/assessments
-modules/audit
-modules/auth
-modules/caseNotes
-modules/client        # components used on client overview page, plus client alerts. could be divided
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -173,3 +173,98 @@ rails driver:hmis:dump_graphql_schema
...
+|
++-- components        # shared components used across the entire application
+|
++-- config            # global configurations, exported env variables etc.
+|
++-- modules           # feature based modules
+|
++-- hooks             # shared hooks used across the entire application
+|
++-- providers         # shared contexts and providers? (added)
```
martha: [to discuss tomorrow] Which contexts and providers will live in here? seems like `AppProvider` will move to `app/`, our Theme Provider lives under config (will it stay there?). The remaining apollo client code, I'm not sure if I think of it as a provider? (but maybe this is just a naming thing? :) ) Maybe that should even live under `api/`?

README.md
```diff
@@ -173,3 +173,98 @@ rails driver:hmis:dump_graphql_schema
...
+
+Important modules
+```
+app
+components             # no components, just a dir
+components/elements    # common components
+components/input       # base input components that are used across the app
+components/layout      # layout components that are used across the app
+components/table       # table components that are used across the app (could be a module or just in elements/ ?)
+components/pages       # ***keep or move to modules or src/app/pages?? these are not re-used
```
martha: aside from `NotFound` I agree that these should be moved out of components into modules.

I also wonder if we should update to use Bulletproof React's naming, i.e. `features` instead of `modules`. I'm not sure, but it might help us remember/enforce that everything going in top-level `components` is meant to be reusable across the app

README.md
```diff
@@ -173,3 +173,54 @@ rails driver:hmis:dump_graphql_schema
 ```
 
 To pick up the local schema changes on the frontend, run the `graphql:codegen` script (see above).
+
+## Project Structure
+
+This project follows the [bulletproof-react](https://github.com/alan2207/bulletproof-react/blob/master/docs/project-structure.md) structure.
```
ttoomey: We're not really 1:1 with bulletproof right? So maybe more accurate to say we are inspired by it?
```suggestion
This project uses a module or feature-based structure which is inspired by [bulletproof-react](https://github.com/alan2207/bulletproof-react/blob/master/docs/project-structure.md).
```

README.md
```diff
@@ -173,3 +173,54 @@ rails driver:hmis:dump_graphql_schema
...
++-- hooks       # hooks scoped to a specific feature
+|
++-- myTypes.ts  # typescript types used within the feature, with appropriate filename
+|
++-- myUtils.ts  # utility functions for a specific feature, with appropriate filename
+
+```
+
+#### Guidelines for categorization
+* Prefer to co-locate code that deals with the same type of thing into a shared `module`, even if the components are rendered in different parts of the application. For example, the Project Services table and Client Services table can live in the same "service" module.
```
ttoomey, line 216: Maybe we should expect some level of documentation
```md
* It should be clear what the purpose of a module is and how to use it. An engineer should be able to quickly understand what the module's entry-points are without needing to understand the internals of the module. If this is not obvious from code structure or storybook, include a brief README.md in the module dir with examples/explanation.
```

ttoomey (approved):
Looks good, thank you!

README.md
gigxz: totally, updated

README.md
gigxz: yes, thanks!

martha (commented):
Sorry for the belatedness, but this looks great to me! Thanks @gigxz !

PR #953, opened 2024-10-08
Minor package upgrades to address security issues

gigxz:
## Description

Resolve https://github.com/advisories/GHSA-593m-55hh-j8gv (@sentry/react) with upgrade
* changelog: https://github.com/getsentry/sentry-javascript/blob/develop/CHANGELOG.md

Resolve https://github.com/advisories/GHSA-pxg6-pf52-xh8x (cookie) with resolution
* this package is used by storybook so its only a dev dep
* changelog: https://github.com/jshttp/cookie/releases
## Type of change
- [x] Dependency update

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,8 @@
     "ws": "^8.17.1",
     "dset": "^3.1.4",
     "express": "^4.20.0",
-    "micromatch": "^4.0.8"
+    "micromatch": "^4.0.8",
+    "cookie": "^0.7.0"
   },
   "dependencies": {
     "@apollo/client": "^3.11.1",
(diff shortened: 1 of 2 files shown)


martha (approved):
Thanks Gig! These aren't coming from dependabot, right? (Checking dependabot PRs here https://github.com/greenriver/hmis-frontend/pulls?q=is%3Apr+dependabot). Otherwise I don't think I know which tools/automated checks we are using to keep an eye on vulnerabilities.

gigxz:
@martha yes they were but I already dismissed them https://github.com/greenriver/hmis-frontend/security/dependabot?q=is%3Aclosed

I would like to turn back on `yarn audit` in CI, but it's disabled right now because it doesn't give you a way to ignore (which we chose to do, in some cases). I opened https://github.com/open-path/Green-River/issues/5890 way back when (nov 2022??  👀 ). Probably a team discussion to have whether we prefer relying on dependabot or adding something back to CI.

PR #926, opened 2024-09-13
Add a few element stories

gigxz:
## Description

Add a few stories for common elements. I had this branch kicking around just forgot to open the PR!

## Type of change
- [x] stories

## Checklist before requesting review
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [x] My code includes comments and/or descriptive variable names to help other engineers understand the intent (or not applicable)
- [x] My code follows the style guidelines of this project (eslint)
- [x] I have updated the documentation (or not applicable)
- [x] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

PR #955, opened 2024-10-08
Add storybook theme and basic component stories

martha:
## Description
GH issues:
- https://github.com/open-path/Green-River/issues/6661
- https://github.com/open-path/Green-River/issues/6660

This PR
- Adds basic theme stories for Typography and Color. I created these in [MDX](https://storybook.js.org/docs/writing-docs/mdx), storybook's markdown + jsx mashup for documentation with inlined components. I did not really love the developer experience of working in MDX (no typescript, ide's auto-comment uses markdown comments instead of js comments). However, Storybook's builtin `ColorPalette` [does not appear to work outside of MDX](https://github.com/storybookjs/storybook/issues/8753).
- I can think of more changes to how I would want these story docs to be structured/what they should talk about in comments/etc, but it seems more useful to work on that when we actually make the code changes indicated in the [figma](https://www.figma.com/design/kqZiUPjXvqSvJVslJCnlTB/OP-HMIS-%2F-2024-%2F-Q3-%2B-Q4-(Aug-1---Dec-31)?node-id=7616-1873&node-type=section&t=x6a0dQwimUpEAFA1-0)?
- Adds the storybook addon for MUI, and replaces our previous storybook theme provider in `preview.tsx` with the suggestion from the [recipe here](https://storybook.js.org/recipes/@mui/material). This appears to accomplish the same thing as before. I was really adding this add-on because I wanted to get the MUI props to show up in the Storybook controls -- see below -- but in fact on closer look I don't think the add-on is necessary for this. I am on the fence about whether to revert or keep this.
  - Edit: I think one thing this does change is that completely unstyled components do now have our default mui theming (like default font) applied, same as they would in the application. (following up from https://github.com/greenriver/hmis-frontend/pull/926#pullrequestreview-2357333503)
- Updates components that are using `forwardRef` to a syntax where storybook can understand and parse their props for the Storybook controls. See https://github.com/storybookjs/storybook/issues/23084#issuecomment-2355951570 , which I got to from https://github.com/storybookjs/storybook/issues/21686 
  - ...This turns on MUI and React Router provided props, etc., which I think could be both a blessing and a curse? The interfaces are a bit unwieldy, but perhaps it's helpful for Storybook to reflect just how unwieldy it is?
  - Per [this tutorial](https://storybook.js.org/recipes/@mui/material) (same as linked above), we could also do a bit more filtering on the props, picking and choosing which node modules to accept props from and which to hide.
  - Also, these changes are making typescript unhappy... so I will spend some time banging my head against those issues. But we should discuss and decide how much head-banging is worthwhile.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


gigxz:
Looks like there are some typescript failures on ci

package.json
```diff
@@ -148,6 +148,9 @@
     "prettier": "^3.3.3",
     "rollup-plugin-visualizer": "^5.6.0",
     "storybook": "^8.3.2",
+    "@storybook/addon-themes": "^8.3.2",
+    "@storybook/blocks": "^8.3.2",
+    "add": "^2.0.6",
```
gigxz: Where are we using this package?

src/components/elements/ButtonLink.tsx
```diff
@@ -1,30 +1,36 @@
...
-      component={Link}
-      ref={ref}
-      target={openInNew ? '_blank' : undefined}
-      rel={openInNew ? 'noopener' : undefined}
-      replace={openInNew ? true : undefined}
-      {...props}
-      role={undefined} // remove the button role
-    />
-  )
+const ButtonLink = forwardRef<HTMLButtonElement, ButtonLinkProps>(
```
gigxz: welp thanks for fixing that

src/components/elements/ButtonLink.tsx
```diff
@@ -1,30 +1,36 @@
...
+  function ButtonLink(
+    { sx, leftAlign, Icon, openInNew, ...props }: ButtonLinkProps,
```
gigxz:
```suggestion
    { sx, leftAlign, Icon, openInNew, ...props },
```

I don't think we need to type it again?

src/components/elements/CommonToggle.stories.tsx
```diff
@@ -0,0 +1,30 @@
...
+export const Default: Story = {
+  args: {
+    items: [
+      { value: 'YES', label: 'Yes' },
+      { value: 'NO', label: 'No' },
+    ],
+    value: 'YES',
+  },
+  // https://github.com/storybookjs/storybook/discussions/21680#discussioncomment-7698496
+  render: function Component(args) {
```
gigxz: Other similar implementations:
https://github.com/greenriver/hmis-frontend/blob/eb66d3a63cfb82a8e316fa0c4580957ff94b5db9/src/components/elements/input/MultiFieldInput.stories.tsx#L10-L15
and
https://github.com/greenriver/hmis-frontend/blob/eb66d3a63cfb82a8e316fa0c4580957ff94b5db9/src/components/elements/input/SsnInput.stories.tsx#L13-L18

On those we define the render function on the default export, rather than on wone of the `Story`s. That lets you make multiple stories without rewriting the render function (as in `MultiFieldInput` which has several). Maybe we should stick to that pattern? And I think that line disabling eslint is unnecessary, lint works fine without it.

src/components/elements/LoadingButton.tsx
```diff
@@ -14,21 +14,26 @@ export interface LoadingButtonProps extends ButtonProps {
...
-      return { pl: 5 };
-    }
-    if (props.loading && props.loadingPosition === 'end') {
-      return { pr: 5 };
-    }
-    return;
-  }, [props.loading, props.loadingPosition]);
+const LoadingButton = forwardRef<HTMLButtonElement, LoadingButtonProps>(
+  function LoadingButton(
+    props: LoadingButtonProps,
```
gigxz: I think making the props optional and adding some defaults here fix the type issues `{ loadingPosition = 'start', loading = false, ...props },`... sorry I completely messed up my `forwardRef`s causing all these type issues 🤦

src/config/Color.mdx
```diff
@@ -0,0 +1,47 @@
...
+import theme from '@/config/theme';
+
+<Meta title='Color' />
+
+# Color
+
+The swatches below are programmatically pulled from our theme, so they reflect the current state of the code,
+not the desired state of the theme discipline. See Figma for desired changes.
+
+From the advanced example here: https://storybook.js.org/docs/6/writing-docs/doc-blocks/doc-block-colorpalette
```
gigxz: This can probably just be a code comment, not visible in the actual storybook?

src/config/Color.mdx
```diff
@@ -0,0 +1,47 @@
+import { Meta, ColorPalette, ColorItem } from '@storybook/blocks';
+import theme from '@/config/theme';
+
+<Meta title='Color' />
+
+# Color
+
+The swatches below are programmatically pulled from our theme, so they reflect the current state of the code,
+not the desired state of the theme discipline. See Figma for desired changes.
```
gigxz: Maybe we could make this comment a little more long-lived. Like simply "The swatches below reflect the default theme of the application" or something.

src/config/Typography.mdx
```diff
@@ -0,0 +1,44 @@
...
+  'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Duis eu diam sed ipsum vulputate condimentum accumsan et massa. Pellentesque sit amet placerat purus. Vestibulum quis risus at lacus scelerisque hendrerit sed a ante. Nulla vel rhoncus arcu. Sed efficitur fringilla aliquam. Etiam sagittis bibendum urna, in imperdiet tellus commodo nec.';
+
+# Typography
+
+The typography below is programmatically pulled from our theme, reflecting the current state of the code,
+not the desired state of the theme discipline. See Figma for desired changes. When we make those changes we should
+consider what updates to make here to most helpfully document our type system.
+
+We haven't used Storybook's built-in Typeset here, which is more useful for displaying a flexible type system
+(different font weights in combination with different sizes).
```
gigxz: also a code comment?

src/config/Typography.mdx
```diff
@@ -0,0 +1,44 @@
...
+          <Typography variant={k}>
+            {formatTypeString(k)}
+            {k.startsWith('body') && ': ' + SampleText}
+          </Typography>
+          <br />
+        </>
+      );
+    } else {
+      // These are MUI values like fontWeightRegular, fontWeightBold
+      console.log(k, v);
```
gigxz: remove these console logs?

src/config/Typography.mdx
```diff
@@ -0,0 +1,44 @@
+import { Meta } from '@storybook/blocks';
+import { ThemeProvider, Typography } from '@mui/material';
+import theme from '@/config/theme';
+
+<Meta title='Typography' />
+
+export const SampleText =
+  'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Duis eu diam sed ipsum vulputate condimentum accumsan et massa. Pellentesque sit amet placerat purus. Vestibulum quis risus at lacus scelerisque hendrerit sed a ante. Nulla vel rhoncus arcu. Sed efficitur fringilla aliquam. Etiam sagittis bibendum urna, in imperdiet tellus commodo nec.';
+
+# Typography
```
gigxz, line 10: Interesting, I have never noticed this random border thats under h2s...
Screenshot 2024-10-10 at 4 09 58 PM

gigxz:
> ...This turns on MUI and React Router provided props, etc., which I think could be both a blessing and a curse? The interfaces are a bit unwieldy, but perhaps it's helpful for Storybook to reflect just how unwieldy it is?

@martha I wouldn't worry about the storybook toggle overwhelm just yet.. we have that issue on lots of other components too. Your `forwardRef` change was fixing an _incorrect_ typing, so I'd say it's definitely worth fixing the type issues that it surfaced.

gigxz (commented):
LGTM after addressing type changes, thanks! *and removing `add` pkg?

And cool that there is a path forward for toggling between themes!

package.json
martha: Oh, yeah this should for sure be removed. Thank you 🤦

src/config/Typography.mdx
martha: yeah! I think we never(?) use h1 or h2 anywhere in the application. The only places where I can find usages of these are component-only, so they use a different variant. E.g. `<Typography component='h2' variant='overline'>`. All part of why Bobby's [proposed changes to our type hierarchy](https://www.figma.com/design/kqZiUPjXvqSvJVslJCnlTB/OP-HMIS-%2F-2024-%2F-Q3-%2B-Q4-(Aug-1---Dec-31)?node-id=7616-1873&node-type=section&t=W2aYBQuRmyQ01UuX-0) make so much sense

src/components/elements/ButtonLink.tsx
```diff
@@ -1,30 +1,36 @@
+import { SvgIconComponent } from '@mui/icons-material';
 import { Button, ButtonProps } from '@mui/material';
-import { forwardRef } from 'react';
+import { forwardRef, Ref } from 'react';
 import { Link, LinkProps } from 'react-router-dom';
 
-export type ButtonLinkProps = Omit<ButtonProps, 'href'> &
+export type ButtonLinkProps = Omit<ButtonProps, 'href' | `on${string}`> &
```
martha, line 6: Unioning `ButtonProps` with `LinkProps` creates a bunch of prop type issues related to the underlying ref types, `HTMLAnchorElement` and `HTMLButtonElement`. The upstream callers of `<ButtonLink>` didn't need to pass any of these `on*` handlers anyway (onCopy, onPaste etc.) so I just omitted them all by way of this handy masking functionality, which I learned about [here](https://stackoverflow.com/questions/53501721/typescript-exclude-property-key-when-starts-with-target-string).

<details>

```

src/components/layout/PrintViewButton.tsx:46:8 - error TS2769: No overload matches this call.
  Overload 1 of 3, '(props: { href: string; } & ButtonOwnProps & Omit<ButtonBaseOwnProps, "classes"> & CommonProps & Omit<Omit<DetailedHTMLProps<AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>, "ref"> & { ...; }, "color" | ... 24 more ... | "startIcon">): Element', gave the following error.
    Property 'href' is missing in type '{ children: string | number | boolean | ReactElement<any, string | JSXElementConstructor<any>> | Iterable<ReactNode> | null; ... 305 more ...; leftAlign?: boolean | undefined; }' but required in type '{ href: string; }'.
  Overload 2 of 3, '(props: { component: ElementType<any, keyof IntrinsicElements>; } & ButtonOwnProps & Omit<ButtonBaseOwnProps, "classes"> & CommonProps & Omit<...>): Element | null', gave the following error.
    Type '{ children: string | number | boolean | ReactElement<any, string | JSXElementConstructor<any>> | Iterable<ReactNode> | null; ... 305 more ...; leftAlign?: boolean | undefined; }' is not assignable to type '{ component: ElementType<any, keyof IntrinsicElements>; }'.
      Types of property 'component' are incompatible.
        Type 'ElementType<any, keyof IntrinsicElements> | undefined' is not assignable to type 'ElementType<any, keyof IntrinsicElements>'.
          Type 'undefined' is not assignable to type 'ElementType<any, keyof IntrinsicElements>'.
  Overload 3 of 3, '(props: DefaultComponentProps<ExtendButtonBaseTypeMap<ButtonTypeMap<{}, "button">>>): Element | null', gave the following error.
    Type 'MouseEventHandler<HTMLAnchorElement>' is not assignable to type 'MouseEventHandler<HTMLButtonElement>'.
      Type 'HTMLAnchorElement' is missing the following properties from type 'HTMLButtonElement': disabled, form, formAction, formEnctype, and 13 more.

46       <Button
          ~~~~~~

  node_modules/@mui/material/ButtonBase/ButtonBase.d.ts:107:12
    107   props: { href: string } & OverrideProps<ExtendButtonBaseTypeMap<TypeMap>, 'a'>,
                   ~~~~
    'href' is declared here.
  node_modules/@types/react/index.d.ts:2489:9
    2489         onClick?: MouseEventHandler<T> | undefined;
                 ~~~~~~~
    The expected type comes from property 'onClick' which is declared here on type 'IntrinsicAttributes & ButtonOwnProps & Omit<ButtonBaseOwnProps, "classes"> & CommonProps & Omit<...>'

```
</details>

src/components/elements/CommonToggle.stories.tsx
martha: Perfect, thanks!

src/components/layout/nav/ToolbarMenu.tsx
```diff
@@ -9,9 +9,12 @@ import { RootPermissionsFilter } from '@/modules/permissions/PermissionsFilters'
 import { Routes } from '@/routes/routes';
 import { RootPermissionsFragment } from '@/types/gqlTypes';
 
-export const TOOLBAR_MENU_ITEMS: (NavItem<RootPermissionsFragment> & {
-  activeItemPathIncludes: string;
-})[] = [
+export const TOOLBAR_MENU_ITEMS: (Required<
```
martha, line 12: Let me know if you think this is too fiddly. `TOOLBAR_MENU_ITEMS` always have `path`, though it's not required on `NavItem`, so I've made it required here. But we could also just pass `|| ''` in on line 81 and 87 of this file

gigxz (approved):
lgtm

PR #959, opened 2024-10-14
Return Not Found on /events when there is no form

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6852; comes out of: https://github.com/open-path/Green-River/issues/6793#issuecomment-2412096127 QA comment

I've targeted release 138 since this issue isn't newly introduced by release 137. (Well, it is newly exposed by adding the published scope on 137, but I guess I am trying to say that 137 is still a strict improvement, not a regression.) However, if it's easier to keep track of QA let's retarget to 137!

To reproduce:
- Temporarily disable any universal rules enabling CE Event in all projects
- Create a new CE Event form, but don't publish it - leave it in draft
- Add an active Rule enabling this new CE Event form in a project
- Navigate to an enrollment dashboard in the project. While you cannot see the CE Event link in the left-hand nav, if you directly type in /events to the url, you can still access the CE Events page and click the button to add an event.
  - Example provided by Angela in QA: https://qa-hmis.openpath.host/client/==VjlBUnZjbzhWaEtHYTg0S2Z2UVJJQT09/enrollments/==WHpNUG9OS0Y3Z2R0dGZxdTN1WnBUdz09/events
- After this PR, you should see "Not Found" instead of being able to access the page

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/components/accessWrappers/EnrollmentRoute.tsx
+++ b/src/components/accessWrappers/EnrollmentRoute.tsx
@@ -6,6 +6,7 @@ import NotFound from '../pages/NotFound';
 import { EnrollmentPermissions } from '@/modules/permissions/types';
 import { useHasPermissions } from '@/modules/permissions/useHasPermissionsHooks';
 import { DataCollectionFeatureRole } from '@/types/gqlTypes';
+import { ensureArray } from '@/utils/arrays';
 import { generateSafePath } from '@/utils/pathEncoding';
 
 /**
@@ -20,10 +21,8 @@ const EnrollmentRoute: React.FC<
 > = ({ permissions, redirectRoute, dataCollectionFeature, children }) => {
   // Use dashboard outlet context that gets set in EnrollmentDashboard
   const { enrollment, enabledFeatures } = useEnrollmentDashboardContext();
-  const hasPermission = useHasPermissions(
-    enrollment?.access,
-    permissions || []
-  );
+  const permissionsArray = ensureArray(permissions);
+  const hasPermission = useHasPermissions(enrollment?.access, permissionsArray);
 
   if (
     dataCollectionFeature &&
@@ -32,7 +31,7 @@ const EnrollmentRoute: React.FC<
     return <NotFound />;
   }
 
-  if (!!permissions && !hasPermission) {
+  if (permissionsArray.length > 0 && !hasPermission) {
     return redirectRoute ? (
       <Navigate
         to={generateSafePath(redirectRoute, {
--- a/src/components/accessWrappers/EnrollmentRoute.tsx
+++ b/src/components/accessWrappers/EnrollmentRoute.tsx
@@ -5,6 +5,7 @@ import NotFound from '../pages/NotFound';
 
 import { EnrollmentPermissions } from '@/modules/permissions/types';
 import { useHasPermissions } from '@/modules/permissions/useHasPermissionsHooks';
+import { DataCollectionFeatureRole } from '@/types/gqlTypes';
 import { generateSafePath } from '@/utils/pathEncoding';
 
 /**
@@ -14,11 +15,20 @@ const EnrollmentRoute: React.FC<
   React.PropsWithChildren<{
     permissions: EnrollmentPermissions | EnrollmentPermissions[];
     redirectRoute?: string;
+    dataCollectionFeature?: DataCollectionFeatureRole;
   }>
-> = ({ permissions, redirectRoute, children }) => {
+> = ({ permissions, redirectRoute, dataCollectionFeature, children }) => {
   // Use dashboard outlet context that gets set in EnrollmentDashboard
-  const { enrollment } = useEnrollmentDashboardContext();
+  const { enrollment, enabledFeatures } = useEnrollmentDashboardContext();
   const hasPermission = useHasPermissions(enrollment?.access, permissions);
+
+  if (
+    dataCollectionFeature &&
+    !enabledFeatures.includes(dataCollectionFeature)
+  ) {
+    return <NotFound />;
+  }
+
   if (!hasPermission) {
     return redirectRoute ? (
       <Navigate
--- a/src/components/accessWrappers/EnrollmentRoute.tsx
+++ b/src/components/accessWrappers/EnrollmentRoute.tsx
@@ -13,14 +13,17 @@ import { generateSafePath } from '@/utils/pathEncoding';
  */
 const EnrollmentRoute: React.FC<
   React.PropsWithChildren<{
-    permissions: EnrollmentPermissions | EnrollmentPermissions[];
+    permissions?: EnrollmentPermissions | EnrollmentPermissions[];
     redirectRoute?: string;
     dataCollectionFeature?: DataCollectionFeatureRole;
   }>
 > = ({ permissions, redirectRoute, dataCollectionFeature, children }) => {
   // Use dashboard outlet context that gets set in EnrollmentDashboard
   const { enrollment, enabledFeatures } = useEnrollmentDashboardContext();
-  const hasPermission = useHasPermissions(enrollment?.access, permissions);
+  const hasPermission = useHasPermissions(
+    enrollment?.access,
+    permissions || []
+  );
 
   if (
     dataCollectionFeature &&
@@ -29,7 +32,7 @@ const EnrollmentRoute: React.FC<
     return <NotFound />;
   }
 
-  if (!hasPermission) {
+  if (!!permissions && !hasPermission) {
     return redirectRoute ? (
       <Navigate
         to={generateSafePath(redirectRoute, {
--- a/src/components/accessWrappers/ProjectRoute.tsx
+++ b/src/components/accessWrappers/ProjectRoute.tsx
@@ -21,9 +21,10 @@ const ProjectRoute: React.FC<
   }>
 > = ({ permissions, redirectRoute, dataCollectionFeature, children }) => {
   const { project } = useProjectDashboardContext();
+  const permissionsArray = ensureArray(permissions);
   const [hasPermission, { loading }] = useHasProjectPermissions(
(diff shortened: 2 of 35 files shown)


src/modules/enrollment/components/dashboardPages/EnrollmentCeEventsPage.tsx
```diff
@@ -62,21 +62,27 @@ const EnrollmentCeEventsPage = () => {
...
-      localConstants,
-      projectId: enrollment?.project.id,
-    });
+  const {
+    onSelectRecord,
+    viewRecordDialog,
+    editRecordDialog,
+    openFormDialog,
+    formNotFound,
+  } = useViewEditRecordDialogs({
```
gigxz: I wouldn't think this is specific to CE Events, is it? There are plenty of other pages that use this pattern: I'm seeing CE Assessments, CLS, Case Notes, HMIS Participation, and CE Participation.

This approach feels a little tricky to maintain, because each component that uses the `useViewEditRecordDialogs` hook needs to realize that it's possible for the hook to fail, and that the component should look at `formNotFound` and render a NotFound component. The fact that the form isn't found feels like something that shouldn't be the concern of the dialog hook. IMO would make more sense if the dialog hook just throw an error if the form wasn't found–– because if the form wasn't found, it cannot render a dialog safely.

I'd propose that we use the same logic that we already use to determine whether the given page shows up in the left nav. For CE Events, that's:
https://github.com/greenriver/hmis-frontend/blob/84fccbcbedc112da322fe8f855f0ea379dccd73a/src/modules/enrollment/hooks/useEnrollmentDashboardNavItems.tsx#L50-L52

I think it would be ideal if the app could say "you're trying to navigate to `/events` but events wasn't included in the nav items, so I'm going to return NotFound [or redirect you to the enrollment dash overview]" without bothering to attempt to load the page itself. If we can do that in a way that is extensible to all dashboard pages, we don't need to worry about it in each pages render logic. (And we can raise an error if the form isn't found, because we don't expect that)


We do already have the enabledFeatures list in the context:
```
const { enabledFeatures } = useEnrollmentDashboardContext();
  if (!enabledFeatures.includes(DataCollectionFeatureRole.CeEvent)) {
    return <NotFound />;
  }
```

Looking at the `protected.tsx` routes, I wonder if we can just introduce a prop to `EnrollmentRoute` to this effect:
```
            element: (
              <EnrollmentRoute
                dataCollectionFeature={DataCollectionFeatureRole.CeEvent|
                redirectRoute={EnrollmentDashboardRoutes.ENROLLMENT_OVERVIEW}
              >
                <EnrollmentCeEventsPage />
              </EnrollmentRoute>
            ),
```
Similar to how `EnrollmentRoute` handles returning NotFound (or redirect) if the user lacks permission, it could likewise return NotFound (or redirect) if the Data Collection Feature is not listed in the enrollment dashboard contexts' `enabledFeatures` list.

martha: Thanks @gigxz  I like that idea, it's really elegant. 

Do you think we should implement something similar in `ProjectRoute`, for HMIS Participation, CE Participation, Current living situation, and maybe external form submissions?

src/modules/form/hooks/useFormDialog.tsx
```diff
@@ -83,7 +83,12 @@ export function useFormDialog<T extends SubmitFormAllowedTypes>({
       projectId:
         projectId || localConstants?.projectId || inputVariables?.projectId,
     },
-    localDefinition
+    localDefinition,
```
martha: I was just fiddling around here, but maybe we should just revert this change. Throwing the error here inside of the hook (as opposed to in a rendered component) just errors in the console. But it will only get thrown visibly to the user when they try to actually render a component that uses the form definition, like on line 130 below.

src/routes/protected.tsx
```diff
@@ -485,13 +486,27 @@ export const protectedRoutes: RouteNode[] = [
           },
           {
             path: EnrollmentDashboardRoutes.CURRENT_LIVING_SITUATIONS,
-            // No perm needed because it only requires enrollment visibility
-            element: <EnrollmentCurrentLivingSituationsPage />,
+            element: (
+              <EnrollmentRoute
+                permissions={[]} // Only requires enrollment visibility
```
martha: Could pass `canViewEnrollments`?

src/modules/form/hooks/useFormDialog.tsx
gigxz: I'm not sure I see the benefit to throwing the error from a callback vs throwing it in the component. Thanks to typescript, I think most of the components that call `useDefinition` already _have to_ deal with the state where `!loading && !formDefinition`--  most of them error in this case, but some render a NotFound component, which I think is wrong (`ViewRecord`, `EditRecord`). In the recent issue that came up-- where a user saw "Not Found" in the case note modal-- it would have been preferable if that had been treated as an error and reported through sentry so we could catch it asap.

I _think_ it would make sense for `useFormDefinition` to just raise an error if the fetch turns up empty. I can't think of any cases where we would expect that, but maybe I'm missing something.

src/modules/enrollment/components/dashboardPages/EnrollmentCeEventsPage.tsx
gigxz: Yes I think it would make sense to have in the project routes as well, since it uses the same concept of data collection features (For Services and Current Living Situations). I don't think the same would apply to HMIS Participation/ CE participation since those aren't "optional" features.

src/routes/protected.tsx
martha: We can't pass an empty array, since it makes the `hasPermissions` check evaluate as false. I made `permissions` optional on the `EnrollmentRoute` and `ProjectRoute` instead.

src/modules/form/hooks/useFormDialog.tsx
martha: Sounds good, I added the throw in `useFormDefinition`. 

I also removed the throw/NotFound returns from the other components, since I thought it's a bit clearer if this situation is only explicitly dealt with in one place. But I wonder if this is a bit messy. I want to think about it more.

gigxz: Hmm I think I see what you mean. I would look at `EditRecord` for example and say "oh, so this component will render null if there is no form" (since it has that `formDefinition && ` conditional render syntax). Which of course isn't true, because that won't get hit. I guess I would almost prefer if BOTH threw an error, so in EditRecord have `if (!formDefinition) throw "form definition not found"` to appease typescript, since at least we're not implying something different with the render. That error wont get hit, of course, because useFormDefinition will have already raised.
I also REALLY like that the query variables are included when raising the error from useFormDefinition, so that will help us track it down. I think that's another "plus" to raising from useFormDefinition.

martha: Ok, this makes sense. I agree with you, the most straightforward thing is for the error to get raised in several places so that we are never confused looking at the code about how it will deal with an absent form definition!

src/components/accessWrappers/ProjectRoute.tsx
```diff
@@ -14,19 +15,29 @@ import { generateSafePath } from '@/utils/pathEncoding';
...
     redirectRoute?: string;
+    dataCollectionFeature?: DataCollectionFeatureRole;
   }>
-> = ({ permissions, redirectRoute, children }) => {
+> = ({ permissions, redirectRoute, dataCollectionFeature, children }) => {
   const { project } = useProjectDashboardContext();
   const [hasPermission, { loading }] = useHasProjectPermissions(
     project.id,
-    ensureArray(permissions)
+    permissions ? ensureArray(permissions) : []
```
gigxz: `ensureArray` should handle the case when permissions is null or undefined- that util is meant to help us avoid ternaries like this!

gigxz (approved):
thanks, looks good to me!

src/modules/form/hooks/useFormDialog.tsx
gigxz: okay, sounds good

src/routes/protected.tsx
gigxz: Ok, sounds good

src/components/accessWrappers/EnrollmentRoute.tsx
```diff
@@ -5,21 +5,34 @@ import NotFound from '../pages/NotFound';
...
+  );
+
+  if (
+    dataCollectionFeature &&
+    !enabledFeatures.includes(dataCollectionFeature)
+  ) {
+    return <NotFound />;
+  }
+
+  if (!!permissions && !hasPermission) {
```
gigxz: Is it intentional that this will always return not found if you pass `permissions=[]`? We could make it more robust to handle that in the same way:
`if (ensureArray(permissions).length > 0 && !hasPermission) {`

src/components/accessWrappers/ProjectRoute.tsx
martha: Sounds good, `ensureArray` just needed a small typescript update to allow. https://github.com/greenriver/hmis-frontend/pull/959/files#diff-d03a2dd2397fbec94efa7e9b0cb974eada2c960a0b3c964fe6b4822c9e809fbbR1

src/components/accessWrappers/EnrollmentRoute.tsx
martha: I agree with your suggestion and made the change to make it more robust. With this, we could also now revert the change I mentioned [here](https://github.com/greenriver/hmis-frontend/pull/959#discussion_r1804664499) making `permissions` an optional param. But I kind of like that change anyway, it seems like a clearer api for the Route components now that they are not only used for checking permissions

PR #967, opened 2024-10-22
Prevent frontend from swallowing displayable error messages

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6886

This is a first stab at solving the bug described in the ticket.

I've tested a couple basic scenarios and I'm happy with the behavior:
- raising `HmisErrors::ApiError` with `display_message` - fixes bug, displays custom error
- raising `HmisErrors::ApiError` with no `display_message` - displays generic error message from [here](https://github.com/greenriver/hmis-warehouse/blob/341367e3474166cd329b3b263ac7dd1212f26d83/drivers/hmis/lib/hmis_errors/api_error.rb#L11)
- raising another runtime error - displays generic error message

However, I don't feel 100% sure that this isn't breaking any carefully considered erroring behavior that I don't know about yet! I also know this isn't compatible with the comment [here](https://github.com/greenriver/hmis-warehouse/blob/341367e3474166cd329b3b263ac7dd1212f26d83/drivers/hmis/lib/hmis_errors/api_error.rb#L10) stating that our goal is to put these error messages into the frontend eventually. So, let's discuss more before merging.

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/modules/errors/components/ApolloErrorAlert.tsx
+++ b/src/modules/errors/components/ApolloErrorAlert.tsx
@@ -116,11 +116,7 @@ const ApolloErrorAlert: React.FC<Props> = ({
       }
       return isString(result) ? [result] : result?.errors || [];
     }
-    // don't show graphql errors to user in prod
-    if (showDeveloperInfo) {
-      return error.graphQLErrors;
-    }
-    return [];
+    return error.graphQLErrors;
   }, [error]);
 
   const isNetworkError = error


ttoomey (approved):
It seems like we were intentionally not showing these errors. 

Displaying them should be okay assuming that most of the errors are suitable for users (non-technical and actionable).

martha:
Thanks @ttoomey for taking a look!

> It seems like we were intentionally not showing these errors.

Hm, my understanding of our usages of the [HmisErrors::ApiError](https://github.com/greenriver/hmis-warehouse/blob/release-139/drivers/hmis/lib/hmis_errors/api_error.rb) is that we don't expect the frontend to hide `display_message`. See [GraphqlController](https://github.com/greenriver/hmis-warehouse/blob/release-139/drivers/hmis/app/controllers/hmis/graphql_controller.rb#L103-L113).

Please let me know if I've misunderstood that!

gigxz:
@martha that is my understanding, too. In fact there are places I _know_ have seen it work before in Staging/Prod (like "Failed to connect to LINK" msg). And we have gotten screenshots from prod users showing the correct messages. I would like to understand if that behavior regressed, or if it is somehow being handled by a different component 🤔

gigxz (approved):

ttoomey:
> Hm, my understanding of our usages of the HmisErrors::ApiError is that we don't expect the frontend to hide `display_message`

Thanks Martha, I think your understanding is correct. I just meant that the condition on `showDeveloperInfo` seemed intentional. I'm all in favor of removing it if graphqlErrors is suitable for end-users.

martha:
Thank you both! After discussing with @gigxz yesterday I did a bit more spelunking to find out what introduced the regression. Here is what I found: 
- It was introduced back in August as part of this [dependency upgrade](https://github.com/greenriver/hmis-frontend/pull/857)
- Specifically it was introduced in `apollo/client` version 3.8.5. (That PR bumps from 3.6.8 to 3.11.1)
- The reason for the change is that, starting in 3.8.5, API errors newly come through on the `error.graphQLErrors` property instead of `error.networkError`. [ApolloErrorAlert.tsx](https://github.com/greenriver/hmis-frontend/blob/release-139/src/modules/errors/components/ApolloErrorAlert.tsx#L111-L123) therefore returns `[]` since `showDeveloperInfo` is false, rather than returning `error.networkError.result.errors`.
- I'm not sure what change in 3.8.5 led to this - [here](https://github.com/apollographql/apollo-client/releases/tag/v3.8.5) are the patch notes, maybe it's the `asyncMap` bug?
- In any case, based on the [Apollo docs on different error types](https://www.apollographql.com/docs/react/data/error-handling#error-types), I think this change is correct? -- i.e., an `HmisErrors::ApiError` should be considered a GraphQL error, not a network error, at least based on the way we are handling it [here in GraphqlController](https://github.com/greenriver/hmis-warehouse/blob/release-139/drivers/hmis/app/controllers/hmis/graphql_controller.rb#L115-L123)? Let's discuss more.

<details>
<summary>How I tested</summary>

- In [GraphqlController](https://github.com/greenriver/hmis-warehouse/blob/release-139/drivers/hmis/app/controllers/hmis/graphql_controller.rb#L103) set `dev_or_test` to false
- Make some mutation or query intentionally error out with a display message. `raise HmisErrors::ApiError.new(msg, display_message: msg)`
- In frontend repo `package.json`, set `"@apollo/client"` version to `"3.8.4"`
- `yarn && yarn build && yarn preview`
- try the mutation/query you changed
</details>

ttoomey:
Thanks so much for tracking this down Martha! Sorry I missed this behavior change in the upgrade

gigxz:
Yes thanks @martha!

I wasn't able to reproduce the correct behavior with the old `@apollo/client` version (tried 3.6.8 and 3.8.4), but, I do see in their issue history that people have complained about this happening. I tested locally and confirmed that the errors are coming through as `graphQLErrors` (Which is good!) and no longer present in `networkError.result.errors`. I wonder if it would make sense to remove the logic of parsing networkErrors, but we can keep it to be safe. I'll go ahead and merge this.

PR #972, opened 2024-10-25
Show data collection features where data is present

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6159
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4867

- Modify the enrollment to directly use the new Data Collection Features property, instead of filtering the Project's features list
- Check for the `legacy` attribute in all places where a feature is used and handle appropriately (usually by hiding the "new record" button)
- Remove some unneeded logic, per https://github.com/open-path/Green-River/issues/6159#issuecomment-2441679112

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/enrollment/components/pages/EnrollmentCurrentLivingSituationsPage.tsx
```diff
@@ -51,14 +52,23 @@ const EnrollmentCurrentLivingSituationsPage = () => {
...
       fieldName: 'currentLivingSituations',
     });
   }, [enrollmentId]);
 
-  const canEditCls = enrollment?.access?.canEditEnrollments || false;
+  const canEditCls =
+    (enrollment?.access?.canEditEnrollments && !clsFeature?.legacy) || false;
 
   const localConstants = useMemo(
     () => ({
```
gigxz, line 73: What I'm not seeing, but was expecting to see, is some change to the logic for how FormDefinitions are fetched when viewing/editing these records. Tracing the code, it should always come back  to `useFormDefinition` hook (from ViewRecord or useFormDialog).

I haven't tested it locally yet, but I would expect an error if I attempt to open, say, a CurrentLivingSItuation if CLS doesn't have any form instances enabled for this project. That's the piece where we want to maybe choose the one that's [managed in version control (tbd)](https://github.com/greenriver/hmis-warehouse/pull/4869) but for now we can just pick the "first". I _think_ we need to make sure that change goes in together. This change may result in more historical data being displayed to users, and we don't want them to face errors when trying to view those records.

gigxz: (Hmm and that's probably a change to the backend w.r.t how form definitions are resolved- just commenting here since I was looking for it )

martha: Hm, yes I agree I need to address that and it should go into this PR. It's missing here because I had held off on it per our discussion last week (https://github.com/open-path/Green-River/issues/6159#issuecomment-2430234508), sorry I did not clarify that! Sounds it is not actually blocked by https://github.com/open-path/Green-River/issues/6763 and I should proceed with just picking the first?

src/modules/form/hooks/useFormDefinition.tsx
```diff
@@ -28,9 +28,8 @@ const useFormDefinition = (
 
   if (error) throw error;
 
-  // TODO(#6159) - turn this explicit error back on?
-  // if (!formDefinition && !loading)
-  //   throw new Error(`Form not found: ${JSON.stringify(queryVariables)} `);
+  if (!formDefinition && !loading)
```
martha: @gigxz do you agree with adding this back here? (followup from https://github.com/greenriver/hmis-frontend/pull/977.) The work in this PR fixes the problem in this specific case. But I'm starting to wonder whether leaving it out is a better idea in general. The bug from the PR was that this caused an explicit error to be thrown when it didn't need to be (in reality, the case notes were always attached to a form processor and could always resolve their own definition when the view modal was opened).

martha: Discussed - should turn this back on with #6113.

martha: Depends how the frontend handles it, sometimes they see not found and sometimes an explicit error. Could be good to send to Sentry.

src/components/accessWrappers/EnrollmentRoute.tsx
```diff
@@ -20,14 +21,15 @@ const EnrollmentRoute: React.FC<
   }>
 > = ({ permissions, redirectRoute, dataCollectionFeature, children }) => {
   // Use dashboard outlet context that gets set in EnrollmentDashboard
-  const { enrollment, enabledFeatures } = useEnrollmentDashboardContext();
+  const { enrollment } = useEnrollmentDashboardContext();
```
gigxz: I'm not _totally_ convinced on the benefit of moving `enabledFeatures` from the context to a new hook. It's not a bad approach but it does cause some more code churn. Just to confirm-- we're never using the enrollment dataCollectionFeatures _outside_ of the enrollment dashboard context, right? (I wouldn't expect that, since I think of them of features that may or may not be enable "on" the Enrollment Dashboard).


What are you seeing as the benefit of moving it into a hook? I don't mean to nit-pick, but it might be helpful to just talk through the options.

_Another_ approach could be for it just to be a simple helper function, which I think would be fine. We lose the `useMemo` from the hook, but it's not an expensive operation (we know that dataCollectionFeatures array is small).



util fn:
```
const { enrollment } = useEnrollmentDashboardContext(); // or could come from a prop

if (getEnrollmentFeature(enrollment, DataCollectionFeatureRole.CaseNote)) { ... }
```

vs keeping it in context:
```
const { getEnrollmentFeature } = useEnrollmentDashboardContext(); // need to consume context

if (getEnrollmentFeature(DataCollectionFeatureRole.CaseNote)) { ... }
```


vs hook:
```
const { enrollment } = useEnrollmentDashboardContext(); // or could come from a prop

const caseNotes = useEnrollmentDataCollectionFeature(enrollment, DataCollectionFeatureRole.CaseNote)
if (caseNotes) { ... }
```

src/modules/caseNotes/components/EnrollmentCaseNotes.tsx
```diff
@@ -119,15 +126,16 @@ const EnrollmentCaseNotes = () => {
         title='Case Notes'
         headerVariant='border'
         actions={
-          canEdit ? (
+          canEdit &&
+          !caseNotesFeature?.legacy && (
```
gigxz: The assumption is that `caseNotesFeature` is present, because this component is being rendered, right?
Maybe we should throw in a `... || !caseNotesFeatures) return <NotFound />;` to make that a little clearer.

Then at this point, we know it is present, so this can be updated to `!caseNotesFeature.legacy` which is easier to understand

src/modules/enrollment/components/EnrollmentQuickActions.tsx
```diff
@@ -28,9 +27,16 @@ const EnrollmentQuickActions = ({
       clientId: enrollment.client.id,
     });
 
-  const canRecordService =
-    enabledFeatures.includes(DataCollectionFeatureRole.Service) &&
-    enrollment.access.canEditEnrollments;
+  const canRecordService = useMemo(() => {
+    const serviceFeature = enrollment.dataCollectionFeatures.find(
+      (f) => f.role === DataCollectionFeatureRole.Service
+    );
```
gigxz: Is there a reason the hook isn't used here?

src/modules/enrollment/components/pages/EnrollmentCeAssessmentsPage.tsx
```diff
@@ -108,15 +115,16 @@ const EnrollmentCeAssessmentsPage = () => {
         title='Coordinated Entry Assessments'
         headerVariant='border'
         actions={
-          canEditCeAssessments ? (
+          canEditCeAssessments &&
+          !ceAssessmentFeature?.legacy && (
```
gigxz: Same question here, I'm a little confused on the logic if it's `!undefined` then we still show the button

gigxz (commented):
looks good!

src/components/accessWrappers/EnrollmentRoute.tsx
martha: Oh, thanks for bringing this up. I like the middle option the best, where we have the context provide a `getEnrollmentFeature` function (whereas before it just used to provide the `enabledFeatures` list).

I don't think my original reasoning for this change was very interesting. It was something like: first I removed `enabledFeatures` because this list is now a property on the `enrollment` itself and doesn't need to be provided separately by the context. Later I was bothered by the repetition of the code to scan the list of enabled features looking for a particular one, so I wanted to factor that out into a separate place. There is no reason that separate place shouldn't be moving it right back into the enrollment context from whence it came, haha.

gigxz: Okay, yeah that makes sense- def agree its helpful to have a function or map to make it easier to access the feature you're looking for. That decision sounds good to me .

src/modules/enrollment/components/EnrollmentQuickActions.tsx
martha: No reason, it can be used here :)

src/modules/enrollment/components/pages/EnrollmentCeAssessmentsPage.tsx
martha: You're right, in all these cases we should be safe to return Not Found if the feature doesn't exist. I think I was trying to minimize the possible problems arising from this change - logic something along the lines of "we weren't resolving the feature before, and now we are, so if it's null and we can't resolve it, we should behave as we would have before when we never tried to resolve it." But you're right, I don't think that logic is correct. We should always be able to resolve these features.

gigxz (approved):
looks good!

PR #973, opened 2024-10-28
Support form duplication for managedInVersionControl forms

gigxz:
## Description

Issue: https://github.com/open-path/Green-River/issues/6763
Backend PR: https://github.com/greenriver/hmis-warehouse/pull/4869

* Hide "edit form" button for forms that are managed in version control
* Add button for new "Duplicate Form" action
* Add tag showing if a form is managed in version control (here in the UI I'm still calling it "System Form" - open to suggestions)

## Type of change
- [x] new feature

## Testing Instructions
* Run `rails driver:hmis:seed_definitions` to set the flag (tested with different clients) and checkout backend
* In config tool you should see:
  * system-managed forms are flagged and sorted to the end
  * when you open a system-managed form, there is an alert saying contact support to edit, and there is a Duplicate Form button
  * the Duplicate button works and creates a new form
  * if there were custom fields in the base form, the new form collects onto a new cded (not the same one) after publish. I tested this by temporarily turning on the "duplicate" button for a non-system form that I knew had CDEs on it.


src/modules/admin/components/forms/FormDefinitionActionsCard.tsx
```diff
@@ -0,0 +1,100 @@
...
+} from '@/modules/hmis/hmisUtil';
+import { RootPermissionsFilter } from '@/modules/permissions/PermissionsFilters';
+import { AdminDashboardRoutes } from '@/routes/routes';
+import { FormIdentifierDetailsFragment, FormStatus } from '@/types/gqlTypes';
+
+interface Props {
+  formIdentifier: FormIdentifierDetailsFragment;
+}
+
+const FormDefinitionActionsCard: React.FC<Props> = ({ formIdentifier }) => {
```
gigxz, line 22: Just refactored this out of `src/modules/admin/components/forms/FormDefinitionDetailPage.tsx` and added the `DuplicateFormButton` logic

src/modules/admin/components/forms/FormDefinitionTable.tsx
```diff
@@ -10,48 +11,54 @@ import {
...
+      {
+        header: 'Form Type',
+        render: ({ displayVersion }) => (
+          <FormTypeChip role={displayVersion.role} />
+        ),
+      },
+      {
+        key: 'status',
+        header: 'Form Status',
+        render: (identifier) => startCase(identifier.displayVersion.status),
```
gigxz, line 39: Added status. This is just the status of the display version, not the full explanatory status. E.g. will show `Published` instead of `Published with unpublished changes`

src/modules/admin/components/forms/FormStatusText.tsx
```diff
@@ -0,0 +1,41 @@
+import { Typography } from '@mui/material';
+import { FormIdentifierDetailsFragment, FormStatus } from '@/types/gqlTypes';
+
+interface Props {
+  identifier: FormIdentifierDetailsFragment;
+}
+const FormStatusText: React.FC<Props> = ({ identifier }) => {
```
gigxz, line 7: Just pulled this out from `src/modules/admin/components/forms/FormDefinitionDetailPage.tsx`, no changes

src/modules/admin/components/forms/FormDefinitionTable.tsx
martha: Nice. Maybe it's obvious but could be handy to add this pr comment as a code comment? That could help call our attention to the 2 different statuses if e.g. we ever get a client question about it

src/modules/admin/components/forms/FormDefinitionTable.tsx
```diff
@@ -10,48 +11,53 @@ import {
...
+        render: ({ displayVersion }) => displayVersion.formRules.nodesCount,
+      },
+
+      {
+        key: 'system',
+        header: 'Form Tags',
+        render: ({ managedInVersionControl }) =>
+          managedInVersionControl && (
+            <Chip
+              label='System Form'
```
martha, line 54: I think it could also be helpful to add a comment here about how we decided not to use "system" to refer to this concept internally in the code, bc we have other places where "system" means something different, but we _did_ decide it was the most client-friendly representation of this concept and that's the reason for the language discrepancy

martha (approved):
I like the refactors, thanks Gig!

ttoomey (approved):
Looks great!

> Add tag showing if a form is managed in version control (here in the UI I'm still calling it "System Form" - open to suggestions)

Naming is tricky, I think "system" is very reasonable and concise; that seems fine. It is a bit generic though. Other names I can think of are "Stock" or "Core", perhaps "HMIS Core Form"? I don't have a strong opinion

src/modules/admin/components/forms/FormDefinitionTable.tsx
gigxz: Good points, added a comment

src/modules/admin/components/forms/FormDefinitionTable.tsx
gigxz: added comment

PR #978, opened 2024-11-05
Stop swallowing assessment errors when invalid type is submitted

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6855

- If a user tries to submit an int value that's unparseable as an int, return it so the backend errors out.
- Same for dates. I didn't find a way to repro this via the frontend, but I made the code change anyway, on the premise that it's overall preferable to send the data back and cause an explicit error rather than swallowing.
- Add todos related to client-side validation (#6713)

## Type of change
- [x] Bug fix - it's kind of a bug fix, or maybe more like a bug exposure
- [ ] Dependency update

### DIFF
--- a/src/modules/form/util/formUtil.ts
+++ b/src/modules/form/util/formUtil.ts
@@ -756,11 +756,11 @@ export const formValueToGqlValue = (
     if (typeof date === 'string') {
       date = parseHmisDateString(value);
     }
-    if (date instanceof Date) return formatDateForGql(date) || undefined;
+    if (date instanceof Date) return formatDateForGql(date) || value;
     // This isn't parseable/formattable into a date, so it may cause a backend validation error.
     // Still, return it, so that it doesn't get swallowed without the user's knowledge
     // TODO(#6713) - Frontend should prevent submission of invalid dates
-    return date;
+    return value;
   }
 
   if ([ItemType.Integer, ItemType.Currency].includes(item.type)) {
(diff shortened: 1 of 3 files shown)


src/modules/form/util/formUtil.ts
```diff
@@ -756,14 +756,17 @@ export const formValueToGqlValue = (
     if (typeof date === 'string') {
       date = parseHmisDateString(value);
     }
-    if (date instanceof Date) return formatDateForGql(date) || undefined;
-    // This isn't parseable/formattable into a date, return undefined to ignore it
-    return undefined;
+    if (date instanceof Date) return formatDateForGql(date) || value;
+    // This isn't parseable/formattable into a date, so it may cause a backend validation error.
+    // Still, return it, so that it doesn't get swallowed without the user's knowledge
+    // TODO(#6713) - Frontend should prevent submission of invalid dates
```
gigxz: This doesn't feel like quite the right place for a TODO, since I don't think this util function logic will _change_ with 6713, but maybe I'm wrong about that

gigxz (approved):
looks good, thanks! If you have a way to repro the date issue, can you add to the ticket instructions? thanks!

martha:
> If you have a way to repro the date issue, can you add to the ticket instructions?

I wasn't able to repro this! I updated the pr comment to be more clear.

src/modules/form/util/formUtil.ts
martha: You're right, I can add a note to the ticket instead. Edit: there already is one

PR #979, opened 2024-11-06
Enable item-level form permissions on frontend

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6843
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4893

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/form/components/DynamicField.tsx
```diff
@@ -96,7 +97,15 @@ const DynamicField: React.FC<DynamicFieldProps> = ({
...
     [linkId, itemChanged]
   );
-  const isDisabled = disabled || inputProps?.disabled;
+
+  const { user: currentUser } = useAuth();
+  const userCanEdit = !currentUser
+    ? false
+    : item.editorUserIds
+      ? item.editorUserIds.includes(currentUser.id)
+      : true;
```
ttoomey: This is fine as is but I do struggle a little bit to read nested ternary operators. Maybe this is a little more readable. I think it's equivalent but I haven't tested it
```suggestion
  const userCanEdit = Boolean(currentUser?.id) && (!item.editorUserIds || item.editorUserIds.includes(currentUser.id));
```

ttoomey (approved):
looks good!

src/modules/form/components/DynamicField.tsx
martha: Thanks, yes this is equivalent (I tested locally) and much nicer. I used `!!currentUser?.id` instead of `Boolean(currentUser?.id)` since it seems to keep typescript happy.

PR #982, opened 2024-11-07
Support file upload in assessments

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6208
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4910

This PR
- Makes schema updates according to the backend PR
- Refactors `Uploader` so that it:
  - can have the `multi` option enabled
  - can display previously uploaded files
- Adds a `FileSummary` component that renders the thumbnail that used to go inside the Uploader box. Now it has two variants, "stacked" and "row", depending on whether the Uploader is single or multi.
- Updates the `DynamicViewField` to use the new `FileSummary`. It was previously using `File` and `Image`, which are now no longer used, I think -- and should perhaps be removed?
- ~Refactors `FileModal` so it provides a variant to render a file that has not yet been saved as a file record (FileFieldsFragment)~ - I removed this per our conversation with design.
- Enables configuration of File and Image types in the Form Builder.

<details>
<summary>Custom assessment with File Upload questions</summary>

```
{
  "item": [
    {
      "text": "date",
      "type": "DATE",
      "initial": [
        {
          "initial_behavior": "IF_EMPTY",
          "value_local_constant": "today"
        }
      ],
      "link_id": "assessment_date",
      "mapping": {
      },
      "assessment_date": true,
      "disabled_display": "HIDDEN"
    },
    {
      "text": "File that doesn't repeat",
      "type": "FILE",
      "link_id": "file_one",
      "mapping": {
        "custom_field_key": "martha_custom_assessment_file_one"
      },
      "repeats": false,
      "helper_text": "And it does have helper text!!",
      "readonly_text": "It has completely different readonly text.",
      "disabled_display": "HIDDEN"
    },
    {
      "text": "File that repeats",
      "type": "FILE",
      "link_id": "file_that_repeats",
      "mapping": {
        "custom_field_key": "martha_custom_assessment_file_that_repeats"
      },
      "repeats": true,
      "disabled_display": "HIDDEN"
    }
  ]
}
```
</details>

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/components/elements/upload/UploaderBase.tsx
```diff
@@ -251,60 +161,191 @@ const Uploader: React.FC<UploaderProps> = ({
...
-    previewUrl: fileImageUrl,
-    dragging: isDragActive,
-    clearFile: handleClear,
-    selectFile: open,
-    rootProps: getRootProps(),
-    inputProps: getInputProps(),
-    accept,
-    maxSize,
-  });
+  const showFileList = useMemo(
```
martha: Ok, even though I made my peace with the `defaultChildren` pattern and agree there's a time and place for it, here's my justification refactoring it out in this particular time and place: I started wanting to memoize things. I know that we could just make `defaultChildren` an actual react component, instead of a function but not a functional component(?). But since it is not actually being used anyway, I'm proposing to remove it. Happy to be persuaded otherwise, however!

src/modules/clientFiles/components/FileModal.tsx
```diff
@@ -2,7 +2,9 @@ import ImageNotSupportedIcon from '@mui/icons-material/ImageNotSupported';
 import {
```
martha: This should probably now be moved out of the ClientFiles module, but if it's helpful, I'm happy to postpone that change until you've reviewed the actual changes within this file, since it's probably easier to read the diff before moving it.

src/modules/form/components/DynamicField.tsx
```diff
@@ -411,20 +428,28 @@ const DynamicField: React.FC<DynamicFieldProps> = ({
     case ItemType.Image:
       return (
         <InputContainer {...commonContainerProps}>
-          <Uploader
-            id={linkId}
-            image
-            onUpload={async (upload) => onChangeValue(upload.blobId)}
-          />
+          <LabelWithContent label={item.text} helperText={item.helperText}>
```
martha, line 414: Note that Uploaders in Assessments were previously not labelled. With this change, the Client Files form will now have the new text "Attachment". It looks like this:
Screenshot 2024-11-21 at 5 18 02 PM

src/components/elements/upload/UploaderBase.tsx
```diff
@@ -88,157 +47,108 @@ const getFileTypesFromAccept = (accept: Accept) => {
...
+  multiple = false,
 }) => {
-  const [currentFile, setCurrentFile] = useState<File>();
-  const [currentUpload, setCurrentUpload] = useState<DirectUpload>();
+  // The uploader accepts a `files` argument which can contain either:
+  // - a STRING which points at a blob ID of a file that has been uploaded within this session, or
+  // - a FileFieldsFragment record which points at a file record in our database, uploaded during a previous session.
+  // `existingFiles` filters the list to only those files that were uploaded some previous time, so we can render them.
+  // The files uploaded during this session, we render this component's internal state, `currentFiles`.
+  const existingFiles = useMemo(
```
martha: Ok, so, save without submitting is the major functionality that is still not working. I want to confirm the current (messy!!) approach before continuing to work on that. But here I'll try to briefly summarize the current behavior:

- Right now if you upload a file then "save without submitting," then reopen the assessment, you don't see the file. However, if you then save, you do see the file -- it successfully saved!
- What's going on under the hood?
  - When you first upload a file, the CDE `value` is a string -- the blob ID. `onChange` callback is called, and the CDE value in memory gets updated to point at that string -- but this component ignores that (filter on line 94), because it's rendering the current upload using the File object coming from dropzone.
  - When you actually _submit_ the assessment, it gets saved as a File record. And then the next time you fetch the assessment, the value is returned as a FileFieldsFragment type.
  - However, when you just _save_ the assessment, the File record doesn't get saved. The value coming back when you reopen is a string, the blob ID, which is filtered out here on line 94. However, we don't have a `currentFile` or `currentUpload` for it either, so it's just not rendered at all.
  - But it's still present on the CDE's `value` here in the frontend in memory, so when you _do_ eventually _submit_ the assessment, it is saved and you can view it when you reopen the assessment. 😵 

Maybe this indicates that something about the approach here is rotten to the core?

src/components/elements/upload/FileSummary.tsx
```diff
@@ -0,0 +1,223 @@
...
+          <Button variant='text' onClick={onRemove}>
+            Delete
+          </Button>
+        )}
+      </Stack>
+    </Stack>
+  );
+};
+
+// CurrentFileSummary and ExistingFileSummary provide wrappers around FileSummary for rendering with a file
```
martha: Let's update the naming here, I think something else would be more clear but haven't quite decided what yet!

src/modules/form/components/DynamicField.tsx
gigxz: Cool, make sense. Is "Attachment" a default label or is it just the value of `item.text` in this example? (i.e. if `text` was left empty, would there be no label)?

src/modules/form/components/DynamicField.tsx
```diff
@@ -97,6 +104,16 @@ const DynamicField: React.FC<DynamicFieldProps> = ({
       itemChanged({ linkId, value, type: ChangeType.User }),
     [linkId, itemChanged]
   );
+  const onChangeFileValue = useCallback(
```
gigxz: It might be a little nearer of the Uploader component's `onChange` had different types depending on the `multiple` value. For individual uploader, the onChange would just be a single file (or string), but for multi-upload it would be an array. That's how the select (MUI Autocomplete) works, which can also handle both single or multiple selection. With that approach we wouldn't need this wrapper, I don't think.

martha: I couldn't manage to get this working with typescript happy ([here's](https://github.com/greenriver/hmis-frontend/pull/982/commits/7693f65121fe4bcb09b0607752c12ea34fa05967) my attempt - maybe there is something simple I am not seeing), but I managed something similar with two different type definitions. Let me know if it looks okay to you.

src/modules/form/components/DynamicField.tsx
martha: It is the value of `item.text` defined here: https://github.com/greenriver/hmis-warehouse/blob/681544cdafe35c9ec96ccc0e706a14cb371382a3/drivers/hmis/lib/form_data/default/records/file.json#L19

Yes, if `text` was left empty there would be no label.

src/components/elements/CommonMenuButton.tsx
```diff
@@ -98,28 +99,35 @@ const CommonMenuButton = ({
...
-                  setAnchorEl(null);
-                  onClick();
-                }
-              }}
-              disabled={disabled}
-            >
-              {title}
-            </MenuItem>
-          )
+        {items.map(
```
martha, line 100: This diff is just adding the `ariaLabel` prop.

src/modules/form/util/formUtil.ts
```diff
@@ -788,6 +788,19 @@ export const formValueToGqlValue = (
     } else if (value) {
       return (value as PickListOption).code;
     }
+  } else if ([ItemType.File, ItemType.Image].includes(item.type)) {
+    // Special case for File types. The frontend receives a FileFieldsFragment
+    // if this file has already been saved, but we don't want to return that whole fragment
+    // to the backend for processing, so just return the ID.
+    if (Array.isArray(value)) {
+      return value.map((file) => (file.hasOwnProperty('id') ? file.id : file));
```
martha: fileOrBlobId

src/components/elements/upload/fileSummary/FileRecordSummary.tsx
```diff
@@ -0,0 +1,50 @@
+import { Typography } from '@mui/material';
```
gigxz, line 1: Move to `clientFiles` module? (And maybe the module should be renamed to just `files`)

src/components/elements/upload/fileDialog/FileDialog.tsx
```diff
@@ -9,19 +9,13 @@ import {
   Stack,
```
gigxz: Oh I see this was moved in the other direction, from modules to src? I think it would be more in line with project structure guidelines for the upload- and file-related code to live in their own module(s), since there are more than a few files associated with the feature.

src/components/elements/CommonMenuButton.tsx
```diff
@@ -39,7 +40,7 @@ const CommonMenuButton = ({
   iconButton,
   MenuProps,
   ...buttonProps
-}: Props) => {
+}: Props & ButtonProps) => {
```
gigxz: If we need to accept all the button props, it might be a bit cleaner to accept a prop `ButtonProps` instead of inheriting them all like this. The reason is that IconButton and Menu components might share a lot of the same props, so it's not clear to the caller which element their prop will be applied to. Clearer is: `<CommonMenuButton ButtonProps={{}} MenuProps={{}} />`
With adding this, it also makes the `variant`,`disabled`,`sx` props above redundant, so those could be removed.

src/components/elements/upload/Uploader.tsx
```diff
@@ -0,0 +1,385 @@
...
+import FilePreviewIcon from '@/components/elements/upload/FilePreviewIcon';
+import FileRecordSummary from '@/components/elements/upload/fileSummary/FileRecordSummary';
+import FileUploadSummary from '@/components/elements/upload/fileSummary/FileUploadSummary';
+import { DirectUpload, FileFieldsFragment } from '@/types/gqlTypes';
+import { ensureArray } from '@/utils/arrays';
+
+const DEFAULT_MAX_BYTES = 3000000;
+const IMAGE_FILE_TYPES = ['.png', '.jpg', '.jpeg', '.gif'];
+
+const getFileTypesFromAccept = (accept: Accept) => {
```
gigxz, line 23: File structure note: these helper methods could go in `modules/uploader/uploaderUtil.ts`

src/components/elements/upload/Uploader.tsx
```diff
@@ -0,0 +1,385 @@
...
+  const [currentUploads, setCurrentUploads] = useState<DirectUpload[]>([]);
+  const [loading, setLoading] = useState<boolean>(false);
+  const [errors, setErrors] = useState<string[]>([]);
+
+  const [uploadFile] = useDirectUpload();
+
+  const uniqueNameValidator = useCallback(
+    (file: File) => {
+      const existingNames = [
+        ...existingFiles.map((f) => (f as FileFieldsFragment).name),
```
gigxz: Do we need to support both `File` and `FileFieldsFragment` as types that can be uploaded? it seems like that is causing some complication with types and needing to cast in some places. I think it could be safe to say that we expect parents of this component to be dealing with `FileFieldsFragment` fragments?

src/components/elements/upload/fileDialog/FileDialog.tsx
martha: Hm, based on bulletproof-react, I was thinking 
- modules import from components
- components should not import from modules
- modules should avoid importing from other modules

and since a lot of File could is used by both the `clientFiles` and the `form` module, it should be refactored into shared `component` code.

reference https://github.com/alan2207/bulletproof-react/blob/master/docs/project-structure.md
> It might not be a good idea to import across the features. Instead, compose different features at the application level. This way, you can ensure that each feature is independent which makes the codebase less convoluted. ...
> You might also want to enforce unidirectional codebase architecture. This means that the code should flow in one direction, from shared parts of the code to the application (shared -> features -> app). This is a good practice to follow as it makes the codebase more predictable and easier to understand.

However, based on the guidelines in our own readme https://github.com/greenriver/hmis-frontend/pull/948/files?short_path=b335630#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5 maybe it makes sense to move all file-related code into one shared `files` module as you've suggested above, which `DynamicForm` is still allowed to import from even though it is in the `form` module?
> Prefer to co-locate code that deals with the same type of thing into a shared module, even if the components are rendered in different parts of the application. For example, the Project Services table and Client Services table can live in the same "service" module.
> If two potential features are highly interdependent and frequently change together, consider combining them into a single module.
> (regarding services) because they are relatively independent implementations.

maybe "relatively independent" indicates that they don't have to be totally independent and not import from one another? 

However, is there a reason for us to diverge from bulletproof-react's suggestion here?

src/components/elements/CommonMenuButton.tsx
martha: Oh thanks, I like this much better!

src/components/elements/upload/Uploader.tsx
martha: same question here as above about importing `modules` code from within `components`

src/components/elements/upload/Uploader.tsx
martha: Hm, `existingFiles` should always be `FileFieldsFragment` and `currentFiles` should always be type `File`. The `files` arg can contain both `FileFieldsFragment`s and string blob IDs, which are filtered out of `existingFiles` on line 94 (because Uploader is a controlled component when used by DynamicField).

This particular cast is unnecessary thanks to the cast on line 95, so I'll take it out. But let's discuss more since I agree it is confusing, maybe we can look at together to see if it can be improved.

martha: I added some more comments to hopefully make it a bit easier to reason through!

src/components/elements/upload/fileDialog/FileDialog.tsx
martha: Dev standup discussion
- Things in modules are domain specific
- Things in src/components are common
- It's ok if this leads to lack of colocation. But they shouldn't be highly interdependent and frequently change together

src/components/elements/input/ClientImageUploadDialog.tsx
```diff
@@ -163,14 +164,12 @@ const ClientImageUploadDialog: React.FC<ClientImageUploadDialogProps> = ({
...
-                    setNewPhotoSrc(URL.createObjectURL(file));
-                    setErrors(emptyErrorState);
-                  }}
-                  onClear={() => {
-                    setNewBlobId(undefined);
-                    setNewPhotoSrc(undefined);
+                  onUpload={(
+                    upload: DirectUpload | undefined,
+                    file: File | undefined
+                  ) => {
```
gigxz:
```suggestion
                  onUpload={(upload, file) => {
```
The explicit types shouldn't be needed here, I don't think

src/components/elements/upload/Uploader.tsx
gigxz: resolved- decided to keep uploaders in src

src/components/elements/upload/fileSummary/FileUploadSummary.tsx
```diff
@@ -0,0 +1,38 @@
...
+  return (
+    <>
+      <FileSummary
+        fileName={file.name}
+        showThumbnail={!!file.type.match(/^image/)}
+        url={url}
+        info={
+          <Typography
+            variant='body2'
+            sx={{ color: theme.palette.warning.main }}
```
gigxz: This contrast is not accessible, "dark" is a little better, but still flagged by storybook. 
Linking https://github.com/open-path/Green-River/issues/7108 which will introduce a warning text color, which _will_ be accessible. Maybe add  a todo?

```suggestion
            color='warning.dark' # TODO(#718) use warning text color
```

src/components/elements/upload/Uploader.tsx
```diff
@@ -0,0 +1,393 @@
...
+      </Box>
+      {showFileList && (
+        <Card sx={{ borderTopLeftRadius: 0, borderTopRightRadius: 0 }}>
+          <Stack aria-live='polite' divider={<Divider />}>
+            {existingFiles.map((file) => {
+              return (
+                <FileRecordSummary
+                  key={file.id}
+                  file={file}
+                  onRemove={() => removeFile(file)}
```
gigxz: Can we add the focus logic here too? When removing a file with keyboard nav I lose focus

src/components/elements/upload/Uploader.tsx
```diff
@@ -0,0 +1,393 @@
...
+                  variant='row'
+                />
+              );
+            })}
+            {currentFiles.map((file) => {
+              return (
+                <>
```
gigxz: Seeing a unique key error logged here, wrap in container with `key={file.name}`

src/modules/form/components/DynamicField.tsx
```diff
@@ -410,20 +411,51 @@ const DynamicField: React.FC<DynamicFieldProps> = ({
...
+            ) : (
+              <Uploader
+                multiple={false}
+                id={linkId}
+                file={value}
+                image
+                onChange={onChangeValue}
+                ariaLabel={item.text}
+              />
+            )}
```
gigxz: Collapse this ternary and just conditionally pass `multiple`?
```suggestion
            <Uploader
              multiple={item.repeats || false}
              id={linkId}
              files={value}
              image
              onChange={onChangeValue}
              ariaLabel={item.text}
            />
```

src/modules/form/components/DynamicField.tsx
```diff
@@ -410,20 +411,51 @@ const DynamicField: React.FC<DynamicFieldProps> = ({
...
       );
     case ItemType.File:
       return (
         <InputContainer {...commonContainerProps}>
-          <Uploader
-            id={linkId}
-            onUpload={async (upload) => onChangeValue(upload.signedBlobId)}
-          />
+          <LabelWithContent label={item.text} helperText={item.helperText}>
+            {item.repeats ? (
```
gigxz: Same here, not sure we need the ternary

src/components/elements/upload/fileDialog/FileRecordDialog.tsx
```diff
@@ -0,0 +1,48 @@
...
+import React, { useMemo } from 'react';
+import { pdfjs } from 'react-pdf';
+import FileDialog from '@/components/elements/upload/fileDialog/FileDialog';
+import useSafeParams from '@/hooks/useSafeParams';
+import ViewRecordDialog from '@/modules/form/components/ViewRecordDialog';
+import { FileFieldsFragment, RecordFormRole } from '@/types/gqlTypes';
+
+pdfjs.GlobalWorkerOptions.workerSrc = pdfWorker;
+
+// Wrapper around FileDialog for representing a FileFieldsFragment, aka a file that has already been saved as a File record in our db
```
gigxz: I was a bit confused by the concept of wrapping one Dialog in another Dialog. But when I looked at the code I noticed that `FileDialog` just renders the label "File Preview" and file preview content -- it doesn't render a Dialog. It also looks like `FileDialog` isn't used anywhere on it's own (currently).

Maybe a structure like this would be more versatile and easier to understand the use-case for each component:


```
// component for viewing a FileFieldsFragment
const FileRecordDialog: React.FC<FileRecordDialogProps> = ({
  file,
  actions,
  ...props
}) => {
  
  return (
    <ViewRecordDialog<FileFieldsFragment>
    >
      <Stack>
         <Typography> File Preview </Typography>
          <FilePreview file={file} /> // formerly called FileDialog
      </Stack>
    </ViewRecordDialog>
  );
};
```

What do you think?

src/components/elements/upload/fileDialog/FileRecordDialog.tsx
```diff
@@ -0,0 +1,48 @@
...
+  actions,
+  ...props
+}) => {
+  const { clientId } = useSafeParams() as { clientId?: string };
+  const pickListArgs = useMemo(() => ({ clientId }), [clientId]);
+
+  return (
+    <ViewRecordDialog<FileFieldsFragment>
+      {...props}
+      record={file as FileFieldsFragment}
```
gigxz: This cast is not necessary
```suggestion
      record={file}
```

src/modules/form/util/formUtil.ts
```diff
@@ -690,7 +690,7 @@ const cleanTypedObject = (o: TypedObject) => {
     o,
     '__typename',
     'dateUpdated',
-    'dateCreated',
+    // 'dateCreated',  // This is probably bad; is there any other way to get the upload data on the file from within the assessment?
```
gigxz: yeah, this is weird. for Files I think we don't want to run `cleanTypedObject` because we want all the metadata to be available in the form. We already have the cleanup process that transforms the file object into a string before submitting it (below). I tested this change out and it seems to work fine even when re-submitting files:

On line 721 (gqlValueToFormValue), inserting:
```js
    case ItemType.File:
      // Use full File object in form to display metadata.
      // The metadata fields wont be submitted because of the `formValueToGqlValue`
      // logic which will transform the file object into an ID before submitting it back.
      return value;
```

That will make it so the metadata is present and shown in the dialog (where currently there is an awkward blank space) >
Screenshot 2024-12-16 at 4 20 41 PM

src/components/elements/upload/Uploader.tsx
gigxz: I see, thank you! I think with my comment I was confusing the `File` type from [@types/node](https://github.com/DefinitelyTyped/DefinitelyTyped) with the generate `File` type from our API.

src/components/elements/upload/Uploader.tsx
```diff
@@ -0,0 +1,393 @@
...
+                    or drag and drop
+                  </Typography>
+                  <Typography variant='body2' color='GrayText' sx={{ mt: 1 }}>
+                    {accept ? getFileTypesFromAccept(accept) : 'Any file type'}{' '}
+                    (max. {getReadableSize(maxSize)})
+                  </Typography>
+                </>
+              )}
+          </Box>
+          <Box aria-live='polite'>
```
gigxz: From here down, there is a lot of logic to reason about for the different scenarios. I would consider refactoring this to reduce the complexity of the render fn component.

I might approach it like this:
```ts
// if it's a single-file upload, render the summary above the error (single thumbnail)
{!multiple && <SingleFileUploadSummary file={currentFiles[0] || existingFiles[0]} onRemove={removeFile} />}}
// upload error
{error}
// if it's a multi-file upload, render the summary below the error (rows)
{multiple && <MultiFileUploadSummary uploadedFiles={[]} stagedFiles={[]} onRemove={removeFile} />}
```
or maybe,

```ts

const fileSummary = (uploadedFiles.length || stagedFiles.length) ?  (<FileUploadSummary uploadedFiles={[]} stagedFiles={[]} onRemove={removeFile} multiple={multiple} />) : null

// if it's a single-file upload, render the summary above the error (single thumbnail)
{!multiple && fileSummary}
// upload error
{error}
// if it's a multi-file upload, render the summary below the error (rows)
{multiple && fileSummary}
```


You can take or leave this feedback. It's very rough. It would help with the complexity of the JSX, but it doesn't help much with reducing the complexity of the component logic overall. There also might be some tradeoffs, like more intermediate components being harder to reason about.

src/components/elements/upload/Uploader.tsx
```diff
@@ -0,0 +1,393 @@
...
+      filesArr = ensureArray(files);
+    } else {
+      const { file } = rest as SingleUploaderProps;
+      filesArr = ensureArray(file);
+    }
+    return filesArr.filter(
+      // Filter out files from the input that are just blob IDs.
+      // These should also be reflected in the currentFiles internal state object.
+      (f) => typeof f !== 'string'
+    ) as FileFieldsFragment[]; // Cast to keep typescript happy; now that we've filtered out all the strings, they should all be FileFieldsFragments
```
gigxz: The types can also be differentiated by using a type guard:


```ts
const isFileFieldsFragment = (
  value: string | FileFieldsFragment
): value is FileFieldsFragment =>
  !isNil(value) && typeof value === 'object' && value.__typename === 'File';

```
```suggestion
    // Filter out files from the input that are just blob IDs.
    // These should also be reflected in the currentFiles internal state object.
    return filesArr.filter((f) => isFileFieldsFragment(f));
```

src/components/elements/upload/Uploader.tsx
```diff
@@ -0,0 +1,393 @@
...
+  `${(maxSize / 1000000).toFixed(1)}MB`;
+
+export type SingleUploaderProps = {
+  multiple?: false;
+  file?: string | FileFieldsFragment;
+  onChange?: (file: string | FileFieldsFragment | undefined) => void;
+  onUpload?: (
+    upload: DirectUpload | undefined,
+    file: File | undefined
+  ) => any | Promise<any>;
```
gigxz: Can use `(upload?: DirectUpload, file?: File)` instead of `| undefined `

src/components/elements/upload/fileSummary/FileUploadSummary.tsx
```diff
@@ -0,0 +1,38 @@
+import { Typography } from '@mui/material';
+import React, { useMemo } from 'react';
+import FileSummary, {
+  FileSummaryProps,
+} from '@/components/elements/upload/fileSummary/FileSummary';
+import theme from '@/config/theme';
+
+// Wrapper around FileSummary for representing a File, aka a file that has just now been uploaded to Active Storage but doesn't yet exist as a File record in our db.
+const FileUploadSummary: React.FC<{
```
gigxz: Suggest giving this component a name that indicates that the file hasn't been uploaded yet- maybe `StagedFileSummary` or `UnsavedFileSummary`? At least to me it wasn't obvious the difference between FileRecord and FileUpload. `SavedFileSummary`/`UnsavedFileSummary` may be clearer

gigxz (commented):
Looks great! Functionality all seems good, mostly left code-style/ts comments!

src/modules/form/components/DynamicField.tsx
martha: Hm let's look at this together, I agree it would be better to get rid of the ternary but I think that it means reverting some of the changes I made related to [this thread](https://github.com/greenriver/hmis-frontend/pull/982#discussion_r1856890379). With the current typescript regime on the Uploader component, getting rid of the ternary leads to unexpected behavior with an existing file on a single-uploader.

src/modules/form/util/formUtil.ts
martha: Oh, good point, lovely!

src/components/elements/upload/fileDialog/FileRecordDialog.tsx
martha: Thanks, I like this. I'll also rename the existing `FilePreviewIcon` to `FileThumbnailIcon` to try to enforce the language distinction that:
- "preview" = when you open the file for viewing
- "thumbnail" = the little tiny view of a file (or in this case the icon) that's displayed in the uploader

martha: Thanks, I like this. I'll also rename the existing `FilePreviewIcon` to `FileThumbnailIcon` to try to enforce the language distinction that:
- "preview" = when you open the file for viewing
- "thumbnail" = the little tiny view of a file (or in this case the icon) that's displayed in the uploader

I also think we can rename `FileRecordDialog` to reduce confusion. I named it that when I also had a `FileUploadDialog` (parallel to `FileRecordSummary` and `FileUploadSummary` which you've commented about below). But I removed `FileUploadDialog` per our conversation with design - it's not needed to display the file you've just uploaded. So this could be renamed (back?) to `FilePreviewDialog` I suppose?

src/components/elements/upload/fileSummary/FileUploadSummary.tsx
martha: Thanks, I like `Saved`/`Unsaved`!

src/components/elements/upload/Uploader.tsx
martha: Cool, thank you

src/components/elements/upload/Uploader.tsx
martha: Oof ok I got stuck on this for some time, but eventually figured out a solution that I think is ok: Getting the ref from the dropzone hook, and then calling it with a setTimeout from within the removeFile callback. I _think_ the downside of `setTimeout` is that it introduces a dependency on the user's browser's speed to execute, but I tried with 20x cpu throttling and the refocus still worked ok. Let me know if you think this looks all right, or have any other suggestions!

src/modules/form/components/DynamicField.tsx
martha: Updated to use generics similar to Autocomplete. Some casting is still needed internally to the component to get typescript happy, but I think the api is way better. https://github.com/greenriver/hmis-frontend/pull/982/commits/9777a416f318f3ba4f998221d6ff0f87e123e892

src/components/elements/upload/Uploader.tsx
martha: Sounds good, I like this idea and took a stab at it here https://github.com/greenriver/hmis-frontend/pull/982/commits/077d08185f8cce3fb996be3864c53742dd99fee8

martha:
Whew ok @gigxz thank you, I think I addressed all your feedback and would love for you to take another look!

src/modules/form/components/DynamicField.tsx
gigxz: Nice improvement, thanks!

src/components/elements/upload/Uploader.tsx
gigxz: cool, looks good!

gigxz (approved):
LGTM! 💯 📎

PR #984, opened 2024-11-18
Propose approach for themeable nav link background color

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6659

This PR proposes the following changes:
- Add a new variable to our Palette options, `lightenCoefficient`. This would allow the amount by which the `primary.light` color is lightened to be configurable per client theme. The prior value, `0.9`, stays the default.
  - This is only a good idea if it's ok to change the `primary.light` and evade the default MUI behavior of having it depend on `primary.main`, which I'm not sure how I feel about. I guess the alternative is to add a new (`tertiary`?) color? 
- Update the background for Toolbar links and Mobile links so that they also use this same background color. Otherwise, they use a very lightened version of the Links color, which ends up being a just-slightly-different shade of blue, and I found it a bit jarring. I realize this is something of a can of worms especially on mobile, so let's discuss before merging.

Script I ran locally, and plan to run in client env:
```ruby
new_color_1 = '#1d5ea2'
new_color_2 = '#c5dfee'
theme = GrdaWarehouse::Theme.where.not(hmis_value: nil).sole
value = theme.hmis_value
value['palette']['primary']['main'] = new_color_1
value['palette']['primary']['light'] = new_color_2
value['palette']['lightenCoefficient'] = 0.6 # to discuss, see below
value['palette']['links'] = new_color_1
theme.hmis_value = value
theme.save!
```

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/components/layout/dashboard/sideNav/ItemBase.tsx
+++ b/src/components/layout/dashboard/sideNav/ItemBase.tsx
@@ -89,7 +89,12 @@ const ItemBase = <T extends object>({
                 }
               : undefined,
           backgroundColor: (theme) =>
-            isSelected ? lighten(theme.palette.primary.light, 0.9) : undefined,
+            isSelected
+              ? lighten(
+                  theme.palette.primary.light,
+                  theme.palette.lightenCoefficient || 0.9
+                )
+              : undefined,
         }}
       >
         {item.title && (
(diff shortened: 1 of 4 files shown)


src/components/layout/dashboard/sideNav/ItemBase.tsx
```diff
@@ -89,7 +89,12 @@ const ItemBase = <T extends object>({
                 }
               : undefined,
           backgroundColor: (theme) =>
-            isSelected ? lighten(theme.palette.primary.light, 0.9) : undefined,
+            isSelected
```
martha, line 92: let's discuss where is a good common place for this logic to live, maybe in the theme?

martha:
Per https://github.com/open-path/Green-River/issues/5746 there's no need for us to give up control of all colors in the app, we want to avoid making the link bg color subject to the user theme.

PR #986, opened 2024-12-01
6913 lock version conflict mitigation

ttoomey:
## Description

[GH Issue ](https://github.com/open-path/Green-River/issues/6913)

Two changes to try and reduce occurrence of users encountering lock version errors. Spot checking suggests this is often due to opening the same record in multiple tabs.

1. add cross-tab sync mechanism. When a tab receives a new version (detected via lock-version), surface that on all other tabs (for clients, enrollments, and assessments) and prompt the user to reload
   * This does not address conflicts due to server side changes
   * users can close snack bar alert
   * known issue: if the user has a form open when the reload, their current form data would be lost. Seems like this could be addressed with improved props handling or maybe some way to temporarily save and re-apply form-state
   * The current implementation uses `BroadcastChannel` which is a newer feature. Looks like it's now supported on evergreen browsers
2. when clicking the button to edit the client record, re-fetch the client data (could extend to other pages)
   * this might catch server side changes
   * we could take this approach on other buttons but the client form is the low hanging fruit

misc fix:
* bug fix for ConfirmationDialog not showing cancelText

##testing 
1. testing cross tab sync: open a two tabs with the same client. Update the client record in one, the other tab should display prompt to reload the client record. Reloading should load new data and close the prompt. Clicking continue should close the prompt but not reload.
2. test client refresh: go to the client dash. In the console or else where, update the client (client.touch in rails). Open client details form and save, there should be no lock-version error

## Type of change
- [x] Bug fix

### DIFF
--- a/src/components/elements/ConfirmationDialog.tsx
+++ b/src/components/elements/ConfirmationDialog.tsx
@@ -73,7 +73,7 @@ const ConfirmationDialog = ({
               variant='gray'
               data-testid='cancelDialogAction'
             >
-              {cancelText || unconfirmable ? 'Close' : 'Cancel'}
+              {cancelText || (unconfirmable ? 'Close' : 'Cancel')}
             </Button>
           )}
           {!unconfirmable && (
--- a/src/components/elements/SnackbarAlert.tsx
+++ b/src/components/elements/SnackbarAlert.tsx
@@ -1,0 +1,40 @@
+import { Alert, AlertProps, AlertTitle, Snackbar } from '@mui/material';
+
+import { ReactNode } from 'react';
+
+interface Props {
+  open: boolean;
+  onClose: VoidFunction;
+  children?: ReactNode;
+  alertProps?: AlertProps;
+  title?: string;
+}
+const SnackbarAlert: React.FC<Props> = ({
+  open,
+  onClose,
+  children,
+  alertProps,
+  title,
+}) => {
+  return (
+    <Snackbar
+      TransitionProps={{ appear: false }} // transition looks kind of junky
+      open={open}
+      onClose={onClose}
+      anchorOrigin={{ vertical: 'top', horizontal: 'center' }}
+      sx={({ shadows }) => ({
+        boxShadow: shadows[6],
+        borderRadius: 2, // matches Alert
+        '.MuiAlertTitle-root': { fontWeight: 600 },
+        marginTop: '100px',
+      })}
+    >
+      <Alert onClose={onClose} {...alertProps}>
+        {title && <AlertTitle sx={{ mb: 0 }}>{title}</AlertTitle>}
+        {children}
+      </Alert>
+    </Snackbar>
+  );
+};
+
+export default SnackbarAlert;
--- a/src/components/elements/SnackbarAlert.tsx
+++ b/src/components/elements/SnackbarAlert.tsx
@@ -18,7 +18,7 @@ const SnackbarAlert: React.FC<Props> = ({
 }) => {
   return (
     <Snackbar
-      TransitionProps={{ appear: false }} // transition looks kind of junky
+      TransitionProps={{ appear: false }} // disabled transition since it looks a bit junky
       open={open}
       onClose={onClose}
       anchorOrigin={{ vertical: 'top', horizontal: 'center' }}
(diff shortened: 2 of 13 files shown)


src/components/elements/ConfirmationDialog.tsx
```diff
@@ -73,7 +73,7 @@ const ConfirmationDialog = ({
               variant='gray'
               data-testid='cancelDialogAction'
             >
-              {cancelText || unconfirmable ? 'Close' : 'Cancel'}
+              {cancelText || (unconfirmable ? 'Close' : 'Cancel')}
```
ttoomey, line 76: looks like a bug where cancelText was not displayed

src/modules/enrollment/components/pages/EnrollmentDashboard.tsx
```diff
@@ -112,6 +113,16 @@ const EnrollmentDashboard: React.FC = () => {
...
+        <LocalVersionCoordinationPrompt
+          recordType='Enrollment'
+          recordId={enrollment.id}
+          currentVersion={enrollment.lockVersion}
+        />
+        <LocalVersionCoordinationPrompt
+          recordType='Client'
+          recordId={client.id}
+          currentVersion={client.lockVersion}
+        />
```
ttoomey, line 125: Could end up with overlapping modals here. We could refactor the LocalVersionCoordinationPrompt to handle that

src/modules/localVersionCoordination/hooks/useLocalVersionCoordination.tsx
```diff
@@ -0,0 +1,68 @@
...
+      }
+    },
+    [currentVersion]
+  );
+
+  // manage channel connection
+  const channelRef = useRef<BroadcastChannel | null>(null);
+  useEffect(() => {
+    const channelName = `record-sync-${recordType}-${recordId}`;
+    channelRef.current = new BroadcastChannel(channelName);
```
ttoomey, line 36: [BroadcastChannel docs here](https://developer.mozilla.org/en-US/docs/Web/API/Broadcast_Channel_API)

src/modules/assessments/components/IndividualAssessmentPage.tsx
```diff
@@ -42,15 +44,23 @@ const IndividualAssessmentPage = () => {
...
-      enrollment={enrollment}
-      client={client}
-      viewingDefinition={assessment.definition}
-      editingDefinition={
-        assessment.upgradedDefinitionForEditing || assessment.definition
-      }
-      assessment={assessment}
-    />
+    <>
+      <LocalVersionCoordinationPrompt
```
ttoomey, line 47: if this approach seems good, we'd also need to add to the household enrollments

src/modules/localVersionCoordination/LocalVersionCoordinationPrompt.tsx
```diff
@@ -0,0 +1,71 @@
...
+    }
+  }, [currentVersion, latestVersion]);
+
+  const handleContinue = () => {
+    setAcknowledgedVersion(latestVersion);
+  };
+
+  // hide prompt if we are on the latest version OR we are on an old version but the user has acknowledged it
+  const showPrompt =
+    latestVersion > Math.max(acknowledgedVersion, currentVersion);
```
ttoomey, line 53: perhaps we should show a snackbar or alert even if the user has acknowledge the stale record

src/components/elements/SnackbarAlert.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { ReactNode } from 'react';
+
+interface Props {
+  open: boolean;
+  onClose: VoidFunction;
+  children?: ReactNode;
+  alertProps?: AlertProps;
+  title?: string;
+}
+const SnackbarAlert: React.FC<Props> = ({
```
ttoomey, line 12: Ideally we'd use this in the existing Apollo Alert but refactoring that felt like a bit too much scope creep for this branch

src/modules/enrollment/components/pages/EnrollmentDashboard.tsx
ttoomey: this doesn't seem to be an issue in practice

gigxz (approved):
This looks great @ttoomey!

gigxz:
Screenshot 2024-12-13 at 12 46 09 PM

https://github.com/greenriver/hmis-frontend/pull/986/commits/5808a545c03411c14ad957783aaeeb18e2b339cc

Linking PR https://github.com/greenriver/hmis-frontend/pull/991 and added a TODO for it, because I will update Alerts to use the grayscale button variant when merged, per design library!

PR #987, opened 2024-12-03
Use signed blob ID when uploading images

martha:
## Description

This PR updates the hmis upload functionality to submit the signed blob ID rather than the plain blob ID. See the backend PR description for more discussion and questions.

GH issue: https://github.com/open-path/Green-River/issues/7077 ( https://github.com/open-path/Green-River/issues/6208 )
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4980

Should go into release-144

## Type of change
- [ ] Bug fix
- [ ] Dependency update

- [ ] I have updated the documentation (or not applicable) - https://github.com/open-path/Green-River/issues/7054
- [x] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue - see backend PR

### DIFF
--- a/src/components/elements/input/ClientImageUploadDialog.tsx
+++ b/src/components/elements/input/ClientImageUploadDialog.tsx
@@ -2,7 +2,6 @@ import DeleteIcon from '@mui/icons-material/Delete';
 import {
   Box,
   Button,
-  ButtonProps,
   CircularProgress,
   DialogActions,
   DialogContent,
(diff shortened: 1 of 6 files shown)


src/components/elements/input/ClientImageUploadDialog.tsx
```diff
@@ -40,7 +49,12 @@ const ClientImageUploadDialog: React.FC<ClientImageUploadDialogProps> = ({
   } = useGetClientImageQuery({
     variables: { id: clientId },
   });
-  const [mutate, { loading: updating }] = useUpdateClientImageMutation();
+  const [errors, setErrors] = useState<ErrorState>(emptyErrorState);
+  const [mutate, { loading: updating }] = useUpdateClientImageMutation({
+    onError: (apolloError) => {
```
martha: I think it'd be ideal style-wise for this to be moved below into the callback at line 77, but I couldn't quite get Typescript working correctly with this approach.

Or else for that callback to be moved up here into an `onCompleted` call, which better matches our general pattern -- but we need the event object for the `handleClose` object, which I think we don't have access to in the mutation-provided callbacks.

But I'd be delighted to learn that I am missing something obvious! 😅

gigxz: Yeah that was a little funky. It's doable by moving the onCompleted steps into `onCompleted` instead of a `then`, and moving handleClose up

```
  const [mutate, { loading: updating }] = useUpdateClientImageMutation({
    onCompleted: (data) => {
      const { errors: validationErrors = [] } = data.updateClientImage || {};
      if (validationErrors.length) {
        setErrors(partitionValidations(validationErrors));
      } else {
        handleClose();
      }
    },
    onError: (apolloError) => setErrors({ ...emptyErrorState, apolloError }),
  });
```

gigxz: Okay a few adjustments on https://github.com/greenriver/hmis-frontend/pull/987/commits/39a56a4939b095e0dddee46c0ade4229632170df:
* Use onCompleted and onError instead of `then(..)` pattern
* Show apollo errors from deleteImage mutation (previously swallowed)
* Stop calling refetchClient after deleteImage. This was unnecessary and caused a weird lag before the dialog get's closed. There is no need to refetch because it gets deleted from the cache right away, I tested.
* Update the `handleClose` callback so the event/reason params are optional.. this is not really needed, since in some cases we are closing it because of some logic rather than an event. 
* Clear out error when dialog is closed

What I tested:
* Success states
* Raised error when deleting client image (appears in snackbar)
* Raised error when updating client image (appears in snackbar)
* Validation error when updating client image (appears in modal)

gigxz (approved):
👍 👏

src/components/elements/input/ClientImageUploadDialog.tsx
martha: Lovely! Thank you!

PR #989, opened 2024-12-05
Update Project Household Enrollments table with new designs

martha:
## Description

Should be merged into the same release as https://github.com/greenriver/hmis-frontend/pull/998

GH issue: https://github.com/open-path/Green-River/issues/6825
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5012 This is not a functional dependency, but is needed for spec tests to pass.

This PR
- Updates the ProjectHouseholdsTable component to render each household member in their own row
- Defaults to the Clients tab instead of Households tab
- Updates the columns rendered in both the ProjectHouseholdsTable and the ProjectClientEnrollmentsTable
- Implements a new `TableRowActions` that will be more widely adopted with https://github.com/open-path/Green-River/issues/6761
- Adds a basic storybook story

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/components/elements/CommonMenuButton.tsx
```diff
@@ -14,18 +14,19 @@ import { To } from 'react-router-dom';
 import RouterLink from './RouterLink';
 import { MoreMenuIcon } from './SemanticIcons';
 
-export type NavMenuItem = {
+export type CommonMenuItem = {
```
martha, line 18: This diff is [copied from the File Upload PR](https://github.com/greenriver/hmis-frontend/pull/982/files#diff-49db4272ddc2c53915ef42e842941e61758a8b7830183f264a219427aeb04cf7R14) which will go into release-145

src/modules/hmis/hmisUtil.ts
```diff
@@ -635,3 +635,24 @@ export const raceEthnicityDisplayString = (race?: Race[]) => {
 
   return race.map((r) => HmisEnums.Race[r]).join(', ');
 };
+
+// TODO @MARTHA discuss - it's probably a bit silly to redefine ranks here,
+// see https://github.com/greenriver/hmis-warehouse/blob/1a170c0e869fd3d69880532f86baa104b800d9ed/drivers/hmis/app/graphql/types/hmis_schema/enums/hud.rb#L723-L733
+const hohPriorityMapping: Record<RelationshipToHoH, number> = {
```
martha: This is for sorting the household members by their relationship to HoH. This definition of the ordering is redundant with the [backend enum definition](https://github.com/greenriver/hmis-warehouse/blob/1a170c0e869fd3d69880532f86baa104b800d9ed/drivers/hmis/app/graphql/types/hmis_schema/enums/hud.rb#L723-L733), so I wanted to discuss whether there's a better way to pass that info to the frontend.

src/modules/projects/components/tables/ProjectHouseholdsTable.tsx
```diff
@@ -215,6 +345,36 @@ const ProjectHouseholdsTable = ({
       }}
       queryDocument={GetProjectHouseholdsDocument}
       columns={columns || defaultColumns}
+      overrideTableBody={true}
+      renderRow={(household, columnKeys) => {
+        return (
+          <>
+            <TableBody
+              // TODO @martha - discuss these TableBodies to achieve the `rowgroup` role, is it worth it?
```
martha: Using the `role='rowgroup'` turned out to be a bit complicated because of MUI expectations around usage of the `TableBody`, `TableRow` and `TableCell` components. We can't directly have `renderRow` return a `<div role='rowgroup'>` which contains all the `TableRow`s within it, because MUI expects that `TableRow` is a direct child of `TableBody`. (Here's a screenshot of what happens:)
Screenshot 2024-11-22 at 12 28 25 PM

Rendering each household as its own `tbody` is the best solution I was able to come up with, but I'm not a huge fan -- both because of the messy markup and because of bloating the already complicated `GenericTable` api. Let me know if you can think of any better solutions. Or, we could skip using `rowgroup`. I think the accessibility is still better than before. We could, for example, give the row header cell additional accessible text specifying the household (HoH name or household ID) that would be read when navigating the screen with a screenreader.

src/modules/projects/components/tables/ProjectHouseholdsTable.stories.tsx
```diff
@@ -0,0 +1,26 @@
+import { Card } from '@mui/material';
```
martha, line 1: We could add more stories here to test different cases, but I was thinking to leave it as is and create more fleshed-out stories for `GenericTable` with 6761, or later. Let me know what you think.

src/modules/projects/components/tables/ProjectHouseholdsTable.tsx
```diff
@@ -158,6 +180,108 @@ export const HOUSEHOLD_COLUMNS: {
...
+
+const CustomDividerRow = ({ colSpan }: { colSpan: number }) => (
+  <TableRow sx={{ height: '8px' }}>
+    <TableCell
+      colSpan={colSpan}
+      sx={(theme) => {
+        return {
+          padding: 0,
+          // TODO - remove after confirming on PR
+          // note: color specified in designs ('#f9f9f9') is slightly lighter than grey[200] ('#eeeeee') and slightly darker than grey[100] ('#f5f5f5')
```
martha: Pending design feedback https://www.figma.com/design/kqZiUPjXvqSvJVslJCnlTB?node-id=8555-31840#1053502385

src/components/elements/table/TableRowActions.tsx
```diff
@@ -0,0 +1,63 @@
+import { Stack } from '@mui/material';
+import { useMemo } from 'react';
+import ButtonLink from '../ButtonLink';
+import CommonMenuButton, { CommonMenuItem } from '../CommonMenuButton';
+
+export type TableRowAction<T> = Omit<CommonMenuItem, 'to' | 'onClick'> & {
+  // getUrl: given a record, return the URL this action should link to.
+  // E.g. for the "View Client" action on an Enrollment table, given the Enrollment record, return the link to the Client.
+  // getUrl could be made optional and onClick (omitted above) could be re-enabled
+  getUrl: (record: T) => string;
```
martha: API changes to this component? 
- Sometimes (like in bulk services) it needs to depend more on the row, be more configurable. 
  - Should be able to pass an action button directly like a React node?
  - getTableRowActions, similar to getColumnDefs ?
- Passing the primaryAction and secondaryActions so that you can have secondary only
- extend our current api to keep using rowLinkTo? I think this is a bit misleading (and also ambiguous what the record linking to the row should be named)

```
interface GenericTableProps<T> {
  ...
  getTableRowActions?: (row: T) => TableRowAction<T>[] | ReactNode;
  // or maybe just get rid of `TableRowAction<T>[]` and have it always return `ReactNode`?
}
```

```

interface TableRowActionsProps<T> {
  record: T;
  recordName?: string;
  primaryAction: TableRowAction<T> | ReactNode;
  secondaryActions: TableRowAction<T>[] | ReactNode[];  // don't need to allow ReactNode here?
}
```

src/components/elements/CommonMenuButton.tsx
gigxz: Is this still applicable? The API of this component is kind of odd, but it seems like `title` is optional because it's ignored if you pass `iconButton`. A more standard API would be to use the children as the button content. Either way, this change doesn't seem related to the rest of the PR?

src/components/elements/table/TableRowActions.tsx
```diff
@@ -0,0 +1,63 @@
...
+    [record.id, recordName]
+  );
+
+  const { primaryAction, secondaryActions } = useMemo(
+    () => getActions(record),
+    [getActions, record]
+  );
+
+  return (
+    <Stack direction='row' alignItems='center' gap={0.5}>
```
gigxz:
```suggestion
    <Stack direction='row' alignItems='center' justifyContent='end' gap={0.5}>
```

I believe this should be right-aligned
Screenshot 2024-12-20 at 1 03 45 PM

src/components/elements/table/GenericTable.tsx
```diff
@@ -62,8 +73,14 @@ export interface Props<T> {
...
   filterToolbar?: ReactNode;
   noData?: ReactNode;
-  renderRow?: (row: T) => ReactNode;
+  renderRow?: (row: T, columnKeys: string[]) => ReactNode;
   condensed?: boolean;
+  // if overrideTableBody is true, GenericTable doesn't render a `tbody` element.
+  // This should only be used by tables that take over rendering using renderRow and render a `tbody` within their custom render fn
+  overrideTableBody?: boolean;
+  injectBelowRows?: ReactNode; // component to inject below all rendered rows, above footer
+  getTableRowActions?: (record: T) => TableRowActionsType;
```
gigxz: I'm wondering about the decision to make this a prop as opposed to just another column definition, can you say more about that? It does add more complexity to the already-complex GenericTable, so just wanting to think that through.

By a column definition, I mean:

```

const columns = [
         ...othercols,
        {
          key: 'actions',
          render: (enrollment) => (
            <TableRowActions
              record={enrollment}
              primaryAction={viewEnrollmentMenuItemConfig(enrollment}
              secondaryActions={[viewClientMenuItemConfig(enrollment)]}
            />
          ),
        },
]


...
<GenericTableWithData columns={columns} />
```

gigxz: Looking at https://github.com/greenriver/hmis-frontend/pull/998/files, I wonder if this would also give us some more flexibility for use-cases when the action button is more complex. I haven't quite wrapped my head around everything yet, but would something like this simplify the changes in the other PR ?

Definitely an early thought, I'm sure there are things I'm missing about the implementation.



```js

const columns = [
         ...othercols,
        {
          key: 'actions',
          render: (enrollment) => (
            <TableRowActions
              record={enrollment}
              // use this prop if the primary action is a simple navigation
              primaryActionConfig={viewEnrollmentMenuItemConfig(enrollment}
              // use this prop if the primary action is more complicated, or if needs some special treatment
              renderPrimaryAction={<AssignServiceButton .... />}
              // maybe secondary actions are always expected to be navigation
              secondaryActionsConfig={[viewClientMenuItemConfig(enrollment)]}
            />
          ),
        },
]


...
<GenericTableWithData columns={columns} />
```

src/modules/projects/components/tables/ProjectHouseholdsTable.tsx
gigxz: using several `tbody`s seems like a fine approach. from [mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tbody#usage_notes),

> It's permitted to use more than one <tbody> per table as long as they are all consecutive. This allows to divide the rows ([<tr>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tr) elements) in large tables into sections, each of which may be separately formatted if so desired.

That seems like basically what we are doing.

`overrideTableBody` prop maybe isn't the clearest as to what it's doing, since it's really omitting the `tbody` tag entirely, not overriding it.  Maybe another idea is passing `TableBodyComponent={React.Fragment}` where the default is `TableBodyComponent = TableBody`, then the passed component can be used directly? Either seems OK.

src/modules/projects/components/tables/ProjectHouseholdsTable.tsx
```diff
@@ -215,6 +347,36 @@ const ProjectHouseholdsTable = ({
       }}
       queryDocument={GetProjectHouseholdsDocument}
       columns={columns || defaultColumns}
+      overrideTableBody={true}
+      renderRow={(household, columnKeys) => {
+        return (
+          <>
+            <TableBody
```
gigxz: I'd love to discuss a couple ideas here:

1. Can we reduce the complexity of `ProjectHouseholdsClientRow` by using column definitions to render the cells or rows directly? This would require some refactoring of GenericTable to pull out some components, so we can use them directly. This would let us rely on existing cell rendering logic and "ColumnDef" implementations. For example:

```js
// A: TableRow render logic is refactored out of GenericTable into a shared component that we can use here
      renderRow={(household, columnKeys) => {
        return (
          <TableBody role='rowgroup'>
            <CustomDividerRow colSpan={columnKeys.length} />
            {household.householdClients.sort(hohSort).map((hc) => (
              <GenericTableRow row={hc} columns={householdClientColumnDefs.filter(is in columnKeys)} />
            ))}
          </TableBody>
        );

```
or 
```js
// B: TableCell render logic is refactored out of GenericTable into a shared component that we can use here

      renderRow={(household, columnKeys) => {
        return (
          <TableBody role='rowgroup'>
            <CustomDividerRow colSpan={columnKeys.length} />
            {household.householdClients.sort(hohSort).map((hc) => (
              <TableRow>
                {householdClientColumnDefs.filter(is in columnKeys).map((columnDef) => (
                  <GenericTableCell columnDef={columnDef} row={hc} />
                ))}
              </TableRow>
            ))}
          </TableBody>
        );
```




2) There would need to be that added filtering on columnKeys, since GenericTableWithData can adjust the visible columns based on the Optional Columns functionality, which this table will need to support.  It seems like this `showAssignedStaff` thing is done a little differently because it's conditionally shown based on `staffAssignmentsEnabled` flag on the project. But we will need to support actual optional columns on this table, though.

Here is an example I was messing with for the column definitions:
```
type OneHouseholdClient = HouseholdFields['householdClients'][number];
// column defs for individual client that are used for rendering cells
const householdClientColumnDefs: ColumnDef<OneHouseholdClient>[] = [
  { header: 'Client Name', render: (hc) => clientBriefName(hc.client) },
  { header: 'Age', render: (hc) => hc.client.age },
  { header: 'Relationship', render: (hc) => hc.relationshipToHoH },
  { header: 'Entry Date', render: () => '' }, //etc
  { header: 'Exit Date', render: () => '' },
  { header: 'Status', render: () => '' },
  {
    key: 'actions',
    header: <Box sx={visuallyHidden}>Actions</Box>,
    render: () => '',
  },
];

// dummy column defs for HOUSEHOLD that are only used for the headers, not for rendering cells
const householdColumnDefs: ColumnDef<HouseholdFields>[] =
  householdClientColumnDefs.map(({ header, key }) => ({
    header,
    key,
    render: () => null,
  }));

```

I think those 2 pieces of feedback go together, but maybe we could resolve (2) without addressing the refactoring desire of (1).

src/components/elements/CommonMenuButton.tsx
martha: The `title` prop of `CommonMenuButton` should probably be optional (or the whole api should change, as you suggested), since it is not used if you pass `iconButton`. But the `title` attribute of the `CommonMenuItem` is what I've changed here. My understanding is that it should not be optional, since menu items always need text; they can't be icon-only. Happy to revert this change though if I'm missing something.

Yes, my original comment is no longer applicable.

src/modules/projects/components/tables/ProjectHouseholdsTable.tsx
martha: Ok, thanks for validating the approach! I like your api suggestion. 👍

martha: Ok, thank you for validating the overall approach! I like your api suggestion, took a stab at implementing that here: https://github.com/greenriver/hmis-frontend/pull/989/commits/3586e74ace3065fa5d782fe4b68a12a6106d67d1

src/components/elements/table/GenericTable.tsx
martha: I didn't think thoroughly about it, but it was based on the idea that, since this is a new pattern implemented consistently across many tables in the app, we want the common table component to guide usage and force consistency. 

Thinking beyond table row actions, the logical conclusion of that train of thought is that all table controls, like search, filter, toggle, add-record button, etc. live as props on the GenericTable. That doesn't seem desirable - it certainly would lead to a bloated GenericTable. But I guess forcing consistency is the one upside. Currently we rely on GenericTable callers to implement search bars and add-one buttons consistently every time, and there are places in the app where we aren't consistent. (For example, search and toggle on enrollments table vs. search on projects/orgs.)

Anyway, I guess those considerations seem out of scope... I don't feel strongly about it and can take a stab at implementing your suggestion.

martha: quick example of the consistency issue I'm mentioning, as actually applied to table row actions: https://github.com/greenriver/hmis-frontend/blob/martha/6825-hh-member-row/src/components/elements/table/GenericTable.tsx#L259-L261 By moving this out of GenericTable and into a regular column definition, we leave the column header (or lack thereof / accessibility thereof) up to the caller

martha: I've taken a shot at the refactor, I'd like to look at it together tomorrow: https://github.com/greenriver/hmis-frontend/compare/martha/6761-tables...martha/6761-tables-refactor?expand=1

gigxz: > we leave the column header (or lack thereof / accessibility thereof) up to the caller

This seems like something that should be addressed globally for all columns. The [ColumnDef](https://github.com/greenriver/hmis-frontend/blob/91e938c0daf2bfa026f2cc45de663d3e20c24812/src/components/elements/table/types.tsx#L8) has the `header` attribute as optional, so any column could be missing a header and become inaccessible. 

Looking at how we implement the Header cells:

https://github.com/greenriver/hmis-frontend/blob/91e938c0daf2bfa026f2cc45de663d3e20c24812/src/components/elements/table/GenericTable.tsx#L225-L236

it could be modified so that it sets a hidden header if no visible `header` is provided:
```
            <HeaderCell>
              {def.header ? (
                <strong>{def.header}</strong>
              ) : (
                <Box sx={visuallyHidden}>
                  {def.key || def.someOtherFieldThatsRequiredIfHeaderIsMissing}
                </Box>
              )}
            </HeaderCell>
```

That might address the concern more widely -- we could add additional attributes to ColumnDef if needed, to ensure that columns always have accessible headers.

src/modules/projects/components/tables/ProjectHouseholdsTable.tsx
martha: Use `renderCellContents`

src/modules/client/components/ClientEnrollmentCard.tsx
```diff
@@ -71,14 +69,21 @@ const RecentEnrollments = ({
...
-        ENROLLMENT_STATUS_COL,
+        ENROLLMENT_COLUMNS.enrollmentStatus,
         {
           key: 'name',
           header: 'Name',
-          linkTreatment: true,
           render: 'projectName',
         },
-        ENROLLMENT_PERIOD_COL,
+        {
```
martha, line 78: Pulled this out to here because this is now the only place where it is being used.

src/modules/hmis/components/DateWithRelativeTooltip.tsx
```diff
@@ -0,0 +1,59 @@
+import {
```
martha: I pulled this change into here, from the other PR, since it's relevant to the Enrollment tables updated in this PR.

src/modules/projects/components/tables/ProjectClientEnrollmentsTable.tsx
```diff
@@ -67,82 +46,78 @@ const isHouseholdWithStaff = (
...
-    ),
-    linkTreatment: true,
+  enrollmentStatus: {
+    header: 'Status',
+    render: (e) => <EnrollmentStatus enrollment={e} />,
   },
-  enrollmentStatus: ENROLLMENT_STATUS_COL,
+};
+
+type WithEnrollment = {
```
martha, line 108: I pulled in this change too from my other PR, again since it's relevant to the two tables touched in this PR. I'll copy over my comment from there:

I'm curious for feedback on this pattern. I was aiming for something a bit equivalent to a HasEnrollments concern, in ruby. In this case the WITH_ENROLLMENTS_COLUMN should be applicable to any model that has an enrollment. ~I don't love using ColumnDef any, and think there might be a better way around it -- so maybe we could look into it together.~ I guess we could consider this an incomplete preview of a more centralized table column organization.

src/components/elements/table/GenericTable.tsx
martha: good idea, I've left a todo [here](https://github.com/greenriver/hmis-frontend/pull/989/files#diff-88d14b02fbc39c3f387add19bc50ecd16c67d9e75418599fca7968427e0713b1R12) that I'd like to address in [the other PR ](https://github.com/greenriver/hmis-frontend/pull/998)

gigxz:
Is this ready for re-review @martha? And can you re-target to 147? thanks!

martha:
@gigxz, yes, I think it's ready for re-review. Thank you!

src/components/elements/table/GenericTable.tsx
```diff
@@ -62,8 +69,12 @@ export interface Props<T> {
   >;
   filterToolbar?: ReactNode;
   noData?: ReactNode;
-  renderRow?: (row: T) => ReactNode;
+  // columnKeys contains the keys of columns currently rendered, so renderRow knows about which optional columns are shown/hidden.
+  renderRow?: (row: T, columnKeys: string[]) => ReactNode;
   condensed?: boolean;
+  // TableBodyComponent can be overridden. This should only be used by tables that take over rendering using renderRow and render a `tbody` within their custom render fn
+  TableBodyComponent?: ComponentType | keyof JSX.IntrinsicElements;
+  injectBelowRows?: ReactNode; // component to inject below all rendered rows, above footer
```
gigxz: I don't love "inject", it feels kinda strong for something that is essentially just a child. What do you think of these alternatives:
* `belowRowsContent`
* `belowRowsChildren`
* `renderBelowRows` (I don't like this as much because "render" to me implies that it is a render function

src/components/elements/table/tableRowActionUtil.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+// TODO(#6761) - enforce header-less columns always have accessible text in GenericTable
+export const BASE_ACTION_COLUMN_DEF: ColumnDef<any> = {
+  header: <Box sx={visuallyHidden}>Actions</Box>,
+  key: 'Actions',
+  tableCellProps: { sx: { p: 0 } },
+  render: '', // gets overridden when used
+};
+
+export const getViewClientAction = (client: ClientNameFragment) => {
```
gigxz: Naming suggestion: `getViewClientMenuItem` and `getViewEnrollmentMenuItem` since these are returning `CommonMenuItem` objects

src/components/elements/table/tableRowActionUtil.tsx
```diff
@@ -0,0 +1,44 @@
...
+};
+
+export const getViewEnrollmentAction = (
+  enrollment: Pick<EnrollmentFieldsFragment, 'id' | 'entryDate' | 'exitDate'>,
+  client: Pick<ClientNameFragment, 'id'> | ClientNameFragment
+) => {
+  return {
+    title: 'View Enrollment',
+    key: 'enrollment',
+    ariaLabel: `View Enrollment, ${client.hasOwnProperty('firstName') ? clientBriefName(client as ClientNameFragment) : ''} ${entryExitRange(enrollment)}`,
```
gigxz: Are there situations where we don't have the client name? Ts locally seemed just as happy with:

```ts
export const getViewEnrollmentAction = (
  enrollment: Pick<EnrollmentFieldsFragment, 'id' | 'entryDate' | 'exitDate'>,
  client: ClientNameFragment
): CommonMenuItem => {
  return {
    title: 'View Enrollment',
    key: 'enrollment',
    ariaLabel: `View Enrollment, ${clientBriefName(client)} ${entryExitRange(enrollment)}`,
    to: generateSafePath(EnrollmentDashboardRoutes.ENROLLMENT_OVERVIEW, {
      clientId: client.id,
      enrollmentId: enrollment.id,
    }),
  };
};

```

src/modules/hmis/components/DateWithRelativeTooltip.tsx
gigxz: Nice. I never noticed that this was part of the design. It does look like we are missing part of the tooltip contents:

> Most accessible text for cells with hover-over interactions should include both cell and tooltip text (2/5/25, 6 days ago’)

gigxz: I think this file and RelativeDateDisplay should go in `src/components/elements` with the rest of the relative date components. (And hopefully we can consolidate later).
The `hmis` module has general hmis-domain-specific components, like `HmisField`,`HohIndicator`, etc. This one isn't domain specific so I think it goes in src.

src/modules/hmis/components/RelativeDateDisplay.tsx
```diff
@@ -7,15 +7,29 @@ import {
...
+    preciseTime ? formatDateTimeForDisplay(date) : formatDateForDisplay(date),
+    formatRelativeDateTime(date),
+  ];
+};
+
 export interface RelativeDateDisplayProps {
   dateString: string;
   prefixVerb?: string;
   suffixText?: string;
+  tooltipSuffixText?: string;
```
gigxz, line 34: I don't see this being used anywhere

src/modules/hmis/hmisUtil.ts
gigxz: I don't think it's redundant with the backend. The enum lists out the values which are the HUD numeric values (Eg child = 2) whereas here you're specifying a custom order for displaying household members. It seems correct to me that this would live on the frontend.

src/modules/hmis/hmisUtil.ts
```diff
@@ -639,3 +635,24 @@ export const raceEthnicityDisplayString = (race?: Race[]) => {
...
+  [RelationshipToHoH.SelfHeadOfHousehold]: 0,
+  [RelationshipToHoH.SpouseOrPartner]: 1,
+  [RelationshipToHoH.Child]: 2,
+  [RelationshipToHoH.OtherRelative]: 3,
+  [RelationshipToHoH.UnrelatedHouseholdMember]: 4,
+  [RelationshipToHoH.DataNotCollected]: 5,
+  [RelationshipToHoH.Invalid]: 6,
+};
+
+export const hohSort = (
```
gigxz: We have another function `sortHouseholdMembers` that sorts by: "current enrollment" first, then by last name, then by ID. That function determines how clients are display on the Edit Household members table, the Enrollment dashboard (I think) and maybe others. I think it is probably better if we just modify that function, rather than introducing a new one.

Let's keep the optional "active enrollment id" , so in other places the "current enrollment" shows up first, to preserve existing behavior.

So something like:

```diff
// Sort household members into "household order" which is HoH > spouse > child > relative > unrelated.
// If activeEnrollmentId is provided, that person will appear at the top regardless of relationship to hoh.
export const sortHouseholdMembers = (
  members?: HouseholdClientFieldsFragment[],
  activeEnrollmentId?: string
) => {
  const sorted = sortBy(members || [], [
    (c) => (c.enrollment.id === activeEnrollmentId ? -1 : 1),
+   (c) => hohPriorityMapping[c.relationshipToHoH], // HoH > spouse > child > relative > unrelated
    (c) => c.client.id, // deterministic tie-breaker
  ]);
  return sorted;
};
```

src/modules/projects/components/tables/ProjectHouseholdsTable.tsx
```diff
@@ -215,6 +181,36 @@ const ProjectHouseholdsTable = ({
...
       columns={columns || defaultColumns}
+      TableBodyComponent={React.Fragment}
+      renderRow={(household, columnKeys) => {
+        return (
+          <TableBody
+            // Render each household as its own `tbody` in order to use the role='rowgroup' for better accessibility markup
+            key={household.id}
+            role='rowgroup'
+          >
+            <CustomDividerRow colSpan={(columns || defaultColumns).length} />
```
gigxz, line 192: should the colspan be determined by `columnKeys` since there may be optional columns?

gigxz (commented):
looks good!

src/modules/projects/components/tables/ProjectHouseholdsTable.tsx
martha: `columnKeys` is tricky because `key` is not required on the ColumnDef type, so that list includes only those columns which do provide keys. We could update GenericTable to so that `columnKeys` includes some representation of every column, either the key or the header, but that is also a bit tricky since the header could be a ReactNode.

We can't use `columnKeys` for the `belowRowsContent` below, unless we update that to accept a function, which seems like too much cruft to me.

It seems ok to use `(columns || defaultColumns)` even with the optional column. It doesn't seem to do any harm for a `colSpan` to be provided that's longer than the width of the table.

gigxz: Ah okay. We should probably make `key` required if its used in this way. But that makes sense to me its OK if the colspan is longer. thanks!

src/modules/projects/components/tables/ProjectHouseholdsTable.tsx
gigxz: I like where we landed 🙂  less churn. Thanks!

src/modules/projects/components/tables/ProjectClientEnrollmentsTable.tsx
gigxz: This makes sense to me, but would be happy to discuss further if you'd like!

gigxz (approved):
📈  looks good! thanks for your work and iteration!!

gigxz:
@martha merge conflicts might be a pain with this and https://github.com/greenriver/hmis-frontend/pull/998? If you want to hold off on squashing this into main until it contains https://github.com/greenriver/hmis-frontend/pull/998, that would be fine w/ me. (I think that would avoid the conflicts)


Another option is a deploy branch, so this can get plenty of review on QA with design (without being tied to regular releases) - lets revisit Monday!

PR #992, opened 2024-12-09
Geolocation input component

gigxz:
## Description
Issue: https://github.com/open-path/Green-River/issues/5726
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4993

This PR adds a GeolocationInput component that requests location from the browser [Geolocation API](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation_API). If allowed, it displays the coordinates on a basic Leaflet map.

It also displays the coordinates on a map when viewing form data in read-only mode.

See issue for testing instructions, including sample forms to use.

### [See component in chromatic](https://66e877f06b978106d75d04bc-hsdtdiuqhn.chromatic.com/?path=/story/modules-geolocation-components-geolocationinput--labeled&args=value.accuracy:25.079;value.latitude:43.6462185;value.longitude:-72.3220551;value.altitude:!null;value.altitudeAccuracy:!null;value.heading:!null;value.speed:!null)

when collecting on CLS:
Screenshot 2024-12-12 at 5 53 04 PM

when viewing collected location on read-only/submitted form:
Screenshot 2024-12-13 at 7 18 22 AM

## Type of change


gigxz:
Builds available at https://www.chromatic.com/builds?appId=66e877f06b978106d75d04bc&branch=gig%2F5726-geolocation-input

src/modules/geolocation/components/BaseMap.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { MapContainer, Marker, TileLayer } from 'react-leaflet';
+import { LatLon } from '@/modules/geolocation/types';
+
+// Fix for default marker icon issue with Vite
+L.Icon.Default.prototype.options.iconUrl = markerIconUrl;
+L.Icon.Default.prototype.options.iconRetinaUrl = markerIconRetinaUrl;
+L.Icon.Default.prototype.options.shadowUrl = markerShadowUrl;
+L.Icon.Default.imagePath = '';
+
+interface BaseMapProps {
```
gigxz: this component will get a significant refactor in my Location Map PR

src/modules/geolocation/components/GeolocationInput.tsx
```diff
@@ -0,0 +1,105 @@
...
+                borderColor: 'borders.dark',
+                borderRadius: 1,
+                borderWidth: 1,
+                borderStyle: 'solid',
+                height: mapHeight,
+              }}
+              aria-live='polite' // Notify screen readers of content updates
+            >
+              {loading ? (
+                <Typography color='links' fontWeight={700}>
```
martha: Maybe a loading spinner would better match our typical pattern in other places in the app? Or keep the text but apply a font weight/size that we use more elsewhere in the app? This color/fontweight combo looks a bit out of place to me.

src/modules/geolocation/components/GeolocationInput.tsx
```diff
@@ -0,0 +1,105 @@
+import {
```
martha: I've been mildly regretting introducing language of "geo" since it's long and complicated compared to simply "location." I just googled to find out if there is actually a semantic difference, and was reminded that "geolocation" refers to getting a device's location (like via satellite or just IP address).

Given that in the future this input will (probably?) also support entering an address and/or dragging and dropping a pin, maybe LocationInput is the more appropriate, as well as simpler, name?

src/modules/geolocation/components/GeolocationInput.tsx
```diff
@@ -0,0 +1,105 @@
...
+            variant='outlined'
+            aria-labelledby={label ? labelId : undefined}
+            disabled={disabled}
+          >
+            Request Location
+          </Button>
+          {coordinates && (
+            <Button
+              variant='gray' // TODO: update to outlined grayscale
+              onClick={() => onChange(null)}
```
martha: a11y note - when this button is clicked it causes its own disappearance. If you don't think it's too much code clutter, we can improve the screenreader ux by adding a `buttonRef` to the Request Location button and then calling `buttonRef.current.focus()` here in the `onClick` callback.

src/modules/geolocation/components/BaseMap.tsx
```diff
@@ -0,0 +1,50 @@
...
+        doubleClickZoom
+        touchZoom
+      >
+        <>
+          <TileLayer
+            url='https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png'
+            attribution='&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors'
+          />
+          {/* todo: add Popup children for location map */}
+          <Marker position={[coordinates.latitude, coordinates.longitude]} />
```
martha: Marker should probably be given a title or alt text? Maybe the coordinates is good enough for now. (I think even better would be a description, like "in San Francisco on Main Street" but we probably don't want to pull in a library to do that)

docker/templates/non_proxy_headers.conf.template
```diff
@@ -1,10 +1,11 @@
...
 
 add_header Referrer-Policy "strict-origin-when-cross-origin";
 add_header Strict-Transport-Security "max-age=63072000; includeSubdomains;" always;
 add_header X-Frame-Options "SAMEORIGIN" always;
 add_header X-XSS-Protection "1; mode=block" always;
 add_header X-Content-Type-Options "nosniff" always;
-add_header Feature-Policy "geolocation 'none'; midi 'none'; notifications 'none'; push 'none'; sync-xhr 'none'; microphone 'none'; camera 'none'; magnetometer 'none'; gyroscope 'none'; speaker 'none'; vibrate 'none'; fullscreen 'none'; payment 'none'; usb 'none';";
+add_header Feature-Policy "geolocation 'self'; midi 'none'; notifications 'none'; push 'none'; sync-xhr 'none'; microphone 'none'; camera 'none'; magnetometer 'none'; gyroscope 'none'; speaker 'none'; vibrate 'none'; fullscreen 'none'; payment 'none'; usb 'none';";
+add_header Permissions-Policy "geolocation=(self)";
```
gigxz, line 11: This change allows our site to request Geolocation from the browser.

`Permissions-Policy` is the modern version of `Feature-Policy`.

I think it's right to put this change here in `non_proxy_headers.conf.template`. My only hesitation is that these allowances are now also included for `/assets`, which I don't _think_ they need to be. (It worked if I just put the add_header in the `location /` block and not `location /assets`). I don't want to grant additional allowances if not necessary. I am not clear on the risk though. Referring to:

https://github.com/greenriver/hmis-frontend/blob/06ba81e97cb1f1903d8f6dd6216d692e979b85ca/docker/templates/default.conf.template#L33-L34

gigxz: @yakloinsteak Can you review this nginx change for me? I confirmed it's working on QA.

In summary:
1. Allow fetching images from `https://*.tile.openstreetmap.org`, to support rendering maps
2. Allow requesting geolocation from the browser

docker/templates/non_proxy_headers.conf.template
```diff
@@ -1,10 +1,11 @@
...
+add_header Content-Security-Policy "default-src 'none'; connect-src 'self' *.sentry.io *.s3.amazonaws.com; script-src 'self' 'unsafe-inline' 'unsafe-eval'; style-src 'self' fonts.googleapis.com 'unsafe-inline'; img-src 'self' data: blob: *.s3.amazonaws.com https://*.tile.openstreetmap.org; font-src 'self' fonts.gstatic.com; object-src 'none';" always;
 
 add_header Referrer-Policy "strict-origin-when-cross-origin";
 add_header Strict-Transport-Security "max-age=63072000; includeSubdomains;" always;
 add_header X-Frame-Options "SAMEORIGIN" always;
 add_header X-XSS-Protection "1; mode=block" always;
 add_header X-Content-Type-Options "nosniff" always;
-add_header Feature-Policy "geolocation 'none'; midi 'none'; notifications 'none'; push 'none'; sync-xhr 'none'; microphone 'none'; camera 'none'; magnetometer 'none'; gyroscope 'none'; speaker 'none'; vibrate 'none'; fullscreen 'none'; payment 'none'; usb 'none';";
+add_header Feature-Policy "geolocation 'self'; midi 'none'; notifications 'none'; push 'none'; sync-xhr 'none'; microphone 'none'; camera 'none'; magnetometer 'none'; gyroscope 'none'; speaker 'none'; vibrate 'none'; fullscreen 'none'; payment 'none'; usb 'none';";
```
gigxz, line 10: Some of these are not valid, like `vibrate`, and they cause warnings. I'll remove them in another PR

src/modules/geolocation/components/GeolocationInput.tsx
```diff
@@ -0,0 +1,105 @@
...
+              ) : (
+                <Typography color='text.secondary'>
+                  Location not collected
+                </Typography>
+              )}
+            </Box>
+          )}
+          {coordinates && (
+            <Box
+              aria-live='polite' // Announce when the map is displayed
```
martha: I think this `aria-live='polite'` won't reliably announce, since aria-live announces that the component's internal text changes, but not when the component is added to or removed from the dom. (See "Live regions that are already "populated" don't work" https://tetralogical.com/blog/2024/05/01/why-are-my-live-regions-not-working/)

Locally I tried out moving the `aria-live='polite'` up to line 63 on the wrapping Box, but the issue with that is that it then announces all internal text changes to the contents, so when the map appears it announces "Leaflet | copyright OpenStreetMap contributors".

martha (approved):
wonderful :)

gigxz:
Thanks for the thorough review @martha, I will take a look at those!

src/modules/geolocation/components/GeolocationInput.tsx
ttoomey: Maybe... "GeoLocation" seems fine to me since "Location" is often used for URL/browser history. So the more descriptive name helps with searchability

ttoomey (approved):
Looks great, thanks!

src/modules/geolocation/components/GeolocationInput.tsx
gigxz: I agree on searchability, "location" alone might be challenging. `clientLocation` could be sufficient, but maybe not worth changing

src/modules/geolocation/components/GeolocationInput.tsx
gigxz: Yeah you're right, it is definitely new/custom typography. I'm going to stick with what was in Figma due to time constraints on dev/design

src/modules/geolocation/components/GeolocationInput.tsx
gigxz: Oh thanks for the reference. I tried to google this but came up with responses saying that it should work (as I interpreted, anyway) but I didn't test. So that is good to know

gigxz: This react explanation that was linked from the article is helpful. https://dev.to/abbeyperini/live-regions-in-react-4dmd
 I'll update this to the container is always rendered

src/modules/geolocation/components/GeolocationInput.tsx
gigxz: Thanks, updated! That helps with keyboard nav too.

src/modules/geolocation/components/GeolocationInput.tsx
gigxz: I got stuck here- moving along for now but I will come back to it

src/modules/geolocation/components/BaseMap.tsx
gigxz: Added a basic title "client location". The enrollment location map (other pr) will have content in the tooltip, but I don't think we need that here. It's probably not helpful to read aloud the coordinates, right?

martha: I guess so yeah. It's too bad though, since without the coordinates in the marker description, the actual location data is info that is only available to sighted users. We can leave it for now though; we could add a ticket for a future improvement, for example when we add additional capabilities like entering an address to drop a pin, that might be a good time to bring in a library that can convert a coordinate pair into a useful verbal description.

src/modules/geolocation/components/GeolocationInput.tsx
martha: I remembered that there is actually somewhere we use the "geolocation" language in a user-facing context, it's in the PIT form. Even if we update that (not in scope for this PR), "geolocation" language is also part of the form definition API, and that for sure doesn't seem worth changing. So I agree that consistently using "geolocation" could help with searchability. I'm good with whatever you decide @gigxz !

src/modules/geolocation/components/GeolocationInput.tsx
gigxz: https://github.com/open-path/Green-River/issues/7122

src/modules/geolocation/hooks/useGeolocation.tsx
```diff
@@ -1,16 +1,23 @@
 import { useCallback, useState } from 'react';
```
gigxz: * Move hook: `src/modules/geolocation/hooks` => `src/hooks`
* Move input component: `src/modules/geolocation/components` => `src/components/elements/input` as well as related components (Help Dialog, Error Alert)
* Move map components: `src/modules/geolocation/components` => `src/components/elements/maps`
  * Additional generic map components from https://github.com/greenriver/hmis-frontend/pull/997 (excluding EnrollmentLocationMap) will also go there
* Remove `modules/geolocation` since it's not domain-specific to anything

Does this sound right base on our conversation @ttoomey @martha?

gigxz: Updated accordingly. I moved the types file to a global location `src/types/geolocationTypes.ts`.

PR #996, opened 2024-12-13
Bug fix for numeric year inputs

gigxz:
## Description

1) Fix new issue where integer fields that represent "year" are displayed with separators. Adjust behavior so we only add separators for currency fields. Future improvement is to add specialized year input https://github.com/open-path/Green-River/issues/7096
Screenshot 2024-12-13 at 7 34 23 AM
=>
Screenshot 2024-12-13 at 7 34 40 AM

2) Fix pre-existing issue where currencies are not displayed right in DynamicView `$150.5` instead of `$150.50`. This is because we store currencies as floats. The input field still has this issue, but that is a little trickier to resolve without using fixed decimals for all currency fields, which we may not want.
Screenshot 2024-12-13 at 7 56 39 AM
=>
Screenshot 2024-12-13 at 8 03 33 AM

Issue: https://github.com/open-path/Green-River/issues/7095

## Type of change
- [x] Bug fix

### DIFF
--- a/src/modules/form/components/viewable/DynamicViewField.tsx
+++ b/src/modules/form/components/viewable/DynamicViewField.tsx
@@ -5,10 +5,7 @@ import React, { useMemo } from 'react';
 
 import { getValueFromPickListData, usePickList } from '../../hooks/usePickList';
 import { DynamicViewFieldProps } from '../../types';
-import {
-  formatNumberWithTrailingZeroes,
-  isDataNotCollected,
-} from '../../util/formUtil';
+import { isDataNotCollected } from '../../util/formUtil';
 import DynamicDisplay from '../DynamicDisplay';
 
 import File from './item/File';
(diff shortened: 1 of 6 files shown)


src/modules/form/components/viewable/DynamicViewField.tsx
```diff
@@ -149,7 +152,12 @@ const DynamicViewField: React.FC<DynamicViewFieldProps> = ({
       }
       return <TextContent {...commonProps} />;
     case ItemType.Currency:
-      return <TextContent {...commonProps} renderValue={(val) => `$${val}`} />;
+      return (
+        <TextContent
+          {...commonProps}
+          renderValue={(val) => `$${formatNumberWithTrailingZeroes(val)}`}
```
ttoomey: Looks like we already have a currency formatter implemented using `Intl.NumberFormat`

Can we just use that instead?
`import { formatCurrency} from '@/modules/hmis/hmisUtil';`
```suggestion
          renderValue={(val) => `${formatCurrency(val)}`}
```

ttoomey (approved):
Looks great, thanks for the year fix!! Sorry I missed that case

src/modules/form/components/viewable/DynamicViewField.tsx
gigxz: Whoops yes! thanks, forgot about that one

PR #998, opened 2024-12-16
Implement Table Actions pattern and default columns

martha:
## Description

I opened this PR against https://github.com/greenriver/hmis-frontend/pull/989, it should be retargeted to the base branch when that is merged. Both PRs should ideally go out in the same release.

Summary of changes:
- Update all in-scope tables (see [sheet](https://docs.google.com/spreadsheets/d/17ZpdB0fH6kUL14KDRWrlukwIH93BNx9grrNZApLpUwg/edit?gid=1987637083#gid=1987637083)) to use the new Table Actions pattern and to have the correct columns displayed. Remove any row-click behavior for these tables.
- (Attempt to) Standardize common columns and language
  - For example, always display an enrollment Entry Date column as an exact date with a relative date tooltip, and the reverse for Last Updated. This involved adding a new component `DateWithRelativeTooltip`, kind of like the inverse of `RelativeDateDisplay`.
  - Always use "View Client" language (as opposed to "Client Profile")
- Consolidate commonly used table row action definitions
- Make some changes to the `TableRowActions` api to enable the Bulk Services use case.
  - Enable the primary action to be a React Node
  - Add the `loading` prop passed from the generic table

Other todos:
- [ ] Co-locate tables that refer to the same model (as opposed to tables that appear on the same dashboard) - I'd like to propose doing this with https://github.com/open-path/Green-River/issues/6848, the diff here is already pretty large.
- [x] Run and update spec tests - done https://github.com/greenriver/hmis-warehouse/pull/5012
- [x] Expand storybook stories - done in this PR https://github.com/greenriver/hmis-frontend/pull/998/commits/7a4dd515938fb9ebf83f1868fbf66c8cc7df40fa where I added a generic story to demonstrate table row actions.

Depends on hmis-warehouse PR: (to do)

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/components/elements/table/GenericTable.tsx
```diff
@@ -44,16 +44,15 @@ import {
...
 
 export interface Props<T> {
   rows: T[];
   handleRowClick?: (row: T) => void;
   rowLinkTo?: (row: T) => To | null | undefined;
-  rowLinkState?: LocationState;
   columns?: ColumnDef<T>[];
   paginated?: boolean;
   loading?: boolean;
+  loadingRegardlessOfData?: boolean; // needed for correctly updating AssignBulkService button
```
martha: To discuss/pair. The `GenericTableWithData` always passes a `loading` prop of `false` to the `GenericTable`, but the `AssignBulkService` button uses a `useEffect` to listen on changes to the `loading` prop. I suspect there is a more graceful and React-y way to solve this problem.

src/components/elements/table/GenericTable.tsx
```diff
@@ -455,7 +451,7 @@ const GenericTable = <T extends { id: string }>({
                                   height: '100%',
                                   alignItems: 'center',
                                   px: 2,
-                                  py: condensed ? 1 : 2,
+                                  py: 2,
```
martha, line 446: I believe `condensed` is no longer used, so I've removed it. 
- `ClientProfileCardTextTable` still accepts a prop called `condensed` but determines its own sx based on that prop, and doesn't pass it along to to `GenericTable`. 
- `HouseholdMemberTable` also accepts a prop `condensed` but uses it to pass specific row sx to `GenericTable`, and doesn't pass the prop itself along.

src/modules/admin/components/services/ServiceTypeTable.tsx
```diff
@@ -11,11 +12,10 @@ import {
 } from '@/types/gqlTypes';
 import { generateSafePath } from '@/utils/pathEncoding';
 
-const columns: ColumnDef<ServiceTypeConfigFieldsFragment>[] = [
+const COLUMNS: ColumnDef<ServiceTypeConfigFieldsFragment>[] = [
```
martha, line 17: We had briefly discussed storing all column definitions centrally, in a code-version of the beautiful column spreadsheet. In the end I didn't decide to do that in this PR, because I was worried it would be too much code churn for now. But I'm still somewhat interested in the idea -- you can see in this PR that we have a mix of approaches, some columns defined inline, some defined and exported and used by other components, and some defined in a util file.

I did try to standardize around only using `useMemo/useCallback` for columns and table actions when they have dependencies in the component props/state, and otherwise defining them in the same file but outside the component, like here. Let me know if you think this is a good approach.

src/modules/hmis/components/RelativeDateDisplay.tsx
```diff
@@ -7,15 +7,29 @@ import {
...
+    preciseTime ? formatDateTimeForDisplay(date) : formatDateForDisplay(date),
+    formatRelativeDateTime(date),
+  ];
+};
+
 export interface RelativeDateDisplayProps {
   dateString: string;
   prefixVerb?: string;
   suffixText?: string;
+  tooltipSuffixText?: string;
```
martha: I added this `tooltipSuffixText` in order to reuse this component for the Last Updated and Uploaded time columns. Tooltip suffix is used for info about user who uploaded/updated.

src/modules/projects/components/ProjectAssessments.tsx
```diff
@@ -23,40 +27,38 @@ export type ProjectAssessmentType = NonNullable<
...
+  const getTableRowActions = useCallback(
+    (record: ProjectAssessmentType) => {
+      return {
+        primaryAction: {
+          ...getViewAssessmentAction(
+            record,
+            record.enrollment.client.id,
+            record.enrollment.id
+          ),
+          state: { backToLabel: project.projectName },
```
martha: Moving this into the table action enables removal of the `rowLinkState` prop on `GenericTable`

src/modules/projects/components/tables/ProjectClientEnrollmentsTable.tsx
```diff
@@ -65,20 +44,68 @@ const isHouseholdWithStaff = (
...
+        );
+    },
+  },
+  enrollmentStatus: {
+    header: 'Status',
+    render: (e) => <EnrollmentStatus enrollment={e} />,
+  },
+};
+
+export const WITH_ENROLLMENT_COLUMNS: {
```
martha, line 115: I'm curious for feedback on this pattern. I was aiming for something a bit equivalent to a `HasEnrollments` concern, in ruby. In this case the `WITH_ENROLLMENTS_COLUMN` should be applicable to any model that has an enrollment. I don't love using `ColumnDef<any>`, and think there might be a better way around it -- so maybe we could look into it together.

I guess we could consider this an incomplete preview of a more centralized table column organization, the idea I mentioned in a comment above.

src/modules/projects/components/tables/ProjectExternalSubmissionsTable.tsx
```diff
@@ -101,20 +101,22 @@ const ProjectExternalSubmissionsTable = ({
...
-              View
-            </Button>
-          ),
-        },
       ];
     },
+    []
+  );
+
+  const getTableRowActions = useCallback(
```
martha: Changing this table is out of scope of the ticket (not in the spreadsheet), but it was an especially easy example to update to the new pattern since it already used an Action Button column, so I pulled it in.

src/modules/search/components/ClientSearch.tsx
```diff
@@ -91,11 +75,6 @@ export const CLIENT_COLUMNS: {
     key: 'age',
     render: (client) => asClient(client).age,
   },
-  linkedName: {
-    header: 'Name',
-    key: 'name',
-    render: (client) => <ClientName client={asClient(client)} linkToProfile />,
-  },
   linkedNameNewTab: {
```
martha, line 85: I removed those that weren't used anymore, but there are still a few here (like `linkedNameNewTab`) that are used in out-of-scope table that I didn't update.

src/modules/services/serviceColumns.tsx
```diff
@@ -6,24 +6,31 @@ import { parseAndFormatDate, serviceDetails } from '@/modules/hmis/hmisUtil';
 import {
```
martha, line 6: Maybe this file should be renamed to `util.ts`?

src/modules/assessments/util.tsx
```diff
@@ -65,18 +65,6 @@ export const generateAssessmentPath = (
   });
 };
 
-export const assessmentRowLinkTo = (
```
martha, line 68: I removed this in favor of keeping the assessment link consistent. I think this means that now, any table column that links to an assessment will open it in a household context if applicable. I'm not 100% confident in this change, maybe there was a good reason for the Client assessments page to link to the individual assessment view? I'm torn about whether it seems better to be consistent or maintain the previous behavior, let me know what you think.

src/components/elements/table/GenericTable.tsx
gigxz: I'm not following why the loading prop can't be passed to the Assign Service Button in the same way that it was before, let's look at this one together

src/components/elements/table/TableRowActions.tsx
```diff
@@ -1,51 +1,88 @@
...
+      ? { primaryActionNode: primaryAction }
+      : { primaryActionTypesafe: primaryAction as CommonMenuItem };
+  }, [primaryAction]);
+
+  const secondariesWithAria = useMemo(
+    () =>
+      secondaryActions?.map((s) => {
+        return {
+          ...s,
+          ariaLabel: s.ariaLabel || `${s.title}, ${recordName}`,
```
gigxz: Do we need to support multiple ways of specifying the aria label? I wonder if we should just make it required on the primary/secondary action definitions, to simplify the logic.  (Instead of inferring them based on `recordName` and `title`)

src/modules/assessments/util.tsx
gigxz: This was a design decision so lets review with design before reverting it

gigxz:
holding off on additional review pending comments on https://github.com/greenriver/hmis-frontend/pull/989

src/modules/hmis/components/RelativeDateDisplay.tsx
martha: Moved to previous PR

src/modules/projects/components/tables/ProjectClientEnrollmentsTable.tsx
martha: Moved into previous PR

src/components/elements/table/TableRowActions.tsx
martha: Hm yeah I agree that it's not worth the extra code complexity. But I also don't necessarily think we should make `ariaLabel` required on `CommonMenuItem`. In general, adding additional aria labels that express something beyond what the button text itself says should be optional.

I'm proposing [this middle ground](https://github.com/greenriver/hmis-frontend/pull/998/files#diff-d00933719217aff188bdc40b93dafcc091dda9af8e9200dd79500c3ff752855fR33-R36) where the primary action's default aria label is set using `title` and `recordName`, if not provided in props, while secondary actions can "fend for themselves" in terms of aria labels. I think it's an ok balance. Let me know what you think, though, I'm happy to reconfigure it.

src/components/elements/table/GenericTable.tsx
martha: Removed this :)

src/components/elements/table/GenericTable.tsx
```diff
@@ -247,7 +243,12 @@ const GenericTable = <T extends { id: string }>({
                 width: def.width,
               }}
             >
-              <strong>{def.header}</strong>
+              {def.header ? (
```
martha, line 246: Continuing the thread from [here](https://github.com/greenriver/hmis-frontend/pull/989#discussion_r1900164832):

> {def.key || def.someOtherFieldThatsRequiredIfHeaderIsMissing}

If you agree, I was thinking that based on the comment [here](https://github.com/greenriver/hmis-frontend/pull/998/files#diff-e192fff2380d8eba4e69c51f1fcb73a656815e79a902758b40ce626c9c304da5L12),
> unique key for element. if not provided, header is used

We probably don't really need to add an extra `someOtherFieldThatsRequiredIfHeaderIsMissing`, because `key` should serve exactly that purpose.

src/modules/assessments/util.tsx
martha: I see, thanks. Let's keep the behavior. I think we can still simplify the code by using `generateAssessmentPath`'s `individualViewOnly`. Existing behavior that this PR replicates per https://github.com/greenriver/hmis-frontend/pull/998/commits/004ea9945bd88bd3adaf1a26ba74d0084ed1a5b4:
- [x] Household Assessments ([qa example](https://qa-hmis.openpath.host/client/==alplMEc4Y3Y5aW5mRVg1Z0VpclEyZz09/enrollments/==bHZoUm4xYlUyQWdPWnFEdERpN1E1Zz09/assessments) -> Household toggle) opens the assessment in household context if applicable
- [x] Enrollment Assessments ([qa example](https://qa-hmis.openpath.host/client/==alplMEc4Y3Y5aW5mRVg1Z0VpclEyZz09/enrollments/==bHZoUm4xYlUyQWdPWnFEdERpN1E1Zz09/assessments)) opens the assessment in household context if applicable
- [x] Client Assessments ([qa example](https://qa-hmis.openpath.host/client/==alplMEc4Y3Y5aW5mRVg1Z0VpclEyZz09/assessments)) always opens the assessment in individual context
- [x] Project Assessments ([qa example](https://qa-hmis.openpath.host/projects/==M3FhbTB3ZWZRMStKQjNrZE1Odm82UT09/assessments)) always opens the assessment in individual context

src/components/elements/RelativeDateDisplay.tsx
```diff
@@ -54,7 +56,7 @@ const RelativeDateDisplay = ({
     <Tooltip
       title={
         <Typography component='span' variant='inherit'>
-          {formattedDate}
+          {formattedDate} {tooltipSuffixText}
```
martha, line 61: Moved over from other PR https://github.com/greenriver/hmis-frontend/pull/989#discussion_r1901304288

src/modules/services/components/ClientServicesPage.tsx
```diff
@@ -36,41 +40,41 @@ const ClientServicesPage: React.FC<{
...
+                recordName={`${getServiceTypeForDisplay(row.serviceType)} on ${parseAndFormatDate(row.dateProvided)}`}
+                primaryActionConfig={getViewServiceMenuItem(
+                  row,
+                  row.enrollment.id,
+                  clientId
+                )}
+                secondaryActionConfigs={[
+                  {
+                    ...getViewEnrollmentMenuItem(row.enrollment, {
+                      id: clientId,
```
martha: This is the use case for https://github.com/greenriver/hmis-frontend/pull/989#discussion_r1901300319

src/components/elements/table/GenericTable.tsx
gigxz: Nice -- sounds good, I see that with the union types you achieved that required-ness of `header` or `key`. Looks good!

src/components/elements/table/tableRowActionUtil.tsx
```diff
@@ -30,15 +38,50 @@ export const getViewClientMenuItem = (client: ClientNameFragment) => {
...
 export const getViewEnrollmentMenuItem = (
   enrollment: Pick<EnrollmentFieldsFragment, 'id' | 'entryDate' | 'exitDate'>,
-  client: ClientNameFragment
+  client: Pick<ClientNameFragment, 'id'> | ClientNameFragment
 ) => {
   return {
     title: 'View Enrollment',
     key: 'enrollment',
-    ariaLabel: `View Enrollment, ${clientBriefName(client)} ${entryExitRange(enrollment)}`,
+    ariaLabel: `View Enrollment, ${client.hasOwnProperty('firstName') ? clientBriefName(client as ClientNameFragment) : ''} ${entryExitRange(enrollment)}`,
```
gigxz: Type cast shouldn't be needed  / I don't think `Pick<ClientNameFragment, 'id'> ` should be permitted -- I think this is same as other comment on the household PR?

martha: Hm, yes it's the same comment; I moved this change over from the other PR, where it wasn't relevant. But I agree with you that it's a bit messy and could be improved. It fixes the typescript issue arising in `ClientServicesPage`, where we don't have a client fragment to pass, only client ID - see https://github.com/greenriver/hmis-frontend/pull/998#discussion_r1901921862. The Client Services Page overrides the aria label that calls for the client ID anyhow, so it's purely a typescript issue.

A couple alternative ideas:
- Just avoid `getViewEnrollmentMenuItem` on `ClientServicesPage` - ok, but leads to some duplication
- Use `client` from dashboard context - seems like a bit of a hassle to keep Typescript happy, but maybe ok?
- edit: or the `getViewEnrollmentMenuItem` function could accept both `clientId` (required, for url) and `client` (optional; if provided, then used in aria label)? Or it could require one or the other?

What do you prefer?

gigxz: Ah I see, thanks. I didn't realize there really was a called that only passed `id`. I think the approach you added with https://github.com/greenriver/hmis-frontend/pull/998/commits/eee782f11fd4276c955c381161b50d7340c10f3d is fine but we should not need the cast `as ClientNameFragment`

gigxz (approved):

martha:
Thanks @gigxz :)

I think this is ready -- what's our typical process for deploy branches? Shall I merge this PR into https://github.com/greenriver/hmis-frontend/pull/989 and then treat that branch as the deploy branch? Should I rename (or merge that branch into a new branch with a clearer name) like `deploy/6761-table-actions` or similar?

gigxz:
@martha either way is fine (using existing branch name or new branch name)! Lets just make sure that we update the backend test branch https://github.com/greenriver/hmis-warehouse/pull/5012 to have the same branch name, and that shouldn't get merged until the frontend branch is merged.

PR #1001, opened 2024-12-17
Fix minor markup accessibility issues

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/7003
Summary of changes:
- Add an aria-label to the `DeleteMutationButton` when it is in icon-only mode. 
  - This was motivated by the Client Alert delete button; the other icon-only usages of this component weren't flagged in the audit because of being part of admin/config screens. 
  - By default the label is just "Delete {recordName}". It can be overridden as part of `ButtonProps` if need be to get more specific, but for AA wcag compliance, which requires that button text make sense in context, I think this is sufficient.
  - I also tidied up usages of this component to consistently end in `/>` if they don't have children
- ~Remove inner `span` from `ButtonTooltipContainer`. Tooltip applies its accessible text to its direct child, so when that's a span, as Bobby noted [here](https://github.com/open-path/hmis-accessibility/issues/42), it won't be correctly parsed by all screen readers.~ See inline comment. This change was reverted
  - Let's discuss, do you remember the reason for this `span` and/or do you see any risk to removing it? I checked the usages in code and confirmed that they are all 1-child only, so I think it should be ok, but I haven't checked every usage in-app to confirm no style regressions. I think it does make a significantly better screen reader experience though (no more ‘group’ to step into - see the Intake/Exit icons on Enrollment Dashboard and compare with QA).
- Remove `describeChild` from the `AssessmentDateWithStatusIndicator` to enable the tooltip to apply its accessible text
  - I found it completely unintuitive that _removing_ `describeChild` causes the accessibility behavior to improve, based on the confusing documentation here https://mui.com/material-ui/react-tooltip/#accessibility. I guess the key is “Note that you shouldn't use `describeChild` if the tooltip provides the only visual label. Otherwise, the child would have no accessible name”

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/components/elements/ButtonTooltipContainer.tsx
+++ b/src/components/elements/ButtonTooltipContainer.tsx
@@ -8,7 +8,16 @@ const ButtonTooltipContainer = ({
   if (!title) return children;
 
   return (
-    <Tooltip title={title} placement='top' arrow {...props}>
+    <Tooltip
+      title={title}
+      placement='top'
+      // describeChild makes the tooltip text become the `title` of the direct child,
+      // which is a span in this case. Otherwise it would become the `aria-label` of
+      // the direct child, which is not valid for spans.
+      describeChild
+      arrow
+      {...props}
+    >
       <span>{children}</span>
     </Tooltip>
   );
--- a/src/components/elements/ButtonTooltipContainer.tsx
+++ b/src/components/elements/ButtonTooltipContainer.tsx
@@ -9,7 +9,7 @@ const ButtonTooltipContainer = ({
 
   return (
     <Tooltip title={title} placement='top' arrow {...props}>
-      <span>{children}</span>
+      {children}
     </Tooltip>
   );
 };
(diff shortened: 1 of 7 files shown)


src/components/elements/ButtonTooltipContainer.tsx
```diff
@@ -9,7 +9,7 @@ const ButtonTooltipContainer = ({
 
   return (
     <Tooltip title={title} placement='top' arrow {...props}>
-      <span>{children}</span>
```
gigxz, line 12: Yes @martha the reason for adding this was because Tooltips won't appear on `disabled` Button elements without it. There are some places where we use a tooltip to explain the reason why a button is disabled (I think on household composition mgmt page, but maybe others).

martha: I see, thanks. I'll revert the change and add accessible text directly to the icon buttons. I will also add a note to https://github.com/open-path/Green-River/issues/7089 if we want to revisit this

gigxz (approved):

martha:
@gigxz do you have a preference btwn merging this into 146 or retargeting to 147?

gigxz:
@martha lets do 147 since 146 has already been deployed. thanks!

PR #1006, opened 2024-12-24
Remove invalid features from Feature-Policy

gigxz:
## Description

There are a few features listed in the Feature-Policy that are not valid features, which causes warnings in production. (See [MDN reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/Permissions_Policy)).

I only looked into the warnings recently when I was enabling geolocation in the `Permissions-Policy` and `Feature-Policy`. (https://github.com/greenriver/hmis-frontend/pull/992) 

Screenshot 2024-12-24 at 9 44 11 AM

## Type of change

### DIFF
--- a/docker/templates/non_proxy_headers.conf.template
+++ b/docker/templates/non_proxy_headers.conf.template
@@ -7,5 +7,5 @@ add_header Strict-Transport-Security "max-age=63072000; includeSubdomains;" alwa
 add_header X-Frame-Options "SAMEORIGIN" always;
 add_header X-XSS-Protection "1; mode=block" always;
 add_header X-Content-Type-Options "nosniff" always;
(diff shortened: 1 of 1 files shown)


ttoomey (commented):
Looks good, thanks!

It sounds like the "Feature-Policy" is obviated by the "Permissions-Policy" header. It may be that we don't need both(?) although I guess older browsers might not be there yet

[Reference](https://developer.chrome.com/docs/privacy-security/permissions-policy#migration_from_feature-policy)

ttoomey (approved):

gigxz:
@ttoomey yeah I figured for old browsers it wouldn't hurt to have both. thanks for review!

PR #1007, opened 2024-12-24
Update HMIS color palette and Button styles, remove secondary color usage

gigxz:
## Description

Issue: https://github.com/open-path/Green-River/issues/7108

- Update color palette values
- Update button styles
- Remove usages of `secondary` buttons

Targeted against 152 since its on QA, but will retarget to 153 and mark as ready-for-review on Thurs when 153 goes to qa
## Type of change
style


gigxz:
waiting to unmark as draft until retargeted against 153

ttoomey (approved):
Looks good, seems like a solid improvement!

Is it intentional that links are still `#1976D2` instead of `#5661A5`?

src/components/elements/table/GenericTable.tsx
```diff
@@ -87,7 +87,10 @@ export interface Props<T> {
 
 export const clickableRowStyles = {
   backgroundColor: 'background.paper',
-  '&:hover': { backgroundColor: 'grayscale.tint' },
+  '&:hover': {
+    // FIXME(#5746) should be primary[100]
```
martha: Maybe this comment should tell us to make it primary[200] to match the hover state in ToolbarMenu? Unless these were intentionally different

src/config/theme.ts
```diff
@@ -106,16 +133,42 @@ export const baseThemeDef: ThemeOptions = {
   },
   palette: {
     primary: {
-      main: '#1976D2',
+      surface: '#F8F9FB',
```
martha, line 136: Related to Theron's comment, it looks like Overline text is another place where we are still using the MUI primary color blue

martha (approved):
Great, thank you Gig!

src/modules/hmis/components/AssessmentDateWithStatusIndicator.tsx
```diff
@@ -15,7 +15,7 @@ const AssessmentDateWithStatusIndicator = ({
 
   return (
     <Stack direction='row' alignItems='center' gap={0.8}>
-      <Typography variant='inherit' color='error'>
+      <Typography variant='inherit' color='error.dark'>
```
martha, line 18: Looks like RequiredLabel is another place we should update to `error.dark` instead of just `error`, to avoid contrast issues

src/config/theme.ts
gigxz: Updated to stop using `links` color for overline text in several places, thanks!

src/components/elements/table/GenericTable.tsx
gigxz: open Q in figma. I updated the header to use primary.100 for selected state, which is its described purpose. 200 is for hover so I think you're right we should be using that for table, but its dark enough to wash out the borders which looks a little funny.

gigxz:
> Is it intentional that links are still #1976D2 instead of #5661A5?


That's an open question to design, may get updated during QA. Thanks!!

src/components/elements/table/GenericTable.tsx
```diff
@@ -85,9 +85,21 @@ export interface Props<T> {
   belowRowsContent?: ReactNode; // component to insert below all rendered rows, above footer
 }
 
+export const stickyCellClassName = 'StickyTableCell';
+
+// primary.100 on top of paper bg, so we don't layer semi-transparent primary.100 twice
+const nonTransparentPrimary100bg = (theme: Theme) =>
+  `linear-gradient(${theme.palette.primary[100]}, ${theme.palette.primary[100]}),${theme.palette.background.paper};`;
```
gigxz, line 92: this is 😭  but it was the only fix I could find, primary 100 is semi-transparent so we couldn't use the previous approach of just setting the background color to make sticky cells non-transparent. Instead here I use linear-gradient to layer the primary.100 over the background paper. @martha

gigxz:
Merging as is, still expecting some adjustments during qa from design

src/components/elements/table/GenericTable.tsx
martha: seems fine to me! Thanks :)

PR #1009, opened 2024-12-24
Check editorUserIds when rendering occurrence point form

gigxz:
## Description

Issue: https://github.com/open-path/Green-River/issues/7093

This PR makes it so that the newly added `editor_user_ids` field on Form Item is respected when rendering "occurrence point" details on the Enrollment Details card.

### How to Test
1. Create and Occurrence Point form with the following content:
    ```
    {
      "item": [
        {
          "text": "Favorite Color",
          "type": "STRING",
          "link_id": "name",
          "editor_user_ids": ["1"] #<=== put your application user id here. can only be found on the backend
        }
      ]
    }
    ```
2. Publish the form
3. Add a rule to enable the form for a project
4. Go to an enrollment at the project
5. "favorite color" should appear in the enrollment details card, and you can click the edit icon and edit the value.  
    Screenshot 2024-12-29 at 11 58 25 AM
6. impersonate a different user who can also access this enrollment
7. that user should be able to view the favorite color, but not edit it. There should be no edit icon.
    Screenshot 2024-12-29 at 11 59 00 AM

Note: if some `item`s are allowed to be edited, but others are not, then the form should still be editable. Screenshot shows an example where a second item is added to the form that doesn't have `editor_user_ids`:
Screenshot 2024-12-29 at 11 54 29 AM

## Type of change

### DIFF
--- a/src/modules/form/util/formUtil.ts
+++ b/src/modules/form/util/formUtil.ts
@@ -1497,6 +1497,8 @@ export const parseOccurrencePointFormDefinition = (
   user?: HmisUser
 ) => {
   let displayTitle = definition.title;
+
+  // Whether this form as any fields that are editable to the current user
   let isEditable = false;
 
   function matchesTitle(item: FormItem, title: string) {
@@ -1505,7 +1507,7 @@ export const parseOccurrencePointFormDefinition = (
     );
   }
 
-  // Modify form definition into a "read only definition" by removing
+  // Created modified "definition for display" that will be used to render values in a DynamicView inside the Enrollment/Client details card
   const definitionForDisplay = modifyFormDefinition(
     definition.definition,
     (item) => {
--- a/src/modules/form/util/formUtil.ts
+++ b/src/modules/form/util/formUtil.ts
@@ -42,6 +42,7 @@ import {
   TypedObject,
 } from '../types';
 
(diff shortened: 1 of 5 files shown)


src/modules/form/util/formUtil.ts
```diff
@@ -1492,9 +1493,12 @@ export const getFieldOnAssessment = (
 };
 
 export const parseOccurrencePointFormDefinition = (
-  definition: FormDefinitionFieldsFragment
+  definition: FormDefinitionFieldsFragment,
+  user?: HmisUser
```
ttoomey: User can't really be undefined here right? Maybe use non-null assertion from the caller?
`parseOccurrencePointFormDefinition(form.definition, user!)` or perhaps it would be helpful to have a variation of `useAuth()` that always returns `Required<HmisAuthState>`

ttoomey (approved):
looks good!

src/modules/form/util/formUtil.ts
gigxz: Thanks good point, added non-null assertion. You're right it should always be present. Previously it was empty in storybook, but I added the decorator so it will always be present there too.

martha (commented):
Thanks, looks great!

PR #1012, opened 2024-12-31
Augment accessible text for MUI Autocomplete

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/7004#issuecomment-2557230787
Summary of changes:
- In our `GenericSelect` wrapper around MUI's `Autocomplete`, add a custom `renderTags`. This allows us to inject additional accessible text into each chip on multiselect dropdowns. This is a partial/temp solution for https://github.com/open-path/hmis-accessibility/issues/37.
  - To discuss: Should we open an issue in MUI's repo?
- In our `FormSelect`, add a visually hidden element explaining the option's group label if it exists. This is a partial/temp solution for https://github.com/open-path/hmis-accessibility/issues/57.
  - This corresponds to an existing issue in the MUI repo, https://github.com/mui/material-ui/issues/42782, which has recent activity, so I think we can hope for a fix soon
  - The temp fix I've made here only helps `FormSelect`, not all places where we use `GenericSelect`.
- Remove `disablePortal` which was recently introduced, per thread here https://github.com/open-path/Green-River/issues/7004#issuecomment-2566690805

## Type of change
- [x] Bug fix - a11y temporary solutions
- [ ] Dependency update

### DIFF
--- a/src/modules/form/components/FormSelect.tsx
+++ b/src/modules/form/components/FormSelect.tsx
@@ -1,5 +1,6 @@
 import { Box, Typography, createFilterOptions } from '@mui/material';
 
+import { visuallyHidden } from '@mui/utils';
 import { useCallback } from 'react';
 import { DynamicInputCommonProps } from '../types';
 import GenericSelect, {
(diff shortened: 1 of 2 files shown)


src/modules/form/components/FormSelect.tsx
```diff
@@ -28,6 +29,11 @@ const renderOption = (props: object, option: Option) => (
     data-testid={`option-${optionId(option)}`}
   >
     <Box sx={{ display: 'flex', justifyContent: 'space-between', width: 1 }}>
+      {/* Insert the group label or group code as a visually hidden element.
+          Temp fix for https://github.com/mui/material-ui/issues/42782 */}
+      {(option.groupLabel || option.groupCode) && (
```
gigxz, line 34: This seems fine to me - can you expand on why it is just a temporary fix?

gigxz (approved):

src/modules/form/components/FormSelect.tsx
martha: I think of it as temporary because it's a fix in our app for an underlying issue that lives in MUI, so we can remove it once the MUI maintainers fix the underlying issue, I hope.

PR #1013, opened 2024-12-31
Fix bugs with Uploader found in QA

martha:
## Description
GH issue: https://github.com/open-path/Green-River/issues/6208
Depends on hmis-warehouse PR: n/a

Summary of changes:
- Fixes the issue where uploaded files aren't downloadable.
  - The fix is to open the file url in a new tab.
  - I wasn't able to repro this problem locally, but I'm almost confident in the fix, because in QA, if you try using command-click the "Download" menu item to open in a new tab, it works correctly, rather than leading to "Page not found."
  - This also adds a new-tab icon to the Download menu item, which I think is acceptable.
  - Unrelated to the fix, but while I was in here, I added a `<li>` enclosure to the MenuItem in order to correct the html, as I found recommended here: https://github.com/mui/material-ui/issues/204#issuecomment-545260260 
- Fixes the issue where previously uploaded images aren't shown when you reopen an assessment for editing.
  - The fix is to include `Image` types along with `File`s in the exception in `gqlValueToFormValue` that keeps the metadata attached to the type.
- Adds the `image` tag to the client profile image uploader?
  - This is not related to a QA bug, I just noticed it while I was working on the fixes above. I figured the client image uploader should only accept images. Currently in QA you are allowed to upload a PDF: https://qa-hmis.openpath.host/client/==TjEzWVNXbndEOFdhcm16aGxuaTNwZz09/profile

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/components/elements/CommonMenuButton.tsx
+++ b/src/components/elements/CommonMenuButton.tsx
@@ -22,6 +22,7 @@ export type CommonMenuItem = {
   divider?: boolean;
   disabled?: boolean;
   ariaLabel?: string;
+  openInNew?: boolean;
 };
 
 interface Props {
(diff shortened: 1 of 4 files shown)


gigxz (approved):
thanks!

PR #1014, opened 2024-12-31
Remove 99 from Relationship to HoH picklist

gigxz:
## Description

Remove `Data Not Collected` as an option when selecting the Relationship To HoH for a new or existing Enrollment. 99 is not a valid selection per the HUD HMIS data dictionary. However it is valid per the CSV Export, so we retain it as a valid option for **display** only.

Testing instructions are on the issue https://github.com/open-path/Green-River/issues/7127

Related Warehouse PR (Does not depend on): https://github.com/greenriver/hmis-warehouse/pull/5028

## Type of change
- [x] Bug fix / HUD Compliance fix


src/modules/household/components/EditHouseholdMemberTable.tsx
```diff
@@ -109,21 +106,6 @@ const EditHouseholdMemberTable = ({
     });
   }, [currentMembers]);
 
-  // If new members have beed added highlight their relationship field if it's DNC
-  useEffect(() => {
```
gigxz, line 113: @gigxz  - test this again did we need this?

src/modules/household/components/EditHouseholdMemberTable.tsx
```diff
@@ -109,21 +106,6 @@ const EditHouseholdMemberTable = ({
     });
   }, [currentMembers]);
 
-  // If new members have beed added highlight their relationship field if it's DNC
```
gigxz, line 112: This is no longer needed, because it's not permitted to add a household member without specifying a valid RelationshipToHoH.

The existing "highlight" behavior is retained whereby when you **change the HoH**, the RelationshipToHoH field for other members is highlighted, to indicate that you should review those and update them if they are no longer accurate.

Screenshot 2025-01-10 at 9 57 09 AM

src/modules/household/components/elements/RelationshipToHohSelect.tsx
```diff
@@ -52,15 +39,12 @@ const RelationshipToHohSelect = ({
...
   return (
     <GenericSelect<Option, false, false>
       options={relationshipToHohOptions}
       disabled={isDisabled}
-      textInputProps={{
-        placeholder: isHoH
-          ? HmisEnums.RelationshipToHoH[value]
-          : 'Select relationship',
-      }}
+      textInputProps={{ placeholder: 'Select Relationship', ...textInputProps }}
```
gigxz, line 47: If an existing Enrollment has `null` or `99` for Relationship To HoH, it appears like this on the Enrollment dashboard Household tab:

Screenshot 2025-01-10 at 9 59 43 AM


And Edit Household:


Screenshot 2025-01-10 at 9 59 08 AM

src/modules/form/util/formUtil.ts
```diff
@@ -115,13 +115,31 @@ export const hasMeaningfulValue = (value: any): boolean => {
...
+ */
 export const localResolvePickList = (
   pickListReference: string,
   excludeDataNotCollected = false
 ): PickListOption[] | null => {
   if (isHmisEnum(pickListReference)) {
     let values = Object.entries(HmisEnums[pickListReference]);
-    if (excludeDataNotCollected) {
+
+    if (excludeDataNotCollected || pickListReference === 'RelationshipToHoH') {
```
ttoomey, line 142: Nice docs! 
Rather than hard-coding this in the utility fn, would it be much more trouble to make this the caller's responsibility?
```js
localResolvePickList(pickListReference, pickListReference === 'RelationshipToHoH')
```

src/modules/household/components/elements/RelationshipToHohSelect.tsx
```diff
@@ -1,46 +1,33 @@
...
 export type Option = { value: RelationshipToHoH; label: string };
 
 export interface Props
   extends Omit<GenericSelectProps<Option, false, false>, 'options' | 'value'> {
   value: RelationshipToHoH | null;
-  showDataNotCollected?: boolean;
 }
+const relationshipToHohOptions: Option[] =
+  // safe to cast because we know that RelationshipToHoH is resolvable
+  (localResolvePickList('RelationshipToHoH') as PickListOption[])
```
ttoomey: do we need the cast?
```suggestion
  (localResolvePickList('RelationshipToHoH') || [])
```

src/modules/projects/components/tables/ProjectHouseholdsTable.tsx
```diff
@@ -106,14 +106,20 @@ export const HOUSEHOLD_COLUMNS: {
...
       <TableCellContainer>
-        {hh.householdClients.map((c) => (
-          <HmisEnum
-            key={c.id}
-            value={c.relationshipToHoH}
-            enumMap={HmisEnums.RelationshipToHoH}
-            whiteSpace='nowrap'
-          />
-        ))}
+        {hh.householdClients.map((c) =>
```
martha: @gigxz could you give more info about why you reverted this? I'm seeing missing "Data Not Collected" info on the table:
Screenshot 2025-01-10 at 2 31 45 PM

Edit: It looks like this would be resolved in the Tables deploy branch, anyway: https://github.com/greenriver/hmis-frontend/pull/1020/files#diff-819ec0ace67cab35cca989ef6e3f3d5a6187dec26e6dcbcb1d4090614788a6c5R43-R53

src/modules/household/components/EditHouseholdMemberTable.tsx
martha: Cool, I like the new inference behavior. There are still some edge cases with inference that I guess we might refine later. For example, in a household like this:
- A, HoH
- B, Spouse
- C, Other Relative

When changing HoH from A to C, B stays marked as "Spouse" but should probably be inferred to now be "Other Relative." I think this is not in scope for this PR though - looks like that would be a backend change.

martha (approved):

ttoomey (approved):
Looks good!

src/modules/projects/components/tables/ProjectHouseholdsTable.tsx
gigxz: I didn't change the current behavior here, enrollments with null/99 relationship to hoh still appear like they do in your screenshot (confirmed in staging environments which appear the same). On a previous version I had removed DataNotCollected from the Enum entirely, which is why you see that commit history, but I reverted that change.

src/modules/form/util/formUtil.ts
gigxz: Yeah, the issue is that when a DynamicForm renders an item with picklist reference `RelationshipToHoH`, we also want to exclude 99 from that list. `DynamicField` uses the `usePickList` hook to resolve pick lists, which in turn uses `localResolvePickList`. So I figured it makes more sense to just hard-code it into the local resolution logic, rather than passing it through from multiple callers.

gigxz: In practice, its' the picklist that appears when you are enrolling a new client: https://github.com/greenriver/hmis-warehouse/blob/24060527d5f707d337e36894f213b181df7dcfc6/drivers/hmis/lib/form_data/default/fragments/new_enrollment_details.json#L53

src/modules/household/components/EditHouseholdMemberTable.tsx
gigxz: @martha Good point, it would be possible to make some additional inferences! Currently the highlight is supposed to nudge the user to review those, and I think it's OK to stick to that existing behavior for now.

PR #1018, opened 2025-01-03
Implement Skip to Content button

martha:
## Description

GH issue: https://github.com/open-path/hmis-accessibility/issues/48

Summary of changes:
- Add a Skip to Content button component
- Insert the button in 4 places:
  - In the top app bar, above the HMIS site title (first content on the site). MainLayout.tsx
  - At the top of the Dashboard left-nav (for Projects, Enrollments, Clients). DashboardContentNav.tsx
  - In the Assessment left-nav sidebar. AssessmentFormSideBar.tsx
  - In the Form Navigation left-nav component that's used when previewing forms to publish, as well as when editing records. FormNavigation.tsx 
- **Refactor forms with navigation** so they all use the same styling/component. This is decidedly scope creep; we discussed on a call and decided to move forward, but I'd be happy to move those changes to a back-burner PR if we start feeling like they're leading to too much churn and iteration this time around.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/components/elements/SkipToContentButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+        padding: 0,
+        margin: 0,
+        border: 'none',
+        position: 'absolute', // Ensure it doesn't take up any space
+
+        '&:focus': {
+          // On focus, restore clip and other styles so the button is visible and styled normally
+          clip: 'auto',
+          clipPath: 'none',
+          position: 'relative',
```
martha, line 44: To discuss: Would it be better to give the button absolute positioning? Maybe optionally, as a prop, in some cases, such as in the app bar? (See e.g. the MUI docs Skip to Content button.)

src/modules/assessments/components/AssessmentFormSideBar.tsx
```diff
@@ -92,9 +93,12 @@ const AssessmentFormSideBar: React.FC<Props> = ({
       }}
     >
       <Box>
-        <Typography variant='body2' component='div'>
-          Assessment Sections
-        </Typography>
+        <Stack gap={2}>
+          <Typography variant='body2' component='div'>
```
martha: There are style inconsistencies (body2 vs h6) between the 2 different form nav components that I haven't addressed here, but would be happy to while we're in here. Maybe:
- consolidate around variant='h6' but component='div'
- consolidate around keeping the divider between the title and items (currently only present on AssessmentFormSideBar, not FormNavigation)?

src/components/elements/SkipToContentButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+  focusTargetId: string;
+  children?: ReactNode;
+};
+
+const SkipToContent: React.FC<SkipToContentProps> = ({
+  focusTargetId,
+  children,
+}) => {
+  const handleClick = () => {
+    const container = document.getElementById(focusTargetId);
```
martha, line 14: There's a tradeoff here with putting focus on the container itself, rather than trying to find the first focusable element in the container. The latter strategy gives a better UX for keyboard navigation, but for voiceover, it skips content that should be included.

The keyboard navigation UX problem is that setting focus to the container sometimes leads to focused content getting scrolled out of view, but I _think_ that it's still worth the tradeoff for better VO experience. Let me know what you think.

To reproduce: Open a long form, keyboard nav to the "skip to content button", and enter. The form gets focus. Continuing to hit tab navigates the form correctly, but the elements you're navigating are scrolled out of view. (Observed on Chrome)

src/modules/assessments/components/AssessmentForm.tsx
```diff
@@ -338,7 +338,12 @@ const AssessmentForm: React.FC<Props> = ({
   return (
     <Grid container spacing={2} sx={{ pb: 20, mt: 0 }}>
       {showNavigation && navigation}
-      <Grid item xs={showNavigation ? 9.5 : 12} sx={{ pt: '0 !important' }}>
+      <Grid
+        id='focusable-form'
```
gigxz: maybe we can pass this as a prop instead so its not a hard-coded string? it's a little confusing to search the codebase and find four usages of `focusable-form`, but I _think_ they are actually 2 separate pairs that have nothing to do with each other.

could this be implemented as:

```
const formContainerId = 'assessment-form-container'

..
<AssessmentFormSideBar formFocusTargetId={formContainerId} />
...
<Grid id={formContainerId}>

File: src/modules/form/components/FormNavigation.tsx [hunk 5]
```diff
@@ -28,17 +29,16 @@ const FormNavigation = ({
...
+            {title && <Typography variant='h6'>{title}</Typography>}
+            <SkipToContentButton focusTargetId='focusable-form' />
+            <FormStepper items={items} />
+          </Stack>
         </Paper>
         {contentsBelowNavigation}
       </Box>
     </Grid>
-    <Grid item xs={9} sx={{ pt: '0 !important' }}>
+    <Grid id='focusable-form' item xs={9} sx={{ pt: '0 !important' }}>
```
gigxz: Can we pull this string out into a variable inside this file or component?

src/modules/assessments/components/AssessmentFormSideBar.tsx
gigxz: Good call-out, lets standardize to what the Assessment nav currently looks like, since that is more widely used and the design was more recently updated. I am sure we just forgot to update the `FormNavigation` (used for EditRecord, maybe only impacts the Edit Project form) when we made that change.

src/components/elements/SkipToContentButton.tsx
gigxz: I'm not seeing a skip to content button appear on the MUI docs (eg https://mui.com/material-ui/getting-started/) when I tab the page. Or are you referencing a specific piece of documentation about how to implement Skip to Content?

src/components/elements/SkipToContentButton.tsx
gigxz: It looks like using `container.focus({ preventScroll: true });` can prevent this issue. Not sure if that's the desirable behavior in all cases though - it looks fine on assessments in chrome.

Another option is to scroll, but the top offset will depend on where we are. For assessments, this looks ok:
```
      container.focus();
      scrollToElement(container, STICKY_BAR_HEIGHT + CONTEXT_HEADER_HEIGHT);
```

src/components/elements/SkipToContentButton.tsx
martha: This is what I see when I open https://mui.com/material-ui/getting-started/ and immediately hit "tab":
Screenshot 2025-01-06 at 4 58 44 PM
It has absolute positioning at the top left.

src/modules/assessments/components/AssessmentForm.tsx
```diff
@@ -422,26 +436,24 @@ const AssessmentForm: React.FC<Props> = ({
...
-          description={
-            <Typography variant='body2' sx={{ my: 2 }}>
-              Select a previous assessment to populate the current assessment.
-              Any changes you have made will be overwritten.
-            </Typography>
-          }
-        />
-      )}
-    </Grid>
+      <RecordPickerDialog
```
martha, line 442: This is unchanged, I just removed `definition &&` since `definition` is always present -- it's either `viewingDefinition` or `editingDefinition`, both of which are required props.

src/modules/assessments/components/AssessmentFormActions.tsx
```diff
@@ -0,0 +1,103 @@
...
+interface Props {
+  enrollment: EnrollmentFieldsFragment;
+  assessment?: FullAssessmentFragment;
+  onAutofill: VoidFunction;
+  isPrintView: boolean;
+  locked: boolean;
+  showAutofill?: boolean;
+}
+
+const AssessmentFormActions: React.FC<Props> = ({
```
martha, line 26: This code isn't new, it is what became of `AssessmentFormSidebar` after I factored the `FormStepper` out into `FormNavigationContainer`.

src/modules/client/components/pages/CreateClientPage.tsx
```diff
@@ -39,7 +38,6 @@ const CreateClientPage: React.FC = () => {
         formRole={RecordFormRole.Client}
         onCompleted={onCompleted}
         FormActionProps={{ submitButtonText: 'Create Client' }}
```
martha, line 40: Removed to fix (I think) a visual bug with scrolling the form stepper.

Before: (https://qa-hmis.openpath.host/client/new)
Screenshot 2025-01-08 at 11 52 41 AM

After:
Screenshot 2025-01-08 at 1 38 00 PM

src/modules/client/components/pages/EditClientPage.tsx
```diff
@@ -42,7 +37,6 @@ const EditClientPage = () => {
       record={client}
       localConstants={localConstants}
       onCompleted={onCompleted}
-      top={STICKY_BAR_HEIGHT + CONTEXT_HEADER_HEIGHT}
```
martha, line 45: Removed since this is the default, no need to pass it in

src/modules/form/components/EditRecord.tsx
```diff
@@ -143,15 +140,13 @@ const EditRecord = <RecordType extends SubmitFormAllowedTypes>({
...
-            top={top}
-            {...FormNavigationProps}
-          >
-            {form}
-          </FormNavigation>
-        </Grid>
+        <FormNavigationContainer
+          navItems={leftNavItems}
+          navScrollOffset={top}
+          {...FormNavigationProps}
```
martha, line 146: One visual result of this refactor is moving the `EditRecord` after-nav action buttons into the FormStepper's Paper, like this:

Before:
Screenshot 2025-01-08 at 11 38 16 AM

After:
Screenshot 2025-01-08 at 11 38 22 AM

I think it's an OK change, even good since it's now consistent with Assessments. But let me know if you would prefer to keep separate or loop in design!

src/modules/form/components/FormNavigationContainer.tsx
```diff
@@ -0,0 +1,90 @@
...
+}
+const FormNavigationContainer = ({
+  navItems,
+  navTitle = 'Form Navigation',
+  navScrollOffset = STICKY_BAR_HEIGHT + CONTEXT_HEADER_HEIGHT,
+  contentsBelowNavigation,
+  useUrlHash = true,
+  children,
+}: FormNavigationProps) => {
+  const isTiny = useIsMobile('sm');
```
martha, line 28: I had some hopes of making this depend on `useDashboardContext`'s `desktopNavIsOpen` for improved mobile-friendliness, but I didn't get that working in the timebox I gave myself for this refactor. The form stepper is moderately more mobile-friendly now though.

src/modules/form/util/formUtil.ts
```diff
@@ -1546,7 +1546,7 @@ export const getFormStepperItems = (
   localConstants: LocalConstants,
   minGroupsToDisplay: number = 3
 ) => {
-  if (!formDefinition || !itemMap) return false;
+  if (!formDefinition || !itemMap) return undefined;
```
martha, line 1549: Changed for typescript friendliness

src/modules/assessments/components/AssessmentFormSideBar.tsx
martha: Thanks! These are the places I've been checking in dev as I'm working on the changes. Let me know if you think of anything else. I'll incorporate them in the QA instructions checklist.
- Form Preview in form builder (qa link https://hmis.dev.test:5173/admin/forms/base-intake/20/preview)
- Individual assessment (qa link https://qa-hmis.openpath.host/client/==RXhXcTVoWEcvLzJ2WnU0UVUrWWllUT09/enrollments/==NVI5aTFrb1c5Q1pIUnFhZVc0TCs0dz09/assessments/new/==UmRqVEVoNVV6VWw3bWh1R1BLYTJ5UT09)
- Household assessment (qa link https://qa-hmis.openpath.host/client/==UGUxMVI0NXJzbmhIUnFhZVc0TCs0dz09/enrollments/==RHhmUEtuWjhCMnhIUnFhZVc0TCs0dz09/intake)
- Edit Project (qa link https://qa-hmis.openpath.host/projects/==ZU84bkZVaUpLb2krK0wweTVteW9MZz09/overview/edit)
- Create Client (qa link https://qa-hmis.openpath.host/client/new)
- Edit Client (qa link https://qa-hmis.openpath.host/client/==emtRY3YzeEt0R1ZIUnFhZVc0TCs0dz09/profile/edit)

src/modules/assessments/components/AssessmentForm.tsx
martha: Yes agreed, thanks! In https://github.com/greenriver/hmis-frontend/pull/1018/commits/d127879d3dda7a86ce59a03eb3ec6d5fcf7d4374 I factored this into a shared variable to avoid repeating the hardcoded string. Otherwise I think your comment is partly addressed with the refactor, since there are no longer 2 separate pairs

martha: This same feedback applies to `focusable-main`, which has 2 instantiations, 1 in dashboard context and 1 in non-dashboard context (such as all clients). I've made a maybe-satisfactory attempt here https://github.com/greenriver/hmis-frontend/pull/1018/commits/c39393ad6ac69c17cd8e89da27cc7ade761f8b4e, let me know what you think

Edit: Make that 3 instantiations: https://github.com/greenriver/hmis-frontend/pull/1018/commits/9bce44cc7c076c53a568b40ce418ba909604cb90

src/components/elements/SkipToContentButton.tsx
martha: Thanks, cool! I think I prefer `preventScroll: true`. It seems simpler to reason about: In skip-to-content navigation situations, we always want to keep scroll at the top of the page. (Rather than reasoning through the different scroll offsets in different contexts within the app.)

src/modules/form/components/EditRecord.tsx
gigxz: looks fine to me, since its consistent with assessments I think it makes sense

src/components/elements/SkipToContentButton.tsx
gigxz: I see. I have no preference here I think they way you implemented it works

gigxz (approved):
looks great!

PR #1020, opened 2025-01-06
Update tables with new Action Column pattern (Deploy Branch) (#989)

martha:
## Description

This is a deploy branch for https://github.com/open-path/Green-River/issues/6825 and https://github.com/open-path/Green-River/issues/6761, so these features can get plenty of review on QA with design (without being tied to regular releases).

[//]: # 'remove if not applicable'
Should merge with warehouse PR to update tests: https://github.com/greenriver/hmis-warehouse/pull/5039

## Type of change
[//]: # 'remove options that are not relevant'
- [ ] Bug fix
- [x] New feature (adds functionality)
- [ ] Code clean-up / housekeeping
- [ ] Dependency update

## Checklist before requesting review
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [x] My code includes comments and/or descriptive variable names to help other engineers understand the intent (or not applicable)
- [x] My code follows the style guidelines of this project (eslint)
- [x] I have updated the documentation (or not applicable)
- [x] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

PR #1021, opened 2025-01-13
Follow up on QA items for Table Actions deploy branch

martha:
## Description

GH issues: https://github.com/open-path/Green-River/issues/6761 and https://github.com/open-path/Green-River/issues/6825

Since the changes on the deploy branch (https://github.com/greenriver/hmis-frontend/pull/1020) have already been reviewed, I'll plan to open PRs against that branch for further changes.

Summary of changes:
- ~Restore blank cells for `DataNotCollected` relationship in the Project Households table. This is the current prod behavior, which my previous PR changed unintentionally. https://github.com/greenriver/hmis-frontend/pull/1014#discussion_r1911307027~ Edit: reverted this reversion, per inline comment thread
- Memoize `AssignServiceButton`. I was seeing "High memory usage" warnings from my tab locally when checking and unchecking the "select all" button on the Bulk Services page multiple times in a row. This seems to have improved it somewhat.
- Restore styles on the Assign Service action tab. The '180px' was removed unintentionally. The <Box> wrapper fixes a problem introduced by the TableRowActions's internal `<Stack>`. You can see the problem on QA if you go to Bulk Services and select/deselect; the button width jumps around.
- Rename Name column to Client Name -- I either missed this or it changed, but per our spreadsheet and mocks it should read "Client Name" always

## Type of change
QA followups
Bug fixes

### DIFF
--- a/src/modules/bulkServices/components/AssignServiceButton.tsx
+++ b/src/modules/bulkServices/components/AssignServiceButton.tsx
@@ -3,7 +3,7 @@ import WarningAmberRoundedIcon from '@mui/icons-material/WarningAmberRounded';
 import { LoadingButton } from '@mui/lab';
 import { ButtonProps, DialogActions, DialogContent } from '@mui/material';
 import DialogTitle from '@mui/material/DialogTitle';
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import { memo, useCallback, useEffect, useMemo, useState } from 'react';
 
 import * as React from 'react';
 import { useBulkAssignMutations } from '../hooks/useBulkAssignMutations';
@@ -161,4 +161,4 @@ const AssignServiceButton: React.FC<Props> = ({
   );
 };
 
-export default AssignServiceButton;
+export default memo(AssignServiceButton);
--- a/src/modules/projects/components/tables/ProjectHouseholdsTable.tsx
+++ b/src/modules/projects/components/tables/ProjectHouseholdsTable.tsx
@@ -1,5 +1,4 @@
-import { Box, TableBody, TableCell, TableRow } from '@mui/material';
-import { visuallyHidden } from '@mui/utils';
+import { TableBody, TableCell, TableRow } from '@mui/material';
 import React, { useMemo } from 'react';
 import { renderCellContents } from '@/components/elements/table/GenericTable';
 import TableRowActions from '@/components/elements/table/TableRowActions';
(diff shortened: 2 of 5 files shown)


src/modules/bulkServices/components/AssignServiceButton.tsx
```diff
@@ -161,4 +161,4 @@ const AssignServiceButton: React.FC<Props> = ({
   );
 };
 
-export default AssignServiceButton;
+export default memo(AssignServiceButton);
```
gigxz, line 164: > I was seeing "High memory usage" warnings from my tab locally when checking and unchecking the "select all" button on the Bulk Services page multiple times in a row. This seems to have improved it somewhat.

Hmm, memory usage warnings are concerning. Did you have react dev tools running at the time?

I'm looked a bit and I didn't see a difference in rerenders between using `memo` vs not. The buttons all _do_ re-render a lot, since the loading and disable states change as bed nights are assigned, but `memo` shouldn't affect that since those are rerenders due to prop changes. I'm sure there is something I am missing though, I didn't dig too far. 
So this seems like a fine change to me, bed night assignment still appears to work as expected!

src/modules/projects/components/tables/ProjectHouseholdsTable.tsx
```diff
@@ -43,14 +45,24 @@ const BASE_COLUMNS: ColumnDef<OneHouseholdClient>[] = [
...
   {
     header: 'Relationship',
-    render: (householdClient) => (
-      <HmisEnum
-        key={householdClient.id}
-        value={householdClient.relationshipToHoH}
-        enumMap={HmisEnums.RelationshipToHoH}
-        whiteSpace='nowrap'
-      />
-    ),
```
gigxz: I think I prefer keeping DNC visible, in this new table design.  I would stay stick with what you had! I was just avoiding making design changes with https://github.com/greenriver/hmis-frontend/pull/1014 since that wasn't the focus.

Screenshot 2025-01-15 at 12 36 45 PM

gigxz (approved):

src/modules/bulkServices/components/AssignServiceButton.tsx
martha: Yes, I did have react devtools running, maybe that was why.

src/modules/projects/components/tables/ProjectHouseholdsTable.tsx
martha: Gotcha, thank you! I reverted this.

PR #1022, opened 2025-01-13
Enable user to Join Households

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/5752

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5047

Summary of changes:
- Implement frontend workflow for joining household ([Figma](https://www.figma.com/design/kqZiUPjXvqSvJVslJCnlTB/OP-HMIS-%2F-Tickets-%2F-Figma?node-id=8258-2&p=f&t=2ziXXNWOnwbyJIvv-0))
- Add StepDialog component. For the time being, this just uses MUI's out-of-the-box tabber, not the newly designed one.
- Add ability to use GenericTable's selection status as a controlled prop.
- Add option to use `useClientAlerts` hook with an existing household from memory.
- Add ability for the DynamicForm caller to both
  - filter out certain errors, in order to takeover rendering separately outside of the ErrorAlert
  - and listen on error change, in order to respond appropriately
- Add a "disabled reason" on the FormDialogActionContent component, which is displayed as a tooltip when the button is disabled

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/household/components/elements/AddToHouseholdButton.tsx
```diff
@@ -56,6 +66,11 @@ const AddToHouseholdButton = ({
       inputVariables: { projectId, clientId },
       pickListArgs: { projectId, householdId },
       localConstants: { householdId, projectCocCount: cocCount },
+      onFieldChange: (linkId?: string, value?: any) => {
```
martha: @gigxz as you can see, this is still a draft with many todos, but this is the key part I'd like early feedback about!

gigxz: Thanks @martha I will take a closer look. I agree with you that RHF may make this a bit different, for example maybe we could hook into the RHF `watch` functionality.



I'll explain the backend alternative I was thinking of, too, just to give some more detail. This is an approach where the backend would be solely responsible for determining whether there are conflicts:
* Similarly to how the [Delete Client](https://github.com/greenriver/hmis-warehouse/blob/16c79eace4f4f7c44749a57ab49a71013ae15228/drivers/hmis/app/graphql/mutations/delete_client.rb#L64-L73) returns some structured JSON data related to a warning, the SubmitForm mutation could return structured data attached to the validation error `Client has another enrollment in this project that conflicts with this entry date.` that it already returns. That could be implemented [here](https://github.com/greenriver/hmis-warehouse/blob/16c79eace4f4f7c44749a57ab49a71013ae15228/drivers/hmis/app/models/hmis/hud/validators/enrollment_validator.rb#L77-L78) by adding something like `data : { conflicting_enrollment_id: 5 }` to the validation error.
* If -- like with DeleteClient -- we included ALL the relevant information about conflicting enrollment(s), the validation error itself would have sufficient data on it for us to show the `Conflicting Enrollment` alert.
* With this approach,
  * There wouldn't be a need to do additional background queries based on the currently selected Entry Date
  * The user wouldn't be made aware of the conflict (and the option to "join") until they have already tried to submit the form
  * The logic for whether enrollments conflict or not would remain in one place (the Enrollment Validator [here](https://github.com/greenriver/hmis-warehouse/blob/16c79eace4f4f7c44749a57ab49a71013ae15228/drivers/hmis/app/models/hmis/hud/validators/enrollment_validator.rb#L38-L40), which relies on the somewhat complicated [`with_conflicting_dates`](https://github.com/greenriver/hmis-warehouse/blob/16c79eace4f4f7c44749a57ab49a71013ae15228/drivers/hmis/app/models/hmis/hud/enrollment.rb#L214-L239) scope.) Since the logic is only in one place, we don't have to worry about making sure that the Backend and Frontend equivalents of "with conflicting dates" have perfectly matched logic. (Mis-matched logic would result in the backend error being surfaced but the frontend not showing the workflow option, or vice versa).
  * Once the `Conflicting Enrollment` alert is shown, the user would have the option to initiate the join, or change the Entry Date to something valid and re-submit.

src/api/operations/client.queries.graphql
```diff
@@ -74,6 +74,31 @@ query GetClientEnrollments(
   }
 }
 
+query GetClientEnrollmentWithHousehold(
+  $id: ID!
+  $limit: Int = 1
+  $offset: Int = 0
+  $filters: EnrollmentsForClientFilterOptions
```
gigxz: (If we end up keeping this approach) I'd suggest making `openOnDate` and `projectId` required variables, since that's the only intended use-case for this query. This resolves so much, it could get really expensive if it was resolved on a larger set of enrollments. I think we are _assuming_ that it shouldn't return a ton of enrollments, because a client shouldn't have multiple open enrollments at a given project on a given date-- but thats only true if those args are passed.  (Even though I see this defaults to `limit: 1`, it could still be used in other ways)

src/modules/household/components/elements/AddToHouseholdButton.tsx
martha: Thank you for the feedback @gigxz. I've made a start on your idea -- it's a bit rough, but I think something like [this](https://github.com/greenriver/hmis-frontend/pull/1022/files#diff-b8a7b13b6a60eed06b9710f2a9ee4e1cacfcef8aa160c980a83bceb8b8f142fbR70-R84) will be necessary to selectively override the FormDialog's default rendering of its error messages. 

I also have a question about this part:

> If -- like with DeleteClient -- we included ALL the relevant information about conflicting enrollment(s), the validation error itself would have sufficient data on it for us to show the Conflicting Enrollment alert.

Am I understanding the following correctly?
- the `data` attr of an error returned from the backend is just json, but can't be easily typed using `HmisSchema` objects
- you're suggesting the error's `data` attr should return everything needed to render the `ConflictingClientAlert`, but *not* return all the data needed for the whole `JoinHouseholds` workflow, right?

If I got that right, then, I think I'd propose removing (descoping) the sentence "There [is/are] also [n] other household member[s] associated: [Additional Clients]" from the mocks [here](https://www.figma.com/design/kqZiUPjXvqSvJVslJCnlTB/OP-HMIS-%2F-Tickets-%2F-Figma?node-id=9454-22285&t=KCcHKAvMGEnPJorc-0). It seems preferable to me to rely on our existing logic that lives in the schema objects for determining whether the current user has access to those clients, and/or access to view their names. The user will see those clients' names on the next step anyway, if they initiate the join workflow, so including them here seems more complicated than it's worth to me. (But maybe I am missing an easy way to do it.)

If I got that wrong, and you really did intend for the conflicting enrollment error message to include all the data necessary for the `JoinHouseholds` workflow -- including client alerts of the associated household members, nested data that is expected in a certain format/type in the common column definitions we are using, etc. -- then I'll have some followup questions about how the typing should work (hand-pick the data we need in the enrollment validator, then cast on the frontend to the relevant gql typescript types? is there a better way?)

Thanks for your help!

gigxz: @martha good question, I didn't mean typing the `data` field but rather just including enough information for the explanatory alert. If it makes sense to descope some content from the alert, I think that's OK. And it seems fine to me that the "join household workflow" itself would need to fetch additional data.

It _is_ a bit brittle to have this un-typed contract between the backend and the frontend regarding what the supplemental error data looks like. The frontend should be as gracious as possible of course, so if the `data` is missing or malformed it just shows the default message.

src/modules/household/components/elements/JoinHouseholdDialog.tsx
```diff
@@ -0,0 +1,264 @@
...
+  const onSubmit = useCallback(() => {
+    if (missingRelationshipsCount > 0) return; // This should never happen; the button will be disabled
+
+    joinHousehold({
+      variables: {
+        input: {
+          receivingHouseholdId: receivingHousehold.id,
+          joiningEnrollmentInputs: joiningClients.map((hc) => {
+            return {
+              // todo @martha - discuss, should we use enrollmentLockVersion?
```
martha: Let's discuss whether to use `enrollmentLockVersion` here. I think I would propose de-scoping it and maybe ticketing it for a future improvement.

src/modules/household/components/elements/JoinHouseholdAddRelationships.tsx
```diff
@@ -0,0 +1,126 @@
...
+  return (
+    <Stack gap={2}>
+      <Box>
+        <Typography variant='overline'>Step 2</Typography>
+        <Typography variant='h3'>Add Relationships</Typography>
+      </Box>
+      <Typography variant='body1'>
+        Update joining clients' relationships{' '}
+        {receivingHohName && <>to {receivingHohName}</>}
+        {/* todo @martha - add warning here about entry dates, pending conversation with design */}
```
martha: Pending https://www.figma.com/design/kqZiUPjXvqSvJVslJCnlTB?node-id=9475-76147#1096321325

src/components/elements/DateWithRelativeTooltip.tsx
```diff
@@ -54,7 +54,10 @@ const DateWithRelativeTooltip = ({
       >
         {formattedDate}
         {/* Include the tooltip text as visually hidden for accessibility */}
-        <Box sx={visuallyHidden}>, {formattedDateRelative}</Box>
+        {/* Add position: fixed to address visual bug when visuallyHidden is used inside dialog */}
```
martha, line 57: Seems like when `visuallyHidden` is used inside a dialog, it causes some surprising scroll behavior. I wasn't able to repro it on stackblitz using the latest mui version. I would propose cutting a separate ticket to track this down later (and see if it impacts anywhere else in our app, since we use `visuallyHidden` all over the place)

Edit: Created https://github.com/open-path/Green-River/issues/7233

src/modules/household/components/elements/AddToHouseholdButton.tsx
```diff
@@ -84,10 +140,38 @@ const AddToHouseholdButton = ({
       {renderFormDialog({
         title: <>Enroll {clientBriefName(client)}</>,
         submitButtonText: `Enroll`,
-        preFormComponent: clientAlerts.length > 0 && (
-          <ClientAlertStack clientAlerts={clientAlerts} />
-        ),
+        // This wrapper is necessary around the preFormComponent Stack,
```
martha: I can create a separate ticket to improve this, as well

Edit: created https://github.com/open-path/Green-River/issues/7234

src/modules/household/components/elements/AddToHouseholdButton.tsx
```diff
@@ -1,38 +1,46 @@
-import { Button } from '@mui/material';
```
martha, line 1: Question about our directory structure here -- what is `src/modules/household/components/elements/`? Should `components/elements` be flattened into `components`?

martha:
@gigxz ok! There is still some work to be done on the backend draft PR (including a system test), but I think this frontend work is now ready for your review. Thank you 🙏

martha:
@gigxz I'm converting this back to draft for now. As I've been working more on the Split workflow, I've been realizing how much code the 2 workflows share, and thinking I can do a better job of factoring this into reusable components to reduce code duplication. 🤔 I want to take a proper stab at that with a fresh brain Monday morning.

src/modules/form/components/FormDialogActionContent.tsx
```diff
@@ -1,6 +1,19 @@
...
+export interface FormDialogActionProps {
+  onDiscard: ButtonProps['onClick'];
+  onSubmit: ButtonProps['onClick'];
+  submitLoading?: boolean;
+  discardButtonText?: string;
+  submitButtonText?: string;
+  disabled?: boolean;
+  disabledReason?: string;
+  otherActions?: ReactNode;
+  PrimaryActionProps?: ButtonProps;
```
gigxz: Maybe `SubmitButtonProps`? Afaik we don't much refer to the submit action as "primary". I imagine it might be confusing with "primary" also being used in theming.

src/modules/hmis/hmisUtil.ts
```diff
@@ -657,3 +657,30 @@ export const raceEthnicityDisplayString = (race?: Race[]) => {
...
+};
+
+export const findHohOrRep = (
+  householdClients: HouseholdClientFieldsFragment[]
+) => {
+  // It's invalid, but possible in case of bad data, to have no HoH. If so (rare), return the first household member
+  return (
+    householdClients.find(
+      (hc) => hc.relationshipToHoH === RelationshipToHoH.SelfHeadOfHousehold
+    ) || householdClients[0]
```
gigxz: suggest making this deterministic, we could use existing sort
```
  const sorted = sortHouseholdMembers(householdClients);
  return (
    sorted.find(
      ({ relationshipToHoH }) =>
        relationshipToHoH === RelationshipToHoH.SelfHeadOfHousehold
    ) || sorted[0]
  );
```

src/modules/household/components/householdActions/JoinHouseholdDialog.tsx
```diff
@@ -0,0 +1,277 @@
...
+  HouseholdFieldsFragment,
+  ProjectAllFieldsFragment,
+  RelationshipToHoH,
+  useGetEnrollmentWithHouseholdQuery,
+  useJoinHouseholdMutation,
+} from '@/types/gqlTypes';
+
+interface Props {
+  open: boolean;
+  conflictingEnrollmentId: string;
```
gigxz: For this component, is it relevant that this enrollment ID conflicted? I'm wondering if this prop can just be called `joiningEnrollmentId` since it represents the enrollment that the user is trying to join to the household.

gigxz: I guess relatedly-- `JoinHousehold` should probably enforce that the 2 households overlap, so that we make sure users can't perform joins of non-overlapping households.

src/modules/household/components/householdActions/JoinHouseholdDialog.tsx
```diff
@@ -0,0 +1,277 @@
...
+      id: conflictingEnrollmentId,
+    },
+    onCompleted: (data) => {
+      if (data?.enrollment?.household) {
+        const household = data?.enrollment?.household;
+        const initiator = household?.householdClients.find(
+          (hc) => hc.enrollment.id === conflictingEnrollmentId
+        );
+
+        if (initiator) {
```
gigxz: What happens if `initiator` is undefined? maybe we should be raising instead if that happens?

src/modules/household/components/householdActions/JoinHouseholdDialog.tsx
```diff
@@ -0,0 +1,277 @@
...
+  RelationshipToHoH,
+  useGetEnrollmentWithHouseholdQuery,
+  useJoinHouseholdMutation,
+} from '@/types/gqlTypes';
+
+interface Props {
+  open: boolean;
+  conflictingEnrollmentId: string;
+  onClose: VoidFunction;
+  onComplete?: (
```
gigxz: `onSuccess` would be more in line with naming conventions for props elsewhere in the application. This clarifies that the callback is only called if the mutation is successful. (Unlike apollo `onCompleted`)

src/modules/household/components/householdActions/JoinHouseholdDialog.tsx
```diff
@@ -0,0 +1,277 @@
...
+    joinedHousehold: HouseholdFieldsFragment,
+    remainingHousehold?: HouseholdFieldsFragment | null
```
gigxz: I commented on the backend PR as well, but can we drop this arg since it's not consumed by anything?

src/modules/household/components/householdActions/JoinHouseholdDialog.tsx
```diff
@@ -0,0 +1,277 @@
...
+        },
+      },
+    });
+  }, [
+    missingRelationshipsCount,
+    joinHousehold,
+    receivingHousehold.id,
+    joiningClients,
+    relationships,
+  ]);
```
gigxz: There is a lot of code here related to the mutation, it seems like a good candidate for pulling into a hook, so that the component logic is easier to follow. I _think_ we could pull the selected lines out so that JoinHouseholdDialog has:

```jsx
  const {
    performHouseholdJoin, 
    joinLoading,
    joinedHousehold, // or just a flag indicating it was successful, since that's all we use
    remainingHousehold, // or just: donorHouseholdHoh , since that's all we use
  } = usePerformHouseholdJoin();

  const onSubmit = () =>
    performJoinHousehold({ receivingHouseholdId, joiningClients });
  
```

src/modules/household/components/householdActions/JoinHouseholdDialog.tsx
```diff
@@ -0,0 +1,277 @@
...
+import SuccessWayfindingStep from '@/modules/household/components/householdActions/SuccessWayfindingStep';
+import {
+  HouseholdClientFieldsFragment,
+  HouseholdFieldsFragment,
+  ProjectAllFieldsFragment,
+  RelationshipToHoH,
+  useGetEnrollmentWithHouseholdQuery,
+  useJoinHouseholdMutation,
+} from '@/types/gqlTypes';
+
```
gigxz, line 24: I was able to submit a join mutation with no clients selected, which caused an error.
Repro:
* start join workflow
* unselect all clients
* click "review join" tab header
* click "join enrollments"

src/modules/household/components/householdActions/JoinHouseholdDialog.tsx
```diff
@@ -0,0 +1,277 @@
...
+          onComplete?.(
+            data.joinHousehold.receivingHousehold,
+            data.joinHousehold.donorHousehold
+          );
+        }
+      },
+    });
+
+  const onSubmit = useCallback(() => {
+    if (missingRelationshipsCount > 0) return; // This should never happen; the button will be disabled
```
gigxz: I'm still able to navigate using the tabs and hit this

src/components/elements/StepDialog.tsx
```diff
@@ -0,0 +1,136 @@
...
+} from 'react';
+import CommonDialog, {
+  CommonDialogProps,
+} from '@/components/elements/CommonDialog';
+import Loading from '@/components/elements/Loading';
+import FormDialogActionContent, {
+  FormDialogActionProps,
+} from '@/modules/form/components/FormDialogActionContent';
+
+export type TabDefinition = {
```
gigxz: Can this be called `StepDialogTabDefinition`?
We already have a TabDefinition type, which should've probably been named AssessmentTabDefinition. https://github.com/greenriver/hmis-frontend/blob/f40f1effb933c3ad7b5861f097a0e6a1dcc0ca18/src/modules/assessments/components/household/util.ts#L36 

~~It would be nice if the tab elements within the StepDialog could be pulled out to a `CommonTabs` component that has a `CommonTabDefinition` type or similar, if we anticipate using tabs outside of a dialog context. It looks like that would require a fair amount of refactoring of this component, though.~~

Actually, it looks like the designs call for a _horizontal stepper_ component, not tabs. Let's discuss. It's a new component in the design library, so should be implemented based on that design reference as a reusable component. It was a nice-to-have for household join, though, so maybe we should defer it.

gigxz: https://www.figma.com/design/bsw8CX4UOq6SIdXYorZDWV/OP-HMIS-%2F-Library-%2F-Design?node-id=2578-665

src/modules/household/components/householdActions/ReviewHouseholdsStep.tsx
```diff
@@ -0,0 +1,45 @@
...
+  description: string;
+  members: HouseholdClientFieldsFragment[];
+};
+
+interface Props {
+  reviewableHouseholds: ReviewableHousehold[];
+  children?: ReactNode;
+}
+
+const ReviewHouseholdsStep = ({ reviewableHouseholds, children }: Props) => {
```
gigxz, line 18: I'm getting a unique key error from this component

src/modules/household/components/householdActions/JoinHouseholdDialog.tsx
```diff
@@ -0,0 +1,277 @@
...
+    });
+  }, [
+    missingRelationshipsCount,
+    joinHousehold,
+    receivingHousehold.id,
+    joiningClients,
+    relationships,
+  ]);
+
+  const tabDefinitions: TabDefinition[] = useMemo(
```
gigxz: Just to help me understand the general approach @martha - is this accurate?
- Each tab has some inputs that update the same pieces of state (`joiningClients` and `relationships`) which live in the `JoinHouseholdDialog` component.  That state is updated live as soon as a user makes a selection within the tab.
- Each tab may have buttons to cancel/go back /proceed to next step. Those buttons have `type='submit'` but they don't submit anything (Except for the last tab's submit button), they are navigational.
  - noting this isn't perfectly correct usage of `button type='submit'` https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#type could be fixed simply with adjusting existing: `!!nextTab ? { endIcon: <ArrowRightIcon />, type: 'button' } : SubmitButtonProps`


It would be great if you can add a header comment to the `StepDialog` explaining some of this flow.

src/components/elements/StepDialog.tsx
```diff
@@ -0,0 +1,136 @@
...
+                {tab.content}
+              </TabPanel>
+            ))}
+          </TabContext>
+        )}
+        {successContent && <Box mt={2}>{successContent}</Box>}
+      </DialogContent>
+      {!successContent && (
+        <DialogActions>
+          <FormDialogActionContent
```
gigxz: To discuss:
It doesn't seem like `FormDialogActionContent` is a great fit for this dialog. It has a pretty specific structure geared around dialogs that contain a single submittable form.

I wonder if it would make more sense to introduce a separate component for specifying DialogActions for a workflow stepper dialog like this one.

gigxz: For example: here we are overloading the `otherActions` prop to hold the "back" button in the workflow. For normal form dialogs, I believe we always use `otherActions` for rendering the delete button. If we want to fiddle with where the delete button lives for form dialogs, then we're also messing with where the back button is on workflow step dialogs, etc.

src/components/elements/table/GenericTable.tsx
```diff
@@ -212,6 +214,11 @@ const GenericTable = <T extends { id: string }>({
 
   // initially undefined so we can early return and avoid state flicker
   const [selected, setSelected] = useState<string[]>();
+  const isSelectControlled = selectedProp !== undefined;
+  const selectedValue = useMemo(
+    () => (isSelectControlled ? selectedProp : selected || []),
+    [isSelectControlled, selected, selectedProp]
+  );
```
gigxz: to discuss: refactor option to move selection logic into a hook? there's a lot going on. I do agree we need the controlled selection behavior you added.

src/components/elements/StepDialog.tsx
martha: Hrmm yes you are right this should be a horizontal stepper, not tabs!

What do you think about the idea of implementing it using MUI's out-of-the-box [horizontal stepper](https://mui.com/material-ui/react-stepper/?srsltid=AfmBOooBpiWLa50U8jEml_xpfy3D48MRRBKvkkjNDDsCJWP1u6UEAtyI#linear) for now, then deferring the new styles to later?

martha: Notes from 1:1: Spend a few minutes to see if it's easy to style the horizontal stepper; otherwise omit it

src/modules/household/components/householdActions/JoinHouseholdDialog.tsx
martha: Hm, I'm trying to think this through... Should we be validating that the _households_ overlap? If so, should we backend-validate that earlier, not just check in the frontend? 

In the backend conflicting enrollment logic, as implemented now, this is not checked. What's being checked is whether the initiator client already has an enrollment that overlaps with the _entry date the user specifies_ in the Enrollment creation dialog.

The current logic _does_ allow joining households that don't overlap, like this:
- Alice has an enrollment with entry date 2/1/25
- Bob has an enrollment with entry date 1/1/25 and exit date 1/30/25
- Go to Alice's Manage Household screen and try to enroll Bob with an entry date of 1/1/25
- This gets you to the Conflicting Enrollment workflow, even though Alice's and Bob's existing enrollments do not overlap.

Note that if Bob did not have a conflicting enrollment, you would be able to enroll. It's a warning, "Entry date is before the HoH's entry date," but not a blocking error.

src/components/elements/StepDialog.tsx
martha: https://www.figma.com/design/kqZiUPjXvqSvJVslJCnlTB?node-id=9471-58317#1119920310

graphql.schema.json
```diff
@@ -32151,6 +32364,12 @@
         "inputFields": null,
         "interfaces": null,
         "enumValues": [
+          {
```
martha: Some changes here are from the base branch

src/components/elements/StepDialog.tsx
```diff
@@ -0,0 +1,189 @@
...
+export type StepDefinition = {
+  title: string; // must be unique, since it is used as the step's key
+  content: ReactNode;
+
+  // if onSubmit is not provided, the default action is to just go to the next step
+  onSubmit?: () => Promise<any>; // after onSubmit promise resolves, go to the next step if there is one
+  submitButtonText?: string;
+  submitLoading?: boolean;
+  ButtonProps?: ButtonProps;
+  omitStepTitle?: boolean;
```
gigxz: For more flexibility, we could replace:
```ts
title: string; // must be unique, since it is used as the step's key
omitStepTitle?: boolean;
```
with
```ts
key: string // unique key for the step
title?: string // in the future we could expand this to ReactNode for more flexibility. it is a presentation prop
```

That would let us expand `title` to accept a ReactNode in the future if we need more flexibility in rendering. It keeps `title` presentation-focused while `key` is logical, and we non longer need a prop for hiding the title.

src/components/elements/StepDialog.tsx
```diff
@@ -0,0 +1,189 @@
...
+ * Note that this means a step with an `onSubmit` action does not necessarily need
+ * to be the last step in the workflow. The last step can, for example, be a
+ * successful response/wayfinding step.
+ *
+ * If no onSubmit is specified, the proceed button just renders the next step
+ * (it is purely navigational and does not submit anything).
+ *
+ * StepDialog only manages the state of which step is currently selected.
+ * If Steps have inputs that update some shared pieces of state, that state
+ * should live in StepDialog's parent (or wherever the StepDefinitions are defined).
```
gigxz, line 59: This is a nice improvement, thanks!

I wonder if we can merge the "submit" and "proceed" concepts into just "proceed", and let the caller specify whether any of the proceed actions constitute a submit. That _might_ also let us use this component in a way where intermediate steps have submit actions, in the future.


Untested, but here's a thought..

```diff
+++ b/src/components/elements/StepDialog.tsx
@@ -21,19 +21,20 @@ export type StepDefinition = {
   title: string; // must be unique, since it is used as the step's key
   content: ReactNode;

-  // if onSubmit is not provided, the default action is to just go to the next step
-  onSubmit?: () => Promise<any>; // after onSubmit promise resolves, go to the next step if there is one
-  submitButtonText?: string;
-  submitLoading?: boolean;
+  // if onProceed is not provided, the default action is to just go to the next step
+  onProceed?: () => Promise<any> | VoidFunction; // after onProceed promise resolves, go to the next step if there is one
+  proceedButtonText?: string;
+  proceedButtonType?: 'submit' | 'button';
+  proceedLoading?: boolean;
   ButtonProps?: ButtonProps;
   omitStepTitle?: boolean;

-  // `disableProceeding` can be used to disable either the onSubmit action, or the default 'next' action
-  disableProceeding?: boolean;
+  // `disabledProceed` can be used to disable the proceed action
+  disabledProceed?: boolean;
   disabledReason?: string;
 };

-interface Props extends Omit<CommonDialogProps, 'onSubmit' | 'onClose'> {
+interface Props extends Omit<CommonDialogProps, 'onProceed' | 'onClose'> {
   title: string;
   stepDefinitions: StepDefinition[];
   onClose: VoidFunction;
@@ -44,13 +45,13 @@ interface Props extends Omit<CommonDialogProps, 'onSubmit' | 'onClose'> {

@@ -84,44 +85,48 @@ const StepDialog = ({
   const {
     title: stepTitle,
     content,
-    onSubmit,
-    submitButtonText,
-    submitLoading,
-    disableProceeding,
+    onProceed,
+    proceedButtonText,
+    proceedButtonType = 'button',
+    proceedLoading,
+    disabledProceed,
     disabledReason,
     ButtonProps,
     omitStepTitle,
   } = thisStep;

   const nextButton = useMemo(() => {
-    if (!onSubmit && !nextStep) return undefined;
+    if (!onProceed && !nextStep) return undefined;

     const handleClick = async () => {
-      if (onSubmit) await onSubmit();
+      if (onProceed) await onProceed();
       if (nextStep) setCurrentStepKey(nextStep.title);
     };

+    const defaultButtonText = nextStep ? 'Next' : 'Finish';
+    const buttonText = proceedButtonText || defaultButtonText;
+
     return (
       <LoadingButton
         onClick={handleClick}
-        type={!!onSubmit ? 'submit' : 'button'}
-        loading={submitLoading}
+        type={proceedButtonType}
+        loading={proceedLoading}
         sx={{ minWidth: '120px' }}
-        disabled={disableProceeding}
+        disabled={disabledProceed}
         endIcon={nextStep ? <ArrowRightIcon /> : undefined}
         {...ButtonProps}
       >
-        {!!onSubmit && (submitButtonText || 'Submit')}
-        {!onSubmit && !!nextStep && (nextStep.title || 'Next')}
+        {buttonText}
       </LoadingButton>
     );
   }, [
```

src/components/elements/table/GenericTable.tsx
```diff
@@ -298,41 +293,14 @@ const GenericTable = <T extends { id: string }>({
...
-  useEffect(() => {
-    if (selected) onChangeSelectedRowIds?.(selected);
-  }, [selected, onChangeSelectedRowIds]);
+  const { selected, selectableRowIds, handleSelectAllClick, handleSelectRow } =
+    useTableSelection({
+      selectable: !!selectable,
+      isRowSelectable,
+      rows,
+      selectedCtrl: selectedProp,
+      onChangeSelectedCtrl: onChangeSelectedRowIds,
```
gigxz: nit: `Ctrl` suffix feels off here because this callback is also used for components that use the selection in an uncontrolled manner. In that case they rely on this onChange handler to know which rows are currently selected

src/components/elements/table/GenericTable.tsx
```diff
@@ -298,41 +293,14 @@ const GenericTable = <T extends { id: string }>({
...
-  );
-
-  // Clear selection when data changes
-  useEffect(() => setSelected([]), [rows]);
-
-  useEffect(() => {
-    if (selected) onChangeSelectedRowIds?.(selected);
-  }, [selected, onChangeSelectedRowIds]);
+  const { selected, selectableRowIds, handleSelectAllClick, handleSelectRow } =
+    useTableSelection({
```
gigxz, line 297: Thanks, nice to cut down on some lines in this render function!

src/components/elements/table/hooks/useTableSelection.tsx
```diff
@@ -0,0 +1,72 @@
+import { without } from 'lodash-es';
+import { useCallback, useEffect, useMemo, useState } from 'react';
+
+export function useTableSelection<T extends { id: string }>({
```
gigxz, line 4: It looks like something here broke the table selection functionality in the household submit summary page.
I'm no longer able to click the checkboxes to select the row:

Screenshot 2025-02-06 at 12 07 14 PM

gigxz: mmm nevermind, this issue is on QA too. Row 543 of GenericTable should change `selectable === 'checkbox' && isSelectable ? ` to just `isSelectable ?`, I believe.. would you mind dropping in that change? thank you!

gigxz: ok sorry amending again: it looks like this is a bug newly introduce on release-151, and not present on release-150. So it should be opened as its own fix against 151!

src/modules/household/components/ManageHousehold.tsx
```diff
@@ -32,21 +32,22 @@ import {
...
+  ProjectAllFieldsFragment,
   SearchClientsDocument,
   SearchClientsQuery,
   SearchClientsQueryVariables,
 } from '@/types/gqlTypes';
 
 interface Props {
   householdId?: string;
-  projectId: string;
+  project: Pick<ProjectAllFieldsFragment, 'id' | 'projectName' | 'access'>;
```
gigxz: Was this an intermediate change that didnt end up getting used?

src/modules/household/components/elements/AddToHouseholdButton.tsx
```diff
@@ -46,25 +54,73 @@ const AddToHouseholdButton = ({
...
-      pickListArgs: { projectId, householdId },
+      inputVariables: { projectId: project.id, clientId },
+      pickListArgs: { projectId: project.id, householdId },
       localConstants: { householdId, projectCocCount: cocCount },
+      errorFilter: (error: ValidationError) => {
+        // If there's an error about a conflicting enrollment, we will show the ConflictingEnrollmentAlert,
+        // so here, we filter that error out of the ErrorAlert displayed by the form dialog.
+        if (!householdId) return true; // Except if this is a new household. Then show the error, since we can't Join Households.
+        if (!error.data) return true;
+        return !error.data?.hasOwnProperty('conflictingEnrollmentId');
```
gigxz: maybe slightly easier to follow?
```suggestion
        // If there's an error about a conflicting enrollment, and we're adding to an existing household,
        // then we will show the ConflictingEnrollmentAlert (so filter it out from the ErrorAlert display)
        if (householdId && error.data?.hasOwnProperty('conflictingEnrollmentId')) {
          return false;
        }
        return true;
```

src/modules/household/components/elements/ConflictingEnrollmentAlert.tsx
```diff
@@ -0,0 +1,106 @@
+import {
```
gigxz: should this go inside the "householdActions" dir which contains the merge/split-related components?

src/modules/household/components/householdActions/JoinHouseholdDialog.tsx
```diff
@@ -0,0 +1,243 @@
...
+  if (fetchError) throw fetchError;
+  if (joinError) throw joinError;
+
+  return (
+    <StepDialog
+      title={'Join Enrollments'}
+      loading={fetchLoading}
+      open={open}
+      fullWidth
+      maxWidth='lg'
```
gigxz: nit: `md` might be sufficient? the wayfinding at the end feels particularly stretched.

src/modules/household/components/householdActions/SuccessWayfindingStep.tsx
```diff
@@ -0,0 +1,68 @@
...
+interface Props {
+  title: string;
+  description: string;
+  primaryClientName: string;
+  secondary?: HouseholdClientFieldsFragment;
+  project: Pick<ProjectAllFieldsFragment, 'id' | 'projectName'>;
+  onClose: VoidFunction;
+}
+
+const SuccessWayfindingStep = ({
```
gigxz, line 24: We may want to pull this out into a generic wayfinding component at some point, but we can wait until next time we add a wayfinding step.

src/modules/household/components/elements/AddToHouseholdButton.tsx
gigxz: Actually it looks like we only show the error if `household` is present,  so we should be checking that instead of `householdId`.
Can you clarify the need for having both props `householdId` and `household`? Are there time when this component receives only the ID but not the household record, and what are the implications?

src/modules/household/components/householdActions/AddRelationshipsStep.tsx
```diff
@@ -0,0 +1,126 @@
...
+                      value={relationships[hc.enrollment.id] || null}
+                      onChange={(_event, selected) => {
+                        updateRelationship(
+                          hc.enrollment.id,
+                          selected?.value || null
+                        );
+                      }}
+                      textInputProps={{
+                        warnIfEmptyTreatment: !relationships[hc.enrollment.id],
+                        placeholder: 'Select Relationship',
```
gigxz: this is not necessary because the component already sets the same placeholder

src/modules/household/components/householdActions/AddRelationshipsStep.tsx
```diff
@@ -0,0 +1,126 @@
...
+                <RequiredLabel
+                  text='Relationship'
+                  TypographyProps={{
+                    fontWeight: 'bold',
+                  }}
+                  required={true}
+                />
+              ),
+              headerCellProps: {
+                id: relationshipHeaderId,
```
gigxz, line 78: Nice. Is it necessary that this ID is on the `th` element itself? I think there is another place where I did something similar with labeling, but the id was inside the th content. The equivalent would be like:
```
              header: (
                <span id={relationshipHeaderId}>
                  <RequiredLabel
                    text='Relationship'
                    TypographyProps={{
                      fontWeight: 'bold',
                    }}
                    required={true}
                  />
                </span>
              ),
```

Would that work just as well for accessible labeling?

gigxz (commented):
Looks good!! And that is a lot of code! Comments left, overall functionality looks excellent !

src/components/elements/table/hooks/useTableSelection.tsx
martha: https://github.com/greenriver/hmis-frontend/pull/1047

src/modules/household/components/ManageHousehold.tsx
martha: It's used in the `useAddToHouseholdColumns` hook args on line 69. It allows the join workflow to not refetch the project, which it needs for name/id (to show the wayfinding button at the end) and access (to know whether the current user has canSplitHouseholds access on this project)

graphql.schema.json
gigxz: This should be updated in the base branch now

src/components/elements/table/hooks/useTableSelection.tsx
gigxz: thanks!

src/modules/household/components/ManageHousehold.tsx
gigxz: Ah I see, my bad

src/modules/household/components/householdActions/JoinHouseholdDialog.tsx
martha: Sure sounds good. I thought it looked a bit cramped at `md`, with placeholder text getting cut off when there is plenty of screen real estate, but agree that the wayfinding step looks much better. I don't necessarily think it is worth making the dialog change size depending on the step, but happy to implement that if you prefer.
Screenshot 2025-02-06 at 1 53 31 PM

src/modules/household/components/householdActions/AddRelationshipsStep.tsx
martha: Yes, I'm pretty sure that would work just as well. I was basing this on the Disability Table approach which I remember having Bobby's stamp of approval :)

I guess the `span` approach would simplify the `ColumnDef` api, but also I don't think the addition of `headerCellProps` is bad, so I'm inclined to leave it.

src/modules/household/components/householdActions/JoinHouseholdDialog.tsx
gigxz: Oh okay, right. I wasn't quite thinking through the complexity related to the Entry Date that the user enters initially. I think what we have is good, no need for changes.

gigxz: And thank you for the explanation!

src/modules/household/components/householdActions/AddRelationshipsStep.tsx
gigxz: Ok gotcha - sounds good! thanks!

src/modules/household/components/householdActions/JoinHouseholdDialog.tsx
gigxz: okay your call, I agree that doesn't seem worth it 👍

src/modules/household/components/elements/AddToHouseholdButton.tsx
martha: Hm, let me try to think it through. The existing code flow is something like this:
- `ManageHousehold` receives the `householdId` from the [`AllEnrollmentDetailsFragment` on the Enrollment dashboard](https://github.com/greenriver/hmis-frontend/blob/release-151/src/modules/household/components/EditHouseholdPage.tsx#L11), meaning that the _household_ itself hasn't been fetched and isn't in memory yet, only the ID.
- `ManageHousehold` then passes that "initial" `householdId` along to the [`useAddToHouseholdColumns` hook](https://github.com/greenriver/hmis-frontend/blob/release-151/src/modules/household/hooks/useAddToHouseholdColumns.tsx)
- `useAddToHouseholdColumns`
  - provides a function for fetching the household;
  - and separately, it provides a renderer for the [`AddToHouseholdButton`, passing it the `householdId`](https://github.com/greenriver/hmis-frontend/blob/release-151/src/modules/household/hooks/useAddToHouseholdColumns.tsx#L79) (not the whole `household`)

Changes introduced in this PR:
- `AddToHouseholdButton` now is the parent of the whole Join workflow, so it needs the whole `household`
- We already have this in `data.household` in `useAddToHouseholdColumns` so I added it as a new prop to `AddToHouseholdButton`.

So, the short answer to your original question,
> Can you clarify the need for having both props householdId and household?

is that I was trying to update the component to get the data about `household` it now needs, without messing with the existing behavior that depends on `householdId`, for fear of introducing new bugs.

However, [`ManageHousehold`](https://github.com/greenriver/hmis-frontend/blob/a028a949d9012265cd12393602bf0864d1b2466e/src/modules/household/components/ManageHousehold.tsx#L129) (which is ultimately the renderer of the `AddToHouseholdButton` returned by `useAddToHouseholdColumns` via `addToEnrollmentColumns` ... whew) returns `Loading` if `household` is undefined but an initial household ID was provided. So, _maybe_, we can rely on the idea that `householdId` and `household` will always match by the time `addToEnrollmentColumns` gets rendered? And thereby get rid of the `householdId` prop on the `AddToHouseholdButton`?

src/modules/household/components/elements/ConflictingEnrollmentAlert.tsx
martha: Ok, sounds good -- I also had a broader question about the directory structure here: https://github.com/greenriver/hmis-frontend/pull/1022#discussion_r1920527491

I remember at one point the floating idea that we want to keep module structure relatively flat, I think it was for the sake of Storybook?

src/modules/household/components/elements/AddToHouseholdButton.tsx
gigxz: Ok, sorry it's a bit of a tangle. I agree it would be preferable to remove the `householdId` prop and only use `household`. 

It think the only time this might cause a blip is in the following scenario:
* Project => Add Enrollment
* Search => "Enroll Client"
  * After the modal closes, the buttons automatically update to "Add to Household" text and have the `householdId` in state. Whereas if we're waiting for the household to fetch (which gets triggered in `useAddToHouseholdColumns` when the householdId changes), we need to wait another moment until those buttons receive the `household` state.
  * This could be addressed by passing a disabled state to `AddToHouseholdButton` where `disabled={loading && !!householdId && !data?.household}`, so those buttons are disabled in those seconds where the household is still loading. (Purpose being to prevent another "enroll client" mutation that spins up a _new_ household).
  * (Also keeping in mind the new enrollment designs have us unmounting the search altogether after an enrollment is added. Taking a peek at the code, that might not be quite as straightforward as it sounds -- otherwise I would suggest including it here.)

Let me know if you'd like to pair on this

martha: Thanks Gig, that's helpful! I was worried about introducing bugs but I played around with it in the browser with the debugger to slow things down, and I think I feel confident in the behavior now.

I've implemented it as you suggested, but I was also wondering, do you think there is any harm in just disabling the button when `loading` is true (regardless of other conditions)? Maybe that logic would be easier to reason thru when we revisit this code in the future. The downside is that the user can't add multiple enrollments to a household as quickly, since it forces you to wait until the household refetches before you can click the "Add to Household" button again.

gigxz: Ok, great! Thank you! I think fewer disabled states is probably more user-friendly because there is less visually changing in the UI for the user to track/interpret.

src/modules/household/components/elements/ConflictingEnrollmentAlert.tsx
gigxz: Yeah, we could probably flatten those out. It looks like several components in that elements directory are not used at all.

src/modules/household/components/elements/AddToHouseholdButton.tsx
```diff
@@ -84,7 +85,7 @@ const AddToHouseholdButton = ({
         }
       },
     }),
-    [project.id, clientId, householdId, cocCount, onSuccess, household]
+    [project.id, clientId, cocCount, onSuccess, household]
```
gigxz: This is probably ok, just noting that now it will rerun when `household` changes as opposed to only when the householdId changes. Not sure if there is any impact in practice, but there might be. Could just watch `household?.id` instead

src/api/operations/access.fragments.graphql
```diff
@@ -85,6 +85,7 @@ fragment ProjectAccessFields on ProjectAccess {
   canManageIncomingReferrals
   canManageOutgoingReferrals
   canManageExternalFormSubmissions
+  canSplitHouseholds
```
gigxz, line 88: Is this still needed here, since we resolve it on the Enrollment fragment?

gigxz (approved):
💯

src/api/operations/access.fragments.graphql
martha: `ManageHousehold` can receive the `project` prop either from the Enrollment fragment (when on the 'Manage Household' page for an existing enrollment) or from the Project dashboard (when on the 'Add Household' page). Scenario:
- go to Add Household page for the project, enroll someone new
- without leaving the page, try to add someone else who has a conflicting enrollment 
  - when `canSplitHouseholds` is resolved on `ProjectAccessFields`: you see the 'Conflicting Enrollment' alert and are able to join the enrollments
  - when `canSplitHouseholds` is not resolved on `ProjectAccessFields`: you see "You don't have permission to join enrollments" even if you actually do

gigxz: Ah gotcha, that makes sense. Thanks!

gigxz:
Can we merge this @martha?

PR #1023, opened 2025-01-13
Implement sticky columns in GenericTable

martha:
## Description

This PR is a follow-up from [this comment thread](https://greenriver.slack.com/archives/C033YNSN0Q1/p1736540407532099?thread_ts=1736448766.234739&cid=C033YNSN0Q1) in Slack. Summary of changes:
- Add ability in GenericTable to define columns as "sticky" on the left or right
- Implement "sticky" columns on all the in-scope tables for 6761
- Update to TableRowActions so that on mobile (smallest screen size), the primary action button appears as an icon. (It sounds like this may become true across the board, even on wider screens pending design decision)

I should note that, while I've once-overed every individual table that's using this new pattern, I haven't tested this PR extremely thoroughly, since the ticket is over estimate. For efficiency, I'm proposing to do more thorough QA in the QA environment once this is merged to the deploy branch.

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/components/elements/CommonMenuButton.tsx
+++ b/src/components/elements/CommonMenuButton.tsx
@@ -102,6 +102,9 @@ const CommonMenuButton = ({
           vertical: 'top',
           horizontal: 'right',
         }}
+        // Bug: Opening the CommonMenu applies padding to the body, which can look weird on mobile.
+        // It's sort of fixable with disableScrollLock, but that seems to introduce other scroll problems.
+        // disableScrollLock={true}
         {...MenuProps}
       >
         {items.map(
--- a/src/components/elements/table/GenericTable.tsx
+++ b/src/components/elements/table/GenericTable.tsx
@@ -81,13 +81,17 @@ const clickableRowStyles = {
   '&:focus': { backgroundColor: 'rgba(0, 0, 0, 0.04)' },
   cursor: 'pointer',
 };
-
-export const getStickyCellStyles = (
-  sticky: 'left' | 'right' | undefined,
-  stickyBorder: boolean,
-  leftOffset: string | number = 0,
-  rightOffset: string | number = 0
-): SystemStyleObject<Theme> => {
+export const getStickyCellStyles = ({
+  sticky,
+  stickyBorder = true,
+  leftOffset = 0,
+  rightOffset = 0,
+}: {
+  sticky?: 'left' | 'right';
+  stickyBorder?: boolean;
+  leftOffset?: string | number;
+  rightOffset?: string | number;
+}): SystemStyleObject<Theme> => {
   if (!sticky) return {};
 
   const base = {
@@ -279,7 +283,7 @@ const GenericTable = <T extends { id: string }>({
           {selectable && (
             <HeaderCell
               padding='checkbox'
-              sx={getStickyCellStyles('left', false)}
+              sx={getStickyCellStyles({ sticky: 'left', stickyBorder: false })}
             >
               <Checkbox
                 color='primary'
@@ -303,12 +307,10 @@ const GenericTable = <T extends { id: string }>({
               <HeaderCell
                 key={key(def) || i}
                 sx={{
-                  ...getStickyCellStyles(
-                    def.sticky,
-                    true,
-                    selectable ? '46px' : 0,
-                    0
-                  ),
+                  ...getStickyCellStyles({
+                    sticky: def.sticky,
+                    leftOffset: selectable ? '46px' : 0,
+                  }),
                   textAlign: def.textAlign,
                   width: def.width,
                   // headerStyles has SxProps type so we can't spread it directly. https://mui.com/system/getting-started/the-sx-prop/#passing-the-sx-prop
@@ -437,7 +439,10 @@ const GenericTable = <T extends { id: string }>({
                       <TableCell
                         padding='checkbox'
                         key='selection'
-                        sx={getStickyCellStyles('left', false)}
+                        sx={getStickyCellStyles({
+                          sticky: 'left',
+                          stickyBorder: false,
+                        })}
                       >
                         <Checkbox
                           color='primary'
@@ -487,12 +492,10 @@ const GenericTable = <T extends { id: string }>({
                           key={key(def) || index}
                           {...tableCellProps}
                           sx={{
-                            ...getStickyCellStyles(
-                              def.sticky,
-                              true,
-                              selectable ? '46px' : 0,
-                              0
-                            ),
+                            ...getStickyCellStyles({
+                              sticky: def.sticky,
+                              leftOffset: selectable ? '46px' : 0,
+                            }),
                             width,
                             minWidth,
                             ...(isLinked ? { p: 0 } : undefined),
--- a/src/components/elements/table/GenericTable.tsx
+++ b/src/components/elements/table/GenericTable.tsx
@@ -17,6 +17,7 @@ import {
   TableRow,
   Theme,
 } from '@mui/material';
+import { SystemStyleObject } from '@mui/system';
 import { visuallyHidden } from '@mui/utils';
 import { compact, get, includes, isNil, without } from 'lodash-es';
 import {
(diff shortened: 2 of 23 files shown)


src/components/elements/table/GenericTable.tsx
```diff
@@ -80,6 +80,47 @@ const clickableRowStyles = {
...
+
+  if (sticky === 'right')
+    return {
+      ...base,
+      right: 0,
+      '&::before': {
+        ...pseudo,
+        left: 0,
+      },
+    } as SxProps<Theme>;
```
martha: I was stumped by this typescript issue and ended up applying the dirty solution of casting to SxProps, here and on line 124, 290, and 471.

src/components/elements/CommonMenuButton.tsx
```diff
@@ -102,6 +102,9 @@ const CommonMenuButton = ({
           vertical: 'top',
           horizontal: 'right',
         }}
+        // Bug: Opening the CommonMenu applies padding to the body, which can look weird on mobile.
```
martha, line 105: I noticed a visual bug where the 3-dots menu causes the table width to jump around, on mobile. It's also observable on qa, for example [here](https://qa-hmis.openpath.host/projects/==T3FGUzVOaDAxdmc5aWJDOVcvbnB4dz09/enrollments). I'd propose to either open another ticket, or fix it separately (maybe as part of [this pr](https://github.com/greenriver/hmis-frontend/pull/1021))

src/modules/caseNotes/components/EnrollmentCaseNotes.tsx
```diff
@@ -52,7 +57,6 @@ export const CASE_NOTE_COLUMNS = {
   NoteContentPreview: {
     key: 'content-preview',
     header: 'Note Content Preview',
-    maxWidth: '450px',
```
martha, line 55: Removed since this wasn't respected (maxWidth is not a valid property on ColumnDef and GenericTable doesn't do anything with it). There is a maxWidth applied to the Box below.

src/modules/bulkServices/components/BulkServicesTable.tsx
```diff
@@ -74,7 +74,7 @@ const BulkServicesTable: React.FC<Props> = ({
         </NotCollectedText>
       );
       return [
-        CLIENT_COLUMNS.name,
+        { ...CLIENT_COLUMNS.name, sticky: 'left' },
```
martha, line 78: Known issue: Checkbox is not sticky. I think my current approach would need a bit more work to accomplish more than 1 sticky column per side. It's not ideal, but I am proposing to descope this and open another ticket, unless you see an easy way around it.

src/components/elements/table/GenericTable.tsx
gigxz: I fiddled around with this a bit, and think I figured out where the issue is coming from. The `SxProp<Theme>` can be an object OR an array, so you can't spread it directly. This is explained in the MUI docs here: https://mui.com/system/getting-started/the-sx-prop/#passing-the-sx-prop

Here is what I found:

* the casts inside the `getStickyCellStyles` function are unnecessary and can be removed. Using the return type is sufficient
* change the return type of `getStickyCellStyles` to be `SystemStyleObject<Theme>` so that we know it's an object, not an array
* instead of casting to `as SxProps<Theme>` when passing the combined sx, pull it out into a variable so that it can be safely type-schecked. This way we won't swallow type errors. Example below that passes type checking and is safe

L282:
```
          {columns.map((def, i) => {
            const headerStyles = headerCellSx ? headerCellSx(def) : {};
            const sx: SxProps<Theme> = {
              ...getStickyCellStyles(def.sticky),
              textAlign: def.textAlign,
              width: def.width,
              // headerStyles has SxProps type so we can't spread it directly. https://mui.com/system/getting-started/the-sx-prop/#passing-the-sx-prop
              ...(Array.isArray(headerStyles) ? headerStyles : [headerStyles]),
            };

            return (
              <HeaderCell key={key(def) || i} sx={sx}>
                {def.header ? (
                  <strong>{def.header}</strong>
                ) : (
                  // If header isn't provided, add a visually hidden header with the column key for accessibility
                  <Box sx={visuallyHidden}>{def.key}</Box>
                )}
              </HeaderCell>
            );
          })}
```

L460:
```
                      const sx: SxProps<Theme> = {
                        ...getStickyCellStyles(def.sticky),
                        width,
                        minWidth,
                        ...(isLinked ? { p: 0 } : undefined),
                        textAlign,
                        whiteSpace: 'initial',
                        ...onClickLinkTreatment,
                        ...tableCellProps?.sx,
                      };
                      return (
                        <TableCell
                          key={key(def) || index}
                          {...tableCellProps}
                          sx={sx}
                        >
```

src/components/elements/table/TableRowActions.tsx
```diff
@@ -21,35 +25,62 @@ const TableRowActions = <T extends { id: string }>({
...
+            tooltip={primaryAria}
+            IconButtonProps={{ size: 'medium' }}
+          />
+        ) : (
+          <Button onClick={primaryActionConfig.onClick} {...buttonProps}>
+            {primaryActionConfig.title}
+          </Button>
+        ))}
       {primaryAction}
       {!!secondaryActionConfigs && secondaryActionConfigs.length > 0 && (
```
gigxz, line 85: Noting that I think there was agreement around including the primary action config inside the secondary menu, so I'd say we can go ahead with that. Maybe it would be as simple as passing `items={[primaryActionConfig , ...secondaryActionConfigs]}` below.

src/modules/bulkServices/components/BulkServicesTable.tsx
gigxz: Ok with me @martha. Not sure how much it would simplify things, but one option is to make the checkbox always be sticky. That seems reasonable to me.

src/components/elements/CommonMenuButton.tsx
gigxz: Good find! agreed to fixing separately

gigxz (commented):
LGTM after suggested Typescript fixes, to ensure continued type safety.

I also think we should include the row-hover-bg-color once we deploy this as well (for review) but I'm not seeing that change on the deploy branch- will that be a separate PR?

src/components/elements/CommonMenuButton.tsx
martha: Opened https://github.com/open-path/Green-River/issues/7220

src/components/elements/table/GenericTable.tsx
martha: Thank you!

martha: > instead of casting to as SxProps<Theme> when passing the combined sx, pull it out into a variable so that it can be safely type-schecked. This way we won't swallow type errors. Example below that passes type checking and is safe

I haven't totally understood this part. Typescript seems happy without pulling it out into a variable, (see https://github.com/greenriver/hmis-frontend/pull/1023/commits/7f50da6300573fd7fd39a5aca6feccf6fdc74a71), but are you indicating that this still carries the risk of swallowing an error, even without the cast?

src/modules/bulkServices/components/BulkServicesTable.tsx
martha: Thanks, Gig, good point! With this suggestion it wasn't so bad, see https://github.com/greenriver/hmis-frontend/pull/1023/commits/273b0594d65fedb1ac13df7470bcb9b119d7ffaa

martha:
> I also think we should include the row-hover-bg-color once we deploy this as well (for review) but I'm not seeing that change on the deploy branch- will that be a separate PR?

Yes, I wanted to keep that on a separate diff since we haven't reached consensus on that change yet, but I opened a draft PR here: https://github.com/greenriver/hmis-frontend/pull/1025

src/components/elements/table/GenericTable.tsx
gigxz: Ah ok, great! It looks like it's happy without it, so it shouldn't be necessary. That must have  just been something I tried that ultimately wasn't necessary.

gigxz (approved):
looks good!

PR #1028, opened 2025-01-17
Follow up on QA for Skip to Content

martha:
## Description

GH issue: https://github.com/open-path/hmis-accessibility/issues/48#issuecomment-2599060192

Summary of changes:
- Make sure "Skip to Content" link has valid destination even when content is loading
- Make sure "Skip to Content" link labels are distinct

To test: See QA instructions https://github.com/open-path/hmis-accessibility/issues/48#issuecomment-2578458354

## Type of change
- Accessibility cleanup

### DIFF
--- a/src/components/layout/dashboard/DashboardContentContainer.tsx
+++ b/src/components/layout/dashboard/DashboardContentContainer.tsx
@@ -19,12 +19,11 @@ interface Props {
   sidebar?: ReactNode;
   header?: ReactNode;
   contextHeader?: ReactNode;
-  navHeader: ReactNode;
+  navHeader?: ReactNode;
   desktopNavIsOpen: boolean;
   mobileNavIsOpen: boolean;
   focusMode?: string;
   navLabel: string;
-  hideLabel?: boolean;
   handleOpenDesktopMenu: VoidFunction;
   handleOpenMobileMenu: VoidFunction;
   handleCloseMobileMenu: VoidFunction;
@@ -46,7 +45,6 @@ const DashboardContentContainer: React.FC<Props> = ({
   handleCloseMobileMenu,
   handleCloseDesktopMenu,
   navLabel,
-  hideLabel,
   contentSx,
 }) => {
   const theme = useTheme();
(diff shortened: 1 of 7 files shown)


src/modules/form/components/FormNavigationContainer.tsx
```diff
@@ -26,7 +26,7 @@ const FormNavigationContainer = ({
   children,
 }: FormNavigationProps) => {
   const isTiny = useIsMobile('sm');
-  const focusTargetId = 'focusable-form';
+  const focusTargetId = 'focusable-form'; // This expects that only 1 form is rendered at a time.
```
gigxz: only 1 form with _form navigation_ enabled, right? On the Enrollment Details we have lots of dynamic-view forms rendered alongside each other, but that doesn't seem like it would cause an issue here.

src/components/layout/dashboard/DashboardContentContainer.tsx
```diff
@@ -22,7 +23,8 @@ interface Props {
   desktopNavIsOpen: boolean;
   mobileNavIsOpen: boolean;
   focusMode?: string;
-  navLabel?: string;
+  navLabel: string;
+  hideLabel?: boolean;
```
gigxz: Is this `hideLabel` complexity worth it? It seems like it's only needed due to a design inconsistency where the Admin dashboard nav doesn't have a nav label. Would adding a nav label there let us remove this complexity?
eg:

Screenshot 2025-01-21 at 2 29 39 PM

==>

Screenshot 2025-01-21 at 2 29 35 PM

(or just keeping the redundant "admin" content too)

src/modules/form/components/FormNavigationContainer.tsx
martha: Yep thanks, I'll clarify!

gigxz: ok. thanks!

src/components/layout/dashboard/DashboardContentContainer.tsx
martha: Yes that's right, it is only needed because of the design inconsistency. I was hesitant to mess with the designs on this unrelated ticket, but if you agree then I will go for it :)

gigxz: Yeah I say go for it!

gigxz (approved):

PR #1032, opened 2025-01-22
RHF mini-refactor for DynamicGroup, bug fix for disabled items

gigxz:
## Description

proposed additions to https://github.com/greenriver/hmis-frontend/pull/612

* Make a change to `DependentFormItemWrapper` to match previous behavior whereby disabled items are set to `null` if they have Protected display type.
  * Reproduction:
    * Select "physical disability" yes
    * Select "physical disability indefinite and impairs" yes
    * Disabling condition should now be autofilled to yes and disabled
    * Select "physical disability" no
    * Disabling condition should now be re-enabled. The bug was that it remained disabled because the "physical disability indefinite and impairs" value was still present in the form state.
* Stop wrapping `DynamicGroup` in `ValueWrapper`. Instead, the groups that need to look at values (FormCard, InputGroup) use handlers from context instead.
* Stop passing `renderSummaryItem` to `DynamicGroup`. Update `DynamicFormFieldAutofillSummary` to work in both View and Edit scenarios for income, and add some comments about this.
* Move `debug` to a hook that is called from FormCard. I think it's worth keeping around.
* Remove special ValueWrapper  for DisabilityTable with `{(values) => cloneElement(group, { values })}`. I don't see any need for it, the components works fine without it, it's able to watch the values it needs.

System tests passed: https://github.com/greenriver/hmis-warehouse/pull/5066

## Type of change
- [x] Bug fix

### DIFF
--- a/src/modules/admin/components/forms/FormPreview.tsx
+++ b/src/modules/admin/components/forms/FormPreview.tsx
@@ -17,11 +17,12 @@ import useSafeParams from '@/hooks/useSafeParams';
 import { useScrollToHash } from '@/hooks/useScrollToHash';
 import SentryErrorBoundary from '@/modules/errors/components/SentryErrorBoundary';
 import DynamicForm, {
+  DynamicFormOnSaveDraft,
   DynamicFormRef,
 } from '@/modules/form/components/DynamicForm';
 import FormNavigationContainer from '@/modules/form/components/FormNavigationContainer';
 import DynamicView from '@/modules/form/components/viewable/DynamicView';
-import { FormValues } from '@/modules/form/types';
+
 import {
   AlwaysPresentLocalConstants,
   createInitialValuesFromSavedValues,
@@ -98,9 +99,11 @@ const FormPreview = () => {
   );
 
   const onSaveFormValues = useCallback(
-    (values: FormValues) =>
+    (values: Parameters<DynamicFormOnSaveDraft>[0]) =>
       formDefinition &&
-      setFormValues(createValuesForSubmit(values, formDefinition.definition)),
+      setFormValues(
+        createValuesForSubmit(values.rawValues, formDefinition.definition)
+      ),
     [formDefinition]
   );
 
@@ -164,9 +167,12 @@ const FormPreview = () => {
       <Stack direction='row' justifyContent='space-between' sx={{ mb: 4 }}>
         <CommonToggle
           value={toggleValue}
-          onChange={(value) => {
+          onChange={(value: PreviewMode) => {
             setToggleValue(value);
-            formRef.current?.SaveIfDirty();
+            // Trigger save to update `formValues` state when toggling away from Edit-mode
+            if (value === 'readOnly') {
+              formRef.current?.SaveDraftForm();
+            }
           }}
           items={toggleItems}
           size='small'
--- a/src/modules/assessments/components/household/HouseholdAssessmentTabPanel.tsx
+++ b/src/modules/assessments/components/household/HouseholdAssessmentTabPanel.tsx
@@ -72,16 +72,18 @@ const HouseholdAssessmentTabPanel = memo(
       if (formState.saving) return;
       // skip save if there are errors. user must do an explicit save to resolve error before proceeding
       if (formState.errors) return;
+      // skip save if form is not dirty
+      if (!formState.dirty) return;
 
       // perform background save or submit
       if (assessmentSubmitted) {
-        formRef.current.SubmitIfDirty(false);
+        formRef.current.SubmitForm();
       } else {
-        formRef.current.SaveIfDirty();
+        formRef.current.SaveDraftForm();
       }
     }, [
       onFormStateChange,
-      formState.saving,
+      formState,
       navigatingAway,
       assessmentId,
       assessmentSubmitted,
@@ -89,7 +91,6 @@ const HouseholdAssessmentTabPanel = memo(
       id,
       refetch,
       updateTabStatus,
-      formState.errors,
     ]);
 
     const { assessment, loading: assessmentLoading } =
@@ -199,7 +200,7 @@ const HouseholdAssessmentTabPanel = memo(
 
       config.push({
         id: 'next',
-        label: 'Next Client',
+        label: nextTab === 'summary' ? 'Next' : 'Next Client',
         action: FormActionTypes.Navigate,
         buttonProps: {
           disabled: !nextTab || disableNavigation,
--- a/src/modules/form/components/DependentFormItemWrapper.tsx
+++ b/src/modules/form/components/DependentFormItemWrapper.tsx
@@ -1,8 +1,9 @@
-import React, { ReactNode } from 'react';
+import React, { ReactNode, useEffect } from 'react';
 
+import { useFormContext } from 'react-hook-form';
 import { useDynamicFieldAutofillSync } from '@/modules/form/hooks/rhf/useDynamicFieldAutofillSync';
 import { useDynamicFieldStatus } from '@/modules/form/hooks/rhf/useDynamicFieldStatus';
-import { FormItem } from '@/types/gqlTypes';
+import { DisabledDisplay, FormItem } from '@/types/gqlTypes';
 
 export interface Props {
   item: FormItem;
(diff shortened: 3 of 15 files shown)


src/modules/form/components/DependentFormItemWrapper.tsx
```diff
@@ -13,9 +14,23 @@ export interface Props {
...
   const { visible, disabled } = useDynamicFieldStatus(item);
   useDynamicFieldAutofillSync(item);
 
+  // If this item is disabled and has a "protected" display, clear the value for this field
+  // (For DisabledDisplay.Hidden, it's cleared automatically when the input is unregistered.)
+  // (For DisabledDisplay.ProtectedWithValue, the value intentionally remains when the item is disabled.)
+  useEffect(() => {
+    if (
+      disabled &&
+      visible &&
```
ttoomey: Doe the `visible` condition need to be here? If we clear the value on a hidden field, does that matter? 
```suggestion
```

src/modules/form/components/DynamicFormFieldAutofillSummary.tsx
```diff
@@ -9,17 +10,37 @@ export interface Props {
   isCurrency: boolean;
 }
 
+/**
+ * Component for rendering the optional "summary item" at the end of a from Group that uses the INPUT_GROUP component.
+ * An item is considered a "summary item" if it has type:DISPLAY, and is the last item in the input group.
+ *
+ * All this really does it implement the currency formatting. If we had a better way to do that for display items,
```
ttoomey:
```suggestion
 * All this really does is implement the currency formatting. If we had a better way to do that for display items,
```

ttoomey (approved):
This looks great! Thanks for all the improvements and the fix to `DependentFormItemWrapper`

src/modules/form/components/DependentFormItemWrapper.tsx
gigxz: Good call-out, I think we _should_ actually be clearing out the value on hidden fields. I updated this to `unregister` disabled items (regardless of visibility) _unless_ the item specifies disabled display "protected with value" which indicates that the value should remain even when disabled (example: Disabling Condition when it's blocked from user-edits).

There is some old logic in `useFormDefinitionHandlers` to translate form state into "cleaned" values, which drops the key for any item that is disabled and not `DisabledDisplay.ProtectedWithValue`. That's the `getValues` function. I have a separate branch that removes it; it should no longer be necessary if disabled items are already unregistered using this wrapper. I think it's a nice cleanup because it's confusing to have both `getValues` (some custom logic for cleaning up the values) and `methods.getValues` (the raw values from rhf). It does add some more risk though. I will open the branch on PR for discussion!

ttoomey: Great! It would be a nice cleanup if we can make do without `handlers.getValues()` or maybe it's something we only use before data is submitted

gigxz:
Hm, a system test `rspec ./drivers/hmis/spec/system/hmis/intake_assessment_spec.rb:67 # Enrollment/household management An active project with wip household can submit an intake assessment` is failing with the latest change, I will look

src/modules/admin/components/forms/FormPreview.tsx
```diff
@@ -98,9 +99,11 @@ const FormPreview = () => {
   );
 
   const onSaveFormValues = useCallback(
-    (values: FormValues) =>
+    (values: Parameters<DynamicFormOnSaveDraft>[0]) =>
       formDefinition &&
-      setFormValues(createValuesForSubmit(values, formDefinition.definition)),
+      setFormValues(
+        createValuesForSubmit(values.rawValues, formDefinition.definition)
```
gigxz, line 105: Fixes pre-existing issue https://github.com/open-path/Green-River/issues/6970

src/modules/assessments/components/household/HouseholdAssessmentTabPanel.tsx
```diff
@@ -72,24 +72,25 @@ const HouseholdAssessmentTabPanel = memo(
...
+      // skip save if form is not dirty
+      if (!formState.dirty) return;
 
       // perform background save or submit
       if (assessmentSubmitted) {
-        formRef.current.SubmitIfDirty(false);
+        formRef.current.SubmitForm();
       } else {
-        formRef.current.SaveIfDirty();
+        formRef.current.SaveDraftForm();
```
gigxz, line 82: Move dirty check from the imperative handler to the `formState`

src/modules/assessments/components/household/HouseholdAssessmentTabPanel.tsx
```diff
@@ -199,7 +200,7 @@ const HouseholdAssessmentTabPanel = memo(
 
       config.push({
         id: 'next',
-        label: 'Next Client',
+        label: nextTab === 'summary' ? 'Next' : 'Next Client',
```
gigxz, line 203: Unrelated ui fix

src/modules/form/components/DynamicForm.tsx
```diff
@@ -91,25 +90,14 @@ export const DynamicFormWithoutHandlers = forwardRef<
...
       () => ({
         SubmitForm: () => {
-          // onSubmit({ values: getCleanedValues(), confirmed: false, });
           onSubmit({ ...getValuesForSubmit(), confirmed: false });
         },
-        SaveIfDirty: () => {
-          if (!handlers.methods.formState.isDirty || props.locked) return;
+        SaveDraftForm: () => {
+          if (props.locked) return;
           handleSaveDraft();
```
gigxz, line 97: I don't _think_ we still need the locked check, but leaving it just in case there is something I'm not catching.

src/modules/form/components/DynamicForm.tsx
```diff
@@ -91,25 +90,14 @@ export const DynamicFormWithoutHandlers = forwardRef<
       ref,
       () => ({
         SubmitForm: () => {
-          // onSubmit({ values: getCleanedValues(), confirmed: false, });
           onSubmit({ ...getValuesForSubmit(), confirmed: false });
         },
-        SaveIfDirty: () => {
-          if (!handlers.methods.formState.isDirty || props.locked) return;
```
gigxz, line 98: This was causing a bug that was caught in system tests.
To reproduce the issue:
* Open a household intake view for a new household (all wip members)
* Enter some responses onto the first intake
* Hit `Next Client`
  * Expected: performs a save with a horizontal loading bar, then proceeds to next tab when complete
  * Bug: does not perform a save, does not proceed to next tab


The fix, which I think is much nicer anyway, is to remove this dirty check and use the one on formState instead which comes from the DirtyObserver. See: https://github.com/greenriver/hmis-frontend/pull/1032/commits/f57119a133d5aff7afbe20eebcb4d93c0d775534#r1934285375

src/modules/form/components/DependentFormItemWrapper.tsx
gigxz: Okay, I pushed one last change https://github.com/greenriver/hmis-frontend/pull/1032/commits/f866f3fdabba347ad2a4d66c3a2a59276f50ec31 which doesn't remove the old `getValues` altogether, but it does rename it and only use it before submission. That should help with confusion between `handlers.getValues()` and `handlers.methods.getValues()`.

src/modules/form/components/DependentFormItemWrapper.tsx
```diff
@@ -1,21 +1,37 @@
...
   useDynamicFieldAutofillSync(item);
 
+  // Unregister disabled item, unless it's configured as "disabled_display: PROTECTED_WITH_VALUE"
+  // which indicates that the field value should remain even when it is disabled.
+  useEffect(() => {
+    if (
+      disabled &&
+      item.disabledDisplay !== DisabledDisplay.ProtectedWithValue
+    ) {
+      unregister(item.linkId);
```
ttoomey, line 31: Do we need to re-register the field if disabled becomes false? Maybe that's already take care of as side-effect.

ttoomey (approved):
Looks great, thanks!

src/modules/form/components/DependentFormItemWrapper.tsx
gigxz: Good question. I tested this and it does appear the field gets re-registered when it's re-enabled. So I think we're good.

gigxz:
Thanks for the review @ttoomey. I will merge in and also update the base branch to release-150.

PR #1037, opened 2025-01-27
Implement Split Households workflow

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/5800
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5089

Summary of changes:
- Add schema updates for Split Household mutation
- Enhance capabilities of the common Menu component
  - Icons for menu items
  - Ability to add a "disabled reason", see inline thread for discussion.
- Add functionality on GenericTable to have disabled selectable rows indicate why they are disabled, in a tooltip on the disabled checkbox
- Implement the Split Households stepper dialog
- Add an action menu, with an item for initiating the dialog, in the Edit Household table
- Update the RelationshipToHohSelect so it has 2 variants: The old one, now called `excludeHoh`, behaves as before, whereas the new one `includeHoh` allows "Self (HoH)" as a selectable item in the dropdown.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/components/elements/CommonMenuButton.tsx
```diff
@@ -105,49 +114,84 @@ const CommonMenuButton = ({
...
+              <Stack direction='row'>
+                {Icon && (
+                  <ListItemIcon>
+                    <Icon />
+                  </ListItemIcon>
+                )}
+                <Stack direction='column'>
+                  {title}
+                  {disabled && disabledReason && (
+                    // We often use tooltips to indicate the reason something is disabled,
```
martha, line 151: Following up from our conversation in the design session about adding disabled reason as a tooltip on action menu items: This turns out to be a bit annoying, since `MenuItem` has to be the direct child of `Menu`, so we can’t wrap it in a tooltip. We could probably create our own disabling mechanism with buttons, but to me, this sounds a bit tricky to get perfect (with good accessibility attributes, tab navigation, etc.). So I’m proposing instead a low lift solution here. Let me know if you think this looks good-enough-for-now:

Screenshot 2025-02-05 at 4 07 23 PM

(Note that I’ve removed “Dense” for this table in particular, which is inconsistent with other table row actions, but I think looks better with disabled reasons)

gigxz: This looks good enough to me, thanks!

src/components/elements/table/GenericTable.tsx
```diff
@@ -71,6 +72,7 @@ export interface Props<T> {
   selectable?: 'row' | 'checkbox'; // selectable by clicking row or by clicking checkbox
   selected?: readonly string[]; // selection can optionally be controlled by the parent
   isRowSelectable?: (row: T) => boolean;
+  getRowSelectDisabledReason?: (row: T) => string | undefined;
```
gigxz: Could these be combined into one function? They seem closely linked, in that `getRowSelectDisabledReason` would only ever be passed if `isRowSelectable` is also passed, and they would have the same internal logic for determining whether the row is selectable or not.


```suggestion
  isRowSelectable?: (row: T) => {
    selectable: boolean; // Whether this specific row can be selected (if `selectable` prop is passed)
    reason?: string; // Reason why this row cannot be selected (displayed in tooltip)
  };
```

src/components/elements/table/tableRowActionUtil.tsx
```diff
@@ -11,8 +11,7 @@ export const BASE_ACTION_COLUMN_DEF: ColumnDef<any> = {
   key: 'Actions',
   tableCellProps: {
     sx: {
-      py: 0,
-      px: 0,
+      p: 1,
```
gigxz, line 17: Slightly unrelated, but I think this `BASE_ACTION_COLUMN_DEF` needs a comment to explain usage. iiuc it should only be used in certain cases where we _can't_ use the default table linking behavior (by passing `rowLinkTo` OR `handleRowClick` OR `rowSecondaryActionConfigs`). For example it's used by the Bulk Services table to control rendering of the table row action menu so it can inject the assign service button. Is that fair?

src/modules/household/components/EditHouseholdMemberTable.tsx
```diff
@@ -199,18 +223,54 @@ const EditHouseholdMemberTable = ({
...
-            householdClient={hc}
-            onSuccess={refetchHousehold}
-            householdSize={currentMembers.length}
-            ariaLabel={`Remove ${clientBriefName(hc.client)}`}
+          <TableRowActions
+            record={hc}
+            recordName={clientBriefName(hc.client)}
+            MenuProps={{
+              MenuListProps: {
+                dense: false,
```
gigxz: I'm not seeing a visual difference when toggling this, what is it doing?

src/modules/household/components/EditHouseholdMemberTable.tsx
```diff
@@ -199,18 +223,54 @@ const EditHouseholdMemberTable = ({
...
+              },
+            }}
+            menuActionConfigs={[
```
gigxz: Now that we don't have the `RemoveFromHouseholdButton` anymore, can we pass these as a `rowSecondaryActionConfigs` prop directly to the table?

src/modules/household/components/householdActions/SplitHouseholdSelectClients.tsx
```diff
@@ -0,0 +1,58 @@
...
+      }
+      getRowSelectDisabledReason={(row: HouseholdClientFieldsFragment) =>
+        row.relationshipToHoH === RelationshipToHoH.SelfHeadOfHousehold
+          ? 'Head of Household may not be split'
+          : undefined
+      }
+    >
+      <Typography variant='body1'>
+        Select which clients from {clientBriefName(donorHoh.client)}’s household
+        to split into a new household.
```
gigxz: This might just be me, but I find it clearer to just read `Select which clients to split into a new household.`. The inclusion of the HoH's name confused me because they are irrelevant to the new household (since they cannot be included in it)

src/modules/household/components/householdActions/SplitHouseholdSelectClients.tsx
```diff
@@ -0,0 +1,58 @@
...
+    <SelectClientsStep
+      donorHousehold={donorHousehold}
+      selectedClients={selectedClients}
+      setSelectedClients={setSelectedClients}
+      isRowSelectable={(row: HouseholdClientFieldsFragment) =>
+        row.relationshipToHoH !== RelationshipToHoH.SelfHeadOfHousehold
+      }
+      getRowSelectDisabledReason={(row: HouseholdClientFieldsFragment) =>
+        row.relationshipToHoH === RelationshipToHoH.SelfHeadOfHousehold
+          ? 'Head of Household may not be split'
```
gigxz: `cannot` for plainer language? that's what we use in other disabled text, I don't see "may not" used anywhere currently

src/modules/household/components/householdActions/SelectClientsStep.tsx
```diff
@@ -70,11 +74,13 @@ const SelectClientsStep = ({
       <Paper>
         <GenericTable<HouseholdClientFieldsFragment>
           rows={donorHouseholdMembers}
+          rowName={(row) => clientBriefName(row.client)}
           columns={MANAGE_HOUSEHOLD_COLUMNS}
           selectable={'checkbox'}
           selected={selectedClientIds}
           onChangeSelectedRowIds={setSelectedClientIds}
           isRowSelectable={isRowSelectable}
+          getRowSelectDisabledReason={getRowSelectDisabledReason}
```
gigxz: This is getting into revisiting design, so feel free to ignore, but just want to throw it out there. I didn't think of it when reviewing the designs, but another approach would be to just exclude the HoH(s) from this table. It seems potentially more straightforward, because the user is only faced with the options they _can_ select, rather than also including options they can't select (with disabled controls and hover actions to understand why). It seems like less cognitive load to me.

src/modules/household/components/householdActions/SplitHouseholdDialog.tsx
```diff
@@ -0,0 +1,191 @@
...
+        ),
+        ...disabledProps,
+        onOpen: () => {
+          if (splittingClients.length === 1) {
+            setRelationships({
+              [splittingClients[0].enrollment.id]:
+                RelationshipToHoH.SelfHeadOfHousehold,
+            });
+          }
+        },
```
gigxz: It feels a little off to me for this data change being tied to an "open" action, when it feels like more of a side-effect of the state having 1 client selected. 

Some alternate ideas:

```
  // If there is only 1 client selected for the new household, auto-mark them as HoH
  useEffect(() => {
    if (splittingClients.length === 1) {
      setRelationships({
        [splittingClients[0].enrollment.id]:
          RelationshipToHoH.SelfHeadOfHousehold,
      });
    }
  }, [splittingClients]);
```
or - probably better since it avoids useEffect:
```
              <SplitHouseholdSelectClients
                donorHousehold={donorHousehold}
                selectedClients={splittingClients}
                // pull this logic into its own callback onChangeSelectedClients
                setSelectedClients={(clients) => {
                  setSplittingClients(clients);
                  if (clients.length === 1) {
                    setRelationships({
                      [clients[0].enrollment.id]:
                        RelationshipToHoH.SelfHeadOfHousehold,
                    });
                  } else { /* clear relationships, maybe */ }
                }}
              />
```

gigxz (approved):
great work!! some minor comments below, but overall LGTM

src/components/elements/table/GenericTable.tsx
martha: I will take a stab at cleaning this up. I think I kept them separate since a disabled reason is not always passed, but maybe it would be good to require.

src/components/elements/table/tableRowActionUtil.tsx
martha: Yeah exactly, I will add comments!

src/modules/household/components/EditHouseholdMemberTable.tsx
martha: Dense (default):
Screenshot 2025-02-11 at 9 56 05 AM

Not dense:
Screenshot 2025-02-11 at 9 55 36 AM

src/components/elements/table/GenericTable.tsx
martha: Looks like we can excise this for now anyway based on https://github.com/greenriver/hmis-frontend/pull/1037#discussion_r1950857673

src/modules/household/components/EditHouseholdMemberTable.tsx
martha: Based on https://github.com/greenriver/hmis-frontend/pull/1037#discussion_r1950839861 this will now be handled through GenericTable's defaults, so I'll leave it as default dense for now. I think it looks ok.

src/modules/household/components/householdActions/SplitHouseholdDialog.tsx
martha: Yeah this is much nicer, thanks!

martha: I could go either way on clearing out the relationships, but I think it makes sense. It's nice to be consistent, so I added the same behavior on Join too: https://github.com/greenriver/hmis-frontend/pull/1037/commits/fa71bf577317af837a66419011b3e2872ca3d066

PR #1040, opened 2025-02-03
Resolve less sensitive data on client search results; fix type issues

gigxz:
## Description

* Only resolve SSN on search results when necessary
  * Column was removed in most tables in https://github.com/greenriver/hmis-frontend/pull/1020
  * The only place it remains is when searching for a client to do a merge
* Fix type issues in lots of places where we do a client search and expect the result to be `ClientFieldsFragment` (FULL fragment for client dashboard) but it's really `ClientSearchResultFieldsFragment` (limited info in search results). This mistake happened because `GenericTableWithData` doesn't enforce that the third generic (row type) is _actually_ the type of data returned from the query. This could probably be improved to catch such errors in the future.
* Fix some Apollo cache issues and incorrect usage of root permissions vs client permissions
* PR to follow for pruning access fragments to improve overall performance. We are resolving a lot of permission on root (and elsewhere) that we don't actually need. https://github.com/greenriver/hmis-frontend/pull/1046

## Type of change
- [x] Bug fix


src/modules/household/components/AssociatedHouseholdMembers.tsx
```diff
@@ -12,14 +12,22 @@ import {
...
 
 export const householdMemberColumns: ColumnDef<
-  ClientFieldsFragment | RecentHouseholdMember
+  | ClientFieldsFragment
+  | ClientSearchResultFieldsFragment
+  | RecentHouseholdMember
 >[] = [
   CLIENT_COLUMNS.name,
-  { ...CLIENT_COLUMNS.ssn, width: '150px' },
+  { ...CLIENT_SSN_COLUMN, width: '150px' }, //??? TODO GIG determine if we need this
```
gigxz, line 30: revisit

src/components/elements/table/GenericTable.stories.tsx
```diff
@@ -3,6 +3,7 @@ import { Meta, StoryFn } from '@storybook/react';
 
 import GenericTable, { Props as GenericTableProps } from './GenericTable';
 import { SsnDobShowContextProvider } from '@/modules/client/providers/ClientSsnDobVisibility';
+import { clientSsnColumn } from '@/modules/clientMerge/components/client/NewClientMerge';
```
martha: Our usual pattern when exporting a column definition is to caps lock and underscore the name. I don't have a strong preference either way but it's nice to be consistent. Do you want to move towards using camel case instead?

src/modules/household/components/ManageHousehold.tsx
```diff
@@ -90,24 +89,25 @@ const ManageHousehold = ({
...
-        ];
-      }
-      return defaults;
-    }, [addToEnrollmentColumns, globalFeatureFlags?.mciId, isMobile]);
+  const columns = useMemo(() => {
+    const defaults: ColumnDef<
+      ClientSearchResultFieldsFragment | RecentHouseholdMember
+    >[] = [CLIENT_COLUMNS.name, CLIENT_COLUMNS.dobAge];
+    if (isMobile) {
+      // On mobile, show enrollment button right next to the client name so user
```
martha, line 97: (no need to make a change in this PR, just mulling) This was an interesting one-off fix, but now that we have sticky columns implemented as a more general pattern, that would be appropriate to apply here? We can check in with design about it, since there is a risk with sticky columns for the total sticky width to get wide enough that the non-sticky columns aren't scrollable at all.

src/modules/household/components/elements/AddToHouseholdButton.tsx
```diff
@@ -10,13 +10,13 @@ import { useFormDialog } from '@/modules/form/hooks/useFormDialog';
...
 import {
+  ClientSearchResultFieldsFragment,
   RecordFormRole,
-  ClientWithAlertFieldsFragment,
   SubmittedEnrollmentResultFieldsFragment,
 } from '@/types/gqlTypes';
 
 interface Props {
-  client: ClientWithAlertFieldsFragment;
+  client: ClientSearchResultFieldsFragment; //ClientWithAlertFieldsFragment;
```
martha, line 19: Clean up comment?

src/modules/search/components/ClientSearch.tsx
```diff
@@ -106,22 +95,23 @@ export const CLIENT_COLUMNS: {
...
-export const SEARCH_RESULT_COLUMNS: ColumnDef<ClientFieldsFragment>[] = [
-  CLIENT_COLUMNS.id,
-  CLIENT_COLUMNS.first,
-  CLIENT_COLUMNS.last,
-  { ...CLIENT_COLUMNS.dobAge, width: '180px' },
-];
+export const SEARCH_RESULT_COLUMNS: ColumnDef<ClientSearchResultFieldsFragment>[] =
+  [
+    CLIENT_COLUMNS.id,
+    CLIENT_COLUMNS.first,
```
martha: Per our [tables sheet](https://docs.google.com/spreadsheets/d/17ZpdB0fH6kUL14KDRWrlukwIH93BNx9grrNZApLpUwg/edit?gid=1987637083#gid=1987637083), I think we had been planning to consolidate around consistently using one Client Name col, instead of sometimes having first/last be separate cols. Do you want to make that change here? (Especially since that's already how the mobile columns are configured.) 

I could also see the justification for keeping first/last separate on desktop search to support some workflows; maybe add a comment about why, if you want to keep it as-is.

martha (approved):
Everything looks great! Thank you for the cleanup!

src/modules/household/components/elements/AddToHouseholdButton.tsx
martha: Oh nvm you caught this one on your other PR

src/components/elements/table/GenericTable.stories.tsx
gigxz: Agreed I will update for consistency. Originally I wasn't exporting it but I realized we have it in storybook, which was probably good to keep.

src/modules/household/components/ManageHousehold.tsx
gigxz: Good question, I was wondering the same. We could add `sticky: 'right',` to `addToEnrollmentColumns`:
Screenshot 2025-02-06 at 12 58 34 PM

Though in the previous designs we wanted to be able to have that "action button" inside the same cell as the menu (re Bed Nights), and I think we descoped it.

src/modules/search/components/ClientSearch.tsx
gigxz: Yes, I got cold feet about that; [slack msg](https://greenriver.slack.com/archives/C033YNSN0Q1/p1738589307247659?thread_ts=1737558629.575329&cid=C033YNSN0Q1). I reverted to reduce the impact of the change, thinking maybe we should follow up on how often people search by DOB. We didn't talk about the fact that you can search by DOB/ID but they never show up in the results anymore.

gigxz: Though replacing this with `[CLIENT_COLUMNS.name, CLIENT_COLUMNS.id, CLIENT_COLUMNS.dobAge]` seems fine to me, it's showing the same information. (Name before ID since it's readable). I will make that change. It is the full removal of DOB and ID that had me slightly concerned about disrupting workflows

PR #1041, opened 2025-02-03
Follow up on QA items for Table Row Actions

martha:
## Description

GH issues: 
- https://github.com/open-path/Green-River/issues/6761#issuecomment-2631902523
- https://github.com/open-path/Green-River/issues/6825#issuecomment-2631767251

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/modules/dataFetching/components/GenericTableWithData.tsx
+++ b/src/modules/dataFetching/components/GenericTableWithData.tsx
@@ -89,7 +89,6 @@ export interface Props<
   applyOptionalColumns?: (columns: string[]) => Partial<QueryVariables>;
   onCompleted?: (data: Query) => void;
   filterRows?: (rows: RowDataType) => boolean; // Client-side row filtering
-  labelRowsPerPage?: string;
 }
 
 function allFieldColumns<T>(recordType: string): ColumnDef<T>[] {
@@ -137,7 +136,6 @@ const GenericTableWithData = <
   paginationItemName,
   filterRows,
   vertical,
-  labelRowsPerPage,
   ...props
 }: Props<
   Query,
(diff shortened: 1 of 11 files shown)


src/modules/dataFetching/components/GenericTableWithData.tsx
```diff
@@ -251,14 +253,16 @@ const GenericTableWithData = <
         setPage(0);
       },
       count: nodesCount,
+      labelRowsPerPage,
```
gigxz: Maybe we can use the existing `recordType` or `paginationItemName` props so that we don't end up with diverging implementations of the `per page:` part.

This should work:
```suggestion
      labelRowsPerPage: `${pluralize(startCase(recordType || paginationItemName || 'Row'))} per page:`,,
```

It think this basically matches what we do for the "displaying 25 of 500 clients" text

martha: Sounds good. It looks like the [pagination text "displaying X of X clients"](https://github.com/greenriver/hmis-frontend/blob/a028a949d9012265cd12393602bf0864d1b2466e/src/modules/dataFetching/components/GenericTableWithData.tsx#L385-L387) uses `paginationItemName` first and then `recordType`, rather than the other way round, so let's do that here too.

gigxz: yes! thanks

src/modules/household/components/ManageHousehold.tsx
```diff
@@ -196,12 +196,7 @@ const ManageHousehold = ({
 
           {searchInput && (
             <SsnDobShowContextProvider>
-              <Box
-                sx={{
-                  border: (theme) => `1px solid ${theme.palette.borders.light}`,
-                  borderRadius: 1,
-                }}
-              >
+              <Paper>
```
martha, line 199: @gigxz I'm proposing to just replace with `Paper` here to avoid custom styling, but was there a purposeful reason for this? (e.g. maybe a Paper within another Paper we consider an antipattern?)

gigxz: looks good, thanks! it was floating border-less before, which I think looked especially off with the action menu.

gigxz (approved):
thanks @martha!

PR #1046, opened 2025-02-06
Prune access fragments

gigxz:
## Description

* Stop resolving permissions that we don't use. This should improve application performance, especially initial page load since we load the root access fragment first.
* Fix some places where we used "root" permission instead of entity-specific
* Remove some dead code
* Add comments pointing out places that could be further improved

system tests: https://github.com/greenriver/hmis-warehouse/pull/5115
issue: https://github.com/open-path/Green-River/issues/7290

## Type of change
- [x] Bug fix


ttoomey (approved):
looks good, thanks!

martha (approved):
This is great Gig, thank you! I'm curious whether you have found the performance improvement on page load to be measurable?

gigxz:
@martha I don't think we measure page loads, but the Sentry instrumentation allows us to track performance of each individual request, so we could use that as a proxy. There is an open issue https://github.com/open-path/Green-River/issues/7179 for digging more into what's available and what instrumentation we are missing

PR #1055, opened 2025-02-12
Follow up on QA items from Join/Split

martha:
## Description

Summary of changes:
- Move `key` out of spread `props` object to avoid console warning
- Fix bug with table selection refactor that impacted Bulk Services
- Avoid unneeded renders caused by `useEffect` in Bulk Services - this is unrelated to join/split, but something I found and fixed while debugging the above
- Lightly edit the language of the workflow instructions (active voice, full stops) per our new [guidelines](https://www.figma.com/design/bsw8CX4UOq6SIdXYorZDWV/OP-HMIS-%2F-Library-%2F-Design?node-id=2680-15812&p=f&t=ectRlT2RZjCxm4L9-0) ✨ 

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/components/elements/table/hooks/useTableSelection.tsx
+++ b/src/components/elements/table/hooks/useTableSelection.tsx
@@ -18,8 +18,9 @@ export function useTableSelection<T extends { id: string }>({
   const [selectedState, setSelectedState] = useState<string[]>();
 
   // Table row selection can be either controlled or uncontrolled:
-  // - controlled: `selectedCtrl` and `onChangeSelectedCtrl` props passed from parent
-  // - uncontrolled: managed internally in `selectedState`
+  // - controlled: `selectedControlled` and `onChangeSelected` props passed from parent
+  // - uncontrolled: managed internally in `selectedState`,
+  // (but we still call `onChangeSelected` so the caller can "listen in" on what rows are selected)
   const isSelectControlled = selectedControlled !== undefined;
 
   const selected = useMemo(
@@ -62,6 +63,13 @@ export function useTableSelection<T extends { id: string }>({
 
   // Clear selection when data changes
   useEffect(() => setSelectedState([]), [rows]);
+
+  // Call onChangeSelected callback if selection changes.
+  // (Usually we call onChangeSelected directly when changing the value, but
+  // it doesn't work for the useEffect above, so we still need this effect too)
+  useEffect(() => {
+    if (selected) onChangeSelected?.(selected);
+  }, [selected, onChangeSelected]);
 
   return {
     selected,
--- a/src/components/elements/table/hooks/useTableSelection.tsx
+++ b/src/components/elements/table/hooks/useTableSelection.tsx
@@ -62,14 +62,10 @@ export function useTableSelection<T extends { id: string }>({
   );
 
   // Clear selection when data changes
-  useEffect(() => setSelectedState([]), [rows]);
-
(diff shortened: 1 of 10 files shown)


src/components/elements/table/hooks/useTableSelection.tsx
```diff
@@ -63,6 +64,13 @@ export function useTableSelection<T extends { id: string }>({
   // Clear selection when data changes
   useEffect(() => setSelectedState([]), [rows]);
 
+  // Call onChangeSelected callback if selection changes.
```
martha: While refactoring, I thought this could be removed, since we now call `onChangeSelected` elsewhere in the hook whenever selection changes. However, it doesn't quite work correctly for the `useEffect` above on line 65, so I'm re-adding this effect. Behavior should be restored to what it was before.

I also investigated an alternative to avoid these effects, such as returning a `clearSelection` callback from this hook. However, the state ownership is a little circuitous here -- it's `GenericTableWithData` and not `GenericTable` that would need to call `clearSelection` when `rows` changes -- so I think the best solution for now is to keep the same `useEffect`s from before.

gigxz: Does this mean onChangeSelected will be called twice for most changes? I worry that might have some unintended side-effects. 


> it doesn't quite work correctly for the useEffect above on line 65, so I'm re-adding this effect

If this effect only needs to be triggered when the effect on line 65 runs, can we just move the contents into the effect on line 65 ?

```
  // Clear selection when data changes
  useEffect(() => { setSelectedState([]);  onChangeSelected?.([])}, [rows]);
```
(or injecting some logic for isSelectControlled if needed, I'm not sure if it would be)

src/modules/household/components/householdActions/ConflictingEnrollmentAlert.tsx
```diff
@@ -59,8 +59,7 @@ const ConflictingEnrollmentAlert = ({
               You have two options:
               <List sx={{ listStyle: 'decimal', pl: 4 }} component='ol'>
                 <ListItem sx={{ display: 'list-item' }}>
-                  The conflicting enrollment can be joined with {hohName}’s
-                  household.
+                  Join the conflicting enrollment with {hohName}’s household.
```
gigxz, line 62: nice, thanks

src/components/elements/table/hooks/useTableSelection.tsx
gigxz: (I didn't test this, so I'm sure there's something I'm missing. Just wanted to offer the idea).

martha: Yes, this was my first thought too -- that we should just call `onChangeSelected([])` inside of the previous `useEffect`. Thanks for pointing me back to it, I think I see a fix for the bug I previously encountered... 🐞

gigxz: ok, great!

gigxz (approved):
lgtm!

martha:
Hm tricky. I didn't QA this thoroughly enough locally. I have to dig a little more to get both Assign Service and Join/Split working correctly. Join/Split is seeing the selection get unexpectedly cleared; but it looks like it could just be an issue of memoizing props

gigxz:
sounds tricky indeed, lmk if you want to pair!

PR #1058, opened 2025-02-13
Address bugs in table selection

martha:
## Description
GH issues: [Join](https://github.com/open-path/Green-River/issues/5752), [Split](https://github.com/open-path/Green-River/issues/5800)

My understanding of the bug:
- Assign Services needs the row selection state to get reset, and the `onChangeSelected` callback to fire, after the bulk assign mutation completes. This allows it to re-enable the individual Assign buttons which are disabled whenever any rows are selected ([here](https://github.com/greenriver/hmis-frontend/blob/release-152/src/modules/bulkServices/components/BulkServicesTable.tsx#L177).) [This PR](https://github.com/greenriver/hmis-frontend/pull/1055/files#diff-682e21260759a46f4f77ae074c6a14ecc7cb21166fb35d4dd6e3d786a128a502R67) yesterday fixed that problem. However,
- This change made selection in Join/Split completely impossible - behavior that is observable on QA right now. It flickers but immediately unselects. The bug is that the passed `clients` prop needs to be memoized in order for the [`useEffect` in `useTableSelection`](https://github.com/greenriver/hmis-frontend/pull/1055/files#diff-682e21260759a46f4f77ae074c6a14ecc7cb21166fb35d4dd6e3d786a128a502R68) not to keep running over and over. _However_,
- Even after fixing that bug, initial selection does not work anymore. I _think_ this is because the GenericTable is first rendered after the initial selection state is already set. ([link to split](https://github.com/greenriver/hmis-frontend/pull/1037/files#diff-d4ec52015eb3476ad89d703af3ed1b770d25545b8a51dbbef38eae66c0bb028fR31-R33), [link to join](https://github.com/greenriver/hmis-frontend/pull/1022/files#diff-9d96cff77d6ee71b82b0fa8139813a2a9b76af074b82c109fcc668cdfb09d72fR46-R48)) The `useEffect` in the `useTableSelection` hook runs on render, and resets selection state to zero.

Edit: Solution this PR implements:
- Only run the `useEffect` in `useTableSelection` on subsequent renders, NOT the first render.

<Details>
<Summary>Alternatives Considered</Summary>

- Add a new prop to `GenericTable`, called `autoResetSelection`. It enables callers to specify whether the table should reset the selection when the data in `rows` changes. It is true by default but set to false for Join/Split.
  - Cons: this seems pretty brittle to me because it means that any caller who wants managed selection on a `GenericTable` has to pass `autoResetSelection`, which they may not always want.
- Use a second effect, as [originally proposed yesterday](https://github.com/greenriver/hmis-frontend/pull/1055/commits/cd0a2e0aef98a1b28692619ed3c1f6c9099a10ff#diff-682e21260759a46f4f77ae074c6a14ecc7cb21166fb35d4dd6e3d786a128a502R68-R73), but _remove_ the other calls to `onChangeSelected` so it is _only_ called from the effect. This more closely mirrors the [existing pre-Join/Split code](https://github.com/greenriver/hmis-frontend/blob/release-151/src/components/elements/table/GenericTable.tsx#L336). But I wasn't able to get it working, encountered `Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.`
- Make the `useTableSelection` hook return a `clearSelection` callback, similar to what I described [in a comment](https://github.com/greenriver/hmis-frontend/pull/1055/commits/cd0a2e0aef98a1b28692619ed3c1f6c9099a10ff#r1952999625) on my PR yesterday. I think this would require "hoisting" selection state up for all components that use table selection. I do think this is the most react-y way to solve the problem here? But it is a lot of code churn, so I didn't pursue it.
</Details>

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/components/elements/table/hooks/useTableSelection.tsx
+++ b/src/components/elements/table/hooks/useTableSelection.tsx
@@ -1,5 +1,5 @@
 import { without } from 'lodash-es';
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 
 export function useTableSelection<T extends { id: string }>({
   selectable = false,
@@ -7,14 +7,12 @@ export function useTableSelection<T extends { id: string }>({
   rows,
   selectedControlled,
   onChangeSelected,
-  autoResetSelection = true,
 }: {
   selectable?: boolean;
   isRowSelectable?: (row: T) => boolean;
   rows: T[];
   selectedControlled?: readonly string[];
   onChangeSelected?: (ids: readonly string[]) => void;
-  autoResetSelection?: boolean;
 }) {
   // Initially set selected to undefined, so we can early return and avoid state flicker
   const [selectedState, setSelectedState] = useState<string[]>();
@@ -36,6 +34,11 @@ export function useTableSelection<T extends { id: string }>({
     return rows.filter(isRowSelectable).map((r) => r.id);
   }, [rows, selectable, isRowSelectable]);
 
+  const deselectAll = useCallback(() => {
+    if (!isSelectControlled) setSelectedState([]);
+    onChangeSelected?.([]);
+  }, [isSelectControlled, onChangeSelected]);
+
   const handleSelectAllClick = useCallback(
     (event: React.ChangeEvent<HTMLInputElement>) => {
       if (event.target.checked) {
@@ -44,11 +47,10 @@ export function useTableSelection<T extends { id: string }>({
         onChangeSelected?.(selectableRowIds);
       } else {
         // deselect all
-        if (!isSelectControlled) setSelectedState([]);
-        onChangeSelected?.([]);
+        deselectAll();
       }
     },
-    [isSelectControlled, onChangeSelected, selectableRowIds]
+    [deselectAll, isSelectControlled, onChangeSelected, selectableRowIds]
   );
 
   const handleSelectRow = useCallback(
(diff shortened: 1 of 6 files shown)


src/components/elements/table/hooks/useTableSelection.tsx
```diff
@@ -63,9 +65,11 @@ export function useTableSelection<T extends { id: string }>({
...
   // Clear selection when data changes
   useEffect(() => {
-    setSelectedState([]);
-    onChangeSelected?.([]);
-  }, [onChangeSelected, rows]);
+    if (autoResetSelection) {
+      setSelectedState([]);
+      onChangeSelected?.([]);
+    }
+  }, [autoResetSelection, onChangeSelected, rows]);
```
gigxz: Thanks for the explanation @martha! I agree this feels a bit brittle. I am wondering a couple things:

1) Elsewhere in this hook, we only call `setSelectedState` if `!isSelectControlled`. (For example when user clicks deselect all). I think this prevents the "setstate in useeffect" error we would otherwise see. So I was wondering why this hook has different behavior (calling set state for controlled)
2) The problem seems to be that this is resetting the selection on initial render. useEffect will run on first render AND on subsequent changes to dependencies. So as-is it still doesn't make a ton of sense, I don't think, because it will clear the selection initially. (Maybe this bug is masked because bed nights doesn't "initially" select anything).

Here's another idea where we just adjust this useEffect to _not_ clear on first render:
https://github.com/greenriver/hmis-frontend/pull/1059/files (tested with: Bed nights; Join household; Split household; global client merge; intake submission)


I don't love it, but it feels a bit closer to desired behavior than adding an "autoResetSelection" prop, which could still result in the selection being cleared on the first render. It would probably be even better if the parent component-- eg Bulk Services Table-- could directly control when to clear the selection.. maybe this could all be cleaned up if we adjusted all uses of `selectable` to be "controlled". But I think that is a bit too much churn for this bug fix.

martha: 1. Good point I agree that should be changed! Thanks!
2. Yes exactly, that's a better articulated/more accurate way to say what I was getting at in my comment above. (that any caller who wants managed selection would have to pass `autoResetSelection` in order to avoid the bug. You are accurately pointing out that it's not just any caller who wants managed selection, but any caller who wants to set an initial selection. Right now (I think) we don't expose a way for non-managed-selection callers to set initial selection so I was collapsing the 2 concepts).

ANYway, yes. I like your proposal pretty well. Thank you!

martha:
Running system tests locally against release-152. Then will merge and re-deploy 152 to frontend stg envs.

PR #1062, opened 2025-02-17
Release 153

gigxz:
https://github.com/greenriver/hmis-warehouse/pull/5141


gigxz:
@martha any idea why the builds are failing on storybook? I'm having a hard time parsing it, maybe chromatic is down or something. Storybook is working for me locally.

martha:
@gigxz not sure, but I wonder if it was a rate limiting thing? Now that we have set chromatic build to happen on every PR we might be hitting a quota. But it looks like it did succeed this morning.

PR #1065, opened 2025-02-20
Updated enrollment workflow

gigxz:
## Description

Issue w/ testing instructions: https://github.com/open-path/Green-River/issues/7209

Streamline enrollment and household management workflows, including:
* Move HoH change and Relationship change into menu items
* Add Previously Associated members to household creation

## Type of change


src/modules/household/components/ManageHousehold.tsx
```diff
@@ -49,171 +49,192 @@ interface Props {
...
+    ClientSearchResultFieldsFragment | RecentHouseholdMember
+  >[] = useMemo(() => {
+    return [
+      { ...CLIENT_COLUMNS.name, sticky: 'left', width: '25%' }, // why does sticky make it so wide?
+      ...(globalFeatureFlags?.mciId
+        ? [externalIdColumn(ExternalIdentifierType.MciId, 'MCI ID')]
+        : []),
+      // On mobile, show enrollment button right next to the client name so user.
+      // This needs fixing, ideally we could use a sticky column but it doesn't work as expected on mobile
+      ...(isMobile ? addToEnrollmentColumns : []),
```
gigxz, line 108: Ran out of time on this one, it seems like an issue with how "sticky" columns are working. I was hoping we could do something similar to Bed Nights, but it has the same issue, it removes the sticky behavior on mobile altogether. I think the "sticky" was added in part _for_ mobile, so we should find a way for it to work there. @martha I think you have a follow-up issue for this somewhere but I can't find it, can you? thanks!

src/components/elements/CommonMenuButton.tsx
```diff
@@ -162,36 +174,49 @@ const CommonMenuButton = ({
...
-                    // close menu before triggering onClick
-                    setAnchorEl(null);
-                    onClick();
-                  }
-                }}
-              >
-                {menuItemLabel}
-              </MenuItem>
-            );
+            if (sectionLabelEl) {
```
gigxz: cant figure out how to add labels and keep accessibility of menu

gigxz: not implementing section labels from mock too much of a time sink. we probably need to implement our own menu

src/modules/household/components/ManageHousehold.tsx
martha: I don't think that we have an open issue for this, but I'm happy to create one!

Is this PR ready for review in general? The functionality looks amazing from what I have poked at, but I saw that it is still in draft so I haven't begun a proper review. Let me know if I should.

martha: Created https://github.com/open-path/Green-River/issues/7338

src/modules/household/components/ManageHousehold.tsx
```diff
@@ -49,171 +49,192 @@ interface Props {
...
-        ...defaults,
-      ];
-    }
-    return defaults;
-  }, [addToEnrollmentColumns, isMobile, globalFeatureFlags?.mciId]);
+  const searchResultColumns: ColumnDef<
+    ClientSearchResultFieldsFragment | RecentHouseholdMember
+  >[] = useMemo(() => {
+    return [
+      { ...CLIENT_COLUMNS.name, sticky: 'left', width: '25%' }, // why does sticky make it so wide?
```
gigxz: I can remove this comment, but would be good to address at the source. Adding `sticky` makes the column take up as much space as possible, it seems, which isn't in line with the mocks

src/modules/household/components/ManageHousehold.tsx
gigxz: Thank you! Yes it's ready for review, I'll mark it now. Sorry it's such a monster, happy to walk through some changes together if that would be helpful!

src/modules/household/hooks/usePreviouslyAssociatedMembers.tsx
```diff
@@ -0,0 +1,74 @@
...
+  useGetClientHouseholdMemberCandidatesQuery,
+} from '@/types/gqlTypes';
+
+/**
+ * Get a unique list of recent household members for this household.
+ * Currently based off HoH only. Could be expanded to look at previously
+ * associated members to all clients in the household, but we should
+ * make a dedicated more performant query before doing so.
+ */
+export function usePreviouslyAssociatedMembers(
```
gigxz, line 21: renamed useRecentHouseholdMembers, with some adjustments

src/components/elements/CommonCard.tsx
```diff
@@ -9,6 +9,10 @@ interface CommonCardProps {
   className?: string;
 }
 
+// TODO(#7191): Replace usages of CommonCard and TitleCard with CommonCollapsibleCard
```
martha: It seems preferable to me to have `CommonCard` and `CommonCollapsibleCard` stay separate components but have them both render the same 'BaseCard' component internally. That way they can gracefully share the things they should have in common (padding styles, title variant) but the props that are irrelevant stay separate (collapsible, actions). What do you think? Is there something I'm missing?

Either way, fine to leave it as a todo for the Design Library epic

src/components/elements/CommonCollapsibleCard.tsx
```diff
@@ -0,0 +1,116 @@
...
+}) => {
+  if (collapsible && actions)
+    throw new Error('Cannot have actions on collapsible card');
+
+  const collapseIcon =
+    collapsible && (open ? <ExpandLessIcon /> : <ExpandMoreIcon />);
+
+  const cardTitle = (
+    <Stack
+      direction='row'
```
martha: This isn't that mobile friendly now, but maybe it's an easy fix, by setting `direction={isTiny ? 'column' : 'row'}` as we do elsewhere?

src/modules/household/components/CreateHouseholdPage.tsx
```diff
@@ -14,40 +18,90 @@ function buttonTextForPath(path?: string) {
...
+  ProjectDashboardRoutes.ADD_HOUSEHOLD,
+];
+
+/**
+ * Page for creating a new household.
+ * This is used for enrolling a new household in a project, including during the bulk services and bed nights workflows.
+ *
+ *
+ *
+ * Back button has special logic for Bulk Services and Bed Nights household creation. To test:
```
martha: Cool, this is great!

Nit, but this comment struck me a little funny. Maybe because it is providing testing instructions (instead of just describing logic/functionality), and also it is somewhat far away from (some of) the logic it describes. Maybe can be simplified here to something like,
> // Back button has special logic for Bulk Services and Bed Nights household creation. It should return to service assignment and auto-fill the search box with the newly added household ID.

src/modules/household/components/HouseholdMemberTable.tsx
```diff
@@ -50,38 +35,59 @@ export const HOUSEHOLD_MEMBER_COLUMNS = {
...
       const isCurrentClient = h.enrollment.id === currentEnrollmentId;
-
       return (
         <ClientName
           client={h.client}
-          linkToProfile={!isCurrentClient && linkToProfile}
-          linkToEnrollmentId={
-            !isCurrentClient && !linkToProfile ? h.enrollment.id : undefined
-          }
+          linkToProfile={false}
```
martha: `linkToProfile` can be removed I think, it looks like it is false by default

src/modules/household/components/HouseholdMemberTable.tsx
```diff
@@ -50,38 +35,59 @@ export const HOUSEHOLD_MEMBER_COLUMNS = {
...
+          linkToEnrollmentId={!isCurrentClient ? h.enrollment.id : undefined}
           bold={isCurrentClient}
         />
       );
     },
   }),
+  entryDate: {
```
martha: Untested, but could we use `WITH_ENROLLMENT_COLUMNS.entryDate` and `.exitDate` here? Then we get tooltips on the dates which I think it is nice to use consistently

src/modules/household/components/HouseholdMemberTable.tsx
```diff
@@ -50,38 +35,59 @@ export const HOUSEHOLD_MEMBER_COLUMNS = {
...
+      </Tooltip>
+    ),
+    key: 'HoH',
+    render: (hc: HouseholdClientFieldsFragment) =>
+      hc.enrollment.relationshipToHoH ===
+        RelationshipToHoH.SelfHeadOfHousehold && <CheckIcon />,
+  },
   relationshipToHoh: {
-    header: 'Relationship to HoH',
+    header: 'Rel. to HoH',
```
martha: Could we use `ProjectHouseholdsTable ENROLLMENT_RELATIONSHIP_COL` here? That gets us consistent use of "HoH" instead of "Self (HoH)".

Also, `ENROLLMENT_RELATIONSHIP_COL` abbreviates the header to "Relationship" instead of "Rel. to HoH." I'm not sure which one I think is more clear but I do want to be consistent.

Maybe if we go with "Rel." we can abbreviate on mobile only (?)

src/modules/household/components/HouseholdMemberTable.tsx
```diff
@@ -50,38 +35,59 @@ export const HOUSEHOLD_MEMBER_COLUMNS = {
...
+      parseAndFormatDate(hc.enrollment.exitDate),
+  },
   enrollmentStatus: {
-    header: 'Status',
+    header: 'Enrollment Status',
     render: (hc: HouseholdClientFieldsFragment) => (
       <EnrollmentStatus enrollment={hc.enrollment} />
     ),
   },
+  hohCheck: {
```
martha, line 60: If it's low enough lift, could we also update the Enrollment Overview to use this HoH check, instead of the HoH indicator?

src/modules/household/components/HouseholdMemberTable.tsx
```diff
@@ -102,86 +108,88 @@ export const HOUSEHOLD_MEMBER_COLUMNS = {
...
-      <EnrollmentDateRangeWithStatus enrollment={hc.enrollment} />
-    ),
-  },
-  mciIds: externalIdColumn(ExternalIdentifierType.MciId, 'MCI ID'),
 };
 
+interface Props {
+  household: HouseholdFieldsFragment;
+  project: ManageHouseholdProject;
+  refetchHousehold: any;
```
martha: Looks like `refetchHousehold` can be typed as `VoidFunction` instead of `any` here

src/modules/household/hooks/useChangeHoh.tsx
```diff
@@ -0,0 +1,108 @@
...
+  // client to highlight for relationship input
+  const [highlight, setHighlight] = useState<string[]>([]);
+
+  const resetState = useCallback(() => {
+    setProposedHoh(null);
+    setErrors(emptyErrorState);
+    setWarningsConfirmed(false);
+  }, []);
+
+  const [setHoHMutate, { loading: hohChangeLoading }] =
```
martha, line 34: This is really minor but, in case it's easy to fix: I noticed that visually when changing the HoH, while the hhm table is reloading, there are 2 HoH checkmarks. I wonder if this might give the user the wrong impression that we add HoH status to another member before removing it from the original HoH. Of course in reality this happens in a transaction so, nothing to worry about.

src/modules/household/hooks/useChangeRelationshipToHoh.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  const handleConfirm = useCallback(() => {
+    if (!member || !relationship) return;
+    updateRelationship({
+      variables: {
+        input: {
+          enrollmentId: member.enrollment.id,
+          enrollmentLockVersion: member.enrollment.lockVersion,
+          relationshipToHoH: relationship,
+          confirmed: errorState.warnings.length > 0,
```
martha: Could you add a comment here explaining why we set confirmed to `errorState.warnings.length > 0`? Is it because this is in the `handleConfirm` callback, so if we had any warnings, the user has now confirmed them; but if there weren't any warnings then we pass `false` since there was nothing to confirm? Is it related to your comment above about handling validations even though the mutation doesn't return any?

src/modules/household/hooks/useHouseholdMenuActions.tsx
```diff
@@ -0,0 +1,250 @@
...
+
+      const manageHouseholdItems = canEdit
+        ? [
+            {
+              key: 'assign hoh',
+              title: 'Make Head of Household',
+              Icon: PersonIcon,
+              onClick: () => onChangeHoh(row),
+              ariaLabel: `Make ${clientBriefName(row.client)} the Head of Household`,
+              disabled: isSoleHoh,
```
martha, line 112: Should we add disabledReason for other disabled actions on HoH (make HoH and change relationship)?

Screenshot 2025-02-21 at 9 26 49 AM

martha: This predates your PR but I'm noticing we should also apply the disabled text more consistently:
- either "HoH" or "Head of Household", but not both
- maybe: either provide the instruction "Change HoH first" on all, or none(?)

src/modules/household/components/ManageHousehold.tsx
martha: Thanks hm yeah it looks like [maxWidth applied](https://github.com/greenriver/hmis-frontend/blob/release-153/src/components/elements/table/GenericTable.tsx#L125) here is not working. And that should probably be a % instead of a px value anyway. I tried updating it to 25% there, which seems to work well here, but doesn't work for e.g. Client Files. 

I will noodle on this a little more and create a ticket if I can't figure it out quickly.

martha (approved):
Thanks Gig this looks awesome overall! And so speedy 🚀

src/modules/household/components/ManageHousehold.tsx
martha: I updated https://github.com/open-path/Green-River/issues/7338 since I think we should address both issues with sticky cols at the same time.

I was seeing some pretty weird/unexpected behavior with width and maxWidth! :/

Anyway, I think the solution you have here is fine for now.

src/components/elements/CommonCard.tsx
gigxz: Thanks Martha I agree completely. I updated as follows so there are 2 components
* `CommonCard` is a card with an optional header, header can optionally be clickable. It also now has the option to _not_ pad content, so that we can use it for tables.
* `CommonCollapsibleCard` uses `CommonCard` internally

Added stories for both. CommonCard should replace usages of TitleCard everywhere, so I left a TODO, I think we still need to align on the header typography style there.

And I did confirm all current usages of CommonCard, there weren't that many.

src/components/elements/CommonCollapsibleCard.tsx
gigxz: Updated to just not render household id on mobile

src/modules/household/components/HouseholdMemberTable.tsx
gigxz: yes! thanks

src/modules/household/components/HouseholdMemberTable.tsx
gigxz: Good questions @martha  I was wondering that too. I'm going to remove this change and keep `Relationship to HoH`.

I think plain old `Relationship to HoH` can continue to be used in contexts where only **one household** is presented.

In situations where multiple households are displayed (project households table only) we abbrev "Relationship ~~to HoH~~" and "~~Self~~ HoH" to make things cleaner/clearer(?) because there is not _one hoh_ that the "relationship to hoh" header would be referring to.

src/modules/household/components/HouseholdMemberTable.tsx
gigxz: It wasn't part of the shove-stuff-in mocks, and would mean adding a new column header to the household overview table (which has very limited width). I have an open question about this in figma but for now I'm going to leave it

src/modules/household/hooks/useChangeHoh.tsx
gigxz: Thanks, thats a good point. There definitely would be ways to do it on the frontend (like: set the new hoh in state and pass it around; or write the updated relationships to the apollo cache) but the cleanest way, I think, would be to adjust the `UpdateRelationshipToHoH` mutation to return the full household, so relationship can be updated in the cache as soon as the mutation completes. I'm going to hold off on that to keep this to frontend-only changes.

src/modules/household/hooks/useChangeRelationshipToHoh.tsx
gigxz: Good callout. It was something along the lines of "If there are warnings present, that means we already performed the initial request, so pass `confirmed: true` to confirm the warnings and perform the operation." 

But that's not actually a good way to do it, because the user could have changed the relationship on the second request, which might have different warnings associated with it.

I'm just going to remove this. (So, reverting to old behavior where changing rel to hoh doesn't support displaying validations)

src/modules/household/hooks/useHouseholdMenuActions.tsx
gigxz: Good question. I want to run this by Ash, I don't think they've seen the new disabled label style (where they're embedded instead of tooltips).

The tooltip limitation is the same limitation that caused me to not be able to add the section labels (there should be labels after each divider). It's the fact that with MUI Menu each child has to be a MenuItem, and they can't be wrapped in divs or keyboard nav gets super wonky.

gigxz: Adding to design session

PR #1066, opened 2025-02-20
Add optional cols to all tables

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6924
Depends on: https://github.com/greenriver/hmis-warehouse/pull/5153

Summary of changes: 
- Large impact to diff, (should be) no impact to functionality: Require `key` prop on `ColumnDef`s and add it on all columns that didn't have it
- Improve ergonomics of applying optional columns with the new `GenericTableWithDataColumnDef`, as discussed [here](https://github.com/greenriver/hmis-frontend/pull/1067)
- Add more optional flags to Graphql queries for table data, such as `$includeOrganizationName`, `$includeMoveInDate` etc.
  - I haven't made flags for `exitDate` and `entryDate`, since even when these are not shown in optional columns, the data itself still needs to be included in the query: either for the row action aria text, or for the status column which is not optional.
- Update the tables in-scope in our [sheet](https://docs.google.com/spreadsheets/d/17ZpdB0fH6kUL14KDRWrlukwIH93BNx9grrNZApLpUwg/edit?gid=1987637083#gid=1987637083), as well as My Dashboard, with the correct optional columns.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/services/components/ClientServicesPage.tsx
```diff
@@ -40,6 +42,13 @@ const columns: ColumnDef<ServiceType>[] = [
     optional: true,
```
martha: Should Service Details now be removed? https://docs.google.com/spreadsheets/d/17ZpdB0fH6kUL14KDRWrlukwIH93BNx9grrNZApLpUwg/edit?disco=AAABeWYJLxU

src/modules/dataFetching/types.ts
```diff
@@ -52,3 +53,12 @@ export type FilterType<I> =
   | EnumFilter<I>
   | PickListFilter<I>
   | DateFilter<I>;
+
+export type GenericTableWithDataColumnDef<RowType, QueryVariables> = {
```
gigxz: Nice!
Maybe this name could be shortened a bit `DataColumnDef` or something?
And comment `"extend ColumnDef type to include additional attributes related to data fetching"`

src/modules/assessments/components/pages/ClientAssessmentsPage.tsx
```diff
@@ -82,6 +96,7 @@ const ClientAssessmentsPage = () => {
           noData='No assessments'
           recordType='Assessment'
           defaultSortOption={AssessmentSortOption.AssessmentDate}
+          showOptionalColumns
```
gigxz: What's the purpose of this prop? It seems like it would be easier if the optional column selection appeared by default if any of the columns have the `optional` attribute. But maybe I'm missing something. Are there cases where we pass optional column definitions, but we _don't_ want the selector to show up? I don't see any cases like that, but if there are, maybe having a `hideOptionalColumnSelection` prop would make more sense.

src/modules/projects/components/tables/ProjectClientEnrollmentsTable.tsx
```diff
@@ -40,65 +45,135 @@ const isHouseholdWithStaff = (
...
+    header: 'Last Contact Date',
+    key: 'lastContactDate',
+    optional: {
+      defaultHidden: true,
+      queryVariableField: 'includeLastContact',
+    },
+    render: (e) => {
+      if ('lastContact' in e && e.lastContact) {
+        return (
+          <Box flexDirection='row'>
```
gigxz: This is probably a design question, but just throwing it out there: in what situation does it make sense to wrap the cell content, vs not wrap the cell content and rely on horizontal scroll?
We did invest time into the horizontal scroll and sticky columns, so maybe it makes sense to remove the wrapping in some instances. Here `whiteSpace='nowrap'` would do that, imo it makes it easier to parse, and we know the content will never be super long (unlike client name).

src/modules/projects/components/tables/ProjectClientEnrollmentsTable.tsx
```diff
@@ -40,65 +45,135 @@ const isHouseholdWithStaff = (
...
+    render: (e) => {
+      if (e.moveInDate) {
+        return (
+          <DateWithRelativeTooltip
+            dateString={e.moveInDate}
+            preciseTime={false}
+          />
+        );
+      }
+    },
```
gigxz: nit: avoiding extra (one-char) enrollment var makes it more succinct
```suggestion
    render: ({ moveInDate }) =>
      moveInDate && (
        <DateWithRelativeTooltip dateString={moveInDate} preciseTime={false} />
      ),
```

src/modules/projects/components/tables/ProjectClientEnrollmentsTable.tsx
```diff
@@ -40,65 +45,135 @@ const isHouseholdWithStaff = (
   return 'household' in e && 'staffAssignments' in e.household;
 };
 
+type EnrollmentQueryVariables = Partial<{
+  includeStaffAssignment?: InputMaybe<Scalars['Boolean']['input']>;
+  includeMoveInDate?: InputMaybe<Scalars['Boolean']['input']>;
+  includeLastContact?: InputMaybe<Scalars['Boolean']['input']>;
+}>;
+
+const renderHouseholdAssignedStaff = (
```
gigxz: This seems complex enough to be in its own component file

src/modules/projects/components/tables/ProjectClientEnrollmentsTable.tsx
```diff
@@ -135,50 +225,75 @@ export const WITH_ENROLLMENT_COLUMNS: {
...
+  organizationName: {
+    header: 'Organization Name',
+    key: 'organizationName',
+    optional: {
+      defaultHidden: true,
+      queryVariableField: 'includeOrganizationName',
+    },
+    render: (e) => {
+      return e.enrollment.organizationName;
+    },
```
gigxz:
```suggestion
    render: ({ enrollment }) => enrollment?.organizationName,
```

src/modules/staffAssignment/components/MyClients.tsx
```diff
@@ -37,31 +39,65 @@ const MY_CLIENTS_COLUMNS: ColumnDef<StaffAssignmentWithClientsFragment>[] = [
...
-      const hoh = memoizedHoh(assignment);
-      return parseAndFormatDateRange(
-        hoh.enrollment.entryDate,
-        hoh.enrollment.exitDate
+      return renderCellContents(
+        memoizedHoh(assignment),
+        WITH_ENROLLMENT_COLUMNS.entryDate.render
       );
     },
+    tableCellProps: undefined, // typescript
```
gigxz: this seems odd, is there an underlying change we can make to the ColumnDef type, or some additional type hints here?

gigxz (approved):
looks great,  nice work!

src/modules/dataFetching/types.ts
martha: Thanks, yeah I was also thinking the name was a bit unwieldy! I like `DataColumnDef`

src/modules/assessments/components/pages/ClientAssessmentsPage.tsx
martha: This was an existing prop that I didn't change, just added it to the tables that newly have optional columns. But thinking about it more, I agree, it would be neater to get rid of it!

martha:
@gigxz I'll leave this open until you get back for one more review, since I did a pretty major refactor of where enrollment column definitions live. I think there is no rush anyway since release-154 won't go to stg until 3/6.

gigxz:
looks good to me @martha!

src/modules/dataFetching/components/GenericTableWithData.tsx
```diff
@@ -149,15 +142,17 @@ const GenericTableWithData = <
...
+  // Careful: Optional column behavior is undefined/unsupported when columns are provided by
+  // getColumnDefs instead of the columns prop. This is because `getColumnDefs` receives `rows`,
+  // but `rows` comes from the graphql query that receives `optionalQueryVariables`,
+  // leading to a circular dependency.
+  const {
+    optionalColumns,
+    includedOptionalColumns,
+    setIncludedOptionalColumns,
+    optionalQueryVariables,
+  } = useOptionalColumns<RowDataType, QueryVariables>({ columns: columnsProp });
```
gigxz, line 155: I had one more follow up on this while reviewing  https://github.com/greenriver/hmis-frontend/pull/1076:

I'm not sure I would understand what (not) to do as a developer when reading this comment. It seems like this hook should just do nothing if `columnsProp` is missing, right?

What happens if you pass a `getColumnDefs` function that returns a column that is specified as optional? Looking at the code down on line 271 I think optional columns would just always be hidden, is that right?

gigxz: It looks like `getColumnDefs` is only used in a handful of cases. Maybe we should look into replacing it, or at least capture future work of how to make it function with optional cols. Looking at the usages:
- BulkServicesTable - used in order to get loading prop. maybe that can be passed to the row render function instead
- EnrollmentCaseNotes - used in order to conditionally include CDE columns based on whether the page has any records with CDEs
- EnrollmentCLS - same as case notes usage
- ProjectExternalSubmissionsTable - similar to CDE use-case

martha: Thanks for looking into `getColumnDefs` usage, and sorry my comment was confusing! I can update it to be a little more clear, but yeah basically the "what to do as a developer" I had in mind when writing this comment was, "if you find yourself needing to use `getColumnDefs` with optional columns, it probably points to the necessity for refactoring `getColumnDefs`." I can open a tech debt ticket so we don't lose track of this.

martha: Created https://github.com/open-path/Green-River/issues/7387

PR #1067, opened 2025-02-20
Draft a suggestion for improving optional column ergonomics

martha:
## Description

This PR suggests an "ergonomic improvement" to the way Optional Columns are handled in GenericTable.

It builds on [this draft PR](https://github.com/greenriver/hmis-frontend/pull/1066), but I thought it would be easiest to discuss the right approach with a smaller diff, so I've opened this separately.

The idea is that the flag telling GraphQL whether to include a certain piece of data in the query lives on the `ColumnDef` that controls whether the data is shown. This allows `GenericTableWithData` to determine which query variables to pass depending on what optional columns are shown, rather than all tables with optional columns needing to define `applyOptionalColumns`.

See inline comments in the code for my concerns/parts that I want to discuss.

GH issue: https://github.com/open-path/Green-River/issues/6924
Warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5153

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/components/elements/table/types.tsx
```diff
@@ -18,6 +18,9 @@ type BaseColumnDef<T> = {
   optional?: boolean;
   defaultHidden?: boolean;
   sticky?: 'left' | 'right';
+  // If the column is optional, then you can optionally provide an optionalFieldFlag, which tells GenericTableWithData how to update the query variables.
+  // Discuss this idea. It's not ideal because ColumnDef is used by GenericTable, but this concept is specific to GenericTableWithData.
+  optionalFieldFlag?: string;
```
gigxz: I think this is a nice improvement! It is much cleaner and more succinct to not have to write out an `applyOptionalColumns` function every time. Thanks!

The fact that it's specific to GenericTableWithData is true for all the optional-column-related attributes. It would be helpful to comment the rest of them so we can better understand what they do. My attempt is something like:
```js
optional?: boolean; // whether this column is optional. if true, it appears in column selector
defaultHidden?: boolean; // whether this column is hidden by default. only makes sense if optional=true
optionalFieldFlag?: string; // name of gql query variables to pass `true` if col is enabled. usually var corresponds to an `include` directive: https://graphql-ruby.org/type_definitions/directives#runtime-directives
```
we could put them into an object, which might be nice especially if we may need to add additional context later on:
```js
  // configuration for making this column Optional. Only works with GenericTableWithData, will be ignored in GenericTable
  optionalColumn: {
    defaultHidden: boolean;
    queryVariableField?: string; // field to set on QueryVariables if col is included. optional if we want/need to keep `applyOptionalColumns` approach, otherwise we could make this required?
    queryVariableValue?: any; // value to set on QueryVariables if col is included (default: true) - we don't strictly need this yet, so could leave it out
  };
```

gigxz: Regarding the GenericTable/GenericTableWithData question - maybe we can try typing it so there is a `GenericTableWithDataColumnDef<RowType, QueryVariables>` that adds the optional col attribute(s), and can even type the query variable field  as a keyof QueryVariable (maybe?). It should still satisfy the normal `ColumnDef` type, so existing static column defs we have floating around should still be usable, I think.

PR #1075, opened 2025-02-24
Make collapsible card accessible

gigxz:
## Description

Make the CommonCollapsibleCard keyboard navigable and accessibly labeled.
Accessibility annotations mimic annotations on `SimpleAccordion`.

## Type of change
- [x] Bug fix

### DIFF
--- a/src/components/elements/CommonCollapsibleCard.tsx
+++ b/src/components/elements/CommonCollapsibleCard.tsx
@@ -1,5 +1,5 @@
 import { Box, Collapse, CollapseProps } from '@mui/material';
-import React from 'react';
+import React, { useId } from 'react';
 import CommonCard, { CommonCardProps } from '@/components/elements/CommonCard';
 import {
   ExpandLessIcon,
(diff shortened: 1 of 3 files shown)


src/modules/household/components/HouseholdMemberTable.tsx
```diff
@@ -65,8 +65,9 @@ export const HOUSEHOLD_MEMBER_COLUMNS = {
     ),
     key: 'HoH',
     render: (hc: HouseholdClientFieldsFragment) =>
+      // Decorative checkmark to indicate HoH. Aria-hidden because Relationship to HoH should be textually represented in another col.
       hc.enrollment.relationshipToHoH ===
-        RelationshipToHoH.SelfHeadOfHousehold && <CheckIcon />,
+        RelationshipToHoH.SelfHeadOfHousehold && <CheckIcon aria-hidden />,
```
martha: Hm, I think this is probably more confusing to screen readers since the HoH column is still part of the accessible dom, so if the checkmark is aria-hidden, all columns are "blank."

Maybe the _best_ screenreader experience would be to hide the whole HoH column, possibly with a new attribute to `ColumnDef`. Another possible ok workaround would be to have the column render function return some `customVisuallyHidden` content that says something more explicit like "Yes this client is HoH" or "This client is not HoH" -- so the screenreader experience navigating the table would be "Mary Jones - Head of Household - Yes this client is HoH"

This also reminds me that client names should probably be marked up as rowheaders. We get that for free if we make the column sticky, I think.

src/components/elements/CommonCollapsibleCard.tsx
```diff
@@ -31,22 +31,28 @@ const CommonCollapsibleCard: React.FC<CommonCollapsibleCardProps> = ({
...
+
   return (
     <CommonCard
       onClickHeader={onClick}
-      actions={open ? <ExpandLessIcon /> : <ExpandMoreIcon />}
+      actions={
+        open ? <ExpandLessIcon aria-hidden /> : <ExpandMoreIcon aria-hidden />
+      }
       padContent={false} // take over content padding, so there is no padding when collapsed
+      cardHeaderProps={{
```
martha, line 44: Cool 🚀 !

martha (approved):

src/modules/household/components/HouseholdMemberTable.tsx
gigxz: Thanks, good point. I implemented the suggestion of hidden text content in the cell: https://github.com/greenriver/hmis-frontend/pull/1075/commits/a4721588bc93b499abe503fc206fb3e18227a6ee

PR #1076, opened 2025-02-24
Save opt cols in search params and local storage

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6773

**Merge notes**
- Merge https://github.com/greenriver/hmis-frontend/pull/1066 to release branch first

**Summary of changes:**
- In `useOptionalColumns`, store optional column state in search params and in local storage.
- In `useSearchParamsState`, add an `initial` param and use an effect to set the initial search param state.
- In `ColumnsMenu` add a workaround for the above ^ which gets around the fact that the currently-shown columns value can now change outside of `ColumnsMenu`'s control.
- Separate the Project Household and Client enrollment pages into 2 separate routes. (following the example from the similar User Audit History pages)

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/hooks/useSearchParamState.ts
```diff
@@ -105,14 +109,44 @@ const getAllCurrentParams = (searchParams: URLSearchParams) => {
...
+ * @param applyDefaults if true, on load apply the default values from
+ * the paramsDefinition to the actual search params that appear the URL bar.
+ */
+const useSearchParamsState = (
+  paramsDefinition: SearchParamsStateType,
+  applyDefaults?: boolean
+) => {
   const [searchParams, setSearchParams] = useSearchParams();
 
+  // `emptyArrays` is a hack to get around the fact that in URL search params,
```
martha: Let's discuss this -- I am not sure I'm happy with the solution I've come up with here.

Here's how to reproduce the issue I'm trying to work around:
- Comment out line 69 of this file (remove my workaround)
- Go to a table with optional columns
- Select some columns
- Navigate away and navigate back to the table. The optional columns selected before should auto-populate (expected)
- Now try to remove the opt col selection, resetting all opt cols to hidden. **Bug:** It can't be removed. If there are no params saved for the `optionalColumn` search param value, then `useSearchParamsState` returns the default value, which is the saved selection, and not an empty array.

An interesting note: This is manifesting with the new opt col persistence, but I think it could be considered a "latent" bug. If we had any columns that were optional but shown by default (`defaultHidden: false`), then this bug would also manifest there. At the moment we don't have any such columns though.

src/modules/dataFetching/components/GenericTableWithData.tsx
```diff
@@ -150,14 +156,55 @@ const GenericTableWithData = <
...
+  // This is so that users can both
+  // - get the same view on their data every time they visit a table
+  // - and share/bookmark URLs that keep the same view into the data.
+  // We plan to eventually add other table controls, like sort/filter.
+  const [paramValues, setParamValues] = useSearchParamsState(
+    {
+      optionalColumns: {
+        type: 'string',
+        multiple: true,
+        default: defaultOptionalCols,
```
martha: This puts the column key in the URL, which can be a spaced string (from the header), or an underscored string (if column key is supplied explicitly). Is it important to clean this up?

Leads to urls that look like this:
`
/projects/:id/enrollments/households?optionalColumns=assigned_staff&optionalColumns=Last+Contact+Date
`

src/hooks/useSearchParamState.ts
martha: default vs. initial state

src/modules/dataFetching/components/GenericTableWithData.tsx
martha: Worth going through to make keys non-optional

src/components/elements/tableFilters/filters/ColumnsMenu.tsx
```diff
@@ -36,6 +36,12 @@ const TableColumnsMenu = ({
         setColumnsValue(defaultValue);
         reset();
       }}
+      // When the popover opens, set the state to whatever is currently in columnsValue.
+      // columnsValue may change on renders subsequent to the first one (see useOptionalColumns hook),
+      // so this menu component needs to "catch up" in order to display the up-to-date state!
+      onOpen={() => setState(columnsValue)}
```
martha: Here's a repro of the problem this is trying to solve:
- comment out this line (remove my workaround)
- go to a table and select some optional columns
- navigate away to another page
- navigate back to the same table, not using the back button, using in-app links
- the table will load correctly, and the search params will update to reflect your saved preference for optional columns, BUT the columns popover will not show any columns selected:
Screenshot 2025-02-25 at 4 01 13 PM

src/hooks/useSearchParamState.ts
```diff
@@ -105,9 +106,40 @@ const getAllCurrentParams = (searchParams: URLSearchParams) => {
...
+ * from https://github.com/jschwindt/react-use-search-params-state/tree/main
+ *
+ * @param paramsDefinition maps search param names to their definitions,
+ * indicating their type and default value
+ * @param initial initial values to set in the search params
+ */
+const useSearchParamsState = (
+  paramsDefinition: SearchParamsStateType,
+  initial?: URLSearchParamsInit
+) => {
```
gigxz: Suggest updating this hook to accept a typed object, so it's easier to understand what the params are from the calling components

src/modules/dataFetching/hooks/useOptionalColumns.tsx
```diff
@@ -9,42 +15,72 @@ export function useOptionalColumns<T extends { id: string }, QueryVariables>({
...
-  // (Internal to this hook) Which optional column defs are currently shown
-  const [currentOptColDefs, setCurrentOptColDefs] = useState<
-    DataColumnDef<T, QueryVariables>[]
-  >(
-    compact(optionalColumns.filter((col) => !col.optional?.defaultHidden) || [])
+  const currentPath = useCurrentPath();
+
+  // Optional columns to show when the page first loads
+  const initialOptionalColumns = useMemo(() => {
+    const storedOptionalCols = currentPath
```
gigxz: I'm looking at the path of what happens if `columns` is not passed, or if the column set doesn't have any optional columns. I think in that case we should skip hitting local storage, and just return early here.

martha:
Still to do: Need to handle the case where the user's session storage has an optional column saved, but the user opens a table where that column isn't applicable (example: assigned staff column is saved from one project, but user opens another project enrollments table where assigned staff config is not turned on)

src/modules/dataFetching/hooks/useOptionalColumns.tsx
```diff
@@ -9,42 +15,72 @@ export function useOptionalColumns<T extends { id: string }, QueryVariables>({
...
     const queryVariableFields = compact(
-      currentOptColDefs.map((c) => c.optional?.queryVariableField)
+      optionalColumns
+        .filter((c) => includedOptionalColumns.includes(c.key))
+        .map((c) => c.optional?.queryVariableField)
     );
+
     return queryVariableFields.reduce((acc: any, key) => {
       acc[key] = true;
       return acc as QueryVariables;
```
gigxz: Typescript seems happy without this cast, can we remove it?

src/modules/dataFetching/hooks/useOptionalColumns.tsx
gigxz: I think it would address https://github.com/greenriver/hmis-frontend/pull/1076#issuecomment-2701896505

martha: Hm yes you are definitely right we can skip hitting local storage if there are no optional columns.

I think https://github.com/greenriver/hmis-frontend/pull/1076#issuecomment-2701896505 still needs to be addressed separately. Say you are a user working in 2 projects, one with staff assignment and one without. Both projects' enrollment tables have optional columns, so returning early here will not affect either table.

src/components/elements/tableFilters/filters/ColumnsMenu.tsx
gigxz: Thanks for the explanation @martha. I see how this fixes it, but it only does load the column selection on a second render. Visually this is a bit jarring, because the page loads first with the "normal" view, then re-renders with the previous columns. Especially with a slower CPU (can see throttled with chrome dev tools) it's not ideal. I think I'd prefer to adjust this so that the table loads initially with the stored columns.

It seems like the underlying issue is that `includedOptionalColumns` (from the useOptionalColumns hook) is empty on first render, and then receives a value on the second render.
I **think** this is because of how `useSearchParamsState` handles the "initial" value. It first puts the initial value into search params, and _then_ reads them from the search params into state. I'm not totally sure on how it's working, but I see you added some comments into useSearchParamState about this. I think it's worth digging into; ergonomically it feels like unexpected behavior for that hook. Since we describe it as `"...like useState but stores the state in URL search params"` I would expect that if passing an initial value, that initial value would be immediately available (since that's how useState works).



comment on the param hook: https://github.com/greenriver/hmis-frontend/pull/1076#discussion_r1982134555

src/hooks/useSearchParamState.ts
```diff
@@ -105,9 +106,40 @@ const getAllCurrentParams = (searchParams: URLSearchParams) => {
...
+    if (Object.keys(toUpdate).length) {
+      // This unfortunately needs to cause an extra render when setting searchParams to the initial value
+      setSearchParams(toUpdate, { replace: true });
+    }
+
+    // Only run on mount, so that we don't prevent the user from overwriting the initial values.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
   const values = useMemo(
```
gigxz: Poking around at https://github.com/greenriver/hmis-frontend/pull/1076/files#r1982130317 , `values` here should ideally represent the "initial" value when this hook first renders. I didn't road-test this much, but here's how I was proof-of-concepting it (`mounted` as a boolean piece of state that got set in the useEffect)

```js
  const values = useMemo(() => {
    if (mounted) {
      return getValues(paramsDefinition, searchParams);
    }
    if (initial) {
      // if the above useEffect hasn't run yet, set the initial values right away
      return getValues(paramsDefinition, createSearchParams(initial));
    }
    return {};
  }, [mounted, paramsDefinition, searchParams, initial]);

```

martha: Thanks, cool! This seems to work well, I made a small modification to avoid a warning about "changing an uncontrolled input to be controlled"

src/components/elements/tableFilters/filters/ColumnsMenu.tsx
martha: Cool this can now be removed :)

src/hooks/useSearchParamState.ts
martha: Ok sounds good! https://github.com/greenriver/hmis-frontend/pull/1076/commits/e5ff83503c044e6fc50afba68c6eba709cae4f76 

I also wondered if it would be nice to put `initial` in `SearchParamStateType`, but I wasn't able to get this working. I could give it fresh eyes tomorrow if it sounds like a good idea to you.

src/hooks/useSearchParamState.ts
```diff
@@ -105,13 +110,56 @@ const getAllCurrentParams = (searchParams: URLSearchParams) => {
...
+
+    if (!initial) return;
+
+    const toUpdate: Record<string, any> = Object.fromEntries(
+      // Find any key/value pairs that are in initial, but not in searchParams
+      Object.entries(initial).filter(([key]) => !searchParams.has(key))
+    );
+    if (Object.keys(toUpdate).length) {
+      // Spread currentParams to avoid overwriting search params set by other components
+      const currentParams = getAllCurrentParams(searchParams);
```
martha, line 143: The longer I stare at this, the more I think there might be latent bugs with providing `initial` values for param types that are not strings. (Latent because we aren't currently doing this.) I'll check again tomorrow to see if I can get a bit more confident about the behavior. But I also think it might be nice to just merge this into 154 regardless, so that the existing behavior can be QA'd along with the other optional column work.

martha: Ok, after taking another look this morning, this and my other outstanding comment above are related: https://github.com/greenriver/hmis-frontend/pull/1076#discussion_r1982640872

- `initial` is typed as a `URLSearchParamsInit`, which enforces that initial values are strings
- If we were to instead remove the `initial` arg to the hook and add an `initial` prop alongside `default` as part of the `SearchParamStateType`, then we _would_ need to convert any non-strings into string values. (Maybe by adapting/reusing the code [here](https://github.com/greenriver/hmis-frontend/pull/1076/files#diff-730f4ec2e5f9d2f89a0a784e17cc19f00aeeb106526cb26c74af2275db1f4bc7R167-R187))
- I'm not sure that is desirable or more clear, plus am running out of time on this ticket, so I'd propose to merge this and leave any future improvements un-ticketed until the need arises.

gigxz: I guess I'm a little confused about the decision to type `initial` as a `URLSearchParamsInit`, as opposed to typing it as the same type as `value`. Can you explain a little bit?

Again, thinking that we want this hook to be analogous to `const [value, setValue] = useState(initial)`, then we would want the `initial` to have the same type as `value`.

For example:

```tsx
// Type for internal state representation of search params
type ValueType = Record<string, any>; // Future improvement: constrain value type
// Arg type for this hook
type Args = {
  paramsDefinition: SearchParamsStateType; // shape of params/state
  initial?: ValueType; // initial value
};
// Returns state and setState function
type ReturnType = [ValueType, (value: ValueType) => void];

const useSearchParamsState = ({
  paramsDefinition,
  initial,
}: Args): ReturnType => {
...
```

If the `initial` arg represented the initial values in the  'state' representation -- rather then the initial values in a URLSearchParamsInit representation -- then maybe we can also get rid of the transformation `getValues(paramsDefinition, createSearchParams(initial))` which converts `initial` into SearchParams and then back into "values".  Instead we could use `initial` directly, I think, since it's already the right shape. Does that make sense?


> If we were to instead remove the initial arg to the hook and add an initial prop alongside default as part of the SearchParamStateType

I'm not totally following this. API-wise, this doesn't seem desirable to me (again trying to mimic somewhat the useState API)

martha: Hm thank you for the feedback Gig. I added your suggestions https://github.com/greenriver/hmis-frontend/pull/1076/commits/666b8cf25615681449f6dd1fb61950c18898dc79, let me know if it looks ok to you.

> the decision to type initial as a URLSearchParamsInit, as opposed to typing it as the same type as value. Can you explain a little bit?

no reason, I'd just typed it that way in original implementation when I thought it would be passed in to the `useSearchParams` hook as the `initial` arg, but that was removed.

> I'm not totally following this. API-wise, this doesn't seem desirable to me (again trying to mimic somewhat the useState API)

just to explain my thinking a bit more, one thing I thought was messy about the API where usage looks like this:
```
  const [paramValues, setParamValues] = useSearchParamsState({
    paramsDefinition: {
      optionalColumns: {
        type: 'string',
        multiple: true,
        default: [],
      },
    },
    initial: {
      optionalColumns: initialOptionalColumns,
    },
  });
```
is that we separately declare `optionalColumns` on both objects passed to the hook. And I thought it would be better for usage to look more like this:
```
  const [paramValues, setParamValues] = useSearchParamsState({
    paramsDefinition: {
      optionalColumns: {
        type: 'string',
        multiple: true,
        default: [],
        initial: initialOptionalColumns
      },
    },
  });
```

Part of my slowness here was that by "mimicking the useState API" I got stuck on thinking you meant changing the hook to deal with one piece of state at a time, something more like
```
const useSearchParamsState<T> = (initial: T) => {
  ...
} 
```
and the usage would be more like
```
// in bulk services
const [coc, setCoc] = useSearchParamsState<string>();
const [serviceDate, setServiceDate] = useSearchParamsState<date>();
...

// or in optional columns land,
const [optionalColumns, setOptionalColumns] = useSearchParamsState<string[]>(initialOptionalColumns);
```
but eventually I figured out that is a bigger refactor than you are proposing here.

gigxz: Thanks @martha, sorry for my lack of clarity there. The type changes look good to me, can we remove the `createSearchParams` transformation or is that still necessary? Seems like another area that could be potentially bug-prone with other non-string types.

martha: No no worries at all, I'm sure it's me 😅!

Removing `createSearchParams` is a bit tricky since `getValues` accepts a `URLSearchParams` object. I think we could refactor so that it accepts either `UrlSearchParams` or a `Record<string, any>` but I'm not totally sure I see the benefit, since inside of all the `paramsToType` methods we would then need to check the type of `searchParams` every time it's used to figure out how to get what's inside it. That seems like it would add some cruft/lack of code clarity that I am a bit hesitant about.

gigxz: Right but the `initial` now has the same type as "value" (ValueType), so I'm wondering why we need to transform it to search params and then back to a value object? The code is:

```ts
    if (initial && !mounted) {
      // If the above useEffect hasn't run yet, return the initial values right away
      // to avoid flickering
      return getValues(paramsDefinition, createSearchParams(initial));
    }
```

and I'm wondering whether it can be
```diff
    if (initial && !mounted) {
      // If the above useEffect hasn't run yet, return the initial values right away
      // to avoid flickering
+      return initial;
    }
```

gigxz (approved):

PR #1077, opened 2025-02-24
Fix back to enrollment path from assessment, fix protected IDs bug

gigxz:
## Description

How to test Assessment linking issue:
* Create a new household in a project (route becomes `https://hmis.dev.test:5173/projects/95/add-household/360cfcb549d9449887a4f6e02a7b6a9c`
* Click "Intake Assessment" or "Household Intakes" (depending if single or multi household -- test both)
* Route become `https://hmis.dev.test:5173/client/58740/enrollments/60397/assessments/245027`
* Click "Back to [client name]"
* Should navigate to enrollment dashboard, not back to project

How to test url hashing issue:
* set `PUBLIC_PROTECTED_IDS='true'` in .env.development or .env.development.local
* test enrolling household, very broken because `householdId` is non-numeric, so the "safe" param functionality doesn't work for it. Change param name to `household` to avoid hashing. (By default any param ending in `id` gets hashed).

How to test not found bug
* navigate to bad url like `add-household/foo`
* it should show NotFound instead of an empty household management screen

## Type of change
- [x] Bug fix


gigxz:
This is up on QA already btw @martha!

src/modules/household/components/ManageHousehold.tsx
```diff
@@ -123,6 +129,9 @@ const ManageHousehold = ({
     [onSuccess, refetchHousehold]
   );
 
+  // could happen if user entered bad URL for /projects/:id/add-household/:householdId
+  if (householdNotFound) return <NotFound />;
```
martha, line 134: When the household is not found, the page flickers for a second before resolving on Page not found.

Untested, but would it be possible to avoid the flicker by moving what's now line 151 up here to before line 133? `if (householdId && !household && loading) return <Loading />`?

martha (approved):
Looks good, thank you Gig!

src/modules/household/components/ManageHousehold.tsx
gigxz: thanks good point, updating

PR #1079, opened 2025-02-25
Fix loading flicker on service form

gigxz:
## Description

QA fix for issue https://github.com/open-path/Green-River/issues/6043 

Bug repro 1 (no remote pick lists):
- add a new service to an enrollment
- select a custom service that uses a basic form that doesn't have any remote pick lists
- **bug**: you'll see the dialog flicker to a loading spinner state before rendering the form. the spinner is being rendered by DynamicFormEnrichedDataLoader.
- **updated behavior**: loading state does not occur at all, because the form has no pick lists to fetch

Bug repro 2 (yes remote pick lists):
- add a new service to an enrollment
- select a HUD service. the default HUD service form has remote pick lists. (tbh we should remove them and hard-code them since they don't _need_ to be remote, but that's for another day).
- **bug**: you'll see the dialog flicker to a loading spinner state while the pick lists are fetched. the spinner is being rendered by DynamicFormEnrichedDataLoader.
- **updated behavior**: the service dialog passes a `Skeleton` to DynamicFormEnrichedDataLoader to avoid the loading spinner flickering while the pick lists load.

## Type of change
- [x] Bug fix

### DIFF
--- a/src/modules/form/hooks/rhf/useEnrichedFormData.tsx
+++ b/src/modules/form/hooks/rhf/useEnrichedFormData.tsx
@@ -6,6 +6,7 @@ import usePreloadPicklists from '@/modules/form/hooks/usePreloadPicklists';
 import { LocalConstants, PickListArgs } from '@/modules/form/types';
 import {
   autofillValues,
+  formHasAnyRemotePicklists,
   getInitialValues,
   getItemMap,
 } from '@/modules/form/util/formUtil';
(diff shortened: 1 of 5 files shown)


src/modules/form/hooks/usePreloadPicklists.tsx
```diff
@@ -94,7 +90,10 @@ const usePreloadPicklists = ({
 
   return {
     fetch,
-    loading: loading && Object.keys(data).length === 0,
+    loading:
+      loading &&
+      pickListTypesToFetch.length > 0 && // always loading:false if there was no fetch
```
gigxz, line 95: I thought it would be nicer if the `loading` state defaulted to false, and only got set to true if a fetch started. However that didn't work as expected because of how this component is calling setState from a callback called by a useEffect... it's a different pattern from usual, I didn't want to touch it too much. This gives the same desired result, which is that forms with no pick lists never enter a `loading` state.

src/modules/form/hooks/rhf/useEnrichedFormData.tsx
```diff
@@ -60,9 +68,18 @@ export const useEnrichedFormData = <T extends FieldValues>({
...
   useEffect(() => {
-    if (loading || result) return;
+    // Form has no remote picklists; nothing to enrich
+    if (!hasRemotePicklists) return;
+    // Remote picklists are loading, wait
+    if (loading) return;
+    // Default values have already been enriched
+    if (!!enrichedDefaultValues) return;
+
     const mutation = cloneDeep(defaultValues);
```
martha: Not changed by your PR or related to this bugfix, so feel free to ignore, but, I found this var name `mutation` a bit confusing. `mutation` immediately makes me think graphQL, so it's just a bit of extra cognitive load to scan through the code and understand, oh right, there is no actual backend data mutation happening.

martha (approved):
Cool looks good to me, thanks for your attention to detail 🔎

PR #1080, opened 2025-02-25
Fix assessment print bug, remove adjustValue

gigxz:
## Description

QA fix for issue https://github.com/open-path/Green-River/issues/6043 

Bug repro: try to print a HUD intake assessment. Fields with remote pick lists don't appear, because DynamicViewField is fetching them and displaying a loading indicator.

Changes:
* Now that we have `useEnrichedFormData`, the pick list data is all loaded upfront. So we can rip out the code from DynamicViewField that re-fetches them.
* Remove  `adjustValue` (yay!!!) as it was **only** used for setting enriched pick list values.

Screenshot 2025-02-25 at 12 17 26 PM

Things tested:
- [x] all system tests pass
- [x] readonly autofill still works (inventory bed count example)

## Type of change
- [x] Bug fix

### DIFF
--- a/src/modules/form/components/viewable/DynamicViewFields.tsx
+++ b/src/modules/form/components/viewable/DynamicViewFields.tsx
@@ -1,11 +1,9 @@
 import { Grid } from '@mui/material';
-import React, { ReactNode, useCallback } from 'react';
+import React, { ReactNode } from 'react';
 
 import { FormDefinitionHandlers } from '../../hooks/useFormDefinitionHandlers';
 import {
-  ChangeType,
   FormValues,
-  ItemChangedFn,
   LocalConstants,
   OverrideableDynamicFieldProps,
   PickListArgs,
@@ -19,7 +17,6 @@ import { FormItem, ItemType } from '@/types/gqlTypes';
 export interface Props {
   horizontal?: boolean;
   visible?: boolean;
-  pickListArgs?: PickListArgs;
   handlers: FormDefinitionHandlers;
   nestingLevel: number;
   item: FormItem;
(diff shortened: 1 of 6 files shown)


src/modules/form/components/viewable/DynamicViewFields.tsx
```diff
@@ -113,21 +103,6 @@ const DynamicViewFields: React.FC<DynamicViewFieldsProps> = ({
   const definition = handlers.definition;
   const values = handlers.methods.getValues();
 
-  const itemChanged: ItemChangedFn = useCallback(
-    ({ linkId, value: baseValue, type }) => {
-      const item = handlers.itemMap[linkId];
-      let value = baseValue;
-      if (item) {
-        if (item.type === ItemType.Integer) value = parseInt(value) || 0;
-        if (item.type === ItemType.Currency) value = parseFloat(value) || 0;
```
gigxz, line 122: This is a little concerning because I don't know why this was here. I can pretty clearly trace that `itemChanged` was only used to adjust pick list option values, thought, so this number parsing code seems irrelevant.


Confirmed that other manipulations that occur on read-only views are still working correctly, like `autofillReadonly`, since those use the rhf-backed hooks now.

gigxz: @gigxz look at this again. are there other components that were getting `itemChanged` and using it, like DisabilityTable

martha (approved):
✂️ nice!

src/modules/form/components/viewable/DynamicViewFields.tsx
gigxz: looks like it was just duplicated from DynamicFormFields

gigxz:
thanks for the review!

PR #1091, opened 2025-03-06
6043-rhf-bugs

ttoomey:
## Description
* add memoization, field definitions must have stable references in rhf

Open any form with with a table.
* Before this fix, form would re-render in a tight loop rendering the form unusable and causing performance issues
* After: form should should only render once, should work as expected without performance issues

Depends on hmis-warehouse PR:

## Type of change
- [x] Bug fix

### DIFF
--- a/src/modules/form/components/group/TableGroup.tsx
+++ b/src/modules/form/components/group/TableGroup.tsx
@@ -73,7 +73,6 @@ const TableGroup = ({
       })
     );
   }, [item, viewOnly, baseId]);
-  console.log(tableHeaderInfo);
 
   // Memoize the props configuration for each cell to maintain stable references
   const cellConfigs = useMemo(() => {
--- a/src/modules/form/components/group/TableGroup.tsx
+++ b/src/modules/form/components/group/TableGroup.tsx
@@ -60,7 +60,7 @@ const TableGroup = ({
       item.item[0].type !== ItemType.Group ||
       !item.item[0].item
     ) {
-      throw new Error(`invalid table format form item ${item.linkId}`);
+      throw new Error(`Invalid table format for item ${item.linkId}`);
     }
 
     return item.item[0].item.map(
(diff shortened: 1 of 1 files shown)


src/modules/form/components/group/TableGroup.tsx
```diff
@@ -131,25 +156,7 @@ const TableGroup = ({
...
           </TableRow>
         </TableHead>
-        <TableBody>
-          {item.item?.map((rowItem) => (
-            <TableRow key={rowItem.linkId}>
-              {rowItem.item?.map((cellItem, index) => (
-                <TableCell key={cellItem.linkId}>
-                  {renderChildItem(
-                    // remove helper text, it is shown in table header
-                    { ...cellItem, helperText: null },
```
ttoomey, line 141: this was the line causing the problem, stabilizing cellItem using memoizing seems to fix the issue

gigxz: Nice find, thank you!
The intention I believe is to hide the Helper Text in the table cells, since they are pulled out and displayed in the table row. It looks like that behavior got lost with your change, I added it back https://github.com/greenriver/hmis-frontend/pull/1091/commits/b25d638d71fc5845debe2bf87408cf53d2907e9c using a deep-clone approach to keep a copy of the modified `item`s. It seems to work.



before (Storybook):
Screenshot 2025-03-07 at 7 55 19 AM

after: (back to original behavior)
Screenshot 2025-03-07 at 7 52 20 AM

gigxz (approved):
thanks!

src/modules/form/components/group/TableGroup.tsx
ttoomey: > The intention I believe is to hide the Helper Text in the table cells, since they are pulled out and displayed in the table row

Ah oops, sorry about that, thanks for fixing!

PR #1094, opened 2025-03-10
Fix initialSelected option list behavior

gigxz:
## Description

Fix behavior for items where the pick list options have an option where `initialSelected: true`. When initial selected is true, the option should be initially selected by default. This fixes it for both remote picklists and local picklists, though it looks like previously this behavior only worked for remote pick lists.

I moved the logic to the formUtil file so I could add unit tests.

- [x] deployed and confirmed reported bug in training environment is fixed

## How to test:
1. Set up an item with `pickListOptions` where one option has `initialSelected: true`. It should be initially selected when you open the form. It should not be populated when opening the form as readonly.
2. Open an item with a remote pick list that sets `initial_selected`. I have been using AvailableUnitsForEnrollment. Setup is to seed the ac forms; go to project with units; enroll one client; enroll second client and their unit should be pre-populated. This has already been deployed and tested in the training environment where the bug was reported, so its probably not necessary for the reviewer to set it up locally too.

Screenshot 2025-03-10 at 3 10 07 PM

## Type of change
- [x] Bug fix

### DIFF
--- a/src/modules/form/hooks/rhf/useEnrichedFormData.tsx
+++ b/src/modules/form/hooks/rhf/useEnrichedFormData.tsx
@@ -81,8 +81,6 @@ export const useEnrichedFormData = <T extends FieldValues>({
   >(undefined);
 
   useEffect(() => {
-    // Form has no remote picklists; nothing to enrich
-    if (!hasRemotePicklists) return;
     // Remote picklists are loading, wait
     if (loading) return;
     // Default values have already been enriched
@@ -118,7 +116,7 @@ export const useEnrichedFormData = <T extends FieldValues>({
   ]);
 
   return {
-    defaultValues: hasRemotePicklists ? enrichedDefaultValues : defaultValues,
+    defaultValues: enrichedDefaultValues,
     loading,
   };
 };
--- a/src/modules/form/hooks/rhf/useEnrichedFormData.tsx
+++ b/src/modules/form/hooks/rhf/useEnrichedFormData.tsx
@@ -3,20 +3,15 @@ import { cloneDeep } from 'lodash-es';
 import { useEffect, useMemo, useState } from 'react';
 import { DefaultValues, FieldValues } from 'react-hook-form';
 import usePreloadPicklists from '@/modules/form/hooks/usePreloadPicklists';
-import {
-  isPickListOption,
-  isPickListOptionArray,
-  LocalConstants,
-  PickListArgs,
-} from '@/modules/form/types';
+import { LocalConstants, PickListArgs } from '@/modules/form/types';
 import {
   autofillValues,
   formHasAnyRemotePicklists,
+  getEnrichedValueForChoiceItem,
   getInitialValues,
   getItemMap,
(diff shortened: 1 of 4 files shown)


src/modules/form/hooks/usePickList.tsx
```diff
@@ -18,57 +12,6 @@ import {
...
-    if (!setInitial) return;
-
-    const initial = item.repeats
-      ? data.pickList.filter((o) => o.initialSelected)
-      : data.pickList.find((o) => o.initialSelected);
-    if (initial) {
-      return { linkId, value: initial, type: ChangeType.System };
-    }
-    return;
-  }
```
gigxz, line 48: This was the previous behavior for applying "initial selected" which got lost

src/modules/form/util/formUtil.tsx
```diff
@@ -1543,3 +1544,85 @@ export function findOptionLabel(
...
 }
+
+// Given a CHOICE or OPEN_CHOICE item and an initial form value, provide an 'enriched' representation of the value.
+//
+// Returns an object with the following attributes:
+//  'enrichedValue' is the PickListOption representation, e.g. 'yes' => { code: 'yes', label: 'Yes', groupLabel: 'Something' }
+//  'initialSelectedValue' is the initialSelected option, if any (only present if the defaultValue was missing)
+//
+// Returns an empty object if the value does not need to be enriched
+export const getEnrichedValueForChoiceItem = ({
```
gigxz, line 1555: This is refactored from useEnrichedFormData. The only change is the addition of lines 1600-1609

src/modules/form/hooks/rhf/useEnrichedFormData.tsx
```diff
@@ -89,53 +89,22 @@ export const useEnrichedFormData = <T extends FieldValues>({
     if (!!enrichedDefaultValues) return;
 
     const clonedValues = cloneDeep(defaultValues);
-    Object.values(itemMap || {}).forEach(({ pickListReference, linkId }) => {
```
ttoomey, line 92: Looks like the initial value won't populate unless the form has at least one remote picklist due to early return on line 85.

ttoomey (approved):
testing locally:
⚠️  I couldn't get the initial values to display for inline picklist options (manually editing a form definition json to add initial_selected). My understanding is that this use-case doesn't exist but it might be good to handle it somehow
✅  for remote pick list references, the initial value to displays as expected (I added it by hacking the picklist resolver). It looks like the default value saves as expected. Changing the default also works

src/modules/form/hooks/rhf/useEnrichedFormData.tsx
gigxz: Ah thanks, good catch! I was testing it on a form that also had remote pick lists, whoops. Fixed. We still have the optimization of skipping the _fetch_ if there are no remote pick lists, which was the important part added in https://github.com/greenriver/hmis-frontend/pull/1079

PR #1096, opened 2025-03-10
Add CE workflow progression

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/7309
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5192

Summary of changes:

- Common components:
  - ~Add an icon to the title of `CommonCard`~
  - Refactor some Wayfinding work done for Join/Split so that it's reusable
  - Add a ~`CommonAppBar`~ `CommonStickyBar` component in order to share code and reduce duplication
  - Add the concept of a "full screen" route (could be renamed?) and handle it in `DashboardContentContainer`. This was previously special-cased in the Admin Dashboard, for the Form Builder only.
  - ~Add a new `gray` color for Alerts~
- Referrals:
  - Move referral routes, previously hanging out at the root, under the Project navigation
  - Add some basic shared referral "chrome," (`ReferralPage`), which isn't yet complete or pixel-matching to the mocks
  - Add a new page/route for viewing an individual referral step
  - Add mutations to start and submit referral steps
  - Add wayfinding when the step changes the overall referral's status

Testing instructions -- Here's what I've been doing:
- Run the "starter pack" locally, `rails driver:hmis:ce_starter_pack_20250302`. This creates some Templates, and Steps, one of which even has a form definition
- In the ui, go to the newly created CE test project, and create an Opportunity, pointing at one of the newly created Templates. ("One Task")
- If you go to the opportunity, you will see there are no candidates. At this point I rerun the starter pack, which creates a candidate pool (out of the first 100 candidates in my database) and runs the match engine
- Reload the opportunity and select one of the prioritized clients to activate a referral
- Step through the workflow progression

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/hooks/useDashboardState.tsx
```diff
@@ -38,9 +42,15 @@ export function useDashboardState() {
     setDesktopNavState(true);
   }, []);
 
+  const fullScreen = useMemo(
```
gigxz: Nice. Could you add some comment documentation explaining what `fullScreen` is and when to use it? My understanding from reading the code:
* `fullScreen` means that the DashboardContentContainer has no padding
* `focusMode` means that when the page is opened, the side nav will automatically collapse


I think the naming "full screen" is a little confusing because of how it could be interpreted to mean that the nav is collapsed and the content is "full screen". Maybe it can be called more plainly `disablePadding` or `noPadding`?

src/components/elements/CommonCard.tsx
```diff
@@ -26,6 +28,7 @@ export interface CommonCardProps {
  */
 const CommonCard: React.FC<CommonCardProps> = ({
   title,
+  Icon,
```
gigxz: I'm a little hesitant to add this based on early mocks, when an icon isn't present in the design library layout Card (which this is based on). Can you add a story that has the Icon prop, so we can follow up on this at some point?

src/modules/ce/components/ReferralNav.tsx
```diff
@@ -0,0 +1,150 @@
...
+import { ProjectDashboardRoutes } from '@/routes/routes';
+import {
+  CeReferralFieldsFragment,
+  useGetCeReferralQuery,
+} from '@/types/gqlTypes';
+import { generateSafePath } from '@/utils/pathEncoding';
+
+interface Props {}
+
+const ReferralNav: React.FC<Props> = ({}) => {
```
gigxz: suggest follow convention to  call this a "page" since it's a top level page. `ReferralPage` would make sense to me.

src/modules/ce/components/ReferralNav.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+  const currentPath = useCurrentPath();
+
+  if (loading) return <Loading />;
+  if (error) throw error;
+  if (!referral) return <NotFound />;
+
+  return (
+    <>
+      <CommonAppBar height={REFERRAL_HEADER_HEIGHT}>
```
gigxz: Reading this, it feels a bit weird to me to have three "AppBar" components on the page. The MUI AppBar by default would use a `header` component, but ours is just a div, so I think its just a question of naming. We could call it `CommonStickyBar` or something instead?
And/or maybe add a comment to the component explaining that it's used for several purposes any time we need a sticky bar, including the nav bar containing breadcrumbs, and sticky headers inside dashboard content

src/modules/ce/components/ReferralNav.tsx
```diff
@@ -0,0 +1,150 @@
...
+            })}
+            color='grayscale'
+            variant={
+              currentPath === '/referrals/:referralId' ? 'contained' : 'text'
+            }
+            startIcon={<DetailsIcon />}
+          >
+            Details
+          </ButtonLink>
+          <ButtonLink
```
gigxz:
```suggestion
          </ButtonLink>
          <Divider orientation='vertical' flexItem />
          <ButtonLink
```

src/modules/ce/components/ReferralNav.tsx
```diff
@@ -0,0 +1,150 @@
...
+              </Typography>
+            </Stack>
+            <Typography variant='body1'>
+              {referral.opportunity.projectName}
+            </Typography>
+          </Stack>
+        </Stack>
+      </CommonAppBar>
+      <CommonAppBar
+        top={STICKY_BAR_HEIGHT + CONTEXT_HEADER_HEIGHT + REFERRAL_HEADER_HEIGHT}
```
gigxz:
```suggestion
        top={STICKY_BAR_HEIGHT + CONTEXT_HEADER_HEIGHT + REFERRAL_HEADER_HEIGHT}
        sx={{ py: 1 }}
```

gigxz: (I know this isn't a design-focused iteration but couldnt help myself!)

src/modules/ce/components/ReferralNav.tsx
gigxz: also suggest adding header comment explaining that this uses react router outlet to render sub-pages

src/config/theme.ts
```diff
@@ -16,6 +16,7 @@ declare module '@mui/material/Alert' {
     low: true;
     medium: true;
     high: true;
+    gray: true;
```
gigxz: This seems good to add since there is a `grayscale` Alert in the new design library. Can you rename it to `grayscale` to match that?

src/config/theme.ts
```diff
@@ -511,6 +512,17 @@ const createThemeOptions = (theme: Theme) => ({
...
           }),
         },
+        {
+          // styles for gray alert
+          props: { color: 'gray' },
+          style: theme.unstable_sx({
+            borderColor: theme.palette.grayscale[300],
+            backgroundColor: theme.palette.grayscale.surface,
+            '& .MuiAlert-icon': {
+              color: theme.palette.grayscale.main,
```
gigxz: I'd like to move towards the more succinct language for specifying colors, `'grayscale.300'`, `'grayscale.surface'` etc

src/modules/ce/components/ReferralStepSummary.tsx
```diff
@@ -0,0 +1,134 @@
...
+  CeReferralStepStatus,
+  CeReferralStepSummaryFieldsFragment,
+  useStartCeReferralStepMutation,
+} from '@/types/gqlTypes';
+import { generateSafePath } from '@/utils/pathEncoding';
+
+interface Props {
+  step: CeReferralStepSummaryFieldsFragment;
+}
+const ReferralStepSummary: React.FC<Props> = ({ step }) => {
```
gigxz: Noting that this component has a mix of state/data/mutation logic and presentation logic.  Since this card has complex styling needs (varying colors/icon based on state) it seems like a good candidate to pull out into a stateless presentational component.  That would make this component easier to reason about.

```jsx
// container component
const ReferralStepCard: React.FC<Props> = ({ step }) => {
    // mutation here, define actions
   return (
       //presentational component
      <ReferralStepCardInner status={referral.status} action={action}>
        {/* card content */}
      </ReferralStepCardInner>
    )
});
```

src/modules/ce/components/ReferralStepSummary.tsx
```diff
@@ -0,0 +1,134 @@
...
+        return {};
+    }
+  }, [status]);
+
+  const [startStepMutation, { loading, error }] =
+    useStartCeReferralStepMutation({
+      variables: {
+        referralId: referral.id,
+        stepId: step.stepId || '',
+      },
```
gigxz: I noticed that after clicking "start" the app performs mutation `StartCeReferralStep` and then immediately performs query `GetCeReferralStep` and shows a second loading state. That seems like it can be cleaned up, we should be able to use apollo cache since  StartCeReferralStep returned the exact same fragment.

src/modules/ce/components/ReferralStep.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  return (
+    <CommonCard title={name}>
+      <Stack gap={2}>
+        <ReferralStepAssignee step={step} />
+        <Divider />
+        {status === CeReferralStepStatus.InProgress && (
+          <DynamicForm
+            definition={formDefinition.definition}
+            onSubmit={(event) => {
```
gigxz: Getting the following error when I try to submit without filling in required fields:


```
/app/drivers/hmis/app/models/hmis/workflow_execution/engine.rb:101:in `traverse_node'
/app/drivers/hmis/app/models/hmis/workflow_execution/engine.rb:114:in `visit_node'
/app/drivers/hmis/app/models/hmis/workflow_execution/engine.rb:104:in `block in traverse_node'
/app/drivers/hmis/app/models/hmis/workflow_execution/engine.rb:104:in `each'
/app/drivers/hmis/app/models/hmis/workflow_execution/engine.rb:104:in `traverse_node'
/app/drivers/hmis/app/models/hmis/workflow_execution/engine.rb:40:in `complete_step!'
/app/drivers/hmis/app/graphql/mutations/ce/submit_ce_referral_step.rb:30:in `block in resolve'
/bundle/ruby/3.3.0/gems/activerecord-7.0.8.7/lib/active_record/locking/pessimistic.rb:94:in `block in with_lock'
/bundle/ruby/3.3.0/gems/activerecord-7.0.8.7/lib/active_record/connection_adapters/abstract/transaction.rb:319:in `block in within_new_transaction'
/bundle/ruby/3.3.0/gems/activesupport-7.0.8.7/lib/active_support/concurrency/load_interlock_aware_monitor.rb:25:in `handle_interrupt'
/bundle/ruby/3.3.0/gems/activesupport-7.0.8.7/lib/active_support/concurrency/load_interlock_aware_monitor.rb:25:in `block in synchronize'
/bundle/ruby/3.3.0/gems/activesupport-7.0.8.7/lib/active_support/concurrency/load_interlock_aware_monitor.rb:21:in `handle_interrupt'
/bundle/ruby/3.3.0/gems/activesupport-7.0.8.7/lib/active_support/concurrency/load_interlock_aware_monitor.rb:21:in `synchronize'
/bundle/ruby/3.3.0/gems/activerecord-7.0.8.7/lib/active_record/connection_adapters/abstract/transaction.rb:317:in `within_new_transaction'
/bundle/ruby/3.3.0/gems/activerecord-7.0.8.7/lib/active_record/connection_adapters/abstract/database_statements.rb:316:in `transaction'
/bundle/ruby/3.3.0/gems/activerecord-7.0.8.7/lib/active_record/transactions.rb:209:in `transaction'
/bundle/ruby/3.3.0/gems/activerecord-7.0.8.7/lib/active_record/transactions.rb:290:in `transaction'
/bundle/ruby/3.3.0/gems/activerecord-7.0.8.7/lib/active_record/locking/pessimistic.rb:92:in `with_lock'
/app/drivers/hmis/app/graphql/mutations/ce/submit_ce_referral_step.rb:27:in `resolve'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/schema/resolver.rb:121:in `public_send'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/schema/resolver.rb:121:in `call_resolve'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/schema/mutation.rb:69:in `call_resolve'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/schema/resolver.rb:108:in `block (3 levels) in resolve_with_support'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/execution/interpreter/runtime.rb:751:in `minimal_after_lazy'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/query.rb:424:in `after_lazy'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/schema/resolver.rb:96:in `block (2 levels) in resolve_with_support'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/execution/interpreter/runtime.rb:751:in `minimal_after_lazy'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/query.rb:424:in `after_lazy'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/schema/resolver.rb:87:in `block in resolve_with_support'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/execution/interpreter/runtime.rb:751:in `minimal_after_lazy'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/query.rb:424:in `after_lazy'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/schema/resolver.rb:75:in `resolve_with_support'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/schema/field.rb:736:in `public_send'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/schema/field.rb:736:in `block (2 levels) in resolve'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/schema/field.rb:870:in `with_extensions'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/schema/field.rb:707:in `block in resolve'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/execution/interpreter/runtime.rb:751:in `minimal_after_lazy'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/query.rb:424:in `after_lazy'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/schema/field.rb:705:in `resolve'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/execution/interpreter/runtime.rb:375:in `block (2 levels) in evaluate_selection_with_resolved_keyword_args'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/tracing/trace.rb:52:in `execute_field'
/app/drivers/hmis/app/graphql/concerns/graphql_trace_behavior.rb:9:in `execute_field'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/execution/interpreter/runtime.rb:374:in `block in evaluate_selection_with_resolved_keyword_args'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/execution/interpreter/runtime.rb:696:in `call_method_on_directives'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/execution/interpreter/runtime.rb:363:in `evaluate_selection_with_resolved_keyword_args'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/execution/interpreter/runtime.rb:341:in `block in evaluate_selection_with_args'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/execution/interpreter/runtime.rb:810:in `after_lazy'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/execution/interpreter/runtime.rb:288:in `evaluate_selection_with_args'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/execution/interpreter/runtime.rb:282:in `block in evaluate_selection'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/execution/interpreter/arguments_cache.rb:46:in `block in dataload_for'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/schema/member/has_arguments.rb:272:in `block (3 levels) in coerce_arguments'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/dataloader.rb:276:in `block in spawn_job_fiber'
/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/dataloader.rb:248:in `block in spawn_fiber'
```

src/modules/ce/components/Opportunity.tsx
```diff
@@ -10,7 +10,7 @@ import NotFound from '@/components/pages/NotFound';
 import useSafeParams from '@/hooks/useSafeParams';
 import ActivateReferralButton from '@/modules/ce/components/ActivateReferralButton';
 import GenericTableWithData from '@/modules/dataFetching/components/GenericTableWithData';
-import { ClientDashboardRoutes, Routes } from '@/routes/routes';
+import { ClientDashboardRoutes, ProjectDashboardRoutes } from '@/routes/routes';
 import {
```
gigxz, line 14: Not sure where to put this comment: after denying a referral I see the opportunity status is "locked". I don't think that's quite right, my understanding is the opportunity should be immediately released into "open" state after a referral is fully denied, but we can discuss. 


Screenshot 2025-03-14 at 11 53 51 AM

src/modules/ce/components/ReferralStep.tsx
gigxz: Nit renaming "event", since its not an event object it is a custom `DynamicFormSubmitInput` record

gigxz (commented):
looks great!!

src/hooks/useDashboardState.tsx
martha: Yeah I agree, I had reservations about the naming `fullScreen` too. I like your suggestion of `noPadding`. Thank you!

src/components/elements/CommonCard.tsx
martha: Sure sounds good, I added a story and also a note [here](https://github.com/open-path/Green-River/issues/7372#issue-2885383284) so we don't forget to revisit.

Would you prefer to just remove this for now? That also seems reasonable to me. We can instead just inject the icon into the title on `ReferralStepSummary`, without making any changes to `CommonCard`.

src/modules/ce/components/ReferralNav.tsx
martha: Sounds good! I wasn't sure about Page since it felt more like navigational chrome than a page in its own right. I was thinking about calling it `ReferralDashboard` since it seems a little similar to the Project, Client, etc. dashboards, but hesitated based on our language guide's suggestion to apply the word "dashboard" less leniently ([thread](https://www.figma.com/design/bsw8CX4UOq6SIdXYorZDWV?node-id=3081-241#1130106854)). That's how I landed on Nav, but if you like Page I will go with that

src/modules/ce/components/ReferralNav.tsx
martha: Oh yeah completely agree, I like `CommonStickyBar` much better! Thanks!

src/modules/ce/components/ReferralStepSummary.tsx
martha: Hrm! Yes I agree that this should be clean-up-able, but I'm stuck. I will leave it for now and try to come back Monday morning with a fresh brain. A few notes:
- the reason for the second query `GetCeReferralStep` is because we load another page for `ReferralStep`. So while the step is already in the cache, it is cached as a `CeReferralStep` object, but not cached as a result to the `GetCeReferralStep` query.
  - I tried calling `cache.writeQuery` in the `onComplete` of the start-step mutation, to directly save this step as a result of the `GetCeReferralStep` query, but didn't get this working. It's also not a pattern we use anywhere else in our app so I'm a bit hesitant about it
- I was wondering whether part of the problem was the unusual ID situation on referral steps (they are not ID'd in the cache by the same ID that we use to query them - this is because they can be returned to the frontend when they are still unpersisted ([see schema](https://github.com/greenriver/hmis-warehouse/pull/5013/files#diff-9a2e4184fa85c9a6f6a65383e8986f26ad77ff3c51716c6cb6a66cced4421b4dR22-R25)). But I tried adjusting that logic temporarily and it didn't seem to help with anything

martha: Edit, of course as soon as I finished typing my notes and then returned to it once more, I found the typo in my `cache.writeQuery` approach, so I think this is working as intended now. I'd still like to discuss it and make sure it's acceptable to introduce this new pattern.

src/modules/ce/components/ReferralStep.tsx
martha: Thanks, yes! Sorry, I had been working on this in a separate branch and thought I'd already put up a PR. I just opened https://github.com/greenriver/hmis-frontend/pull/1100 which depends on the backend PR https://github.com/greenriver/hmis-warehouse/pull/5195

src/modules/ce/components/Opportunity.tsx
martha: Thanks! This is a backend issue, I've opened another PR https://github.com/greenriver/hmis-warehouse/pull/5210

src/modules/ce/components/ReferralNav.tsx
gigxz: I would say it is a Page because it is rendered into the Project Dashboard by the react router. But I see what you mean, we could come up with some other word to describe the navigational element. I don't love the idea of calling it a Dashboard because we end up with nested dashboards

src/config/theme.ts
gigxz: I followed up with Bobby on the design language about the introduction of this "grayscale alert" and we landed on _not_ adding it since it doesn't fit into the MUI severity paradigm https://www.figma.com/design/bsw8CX4UOq6SIdXYorZDWV?node-id=2482-931#1173966176

So suggest removing this altogether!

src/modules/ce/components/ReferralWayfinder.tsx
```diff
@@ -0,0 +1,79 @@
...
+      return (
+        <WayfindingDialog
+          open={open}
+          onClose={onClose}
+          title='Referral Complete'
+          alertTitle='Referral Declined'
+          alertText={`${clientBriefName(client)} has been declined from ${opportunity.name}`}
+          AlertProps={{
+            icon: <DeclinedIcon />,
+            color: 'grayscale',
```
gigxz:
```suggestion
            severity: 'info',
```

https://github.com/greenriver/hmis-frontend/pull/1096#discussion_r1999340114

src/modules/ce/components/ReferralStepCard.tsx
```diff
@@ -0,0 +1,124 @@
...
+        stepId: step.stepId || '',
+      },
+      onCompleted: (data) => {
+        if (data.startCeReferralStep?.step) {
+          const step = data.startCeReferralStep?.step;
+
+          // The step returned from the mutation is now auto added to the Apollo cache.
+          // Here we are writing it to the cache specifically for the GetCeReferralStepDocument query,
+          // so that when we navigate to the Step page, we don't do a double-fetch for the data we already have.
+          cache.writeQuery({
```
gigxz, line 47: The explicit write is a bit of a code smell to me, but I am guessing it's needed because `CeReferralStep.id` is the cache key but `CeReferralStep.stepId` is the id we're using to fetch.. a little funky but seems ok, we might need to come back to those cache keys in the future if there are issues

src/modules/ce/components/ReferralStep.tsx
gigxz: all good! great!

src/modules/ce/components/ReferralStepSummary.tsx
gigxz: Hmm thanks. Yeah I do think the explicit write is a bit of a code smell, but I am guessing it's needed because of the IDs, like you point out. `CeReferralStep.id` is the cache key but `CeReferralStep.stepId` is the id we're using to fetch. If we _knew_ the cache key then we could just call `apolloClient.readFragment` to see if we have it, but we don't know the cache key. I think what you added makes sense, and we might need to come back to those cache keys in the future if there are issues.

gigxz (approved):

src/components/elements/CommonCard.tsx
martha: Based on our conversation I ended up just removing it, we can add it back later if need be!

src/modules/ce/components/ReferralStepSummary.tsx
martha: Ok, thanks! I'll add some more comments here

PR #1098, opened 2025-03-13
Patch render bug in ProjectReferralPostingForm

gigxz:
## Description

Patch for bug that was introduced in 153. The way `initialValues` are treated changed with the RHF refactor. I also audited the codebases for other locations where we may have been expecting a change to initialValues; I couldn't find any.

Issue: https://github.com/open-path/Green-River/issues/7415

How to reproduce:
* open any "incoming" assigned referral
* update the note and hit save
* the value saves, but the form appears empty. refreshing the page makes the updated note appear.

## Type of change
- [x] Bug fix

### DIFF
--- a/src/modules/referrals/components/ProjectReferralPostingForm.tsx
+++ b/src/modules/referrals/components/ProjectReferralPostingForm.tsx
@@ -97,6 +97,10 @@ export const ProjectReferralPostingForm: React.FC<Props> = ({
 
   return (
     <DynamicForm
+      // Use the key to force a re-render when the referral posting changes.
+      // Note: the change to initialValues won't not trigger a form re-render,
+      // because useEnrichedFormData doesn't recalculate defaultValues.
+      key={JSON.stringify(referralPosting)}
       definition={definition}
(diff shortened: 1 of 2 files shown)


src/modules/referrals/components/ProjectReferralPostingForm.tsx
```diff
@@ -97,6 +97,10 @@ export const ProjectReferralPostingForm: React.FC<Props> = ({
 
   return (
     <DynamicForm
+      // Use the key to force a re-render when the referral posting changes.
+      // Note: the change to initialValues won't not trigger a form re-render,
```
martha: Maybe typo, should it be `change to initialValues won't trigger a form re-render`?

martha (approved):
Looks good, thank you Gig!

src/modules/referrals/components/ProjectReferralPostingForm.tsx
gigxz: whoops thanks

PR #1101, opened 2025-03-19
Implement basic Opportunity page

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/7322
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5218

Summary of changes:
- Add a new `CommonTabs` component that wraps the MUI Tabs
- Make some improvements to the `BeginReferralButton` (but see outstanding [thread](https://www.figma.com/board/axGI9vwmuBRC4uZ11ZJ3RP?node-id=2697-73498#1173994790))
- Add details to the Opportunity page about which match rules apply
- Refactor `PrioritizedClientsTable` out of `Opportunity` for clarity
- Add `OpportunityBanner`, meant to encapsulate some of the styles [here](https://www.figma.com/design/znYqGn21VZO5zF2LjpnUiL/OP-HMIS-%2F-Coordinated-Entry-%2F-Figma?node-id=3558-27348&t=9w7vL66doIA4FGJ4-0), although it is not pixel perfect yet. I'm open to feedback/thoughts about taking the time to more thoroughly match the mocks at this point!
- Make some also-not-pixel-perfect updates to `ReferralStatusChip`, 

How to test: 
- Run the "starter pack" locally, and see [this caveat](https://github.com/open-path/Green-River/issues/7396#issuecomment-2734633474) if you run into issues
- In the ui, go to the CE test project, and create some Opportunities
- Run the starter pack again to create candidate pools
- Check out the Opportunity page. Step the Opportunity through various statuses to see how it looks in different states

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/src/components/elements/CommonTabs.tsx
+++ b/src/components/elements/CommonTabs.tsx
@@ -15,7 +15,7 @@ interface CommonTabsProps {
 }
 
 // CommonTabs wraps the MUI Tabs component. It can be controlled or uncontrolled,
-// so callers can choose whether to delegate of which tab is selected.
+// so callers can choose whether to delegate controlling the state of which tab is selected.
 // This is a rare case where ariaLabel is a required prop, otherwise the tabs won't be accessible.
 const CommonTabs: React.FC<CommonTabsProps> = ({
   sx,
--- a/src/components/elements/CommonTabs.tsx
+++ b/src/components/elements/CommonTabs.tsx
@@ -1,0 +1,74 @@
+import { Box, SxProps, Tab, Tabs } from '@mui/material';
+import React, { ReactNode, useState } from 'react';
+
+export type TabDefinition = {
+  title: string;
+  contents: ReactNode;
+};
+
+interface CommonTabsProps {
+  tabDefinitions: TabDefinition[];
+  ariaLabel: string;
+  sx?: SxProps;
+  currentTab?: number;
+  onChangeTab?: (tab: number) => void;
+}
+
+// CommonTabs wraps the MUI Tabs component, which requires a lot of boilerplate,
+// allowing the caller to provide a TabDefinition and abstracting away handling
+// of which tab value is currently selected.
+// This is a rare case where ariaLabel is a required prop, otherwise the tabs won't be accessible.
+const CommonTabs: React.FC<CommonTabsProps> = ({
+  sx,
+  ariaLabel,
+  tabDefinitions,
+  currentTab,
+  onChangeTab,
+}) => {
+  const [internalValue, setInternalValue] = useState(0);
+
+  // Use external value if provided; otherwise, use internal state
+  const currentValue = currentTab ?? internalValue;
+
+  const handleChange = (_event: React.SyntheticEvent, newValue: number) => {
+    if (onChangeTab) {
+      onChangeTab(newValue); // Controlled mode: notify parent
+    } else {
+      setInternalValue(newValue); // Uncontrolled mode: manage internally
+    }
+  };
+
+  return (
+    <Box sx={{ width: '100%', ...sx }}>
+      <Box sx={{ borderBottom: 1, borderColor: 'divider' }}>
+        <Tabs
+          value={currentValue}
+          onChange={handleChange}
+          aria-label={ariaLabel}
+        >
+          {tabDefinitions.map((t, i) => (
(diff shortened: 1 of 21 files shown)


src/modules/ce/components/CreateOpportunityButton.tsx
```diff
@@ -20,8 +20,8 @@ interface Props {
 }
 const CreateOpportunityButton: React.FC<Props> = ({ projectId }) => {
   const [open, setOpen] = useState(false);
-  const [name, setName] = useState<string | undefined>(undefined);
-  const [templateId, setTemplateId] = useState<string | undefined>(undefined);
+  const [name, setName] = useState<string | null>(null);
```
martha, line 23: This is fixing a console warning about component changing from uncontrolled to controlled

src/modules/ce/components/Opportunity.tsx
```diff
@@ -50,81 +32,68 @@ const Opportunity: React.FC<Props> = ({}) => {
...
+                  />
+                </Grid>
+              </Grid>
+            ),
+          },
+          {
+            title: 'Eligible Clients',
+            contents: (
+              <Paper>
+                {/*todo @martha - discuss showing current referral and status here */}
```
martha, line 111: One thing from the [mocks](https://www.figma.com/board/axGI9vwmuBRC4uZ11ZJ3RP/OP-HMIS-%2F-Coordinated-Entry-%2F-FigJam?node-id=2481-81423&t=iuUruj0pOsv7VX0j-0) that I haven't addressed at all: for in-progress and completed referrals, displaying the "eligible clients" list with that referral and its special status floated to the top.

I skipped it for now, since it introduces some complexity beyond just relying on `GenericTableWithData`, and the mocks are not finalized yet. Let me know if you think that sounds good, or I should put a little more time into it.

src/modules/ce/components/OpportunityBanner.tsx
```diff
@@ -0,0 +1,129 @@
...
+          </Button>
+        )}
+      </Stack>
+      <Paper
+        sx={{
+          backgroundColor: acceptedReferral ? undefined : 'primary.surface',
+          p: 2,
+        }}
+      >
+        <Stack direction='row' justifyContent='space-between'>
```
martha, line 114: I think could be pretty easily refactored further into a `<ClientCard/>`, but I haven't done so for now since the only place it's used so far is here.

src/components/elements/CommonTabs.tsx
```diff
@@ -0,0 +1,71 @@
...
+  ariaLabel: string;
+  sx?: SxProps;
+  currentTab?: number;
+  onChangeTab?: (tab: number) => void;
+}
+
+// CommonTabs wraps the MUI Tabs component. It can be controlled or uncontrolled,
+// so callers can choose whether to delegate controlling the state of which tab is selected.
+// This is a rare case where ariaLabel is a required prop, otherwise the tabs won't be accessible.
+const CommonTabs: React.FC<CommonTabsProps> = ({
```
gigxz, line 20: I didn't think to spec this before, but I do think we will want to adjust this at some point so it can support using hash params so that you can navigate directly to a tab (and have the current tab reflected in the URL), like `/projects/933/opportunities/4#overview` vs `/projects/933/opportunities/4#clients`. We have a implementation of that in the household assessments tab (maybe weird implementation, shouldn't necessarily copy).

Especially looking at the "referral queue" page which _also_ consists of tabbed tables, I think we will want this. But happy to defer to a separate ticket! (maybe it can be done as part of https://github.com/open-path/Green-River/issues/7350)

gigxz, line 20: Can you add a basic Story for this? thanks!

martha: Yeah I was thinking this too. I like the idea of including it in 7350's scope. I'll make a note on that ticket.

martha: Oh, this is a lot easier than I thought; we can just use `href`. No need for a fancy `useHashParamsState` hook or anything like that.

I think in this case I will take the time to split this PR out into 2 PRs, one with the CommonTabs implementation and the other with the Opportunity Page changes.

PR #1102, opened 2025-03-20
Prevent silent failure on assessment when key is not resolved

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/6868
Addresses part (3) from the linked issue:
> Throw an error if the record doesn't have that attribute at all, because that means we made a mistake and didn't resolve the field.

How to test:
- Locally, make a schema change that fails to resolve a field. I commented out `viralLoadSource` [here](https://github.com/greenriver/hmis-frontend/blob/0a5b76269a9dd36fd887f8dc9fd9c5e9db24b21f/src/api/operations/assessmentRecords.fragments.graphql#L131). 
- Try to load an assessment that will display that field. (For `viralLoadSource` this means an assessment where “Viral Load Information Available” has been answered “Available”)
- Before: The field would silently fail to load
- Now: You should see an error

## Type of change
- [x] Bug fix


src/modules/form/util/formUtil.tsx
```diff
@@ -1290,7 +1290,36 @@ const getMappedValue = (record: any, mapping: FieldMapping) => {
...
-    return get(record, compact([relatedRecordAttribute, mapping.fieldName]));
+    const keys = compact([relatedRecordAttribute, mapping.fieldName]); // for example: ['disabilityGroup', 'viralLoadSource']
+
+    // FIXME - do these special cases make sense or should they be fixed instead of carved out here?
+    // Special cases where we DON'T want to raise an error if the field can't be resolved on the record:
+    // 1. The only key in `record` is `enrollment`. This happens in AssessmentForm
+    // around line 212 when we are creating a new assessment and populating it from the enrollment, but other
+    // fields that are normally resolved on an Assessment, like incomeBenefit, disabilityGroup, etc. are not resolved.
+    const recordKeys = Object.keys(record);
+    if (recordKeys.length === 1 && recordKeys[0] === 'enrollment') {
```
martha: Notes from dev standup:
- add a flag `dontRaise` to the method
- see custom fields that we process onto something else - like MCI
- avoid huge investment if we are moving into backend
- replace error with reporting to Sentry

src/modules/errors/util.ts
```diff
@@ -76,3 +77,10 @@ export class NotFoundError extends Error {
     Object.setPrototypeOf(this, new.target.prototype);
   }
 }
+
+export const sendToSentry = (message: string) => {
+  if (['test', 'development'].includes(import.meta.env.MODE)) {
+    console.error(message);
```
martha, line 83: To discuss: Should this actually just throw an error in dev (stronger than just `console.error`) to make *sure* we don't miss it?

martha: (This is lightly modified from the former private method `handleError` in `useEnrichedFormData.tsx`. I moved it here to be used in common, but if the behavior should actually diverge, I can revert that change.)

gigxz: I think this is ok. Once we address https://github.com/open-path/Green-River/issues/7051 we should have better practices around checking for console errors and having system tests fail on them.

src/modules/form/util/formUtil.tsx
```diff
@@ -1290,7 +1289,27 @@ const getMappedValue = (record: any, mapping: FieldMapping) => {
...
+    // This would indicate we're not resolving a field that we should be resolving.
+    return keys.reduce((result, key) => {
+      if (!(key in result)) {
+        sendToSentry(
+          `Property "${key}" is missing in record ${record.__typename}:${record.id}`
+        );
+        return undefined;
+      }
+      return result[key];
+    }, record);
```
gigxz: I think it's a little confusing that there are `return` statements conditional upon the `raiseOnMissing` argument, when that argument really should only affect logging. It shouldn't affect actual behavior. I think this code could be adjusted to make that more obvious. For example:


```jsx
// handleMissingField = sendToSentry

const keys = compact([relatedRecordAttribute, mapping.fieldName]);
const value = get(record, keys); // get attribute value from record

if (!specialCaseFieldNames.includes(mapping.fieldName)) {
   const isMissingField = !has(record, keys); /* logic to determine if attribute is not present */
   if(isMissingField) handleMissingField(`Field "${keys.join('.')}" is missing`);
}

return value
```

In that example it is clear to the reader that the `value` returned is **always** the same, regardless of how missing attributes are handled.

src/modules/form/util/formUtil.tsx
```diff
@@ -1268,18 +1269,16 @@ export const transformSubmitValues = ({
 };
 
 // record could be an Assessment
-const getMappedValue = (record: any, mapping: FieldMapping) => {
+const getMappedValue = (
+  record: any,
+  mapping: FieldMapping,
+  raiseOnMissing: boolean
```
gigxz: The passing around of `sentryOnMissing` and `raiseOnMissing` feels a little too prescriptive. Maybe we could make it more flexible by just accepting a callback method for handling missing fields? For example:

* `createInitialValuesFromRecord({ handleMissingField = sendToSentry })`  defaults to passing `sendToSentry` for handling missing attributes
* `createInitialValuesFromRecord({ ..., handleMissingField: () => ()})` in the one place in AssessmentForm where we want to override this behavior, we can pass a void function

What do you think?

martha: Thanks, that seems reasonable to me!

src/modules/form/util/formUtil.tsx
```diff
@@ -1290,7 +1290,27 @@ const getMappedValue = (record: any, mapping: FieldMapping) => {
...
     return customDataElementValueForKey(mapping.customFieldKey, cdes);
   } else if (mapping.fieldName) {
-    return get(record, compact([relatedRecordAttribute, mapping.fieldName]));
+    const keys = compact([relatedRecordAttribute, mapping.fieldName]); // for example: ['disabilityGroup', 'viralLoadSource']
+
+    const value = get(record, keys); // lodash's `get` will return undefined if the field doesn't exist
+
+    // Special cases where we DON'T want to alert Sentry if the field can't be resolved on the record.
+    // - imageBlobId: we pass this field when first saving a client image, but it isn't used for returning a saved client image.
+    // - mciId: similarly, we pass this field to save an MCI ID or indicate that a new one would be created, but it's not a persisted or resolved field on the client record.
```
gigxz:
```suggestion
    // - mciId: similarly, we pass this field to save an MCI ID or indicate that a new one would be created, but it's resolved on `externalIds` on the client record
```

src/modules/form/util/formUtil.tsx
```diff
@@ -1290,7 +1290,27 @@ const getMappedValue = (record: any, mapping: FieldMapping) => {
     );
     return customDataElementValueForKey(mapping.customFieldKey, cdes);
   } else if (mapping.fieldName) {
-    return get(record, compact([relatedRecordAttribute, mapping.fieldName]));
+    const keys = compact([relatedRecordAttribute, mapping.fieldName]); // for example: ['disabilityGroup', 'viralLoadSource']
+
+    const value = get(record, keys); // lodash's `get` will return undefined if the field doesn't exist
+
+    // Special cases where we DON'T want to alert Sentry if the field can't be resolved on the record.
```
gigxz, line 1297: Did you try client address/phone/email fields on the client form? those look like others where we have special processin glogic https://github.com/greenriver/hmis-warehouse/blob/120731182c2eca7c0175b064006270d2ff7b8e56/drivers/hmis/app/models/hmis/hud/processors/client_processor.rb#L43-L48

src/modules/form/util/formUtil.tsx
martha: ... but there is some problem with passing `sendToSentry` as a default param, storybook build chokes on it at the react docgen step, I think. https://github.com/greenriver/hmis-frontend/actions/runs/14174457600/job/39705834696#step:9:631. I'll keep digging.

src/modules/form/util/formUtil.tsx
gigxz: Nvm just tested and these fields are resolved so it's happy

gigxz (approved):

src/modules/form/util/formUtil.tsx
gigxz: Hmm odd. I looked at this a bit. I wouldn't expect it to run docgen on that file since it's a util file. It doesn't need to be a `tsx` file since it doesn't contain any jsx. changing it to `ts` seems to fix it

martha: Cool, thank you

PR #1109, opened 2025-03-24
Implement CE Project Page

martha:
## Description

GH issue: https://github.com/open-path/Green-River/issues/7350
Depends on warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5235

Summary of changes:
- Add `GetProjectCeReferrals` query
- Make both Project Referrals and Opportunities filterable on status
- Enable `RelativeDateDisplay` to accept `preciseTime` flag, similar to `DateWithRelativeTooltip`. 
- Update the Project CE base page to have 2 tabs, Opportunities and Referrals.
- Factor out the `ProjectOpportunitiesTable` into its own component and update the columns to match wip mocks
- Add a `ProjectReferralsTable`

How to test:
- With the backend branch checked out, and a few Opportunities created, run the starter pack script.
- This lives at `/project/:id/opportunities` for now, but let's discuss inline whether to change the route

## Type of change
- [ ] Bug fix
- [ ] Dependency update


src/modules/ce/components/ProjectCePage.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import CommonTabs from '@/components/elements/CommonTabs';
+import PageTitle from '@/components/layout/PageTitle';
+import useSafeParams from '@/hooks/useSafeParams';
+import CreateOpportunityButton from '@/modules/ce/components/CreateOpportunityButton';
+import ProjectOpportunitiesTable from '@/modules/ce/components/ProjectOpportunitiesTable';
+import ProjectReferralsTable from '@/modules/ce/components/ProjectReferralsTable';
+
+interface Props {}
```
martha: Should we make some changes to the established routes? This is currently at `/project/:id/opportunities`, and with the introduction of the CommonTabs and the tab keys I chose, we now have
- `/project/:id/opportunities#opportunities` 
- `/project/:id/opportunities#referrals` 

We can change around the tab keys I chose to avoid `opportunities#opportunities`, but maybe it would be nice to change the base URL from `/project/:id/opportunities` to `/project/:id/ce`? That gives us
- `/project/:id/ce#opportunities` 
- `/project/:id/ce#referrals`

gigxz: Yes I think we should change this page route to be "ce" and use `Coordinated Entry` as the page title. thanks!

src/modules/ce/components/ProjectCePage.tsx
```diff
@@ -0,0 +1,41 @@
...
+const ProjectCePage: React.FC<Props> = ({}) => {
```
gigxz:
```suggestion
const ProjectCePage: React.FC = () => {
```

src/components/elements/RelativeDateDisplay.tsx
```diff
@@ -34,6 +35,7 @@ export interface RelativeDateDisplayProps {
   tooltipSuffixText?: string;
   TooltipProps?: Omit<TooltipProps, 'title' | 'children'>;
   TypographyProps?: TypographyProps;
+  preciseTime?: boolean;
```
gigxz: This might be worth a comment, I'm not sure I totally understand the behavior.

My expectation for this component would be:
- If `dateString` is a  DateTime string, then the date would display formatted as DATETIME_DISPLAY_FORMAT and the relative date may include the time (e.g. "5 minutes ago")
- If `dateString` is a  Date string, then the date would display formatted as DATE_DISPLAY_FORMAT and the relative date can't include the time


Is `preciseTime` introduced so that the caller can pass a datetime `dateString` but specify that they _dont_ want the time to show?
A.k.a. it would be the equivalent of passing:

```tsx
<RelativeDateDisplay
   dateString={dateTimeString.substring(0, 10)} # trim to only date, because we dont want to show the time
/>
```

am I getting that right?

gigxz (approved):
👍

src/components/elements/RelativeDateDisplay.tsx
gigxz: If the behavior doesn't match the expected behavior, I think we should update it. It should be pretty simple to inspect the `dateString` to see if it has a time or not. Simple way would be to just check if the length is > 10 chars.

gigxz:
Looks good pending some schema comments on the backend pr!

src/components/elements/RelativeDateDisplay.tsx
martha: I think, given your backend feedback that [this can actually just be a DateTime](https://github.com/greenriver/hmis-warehouse/pull/5235#discussion_r2010709533), I'm going to revert the change to RelativeDateDisplay in this PR. I'll make a ticket in the [Align Design](https://github.com/open-path/Green-River/issues/7191) epic to revisit our date components and make sure they match up with the date-related recommendations under "Language" in [figma](https://www.figma.com/design/bsw8CX4UOq6SIdXYorZDWV/OP-HMIS-%2F-Design-Library?node-id=3081-241&t=S9Psc4ZRyqxnHh81-0).

gigxz: sounds good, thanks!

gigxz (approved):
nice nice 💯

PR #1110, opened 2025-03-24
Fix aria-live for column level toggle

martha:
## Description

While looking around for a good example of `aria-live` for an upcoming internal accessibility presentation, Bobby and I found this bug.
- https://github.com/open-path/hmis-accessibility/issues/55 is the original bug ticket and links the original PR, which introduced a proposed fix, but,
- https://github.com/greenriver/hmis-frontend/pull/935#discussion_r1955263159 is where the bug was re-introduced when we added axe automated testing and the previous fix caused a warning there.
- This PR fixes in a way that's both sensible to screenreaders and doesn't cause axe warnings, Whew! 🙈 Sorry for the churn!

## Type of change
- [x] Bug fix
- [ ] Dependency update


gigxz (approved):
thanks!

PR #1111, opened 2025-03-25
Include HMIS login history on audit page; audit history page improvements

gigxz:
## Description

* Expose Login History on HMIS User audit page
* General cleanup to audit page, including...
  * Make tables more accessible by removing linked cells and adding secondary action menus instead
  * Replace toggle navigation with `CommonTabs`, use more descriptive names for them
  * Update alert severity on access history page
  * Show the user's name in the breadcrumbs
* Add `Edit Account` link to the user menu that brings the user to the warehouse account management page
* Fix bug in `DateWithRelativeTooltip` that had some weird formatting for dates (it would display date as a date, but relative date would always be time-specific -- i.e. "4 hours ago" for something that had a YYYY-MM-DD date). Remove `preciseTime` prop in favor of inferring from the `dateString` whether it includes a time or not.

Screenshot 2025-03-28 at 3 59 02 PM

Screenshot 2025-03-28 at 3 59 26 PM

How to test:
- Test all functionality on the User Audit page
- Test Enrollment Audit and Client Audit pages (styles only, some change to wrapping behavior)
- Test existing relative date displays (eg Entry Date on project enrollment table)

## Type of change


src/components/elements/RelativeDateDisplay.tsx
```diff
@@ -11,24 +11,24 @@ import { customVisuallyHidden } from '@/config/theme';
...
-  const date = parseHmisDateString(dateString);
+export const getFormattedDates = (dateString: string) => {
+  const date = parseHmisDateString(dateString); // parse Date or DateTime string into Date object
   if (!date) return [];
+
+  const isDateTime = dateString.length > 10;
   return [
-    preciseTime ? formatDateTimeForDisplay(date) : formatDateForDisplay(date),
-    formatRelativeDateTime(date),
+    isDateTime ? formatDateTimeForDisplay(date) : formatDateForDisplay(date),
```
martha, line 25: Cool, this looks like a nice improvement.

I keep getting mixed up about these components, `RelativeDateDisplay` and `DateWithRelativeTooltip`. They seem kinda like inverses of each other -- `RelativeDateDisplay` shows a relative date, with a date as a tooltip, while `DateWithRelativeTooltip` shows the reverse. But they also arose with different needs and have different "chrome" inside them, like prefix/suffix logic that's present in `RelativeDateDisplay` but not `DateWithRelativeTooltip`.

Anyway, I had created https://github.com/open-path/Green-River/issues/7466 (in the Align Design epic) to make sure we are aligned with the design library's recommendations around dates, so maybe we can also take that opportunity to consolidate and reduce code duplication among the components.

src/modules/admin/components/users/UserAuditHistory.tsx
```diff
@@ -93,10 +104,9 @@ const UserAuditHistory = () => {
           paginationItemName='event'
           queryDocument={GetUserAuditEventsDocument}
           queryVariables={{ id: userId }}
-          rowSx={() => ({ whiteSpace: 'nowrap' })}
-          tableProps={{ sx: { tableLayout: 'fixed' } }}
           recordType='ApplicationUserAuditEvent'
           filters={filters}
+          rowSecondaryActionConfigs={rowSecondaryActionConfigs}
```
martha, line 116: Adding `rowName` here would fix up the aria-label so it says something interesting instead of just the ID of the audit record. Maybe something like
```tsx
rowName={(row) => `${row.clientName}'s ${row.recordName}, ${row.createdAt}`} // createdAt date needs formatting
```

src/modules/admin/components/users/UserAuditHistory.tsx
```diff
@@ -77,6 +50,44 @@ const UserAuditHistory = () => {
     type: 'UserAuditEventFilterOptions',
   });
 
+  const rowSecondaryActionConfigs = useCallback(
+    ({ projectName, clientId, enrollmentId, projectId }: AuditHistoryType) => {
+      const viewClient = clientId && {
+        title: 'View Client',
+        key: 'client',
+        ariaLabel: `View Client ${clientId}`,
```
martha: Maybe `clientName` instead of `clientId` here and on line 68?

src/modules/admin/components/users/UserAuditHistory.tsx
```diff
@@ -1,70 +1,43 @@
...
+} from '@/routes/routes';
+import {
   GetUserAuditEventsDocument,
   GetUserAuditEventsQuery,
   GetUserAuditEventsQueryVariables,
+  UserAuditEventFilterOptions,
 } from '@/types/gqlTypes';
 import { generateSafePath } from '@/utils/pathEncoding';
 
 type AuditHistoryType = NonNullable<
```
martha: This isn't related to your changes, but while we're in here, maybe we can remove `AuditHistoryType` and replace it with `UserAuditEventFields`? I think this is from when I was brand-new, and probably pattern-matching from a table where we don't have a fragment to use neatly.

src/modules/admin/components/users/UserAuditHistory.tsx
```diff
@@ -93,10 +104,9 @@ const UserAuditHistory = () => {
           paginationItemName='event'
```
martha, line 108: It could be nice to give the "fields changed" column a fixed width, so the table doesn't jump around when it is expanded/contracted. (See "Client Name" col in the screencap)

https://github.com/user-attachments/assets/ec157a12-23af-4d15-8546-811e9837429b

src/modules/admin/components/users/UserAuditPage.tsx
```diff
@@ -67,33 +60,57 @@ const UserAuditPage: React.FC<Props> = ({
...
-            sx={{ mb: 3 }}
-            value={userHistoryType}
-            onChange={handleHistoryTypeToggleChange}
-            items={historyTypeToggleItems}
-            aria-labelledby={
-              (labelElement && labelElement.getAttribute('id')) || undefined
-            }
-          />
-        )}
+      <CommonTabs
```
martha, line 108: Hrm, you probably already noticed this, but I think the addition of `CommonTabs` here is making the navigation a little funny here. `CommonTabs` controls hash params, but here we are controlling navigation with routes, so the resulting URLs look like,
- /admin/users/2/enrollmentAccessHistory#access
- /admin/users/2/clientAccessHistory#access
- /admin/users/2/editHistory#edits
- /admin/users/2/loginActivity#logins

Maybe the simple solution is to add a new flag prop on `CommonTabs` telling it _not_ to control the hash params?

src/modules/admin/components/users/UserAccessHistory.tsx
```diff
@@ -16,16 +16,16 @@ import { AdminDashboardRoutes } from '@/routes/routes';
...
-    value: 'clients',
-    label: 'Clients',
-    Icon: ClientIcon,
-  },
   {
     value: 'enrollments',
     label: 'Enrollments',
     Icon: EnrollmentIcon,
   },
+  {
```
martha, line 24: Along with flipping the order of the toggle around, should we change the `UserActionsMenu` link to point at Enrollments instead of Clients? https://github.com/greenriver/hmis-frontend/pull/1111/files#diff-0a8e81d2b74f0bc9e0ccdf728096d1c07eb636bf71613323f2a6f55d97fde75aR35

src/modules/admin/components/users/UserActionsMenu.tsx
```diff
@@ -1,20 +1,28 @@
...
 
 interface Props {
   onClickImpersonate: VoidFunction;
   isCurrentUser: boolean;
-  userId: string;
+  user: UserAdminFieldsFragment;
   rootAccess: RootPermissionsFragment;
 }
 
 const UserActionsMenu: React.FC<Props> = ({
```
martha, line 22: It occurs to me that we could update this `UserActionsMenu` to be a 3-dots menu matching our usual pattern for tables? I'm guessing this already probably occurred to you too and you decided against it, maybe because it's a change that might have some training burden?

src/modules/admin/components/users/UserAuditHistory.tsx
```diff
@@ -77,6 +50,44 @@ const UserAuditHistory = () => {
...
+          to: generateSafePath(EnrollmentDashboardRoutes.ENROLLMENT_OVERVIEW, {
+            clientId,
+            enrollmentId,
+          }),
+        };
+
+      const viewProject = projectId && {
+        title: 'View Project',
+        key: 'project',
+        ariaLabel: `View Project ${projectId}`,
```
martha: Same here, maybe `projectName` instead of `projectId`

martha (approved):
Looks good to me! Thanks for updating to match the new table pattern

src/modules/admin/components/users/UserAuditHistory.tsx
gigxz: thanks!

src/modules/admin/components/users/UserAuditHistory.tsx
gigxz: thanks for some reason I thought we didnt have names, updating!

src/modules/admin/components/users/UserActionsMenu.tsx
gigxz: Yeah I didn't bother in case it would affect any training screenshots, and it's still using the common menu so its not something super bespoke. We will  revisit this if/when we move the user management piece more fully into the HMIS

src/modules/admin/components/users/UserAuditPage.tsx
gigxz: Updated to add a flag to skip the hash/href behavior. The way the audit page is implemented still doesn't feel ideal because the tab buttons are buttons not nav links, yet they do nav you to a different route. It might be even better if CommonTabs could support using `a` links  on tab definitions. But leaving as-is since that felt like too big of a change

src/modules/admin/components/users/UserAuditHistory.tsx
gigxz: Hmm, I had to make some changes to get the right sticky bar to work, and I didn't realize it had this side-effect. Below is what it looks like without those changes (`tableLayout: 'fixed'`). The jumping around isn't ideal, but I think it's better than having the action menu take up a lot of space. I am going to opt to merge this even with the jumping-around-ness.

Screenshot 2025-04-01 at 3 48 55 PM

src/components/elements/RelativeDateDisplay.tsx
gigxz: Thanks agreed I am looking forward to standardizing these.. I added stories in the meantime so we have something to review with design https://github.com/greenriver/hmis-frontend/pull/1111/commits/8745eda842e6df8cbb727a4d7f5cf9e291870a38

PR #1123, opened 2025-04-04
Reduce noise from missing field Sentry warnings

martha:
## Description

This PR aims to reduce Sentry noise introduced in https://github.com/greenriver/hmis-frontend/pull/1102. It targets **release-158**.

Examples of noise on QA:
- An assessment's related record is null, like `employment_education`
  - **Example**: https://green-river.sentry.io/issues/6505375313/?referrer=slack&notification_uuid=93170c43-606d-4f7c-bf6d-d64da33e6f6d&environment=staging&alert_rule_id=14743218&alert_type=issue
  - **Cause**: I _think_ this is happening because the form definition itself changed. The [assessment](https://qa-hmis.openpath.host/client/==RHBHYzd6RzRqcDRNNUkzNk80alRzdz09/enrollments/==NXJ6SllLK04xbGtNNUkzNk80alRzdz09/intake) was first filled out before this [enable_everything QA patch](https://github.com/greenriver/hmis-warehouse/blob/24aac3afd301147eaa8617d358ad7b2e59c5458b/drivers/hmis/lib/form_data/qa_hmis/fragments/patches/enable_everything_patch.json#L4) was introduced. I reproduced this locally and found that reopening and resubmitting the assessment (without touching the missing `employment_education` data in the form) caused creation of an empty `EmploymentEducation` record for this assessment, so on subsequent reloads, no errors are raised.
  - **Proposed changes**: It seems valid to me that the whole related record be nil. What [the new code that's triggering this warning](https://github.com/greenriver/hmis-frontend/pull/1119/files#diff-7d33d3e9a02ac3ebc1f517b289868d09de568774095409317f30b95d4c1d1341R1302) really wants to confirm is that we aren't forgetting to resolve a field _on_ a related record. Therefore, this PR introduces logic to _not_ consider fields "missing" if the parent related record is null.
- New referral form fields are missing
  - **Example**: https://green-river.sentry.io/issues/6506156446/?referrer=slack&notification_uuid=3f1c65f4-3e8a-425b-872c-bd1657b928cb&environment=staging&alert_rule_id=14743218&alert_type=issue
  - **Cause**: In the project Outgoing Referrals form, we use `useDynamicFormHandlersForRecord` with a [record of `{}`](https://github.com/greenriver/hmis-frontend/pull/1123/files#diff-9c914c598537ebdc0ba351424f0e318dbcf545706d77a45cb16bf2d155241ef0R46). I think this case is a bit similar to [`AssessmentForm`](https://github.com/greenriver/hmis-frontend/blob/release-158/src/modules/assessments/components/AssessmentForm.tsx#L212-L218) so I've added a similar override. 

## Type of change
Bug fix

### DIFF
--- a/src/modules/form/hooks/useDynamicFormHandlersForRecord.tsx
+++ b/src/modules/form/hooks/useDynamicFormHandlersForRecord.tsx
@@ -42,6 +42,7 @@ export interface DynamicFormHandlerArgs<T> {
   localConstants?: LocalConstants;
   inputVariables?: SubmitFormInputVariables;
   errorRef?: RefObject<HTMLDivElement>;
+  handleMissingField?: (message: string) => void;
 }
 
 export function useDynamicFormHandlersForRecord<
@@ -53,6 +54,7 @@ export function useDynamicFormHandlersForRecord<
   inputVariables,
   localConstants,
   errorRef,
+  handleMissingField,
 }: DynamicFormHandlerArgs<T>) {
   const [errors, setErrors] = useState<ErrorState>(emptyErrorState);
 
@@ -90,6 +92,7 @@ export function useDynamicFormHandlersForRecord<
     itemMap,
     definition: formDefinition?.definition,
     localConstants,
+    handleMissingField,
   });
 
   const onSubmit: DynamicFormOnSubmit = useCallback(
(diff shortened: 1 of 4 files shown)


gigxz (approved):
Looks good! Nice find & thank you for the detailed comment! 🙏

PR #1124, opened 2025-04-07
Fix bug with history

martha:
## Description

Fix bug with back-navigation by ensuring that react-router-dom's "replace: true" option is used where appropriate.
- When updating the hash from within `CommonTabs`. (This isn't currently manifesting, since the only tabs we have other than on the CE feature branch are on the User Audit page, which don't update hash)
- When updating the search params from within `useSearchParamsState`. (This is the source of the bug that's described in the ticket.)

## Type of change
Bug fix


gigxz (approved):
looks great, good find! can you adjust this to go into release-158? I think thats worthwhile since its active an bug on prod, and we should have enough time to QA it tomorrow or wednesday.

PR #1128, opened 2025-04-09
Implement UI for unit availability

martha:
## Description

Summary of changes:
- On Unit Management table, add actions for managing unit availability:
  - Mark available
  - Mark unavailable
  - View opportunity 
- And add columns for viewing current opportunities/referrals
- move Delete Unit into the action menu. To do this, I refactored the `DeleteMutationButton` component so that `DeleteMutationConfirmationDialog` is usable separately.
- Add a `loading` prop to `GenericTableWithData` so that the parent can indicate loading state even if the fetch that the table performs isn't currently loading. My intention is for this to not impact the internal loading logic.

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5275

How to test:
- if you are using the starter pack, navigate to Unit Management page
- You should be able to single and bulk-delete units, unless they are occupied
- You should be able to mark a unit as available, and view the opportunity this generates
- You should be able to mark a unit unavailable, unless there is a referral in progress on that opportunity 

## Type of change
New feature


src/api/operations/units.fragments.graphql
```diff
@@ -28,4 +28,10 @@ fragment UnitFields on Unit {
       ...ClientName
     }
   }
+  latestOpportunity {
+    ...CeOpportunitySummaryFields
+    referral {
+      ...CeReferralSummaryFields
+    }
+  }
```
gigxz: This is the first time we'd have production users making API request that hit the new CE api types. It does "lock us in" a bit so we need to be more careful about not making backwards-incompatible changes.

Since CE isn't going to go live for a long while, it might be preferable to, add a directive here so we only query for it if CE is enabled. That way, if we did make a breaking change to this shape (like rename `latestOpportunity`, it wouldn't affect users.). Maybe I'm being overly cautious, it's also not a huge deal to start paying closer attention to backwards compatibility.

src/modules/ce/components/OpportunityBanner.tsx
```diff
@@ -30,13 +30,7 @@ const OpportunityBanner: React.FC<Props> = ({
...
   const header = useMemo(() => {
     if (referral?.status === CeReferralStatus.Accepted) return 'Filled By';
-    if (
-      referral &&
-      [CeReferralStatus.Initialized, CeReferralStatus.InProgress].includes(
-        referral.status
-      )
-    )
-      return 'In-Progress Referral';
+    if (referral && referral.active) return 'In-Progress Referral';
```
gigxz, line 33: nice improvement to add this to the api!

src/modules/dataFetching/components/DeleteMutationButton.tsx
```diff
@@ -108,32 +61,11 @@ const DeleteMutationButton = <Mutation, MutationVariables>({
...
-              <Typography>
-                {`Are you sure you want to ${verb} this ${recordName}?`}
-              </Typography>
-              <Typography>This action cannot be undone.</Typography>
-            </>
-          )}
-        </>
-      </ConfirmationDialog>
+        onClose={() => setShowDialog(false)}
+        {...rest}
```
gigxz, line 68: should pass through `recordName` as well, otherwise deletion confirmation models just show "
are you sure you want to delete this record"

src/modules/units/components/UnitManagementTable.tsx
```diff
@@ -110,51 +58,170 @@ const UnitManagementTable = ({
...
-              render: (unit: UnitFieldsFragment) =>
-                renderDeleteButton([unit.id], unit.occupants.length > 0),
+              header: 'Accepting Referrals?',
+              key: 'available',
+              render: (unit: UnitFieldsFragment) => {
+                if (
+                  unit.latestOpportunity &&
+                  !unit.latestOpportunity.referral &&
+                  unit.occupants.length === 0
+                ) {
```
gigxz: This seems like logic that could be moved to the backend. (eg `accepting_ce_referrals` on the unit). That might make things easier for testing and to have this "status" displayed consistently throughout the application.

src/modules/units/components/UnitManagementTable.tsx
```diff
@@ -110,51 +58,170 @@ const UnitManagementTable = ({
...
         : []),
     ];
-  }, [allowDeleteUnits, renderDeleteButton]);
+  }, [canViewCoordinatedEntry]);
 
   const filters = useFilters({
     type: 'UnitFilterOptions',
   });
 
+  const hasSecondaryActions = useMemo(() => {
```
gigxz: This is a lot of CE-specific logic added to this component. I would suggest moving it into a hook. That way we can keep CE code more isolated. We have other similar hooks for `use<thing>Actions`.

src/modules/units/hooks/useDeleteUnits.tsx
```diff
@@ -0,0 +1,108 @@
...
+  setUnitToDelete: (unitId: string) => void;
+  renderSingleDeleteDialog: () => JSX.Element;
+  renderBulkDeleteButton: (
+    unitIds: string[],
+    disabled?: boolean
+  ) => JSX.Element;
+} => {
+  const deleteDialogProps = useMemo(() => {
+    return {
+      idPath: 'deleteUnits.unitIds[0]',
```
gigxz, line 29: This is a bit unusual, I don't think we've used this component for deleting multiple records. Do we need to change the API at all? WiIl this work for retrieving and displaying validation errors?

gigxz: Oh I guess we were already doing this before

gigxz (approved):
looks good! just a few suggestions for encapsulating the CE logic a bit more

src/modules/units/components/UnitManagementTable.tsx
```diff
@@ -110,51 +58,170 @@ const UnitManagementTable = ({
...
+          title: 'Mark as Unavailable for Referrals',
+          key: 'markUnavailable',
+          ariaLabel: `Mark Unit ${unit.id} as Unavailable for Referrals`,
+          onClick: () => {
+            markUnitsUnavailable({ variables: { unitIds: [unit.id] } });
+          },
+          disabled:
+            unit.latestOpportunity.status === CeOpportunityStatus.Locked,
+          disabledReason:
+            'Unit with in-progress referral cannot be marked as unavailable',
```
gigxz, line 142: I think `Unit has an in-progress referral'`  would be sufficient

src/modules/dataFetching/components/DeleteMutationButton.tsx
martha: Thanks, good catch!

src/api/operations/units.fragments.graphql
martha: Thanks, yeah I totally agree with this. Is https://github.com/greenriver/hmis-frontend/pull/1128/commits/b65071601c4a414c41a3dc4600bc52f37ce2c58d along the lines of what you had in mind?

src/modules/units/components/UnitManagementTable.tsx
martha: Sounds good, I've done so here https://github.com/greenriver/hmis-warehouse/pull/5275/commits/2de32fe3613a08020359abf2311e3209d6c3a66b. 

I also want to note that I removed the `occupants`-related logic there. So it's now checking whether the unit is accepting referrals based only on the CE entities opportunity/referral. That change was based on my memory of our conversations last week/a couple weeks ago about the idea that a referral can be in-progress while an occupying household is getting ready to move out. (See https://github.com/open-path/Green-River/issues/7537)

gigxz (approved):

PR #1131, opened 2025-04-11
Determine nav item selection using regex

martha:
## Description

On QA, visit a project's CE Participations page. 
Bug: Both "Coordinated Entry" and "CE Participations" appear selected in the project nav.
Screenshot 2025-04-11 at 9 47 44 AM

I targeted release-159 with this PR, since it's fixing a bug, but I think it would be ok to go on release-160 instead since the bug is only on our QA environment (not affecting any clients).

## Type of change
Bug fix

### DIFF
--- a/src/components/layout/dashboard/sideNav/useItemSelectionStatus.tsx
+++ b/src/components/layout/dashboard/sideNav/useItemSelectionStatus.tsx
@@ -1,8 +1,8 @@
-import { escapeRegExp } from 'lodash-es';
 import { useMemo } from 'react';
 import { useLocation } from 'react-router-dom';
 
 import { NavItem } from './types';
+import { addTrailingSlash } from '@/components/layout/dashboard/sideNav/navUtil';
 import useCurrentPath from '@/hooks/useCurrentPath';
 
 export type Options<T> = {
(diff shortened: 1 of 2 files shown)


src/components/layout/dashboard/sideNav/useItemSelectionStatus.tsx
```diff
@@ -16,9 +17,12 @@ export const useItemSelectionStatus = <T extends object>({
...
     if (!item.path && !item.activePaths) return false;
-    // If current location starts with item path, then this nav item should be selected.
+    // If current location matches item path, then this nav item should be selected.
     // (e.g. item path is /enrollments/1/services and current location is /enrollments/1/services/2)
-    if (item.path && currentLocation.startsWith(item.path)) {
+    // Use a regex to avoid false positives (e.g. item path is /projects/1/ce and current location is /projects/1/ce-participation)
+    const escapedPath = item.path ? escapeRegExp(item.path) : '';
+    const regex = new RegExp(`^${escapedPath}(?:[#/\?]|$)`);
+    if (item.path && regex.test(currentLocation)) {
       return true;
```
gigxz: Good find! I'm unsure we need the complexity of the regex. There should always be a `/` between the "base" path and the rest of the path. I _think_ we could resolve the bug of comparing `/projects/1/ce` and `/projects/1/ce-participation` could be resolved by ensuring that the path starts with `item.path + '/'`?

Does this achieve the same result? Maybe there is something I am missing about hashes.
```jsx
    // move to util
    const addTrailingSlash = (path: string) =>
      path.endsWith('/') ? path : path + '/';

    const matchesPath =
      currentLocation === item.path || // matches path exactly
      currentLocation.startsWith(addTrailingSlash(item.path)); // starts with path

    if (matchesPath) return true;
```

martha: Thanks! Yes I think it achieves the same result, even for pages with hash params, since `currentLocation` returns just the pathname, not the hash.

gigxz (approved):

PR #1132, opened 2025-04-14
Fix bugs with back navigation

martha:
## Description

Summary of changes:
- Update the `useSearchParamState` hook to call `navigate`, not `setSearchParams`, when populating params on initial render.
  - Why: I traced the Common Tabs bug described in the linked ticket to this `useEffect`. I'm not sure _exactly_ why `setSearchParams` results in weird behavior, because we were using `replace: true` anyway; but calling `navigate` seems to improve the situation.
- Update to the `CommonTabs` component to factor hash state logic out into a new `useHashState` hook.
  - Why: Once the above `useSearchParamState` is fixed, back/forward buttons correctly update the URL bar. But there was nothing in CommonTabs that's listening to `hash`, so the content of the page did not actually change. This fixes that.
  - See internal comment for more discussion
- Add a `useEffect` in `useIntermediateState` to sync the "intermediate" value with external changes to `value`
  - Why: This fixes the bug with Optional Columns described in the linked ticket
  - Supporting changes:
    - Memoize `TableColumnsMenu` to make sure that extra renders don't cause this effect to run when `value` actually _hasn't_ changed, which leads to buggy behavior
    - Memoize props passed from `GenericTableWithData`, for the same reason
- Rename `TableFilters` component to `TableControls`

Whew! Fixing these bugs sometimes felt like playing whack-a-mole -- all the effects make it somewhat fragile, I think -- but I _think_ that the current behavior looks correct to me. Please let me know if you see any behavior that looks off!

## Type of change
Bug fix

- [x] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue - yes, see the linked ticket


src/components/elements/tableFilters/filters/ColumnsMenu.tsx
```diff
@@ -1,6 +1,6 @@
 import ViewColumnIcon from '@mui/icons-material/ViewColumn';
```
martha, line 1: This component is now wrapped in a call to `memo` but not otherwise changed

src/components/elements/CommonTabs.tsx
```diff
@@ -14,60 +14,42 @@ interface CommonTabsProps {
   sx?: SxProps;
   currentTab?: string;
   onChangeTab?: (tab: string) => void;
-  updateUrlHash?: boolean; // If true, the URL hash will be updated to match the selected tab.
 }
 
-// CommonTabs wraps the MUI Tabs component. It can be controlled or uncontrolled,
-// so callers can choose whether to delegate controlling the state of which tab is selected.
+// CommonTabs wraps the MUI Tabs component. It can be controlled or uncontrolled.
```
martha, line 19: Some ideas I had about `CommonTabs`, which previously supported two dichotomies: controlled/uncontrolled and hash/no-hash.
- I think it doesn’t make sense for the component to both be “controlled” and use hash params. If a parent or ancestor needs to update the state on a CommonTabs that uses hash, that parent should just use `navigate`.
- By the same token, I think it doesn’t make sense to have a controlled component that _does_ use hash params. 

So I’m proposing to collapse the 2 dichotomies into: controlled by parent / controlled by hash params.

Maybe in the future we will need to add a use case for “controlled internally but NOT with hash params” but I think it’s okay to leave out support for this for now. Let me know what you think!

src/hooks/useSearchParamState.ts
```diff
@@ -106,6 +106,22 @@ const getAllCurrentParams = (searchParams: URLSearchParams) => {
   return allUrlParams;
 };
 
+const populateParams = (
+  params: Record<string, any>,
+  searchParams: URLSearchParams
+) => {
+  Object.entries(params).forEach(([key, value]) => {
+    if (!isEmpty(key)) {
+      // Don't add this key to params if the value is empty
```
gigxz: I'm having a hard time understanding this comment -- it doesn't look like we're checking anywhere whether the value is empty, just whether its an array. Did you mean to be commenting about checking for an empty key above? Or is the key check really supposed to be checking for an empty value? I don't see how the key can be empty

src/hooks/useSearchParamState.ts
```diff
@@ -138,11 +159,23 @@ const useSearchParamsState = ({
...
       Object.entries(initial).filter(([key]) => !searchParams.has(key))
     );
-    if (Object.keys(toUpdate).length) {
-      // Spread currentParams to avoid overwriting search params set by other components
-      const currentParams = getAllCurrentParams(searchParams);
-      setSearchParams({ ...currentParams, ...toUpdate }, { replace: true });
-    }
+
+    const accumulator = new URLSearchParams();
+    populateParams(toUpdate, accumulator);
```
gigxz: Maybe `toUpdate` can have a more descriptive name. It looks like it is the same as `initial` but we filtered out _keys_ that are already present in searchParams. Maybe `initialParamsToUpdate`?

src/hooks/useSearchParamState.ts
```diff
@@ -138,11 +159,23 @@ const useSearchParamsState = ({
...
+    if (accumulator.entries.length === 0) return; // If we don't have anything to update, return early
```
gigxz: `accumulator.entries` [is a function](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/entries) so I think this is always returning zero and not doing anything. I'm not sure how to even reproduce a bug here though.

If the intention is to return early when there are no params to update, can we instead base that on `Object.keys(toUpdate).length` up on line 162? Or is the population logic relevant here? (I had a question on that method logic too)

src/hooks/useSearchParamState.ts
```diff
@@ -138,11 +159,23 @@ const useSearchParamsState = ({
...
+    populateParams(currentParams, accumulator);
+
+    navigate(
+      {
+        pathname,
+        hash,
+        search: accumulator.toString(),
+      },
+      { replace: true }
+    ); // replace so that browser history acts normal
```
gigxz:
```suggestion
    ); // replace so that this doesn't create a new entry in browser history
```

src/hooks/useSearchParamState.ts
```diff
@@ -179,7 +212,7 @@ const useSearchParamsState = ({
           }
         }
       }
-      setSearchParams({ ...currentParams, ...newValues });
+      setSearchParams({ ...currentParams, ...newValues }, { replace: false });
```
gigxz: This change shouldn't be needed, non-replacement  is the default behavior

src/modules/dataFetching/components/GenericTableWithData.tsx
```diff
@@ -154,11 +154,35 @@ const GenericTableWithData = <
...
+            value: col.key,
+            header: col.header,
+            defaultHidden: !!col.optional?.defaultHidden,
+          })),
+          columnsValue: includedOptionalColumns,
+          setColumnsValue: setIncludedOptionalColumns,
+        }
+      : undefined;
+  }, [includedOptionalColumns, optionalColumns, setIncludedOptionalColumns]);
+
```
gigxz: This is fine, but it does seem like something that can be moved into the `useOptionalColumns` hook. It only depends on data that is in that hook, so it really doesn't need to be inside GenericTableWithData. Inside `useOptionalColumns` you could add:
```tsx
  const optionalColumnsMenuProps: TableColumnsMenuProps | undefined =
    useMemo(() => {
      if (optionalColumns.length === 0) return undefined;

      return {
        columns: optionalColumns.map((col) => ({
          value: col.key,
          header: col.header,
          defaultHidden: !!col.optional?.defaultHidden,
        })),
        columnsValue: includedOptionalColumns,
        setColumnsValue: setIncludedOptionalColumns,
      };
    }, [optionalColumns, includedOptionalColumns, setIncludedOptionalColumns]);

```

Then return optionalColumnsMenuProps from the hook. (And remove `setIncludedOptionalColumns` then it wouldnt be needed anymore). I think that would be a nice cleanup

src/modules/dataFetching/components/GenericTableWithData.tsx
```diff
@@ -154,11 +154,35 @@ const GenericTableWithData = <
...
 
+  const filterProps = useMemo(() => {
+    return !isEmpty(filters)
+      ? {
+          filters,
+          filterValues,
+          setFilterValues,
+        }
+      : undefined;
+  }, [filterValues, filters]);
```
gigxz, line 170: Moving this out of the jsx is nice, we don't need to use a ternary anymore
```
  const filterProps = useMemo(() => {
    if (!filters || Object.keys(filters).length === 0) return;

    return {
      filters,
      filterValues,
      setFilterValues,
    };
  }, [filterValues, filters]);
```

gigxz (commented):
Thanks @martha & sorry for the late review on this one! The behavior all looks good when I'm testing it out. I just had some questions on what the code is doing / readability.

src/hooks/useSearchParamState.ts
martha: Thanks, good catch. I _think_ that my intention was to check for an empty value (not key), but I didn't catch the bug because I've been working with search params whose values are arrays, and the empty array case is working. I'll add a guard in case of empty value too.

src/modules/dataFetching/components/GenericTableWithData.tsx
martha: Thanks, sounds good! 

In this diff I'm also returning the "default optional columns" calculation as part of the component props returned by that hook. It's already computed by the hook, so it doesn't need to be computed again in the `ColumnsMenu`.

src/hooks/useSearchParamState.ts
martha: Hm, yes you're right! And I'm not sure why I thought this was working before -- it's not. (Previously saved optional columns were not getting re-populated when they should be.)

I think that we do need to check `accumulator`'s size separately from `Object.keys(toUpdate).length`. In this example, `optionalColumns` may be present as a key, but the value is an empty list, so no redirect is needed.

Sorry about the missed bugs here. I was getting pretty frustrated with bug whack-a-mole, but I definitely should have spent more time on the final review.

src/hooks/useIntermediateState.tsx
```diff
@@ -1,11 +1,34 @@
-import { useCallback, useMemo, useState } from 'react';
+import { isEqual } from 'lodash-es';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+
+// A custom alternative to useEffect that uses deep comparison to compare
```
martha: @gigxz this is newly added since you reviewed. Speaking of bug whack-a-mole, this (should!) fix this bug:
> If you click "Apply" then quickly re-open the columns menu and change the selection, you shouldn't see any flickering. The checkboxes should only respond to your user input, and not randomly reset themselves

, a bug I thought I had fixed before, but which was still reproducible on the version of the PR you reviewed. :(

This approach was suggested to me by gpt when I asked if there is a way to use `useEffect` with an option similar to React `memo`'s `arePropsEqual`.

gigxz: Thanks @martha. I recognize the problem, and I think the chatGpt hook is good. I'd prefer pulling it into its own hook like `useDeepCompareEffect`. (Deep compare because it uses lodash's isEqual).

FWIW, we solve this problem elsewhere by using `usePrevious`  with a deep equal in `useEffect` to find out whether a prop has changed. Here is an [example](https://github.com/greenriver/hmis-frontend/blob/3b0dbaae0b07ce73d0b3a66ff97eb3dffe881867/src/modules/dataFetching/components/GenericTableWithData.tsx#L199-L203). It would look like:

```jsx
const previousValueArg = usePrevious(value);

useEffect(() => {
  // deep-equal check to see whether the `value` argument passed has changed.
  // if it has changed, set state to synchronize internal state, ex user navigation etc.....
  if (!isEqual(previousValueArg, value)) {
    setState(value);
  }
}, [previousValueArg, value]);
```

Again, I think it's essentially the same as your proposed solution. I don't have a preference between using usePrevious vs adding a new useDeepCompareEffect. (I do think its nicer to pull the logic out of useIntermediateState file though).

These are both described in https://dev.to/vsramalwan/ways-to-handle-deep-object-comparison-in-useeffect-hook-1elm

gigxz (approved):

src/hooks/useSearchParamState.ts
gigxz: ok. its a tricky one!

src/hooks/useIntermediateState.tsx
martha: Thanks! I'd like to go with `usePrevious` to match our existing pattern.

gigxz (approved):

PR #1133, opened 2025-04-15
Add Client Referrals page

martha:
## Description

Summary of changes: Add client Referrals page containing 2 new tables:
- Client Referrals, backed by new query `GetClientCeReferrals`
- Client Eligible Opportunities, backed by `GetClientEligibleOpportunities`

How to test:
- With CE turned on locally, visit a client who has ongoing referral(s) and/or is eligible for several opportunities. 
- You should see the "Referrals" nav item on the client left nav, and be able to navigate to it. 
- On the page, you should see a table displaying info about the ongoing referrals
- You should be able to filter by project, project type, and status. 
- They should initially be filtered by status (ongoing referrals, not accepted or rejected ones)
- On the same page, you should be able to navigate to the other tab "Eligible Opportunities. You should see a table displaying info about opportunities this client is eligible for
- You should be able to filter by project and project type

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5294

## Type of change
New feature


src/modules/ce/components/client/ClientOpportunitiesTable.tsx
```diff
@@ -0,0 +1,88 @@
...
+    key: 'projectName',
+    render: (opportunity: ClientCeOpportunitySummaryFieldsFragment) =>
+      opportunity.project.projectName,
+  },
+  {
+    header: 'Project Type',
+    key: 'projectType',
+    render: (opportunity: ClientCeOpportunitySummaryFieldsFragment) => (
+      <ProjectTypeChip projectType={opportunity.project.projectType} />
+    ),
```
martha, line 37: Should we include the Status column? I didn't include it because my understanding is that it would always be "eligible/on waitlist"

src/modules/ce/components/client/ClientReferralsPage.tsx
```diff
@@ -0,0 +1,31 @@
...
+  return (
+    <>
+      <PageTitle title={'Referrals'} />
+
+      <CommonTabs
+        ariaLabel={'Client Referrals Tabs'}
+        tabDefinitions={[
+          {
+            title: 'All Referrals',
+            key: 'referrals',
```
martha, line 17: `clients/:id/referrals#referrals` is a little too weird, needs improvement.

I'm tempted to put `ce` in the URL (matching `projects/:id/ce`), even though I think we've been moving towards removing "ce" from user-facing language (e.g. "Referrals" instead of "Coordinated Entry" in the side nav item).

I'm proposing
`clients/:id/ce#referrals`
`clients/:id/ce#opportunities`

What do you think?

src/modules/ce/components/client/ClientOpportunitiesTable.tsx
```diff
@@ -0,0 +1,88 @@
...
+  });
+
+  return (
+    <>
+      <Paper sx={{ p: 2, mb: 2 }}>
+        <Typography variant='h5' component='h2' pb={1}>
+          About Eligible Opportunities
+        </Typography>
+        <Typography variant='body2'>
+          {clientName} qualifies for several open positions. These positions are
```
martha: I think we should probably workshop this language; I took it from the figma mocks but I'm not sure all of it is applicable. Also, the ticket spec states we wanted to add language about the eligibility pools being determined on a schedule. That schedule isn’t determined yet (7544), so maybe we should just leave a todo for now?

src/api/operations/ce.fragments.graphql
```diff
@@ -21,6 +21,15 @@ fragment CeOpportunityFields on CeOpportunity {
   }
 }
 
+fragment ClientCeOpportunitySummaryFields on CeOpportunity {
+  ...CeOpportunitySummaryFields
+  project {
+    id
+    projectName
```
gigxz: The fields that were deprecated are not  removed from `CeOpportunitySummaryFields` above

src/api/operations/ce.fragments.graphql
```diff
@@ -21,6 +21,15 @@ fragment CeOpportunityFields on CeOpportunity {
   }
 }
 
+fragment ClientCeOpportunitySummaryFields on CeOpportunity {
+  ...CeOpportunitySummaryFields
```
gigxz, line 27: Seems like this resolves the eligibility requirements and priority scheme too, and active referral, do we need all that? In the UI it is just showing project name and type. We shouldn't resolve things we don't need since it will slow down the page load unnecessarily.

gigxz: That said I do think it makes sense to display the opportunity name in that table, maybe as the first column. Since currently there are duplicate rows with no visible differentiators indicating that they are different units.

src/modules/ce/components/client/ClientOpportunitiesTable.tsx
gigxz: Sure. "Position" is a new word not present anywhere else though, maybe we can just replace it with `opportunity` for now? Knowing that later we may want to find=>replace "opportunity"  with another word, for user-facing text.

src/modules/ce/components/client/ClientReferralsPage.tsx
gigxz: I don't think we should make any changes to this yet, it feels too early to be fiddling with the routes. We have yet to really decide, UX-wise, whether we want to commit to language of "referrals" or to also call things "ce".

src/modules/ce/components/client/ClientOpportunitiesTable.tsx
gigxz: I think its fine to leave it  out

src/modules/ce/components/ProjectReferralsTable.tsx
```diff
@@ -16,36 +16,58 @@ import {
...
   },
-  {
-    header: 'Started',
-    key: 'started',
-    render: (referral) => (
+  date: {
+    header: 'Referral Date',
+    key: 'date',
+    render: (referral: CeReferralTableFieldsFragment) => (
       <RelativeDateDisplay dateString={referral.createdAt} />
```
gigxz: "Referral Date" doesn't feel like the right header for a relative date description string. Another thing to align on with design. I would suggest just using the date-only format (`parseAndFormatDate`) to be consistent with the other tables on the Client dashboard. (Services, Assessments, Case notes -- they all have a formatted date as the first column.)

src/modules/ce/components/ReferralPage.tsx
```diff
@@ -80,7 +81,17 @@ const ReferralPage: React.FC<Props> = ({}) => {
             <Stack direction='row' alignItems='center' gap={0.5}>
               <Person sx={{ color: 'grayscale.main' }} />
               <Typography variant='body1'>
-                {clientNameFromRecordWithOptionalClient(referral)}
+                {referral.client ? (
+                  <RouterLink
+                    to={generateSafePath(ClientDashboardRoutes.REFERRALS, {
+                      clientId: referral.client.id,
```
gigxz: This feels kind of unnecessarily tricky because we will need to check whether the user has permission to view this page in particular. Lets just link to the client dashboard overview page, which we know the user can see if the `referral.client` is present.

src/types/gqlEnums.ts
```diff
@@ -75,18 +75,18 @@ export const HmisEnums = {
...
+  CeOpportunityStatus: { closed: 'Closed', locked: 'Locked', open: 'Open' },
   CeReferralStatus: {
-    accepted: 'accepted',
-    in_progress: 'in_progress',
-    initialized: 'initialized',
-    rejected: 'rejected',
+    accepted: 'Accepted',
+    in_progress: 'In progress',
+    initialized: 'Initialized',
+    rejected: 'Rejected',
```
gigxz, line 83: Just noting the discrepancy between these and the `ReferralStatusChip`. If we are using these as labels in some places, it will be inconsistent.

gigxz (approved):
looks good! some minor comments.

src/modules/ce/components/client/ClientOpportunitiesTable.tsx
martha: Sounds good, I'm happy to replace "position" with "opportunity," but I guess my initial question was broader -- Should we include this text at all? I'm not really sure where this text came from (copy/pasted from CAS maybe?) but some parts of it are unfamiliar to me and I'm wondering if they aren't applicable to ce-in-hmis, such as 
- "CoC Initial Review stage" -> maybe that is a specific step? but in that case this text isn't generic
- "Activating a match for any of these opportunities will cancel any existing match for that opportunity." -> that isn't implemented or allowed right now, I think

gigxz: Oh I see, sorry I didn't read it thoroughly. I does look copied from CAS, and not really relevant. Can you take a stab at some more descriptive text? We could leave a TODO but feels like one of those things we won't come back to. I think we want to hit:
- this client is eligible for the below opportunities
- these opportunities are all available
- the client may have any position on the waitlist. click in to the opportunity to find their waitlist position (will be possible once we introduce a search into the client waitlist)

src/api/operations/ce.fragments.graphql
martha: I un-deprecated them, per conversation https://github.com/greenriver/hmis-warehouse/pull/5294#discussion_r2052893847

src/api/operations/ce.fragments.graphql
martha: It is resolving these fields (copied from `gqlTypes.ts` line 16009):
```
export type ClientCeOpportunitySummaryFieldsFragment = {
  __typename?: 'CeOpportunity';
  projectType: ProjectType;
  id: string;
  name: string;
  categories: Array<string>;
  status: CeOpportunityStatus;
  expiresAt?: string | null;
  projectId: string;
  projectName: string;
};
```
I don't think it is resolving eligibility, priority, or referral. Those are resolved on `CeOpportunityFields` as opposed to `CeOpportunitySummaryFields`.

Yes, I can add the opportunity name to the table!

gigxz: Whoops my bad, I must have been looking at CeOpportunityFields by accident. Thanks!

src/types/gqlEnums.ts
martha: Yeah that's a good point. If it sounds good to you, I think we should be consistent around just using what's in the code for now, then we can discuss more as the designs finalize whether we want to diverge from that for user-facing language.

gigxz: agreed, thanks!

PR #856, opened 2024-07-24
Add staff assignment management to enrollment overview

martha:
## Description

[//]: # 'remove if not applicable'
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/4564
GH issue: https://github.com/open-path/Green-River/issues/6335
Figma: https://www.figma.com/design/F9EHR2IZYuMAlOJ9aOG2IL/OP-%2F-HMIS-%2F-2024-%2F-Q1-%2B-Q2-(Jan-1---June-30)?node-id=417-1961&t=go59iBv2Tl8nSyW9-0

## Type of change
[//]: # 'remove options that are not relevant'
- [ ] Bug fix
- [x] New feature (adds functionality)
- [ ] Code clean-up / housekeeping
- [ ] Dependency update

## Checklist before requesting review
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [x] My code includes comments and/or descriptive variable names to help other engineers understand the intent (or not applicable)
- [x] My code follows the style guidelines of this project (eslint)
- [x] I have updated the documentation (or not applicable)
- [x] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

PR #1134, opened 2025-04-15
Update staff assignment queries/fragments to fix bug & improve performance

gigxz:
## Description

Issue:  https://github.com/open-path/Green-River/issues/7563
Backend PR: https://github.com/greenriver/hmis-warehouse/pull/5297

See issue for testing instructions

## Type of change
- [x] Bug fix

### DIFF
--- a/src/api/operations/staffAssignment.fragments.graphql
+++ b/src/api/operations/staffAssignment.fragments.graphql
@@ -1,6 +1,7 @@
 fragment HouseholdWithStaffAssignments on Household {
   id
   householdClients {
+    # this is included because of "My Dashboard" query I think it should be split to 2 fragments
     id
     relationshipToHoH
     client {
@@ -25,11 +26,9 @@ fragment StaffAssignmentDetails on StaffAssignment {
   staffAssignmentRelationship
   assignedAt
   unassignedAt
-  household {
-    id
-  }
 }
 
+# Fragment to resolve details about each household assigned to the current user on My Dashboard
 fragment StaffAssignmentWithClients on StaffAssignment {
   id
   user {
--- a/src/api/operations/staffAssignment.fragments.graphql
+++ b/src/api/operations/staffAssignment.fragments.graphql
@@ -1,19 +1,9 @@
+# This fragment is resolved both on an individual household (staff assignment card)
+# AND on a table of households (staff assignment optional column)
 fragment HouseholdWithStaffAssignments on Household {
   id
-  householdClients {
-    # this is included because of "My Dashboard" query I think it should be split to 2 fragments
-    id
-    relationshipToHoH
-    client {
-      id
-      ...ClientName
-    }
-  }
-  staffAssignments(limit: 100) {
-    nodesCount
-    nodes {
-      ...StaffAssignmentDetails
-    }
+  currentStaffAssignments {
+    ...StaffAssignmentDetails
   }
 }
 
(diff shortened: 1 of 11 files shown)


src/modules/enrollment/components/pages/ClientEnrollmentsPage.tsx
```diff
@@ -126,6 +126,18 @@ const ClientEnrollmentsPage = () => {
     type: 'EnrollmentsForClientFilterOptions',
   });
 
+  const rowLinkTo = useCallback(
+    (enrollment: ClientEnrollmentTableFields) => {
+      if (!enrollment.access.canViewEnrollmentDetails) return;
```
gigxz, line 131: Table looks a little funny when some enrollments are viewable (clickable) and others aren't. Could use some improvement to communicate why to the user. Maybe like we show the menu but the View Enrollment menu items is disabled with a lock icon or something?

src/modules/enrollment/columns/HouseholdStaff.tsx
```diff
@@ -1,6 +1,9 @@
 import { Box, Chip, Tooltip } from '@mui/material';
 import React from 'react';
-import { HouseholdWithStaffAssignmentsFragment } from '@/types/gqlTypes';
+import {
+  HouseholdWithStaffAssignmentsFragment,
+  StaffAssignmentDetailsFragment,
+} from '@/types/gqlTypes';
 
 export const hasHouseholdWithStaff = (
```
martha: I think this is now unused and can be removed

src/modules/enrollment/components/pages/ClientEnrollmentsPage.tsx
martha: That seems reasonable to me! I don't think we have a pattern yet for the row primary action being disabled. But it does seem worth introducing something like that (rather than doing something hacky where non-viewable rows have a secondary action menu with the only item being a disabled version of the primary action)

src/modules/staffAssignment/components/StaffAssignmentCard.tsx
```diff
@@ -49,6 +51,14 @@ const StaffAssignmentCard: React.FC<StaffAssignmentCardProps> = ({
     return Object.entries(staff).sort(([a], [b]) => a.localeCompare(b));
   }, [household]);
 
+  const headOfHousehold = useMemo(() => {
+    if (!household) return;
+
+    return household.householdClients.find(
```
martha, line 57: We could use the `findHohOrRep` helper here. (But it would have slightly different behavior because it would use the first hh member, not the one whose enrollment dashboard we're on right now)

src/api/operations/staffAssignment.fragments.graphql
```diff
@@ -25,11 +16,9 @@ fragment StaffAssignmentDetails on StaffAssignment {
   staffAssignmentRelationship
   assignedAt
   unassignedAt
-  household {
```
martha, line 28: I was wondering why I even included this originally, since `StaffAssignmentDetails` is only ever resolved on a household or enrollment, where we already have the household ID. I thought maybe it was related to graphql caching. (see https://github.com/greenriver/hmis-frontend/pull/856#discussion_r1693010151?) But I clicked around, including assigning and unassigning staff, and wasn't able to create any weird behavior, so I think it makes sense to remove.

martha (approved):
All looks good, thank you Gig!

src/modules/staffAssignment/components/StaffAssignmentCard.tsx
gigxz: Good point. I copied this code from an existing file. That helper requires `enrollment { id }` to be present, which we don't resolve here - doesn't feel worth it for the bugfix so I am going to leave as-is

src/api/operations/staffAssignment.fragments.graphql
gigxz: Ah okay, thanks for poking around at that. I agree I don't see that it's necessary. I removed it as a performance improvement.

PR #1136, opened 2025-04-17
Add modal for assigning referral participants

martha:
## Description

Summary of changes: 
- Update the CE Referral object fetched by the Referral Context to additionally fetch Referral Participants
- Add modal for assigning contacts to a referral
- Update the step so that if there are participants for this step's swimlane, it will show their names

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5305

How to test:
- In local environment, create a referral (use starter pack)
- On the referral page, you should see a "Contacts" button
- You can click into the button to open a modal
- You should be able to assign contacts to the referral
- You should be able to unassign contacts

## Type of change
New feature


src/modules/ce/components/AssignContactsButton.tsx
```diff
@@ -0,0 +1,124 @@
...
+  });
+
+  if (error) throw error;
+
+  return (
+    <>
+      {swimlanes.length === 0 ? (
+        // If this referral has no swimlanes, disable the button.
+        // For now this will only happen if the referral also has no tasks,
+        // because the schema requires swimlane to be non-null on tasks.
```
martha, line 76: Let's discuss this. To expand on the comment a bit: The _database_ does not require swimlanes to exist for a template. But the _schema_ does require swimlane to exist on a task. So if the workflow has any tasks, this code would never run, because if it had tasks but no swimlanes, you'd see a schema error "cannot return null for non-nullable field."

However, you can see this code in action if you create a workflow with the template "No Tasks" (succeeds immediately).

src/modules/ce/components/ReferralStepAssignee.tsx
```diff
@@ -1,18 +1,43 @@
...
+            </Fragment>
+          );
+        });
+    }
+
+    if (assignees && assignees.length > 0) return assignees;
+
+    return <Chip size='small' label={swimlane} />;
+  }, [participants, swimlane]);
+
```
martha: TBD based on designs -- probably we actually want to show _both_ the assignees' names and the name of the swimlane?

src/modules/ce/components/AssignContactsButton.tsx
gigxz: This seems OK to me.  Now that the schema is updated, the check on line 65 could be simplified to `referral.swimlanes.length === 0`.

src/modules/ce/components/ReferralStepAssignee.tsx
```diff
@@ -1,18 +1,45 @@
...
 }
 const ReferralStepAssignee: React.FC<Props> = ({ step }) => {
-  const { swimlane } = step;
+  const { swimlane, assignees = [] } = step;
+  const { participants } = swimlane || {};
+
+  const content = useMemo(() => {
+    if (!swimlane) return null;
+
+    if (assignees && assignees.length > 0) {
```
gigxz:
```suggestion
    if (assignees.length > 0) {
```

src/modules/ce/components/ReferralStepAssignee.tsx
```diff
@@ -1,18 +1,45 @@
...
+            <Chip size='small' label={participant.name} />{' '}
+          </Fragment>
+        );
+      });
+    }
+
+    return <Chip size='small' label={swimlane.name} />;
+  }, [participants, swimlane, assignees]);
 
   if (swimlane)
```
gigxz: This can be removed, to make it clearer that this component always renders something

src/modules/ce/components/ReferralStepAssignee.tsx
```diff
@@ -1,18 +1,45 @@
 import Chip from '@mui/material/Chip';
 import Typography from '@mui/material/Typography';
-import React from 'react';
+import React, { Fragment, useMemo } from 'react';
 import { CeReferralStepSummaryFieldsFragment } from '@/types/gqlTypes';
 
 interface Props {
   step: CeReferralStepSummaryFieldsFragment;
 }
 const ReferralStepAssignee: React.FC<Props> = ({ step }) => {
```
gigxz, line 9: A comment about the difference between `assignee` and `participant` would be helpful on this component. Or is there a comment on the API/engine explaining it already?

I am not 100% sure of the intended behavior. I was testing out a "Review denial" workflow and I saw the following behavior:
1. Assign 1 user (not me) to "case managers"
2. Assign 2 users (not me) to "admins")
3. In the UI at this point, all tasks are unstarted, and I see the assigned users on the task cards
4. I click "Start" on the first task (assigned to CMs). The **Assigned** label switches to my name only.
5.  I submit that task, the next one appears (assigned to admins). Again admin task is assigned to the 2 users I chose.
6. I start the admin task. Now I see 3 names at the **Assigned** label -- the 2 users I picked, and me.

I'm not sure I understand the difference in behavior between step (4) and (6). In one case it replaced the old participants with the user who performed the task, in the other case it just added that person to the list.

gigxz (approved):
Looks good!
1 outstanding question about the assignee/participant behavior, but that probably belongs on the backend moreso than the frontend code.

src/modules/ce/components/ReferralStepAssignee.tsx
martha: Assignee vs. participant schema comment: https://github.com/greenriver/hmis-warehouse/pull/5305/files#diff-9a2e4184fa85c9a6f6a65383e8986f26ad77ff3c51716c6cb6a66cced4421b4dR25-R26

I'll add a comment here as well.

The behavior you are describing here is unexpected to me, though. I would have expected that (4) is the consistent behavior and (6) does not happen. I'll look into it more.

martha: Aha so, I figured out what is going on, and I'm grateful to you for pointing it out since I think it points to a bigger refactor that maybe should be made here. Here's what leads to the behavior you described above:
- When the referral is first started, only one task is available. The [engine](https://github.com/greenriver/hmis-warehouse/blob/release-162/drivers/hmis/app/models/hmis/workflow_execution/engine.rb#L29-L32) does `start_workflow`, `visit_node`, `assign_task` on that step, but there's no one to assign to, since there aren't any `ReferralParticipants` yet.
- You go in to the UI and create some `ReferralParticipants` that should apply to both steps, but no Assignments are created.
- You start the first task, which doesn't have any Assignments yet, only a Participant. Per the [recent change that's still in PR](https://github.com/greenriver/hmis-warehouse/pull/5305/files#diff-6ef5468bcbb80f7416024dddaabc3f32cbe0bd3587d0291d9ac9a38ef466115eR36), that task now is assigned to you. And in the frontend [`ReferralStepAssignees`](https://github.com/greenriver/hmis-frontend/pull/1136/files#diff-c7815e6b24fadc38ad627643932c18d034a35ccaf3d388e3f2d2a69d25a17839) component, Assignees take precedence over Participants, so your name replaces the Participant's name.
- You complete the first task, prompting the Engine to `visit_node` on any new tasks that now become available. Since there _are_ now Participants for this task, [this code here](https://github.com/greenriver/hmis-warehouse/blob/release-162/drivers/hmis/app/models/hmis/workflow_execution/engine.rb#L136) turns those Participants into Assignees.
- Now in the UI the second task appears with the users you selected represented as "official" Assignees, which is why when you start the task, your name is _added_ as an Assignee but it doesn't _replace_ the other Assignees (because they got "promoted" to actual Assignees, not just Participants.)

martha: I haven't played around with this yet, but I think I'm going to propose a refactor where the schema _doesn't_ share the Participants on a particular Step, but only the Assignees.

gigxz: Ok interesting, thanks for laying that out. I'm not sure I understand the direction you're going with the schema change. My first thought was that the _engine_ needs to change, and/or AssignReferralParticipants needs to change. 

This is going to be a bit of a word-vomit:

The description for the assignees field is `User(s) currently assigned to this step`. After creating a new referral, when I go to change the swimlane lists, I would expect those users to **become assigned** to any unstarted tasks, but they don't. We add them to the _swimlane_, but we don't add them as assignees to the task. (So presumably they wouldn't get notified?).

Something doesn't sit quite right with me. I think we really want to wait to run `assign_task!` on entry point steps until the swimlane contacts have been assigned/confirmed.

A scenario to consider:
- WorkflowTemplate has  a swimlane where "Gig" is the default assignee for swimlane Case Managers. The first task is assigned to Case Managers swimlane. (post- https://github.com/open-path/Green-River/issues/7538)
- When a new referral is created with this swimlane, Gig is immediately assigned to the first task. (Current engine behavior). She is notified right away.
- Now user goes through the workflow to assign contacts. Seemingly changing the CM swimlane selection doesn't affect the assignment of the first task, since it is already assigned to Gig (the default).

That certainly doesn't seem ideal, we don't want to send out notifications to the "default contacts" before the _actual_ contacts are assigned. I think this points even more strongly to the need for https://github.com/open-path/Green-River/issues/7539 since we really want to defer step assignment until contacts have been selected.

Maybe for 7539 we want `CreateCeReferral` to create the referral, but not start the workflow until `AssignReferralParticipants` is first called.
Is there an intermediate change we can make that will take us in that direction? Or should we just incorporate 7539?

martha: Thanks! Yeah I think we are kinda thinking in the same direction. Maybe the change I just pushed up to the backend https://github.com/greenriver/hmis-warehouse/pull/5305/commits/e478cae3b520c43ed1e49dbf6b995c90867f8c0d could be the "intermediate change" that takes us in the right direction. It proposes to update the referral step assignees whenever the participants are updated.

Behavior after this change would be:
- Referral is created with no default participants. The workflow is started, the first task doesn't have any assignees.
- A user goes in and assigns contacts in the UI. Now the assignees are created (and not yet notified, but I added a todo about that)
- In the future, if a referral is created with default participants, they would be notified right away
- If a user goes in and changes contacts in the UI, then new assignees would be created too, as appropriate. (And we could also delete old assignees, although I haven't implemented this yet.)

What do you think?

martha: The schema change I'm proposing is related, but separate. We added a fair amount of complication in the [backend](https://github.com/greenriver/hmis-warehouse/pull/5305/files#diff-9a2e4184fa85c9a6f6a65383e8986f26ad77ff3c51716c6cb6a66cced4421b4d) supporting Steps returning with both their Participants and their Assignees. Now that I understand the Participants/Assignees models and relationships a bit better, I'm wondering if it is really necessary for the Step to return its Participants at all in the schema, or could it be enough for the frontend to know about the step's Assignees only?

src/modules/ce/components/AssignContactFormItem.tsx
```diff
@@ -0,0 +1,60 @@
...
+  users,
+  setUsers,
+}) => {
+  const {
+    data: { pickList: staffPickList } = {},
+    loading: staffPickListLoading,
+    error: staffPickListError,
+  } = useGetPickListQuery({
+    variables: {
+      pickListType: PickListType.Users,
```
gigxz, line 28: Thinking ahead: Need a `TODO` here for permissions ticket (or thereafter) to introduce a new picklist (like `ELIGIBLE_REFERRAL_PARTICIPANT_USERS`) that respects the referral participation permission(s). We will need to pass a `projectId` to the picklist query here too, when we implement that.

src/modules/ce/components/AssignContactFormItem.tsx
```diff
@@ -0,0 +1,60 @@
...
+    },
+  });
+
+  const selectedUsers = useMemo(
+    () =>
+      users.map((userId) => {
+        return { code: userId };
```
gigxz: With the approach of only tracking `id`s, if the user is not in the picklist then they will appear as just a selected ID. Which the user won't be able to understand. Seems OK for now, but the approach may need to change depending on how we want to handle users who were assigned to a swimlane or task, then became inactive or lost permission to participate in referrals. Ticket is https://github.com/open-path/Green-River/issues/7615  maybe we can add a note about this in the code?

Or, a relatively simple change would be to store `users` in state as PickListOptions rather than `string[]`, so we can retain their names. (Looks like that would be a minor adjustment to `AssignContactsButton`).

src/modules/ce/components/ReferralStepAssignee.tsx
gigxz: I like this simplification a lot! Totally makes sense to me to just list the `assignees` on the task cards.

gigxz (approved):
🥳

src/modules/ce/components/AssignContactFormItem.tsx
martha: thanks, yeah I think that is a nice improvement!

PR #1141, opened 2025-04-21
Minor vite upgrade

gigxz:
## Description

resolve [CVE-2025-32395](https://github.com/advisories/GHSA-356w-63v5-8wf4) with a minor upgrade to vite and vitest

## Type of change
- [x] Dependency update

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2578,105 +2578,105 @@
     estree-walker "^2.0.2"
     picomatch "^2.3.1"
 
-"@rollup/rollup-android-arm-eabi@4.38.0":
-  version "4.38.0"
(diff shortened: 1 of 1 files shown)


martha (approved):
Looks good but I have a question -- I noticed the file changes in this PR are in `yarn.lock` only, with no change to `package.json`. Should we update vite in `package.json` too from `"vite": "^6.2.0",` to `"vite": "^6.3.0"`?

gigxz:
@martha that's because I upgraded the packages via `yarn upgrade`. This upgrades the packages to the latest version that matches our constraints in package.json. So for example, our package json states `"vite": "^6.2.0",`, then `yarn upgrade vite` will upgrade vite to the latest 6.x version (because of semantic meaning of `^`). It does not need to upgrade package.json in this case because the new version still satisfies the stated requirement.
If I had upgraded to 7.x (which doesn't exist), then package.json would need to be updated.

PR #1147, opened 2025-05-01
Display validation errors on bulk service assignment, add CommonSnackbar

gigxz:
## Description

* Display ValidationErrors returned from bulkAssignService mutation in a Snackbar
* Refactor common snackbar styles into `CommonSnackbar` and theme

https://github.com/open-path/Green-River/issues/7528

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5341

How to test:
* manually adjust bulk service assignment mutation to return an error, or trigger a state where it will. (for example: trying to `enroll + assign` client for a date one week ago, when they have an open enrollment starting yesterday)
  * test both assigning 1 person vs assigning multiple. all errors show (screenshot below)
* manually adjust bulk service assignment mutation to **raise** an error, that should still work as before (apollo alert snackbar)

## screenshots
Screenshot 2025-05-01 at 9 04 42 AM

Screenshot 2025-05-01 at 9 04 11 AM

## Type of change
Code cleanup, reduce sentry noise


src/components/elements/CommonSnackbar.tsx
```diff
@@ -0,0 +1,23 @@
...
+  open,
+  ...props
+}) => (
+  <Snackbar
+    TransitionProps={{ appear: false }}
+    open={open}
+    onClose={onClose}
+    anchorOrigin={{ vertical: 'top', horizontal: 'center' }}
+    {...props}
+    sx={{ marginTop: '100px', ...props.sx }}
```
martha, line 17: Maybe it would be possible to override `top` (which seems to be what MUI Snackbar uses internally) instead of adding `marginTop`? But I couldn't get this working locally without `important`

martha (approved):
Everything looks good. Thank you for the work to reduce sentry noise!

I wonder whether users will easily distinguish between these validation errors and other non-actionable errors, given that they are styled very similarly. Maybe that is something we can gather feedback on later.

src/components/elements/CommonSnackbar.tsx
gigxz: Oh odd, I didn't look at that, but maybe it has something to do with our sticky headers. I'm inclined to leave it as-is since it looks OK the way its appearing!

PR #1154, opened 2025-05-14
Implement frontend CE permissions

martha:
## Description

Summary of changes:
- Update schema to resolve new CE permissions
- (Not strictly related to this ticket) Rename schema arg `ceEnabled` to `includeCeFields`, which I think is more descriptive. I think `ceEnabled` might lead to confusion about the idea that the frontend might be telling the backend whether CE is enabled, which it definitely isn't/shouldn't!
- Update the `CommonTabs` component behavior:
  - If no tab definitions are found, return NotFound
  - If only one tab definition is passed, by default, it removes the Tab wrapping and just returns that tab's content.
  - These changes are to more gracefully support interfaces where the user might have permission to see one tab's contents but not another's. For example, on the Opportunity page a user may see opportunity details but not eligible clients.
- For assigning contacts, resolve the new, more restrictive picklist of users with permission to do referral steps in this project
- Across the app, hide content/actions the user doesn't have permission to view based on the newly added CE permissions
- Across the app, remove many usages of the "fake" permission `canViewCoordinatedEntry`. I think our goal is to get rid of this entirely, in favor of real permissions + project config. I've left some todos related to the project config ticket, as well.

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5368

How to test:
- With the backend branch checked out, update permissions and play around with functionality
- In general, users should not be able to view screens or action buttons for things they don't have permission to do. See CE in HMIS spec doc for permission details
- Example scenario: User has permission to "view own referrals" and "perform own referral tasks." If a referral task is assigned to that user,
  - They can see the referral
  - They can see all tasks
  - They can only _do_ their own task. Action buttons don't show up on the other tasks.

## Type of change
Feature build-out

### DIFF
--- a/src/api/operations/access.fragments.graphql
+++ b/src/api/operations/access.fragments.graphql
@@ -29,6 +29,7 @@ fragment RootPermissions on QueryAccess {
   # Application permissions:
   canViewMyDashboard
   canViewCoordinatedEntry
+  canAdministrateCoordinatedEntry
 
   # Root Project permissions:
   canEditOrganization # globally, determines whether user can create organization
--- a/src/api/operations/access.fragments.graphql
+++ b/src/api/operations/access.fragments.graphql
@@ -22,6 +22,9 @@ fragment RootPermissions on QueryAccess {
   canEditClients # globally, determines whether user can add new clients
   canViewDob # globally, determines whether to show col header as "DOB" or "Age"
   canViewClientAlerts
+  canViewClientEligibleOpportunities
+  canViewReferrals
+  canViewOwnReferrals
 
   # Application permissions:
   canViewMyDashboard
@@ -88,6 +91,13 @@ fragment ProjectAccessFields on ProjectAccess {
   canManageOutgoingReferrals
   canManageExternalFormSubmissions
   canSplitHouseholds
+  canViewReferrals
+  canViewOwnReferrals
+  canViewPrioritizedClientLists
+  canStartReferrals
+  canPerformAnyReferralTasks
+  canPerformOwnReferralTasks
+  canAssignReferralTasks
 }
 
 fragment OrganizationAccessFields on OrganizationAccess {
--- a/src/api/operations/access.fragments.graphql
+++ b/src/api/operations/access.fragments.graphql
@@ -22,9 +22,6 @@ fragment RootPermissions on QueryAccess {
   canEditClients # globally, determines whether user can add new clients
   canViewDob # globally, determines whether to show col header as "DOB" or "Age"
   canViewClientAlerts
-  canViewClientEligibleOpportunities
-  canViewReferrals
-  canViewOwnReferrals
 
   # Application permissions:
   canViewMyDashboard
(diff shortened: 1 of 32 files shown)


src/api/operations/access.fragments.graphql
```diff
@@ -22,10 +22,14 @@ fragment RootPermissions on QueryAccess {
   canEditClients # globally, determines whether user can add new clients
   canViewDob # globally, determines whether to show col header as "DOB" or "Age"
   canViewClientAlerts
+  canViewClientEligibleOpportunities
+  canViewReferrals
+  canViewOwnReferrals
```
gigxz: Can you add inline comments about why these need to be resolved globally? It's not immediately obvious to me why these would be global permissions. E.g. `canViewClientEligibleOpportunities` seems to be used to determine whether to show a specific tab on the Client Dashboard, which seems like something that should be resolved on the [client access fragment](https://github.com/greenriver/hmis-frontend/blob/50d10e6d3c3a9b075212163fa334de784e314c67/src/api/operations/access.fragments.graphql#L35-L38). 

We used to have a super bloated `RootPermissions` fragment which really slowed down initial page load times, so that's what I'm trying to avoid. It's also conceivable that you might have `canViewClientEligibleOpportunities` for some clients or not others (in the future, if thats not the case now) so it seems logical to resolve on the Client.

I think the same is true for canViewReferrals and canViewOwnReferrals, we're using them to determine some visibility on the Client dashboard.

Let me know if you have any questions/thoughts on the approach here

gigxz: We can use the `root_can` helper to resolve a global permission on the client, like we did here: https://github.com/greenriver/hmis-warehouse/blob/aeb811c07e13fca7e2bfbdd1cf0ff1178f5b849d/drivers/hmis/app/graphql/types/hmis_schema/client.rb#L180

Meaning its resolved on the client access fragment, but it doesn't actually check access against that client, it's still global. I think the idea was for the API to provide what the frontend needs, where it needs it. (As opposed to shoving everything global into the root fragment, or needing to resolve additional root fragment fields when viewing the client dashboard)

gigxz: Can also consider resolving a composite permission like `can_view_any_referrals` or `can_peform_any_tasks` etc [like we do for files](https://github.com/greenriver/hmis-warehouse/blob/aeb811c07e13fca7e2bfbdd1cf0ff1178f5b849d/drivers/hmis/app/graphql/types/hmis_schema/client.rb#L177), if its helpful. That moves the composite logic onto the backend where its more easily testable, and simplifies the frontend code. Take or leave that, just wanted to make sure you were aware of the pattern

gigxz: (Sorry to spam, last thought on this) thinking through the composite perm idea, it might not be wise to entrench the current UI too much. For example the frontend checks "has any perm: (view referrals, view own referrals, view waitlist)". It might not always be the case going forward that referrals are lumped in with the waitlist in the UI, so I wouldn't recommend lumping them together in the API (yet). So maybe we hold off on throwing in composite perms to keep things more fluid as the UI progresses.

martha: Thanks, cool, this is really helpful!
- Agreed, these perms don't need to be resolved globally. Thanks for the pointer on `root_can`, I updated to use this.
  - I added a small new capability on the `ClientRoute` component. Previously it always used a permission mode of "all"; now it can accept `permissionMode` as a prop.
- Agree about not resolving a composite permission yet

gigxz: great, thanks!!

gigxz (approved):
looks good to me!

gigxz:
Just noting that the "Create Opportunity" button still appears unpermissioned, but that seems ok since we'll be removing it relatively soon

PR #1155, opened 2025-05-15
Update referral workflow design

martha:
## Description

Summary of changes:
- Resolve new fields `updatedAt`, `updatedBy` on referral step
- Add a new `CommonButtonDrawer` component - we can refactor this later for phase 2 when we make the slideout panel.
- Refactor referral routes!
  - `REFERRAL_DETAILS` no longer exists
  - The overarching `REFERRAL` page `/projects/:id/ce/opportunities/:id/referrals/:id` now navigates to `REFERRAL_TASKS`
- Refactor `AssignContactsButton` into `AssignContactsForm` that now lives inside the contacts drawer
- Update to match the new designs - Referral page, ReferralStatusChip, ReferralStep, ReferralSteps, ReferralStepAssignee
- Update ReferralStepCard 
  - I used a Paper here instead of relying on `CommonCard/TitleCard` to match the designs
  - This isn't a click target as specified by designs; I kept the View/Start button from before, per discussion
  - I removed `ReferralStepCardInner`, which didn't seem as necessary to me anymore based on the new designs
  - Add storybook story
- Add `ReferralStepCardStatusChip` 
- Add to `FormActions` api to support right- and left-aligned buttons without a center aligned button

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5395

Create some opportunities and referrals and click around to check out the new designs:
- Assignment of contacts is moved into a drawer
- Navigating among tasks and using the Back button all works as expected
- Tasks should not be navigable if they are unavailable or not started (use the newly added Sequential workflow from the starter pack)

Outstanding questions, marked with todos:
- [ ] where will we display the client name on the referral page?
- [x] Will the whole-card area be clickable and if not (devs' preference), should we update the "locked" step card variant to comply with accessibility guidelines? - Resolved for now; see discussion below

## Type of change
New feature


src/modules/ce/components/ReferralStepCard.tsx
```diff
@@ -23,50 +34,83 @@ const ReferralStepCard: React.FC<Props> = ({ step }) => {
     opportunityId: string;
   };
   const { referral } = useReferralContext();
-  const { name, status } = step;
+  const { name, status, updatedBy, updatedAt } = step;
 
-  const action = useMemo(() => {
```
martha, line 28: instead of 3-dots menu, it should be the same action button as before - view or start

src/modules/ce/components/ReferralStep.tsx
```diff
@@ -97,51 +100,90 @@ const ReferralStep: React.FC<Props> = ({}) => {
...
+          referralId,
+        })}
+      >
+        Back to All Tasks
+      </Paper>
+      <CommonCard title={name}>
+        <Stack gap={2}>
+          <ReferralStepAssignee step={step} />
+          <Divider />
+          {/* TODO - discuss this "Start step" button. Is this status distinction important to the user? */}
```
martha: decision:
- when people navigate into the referral, they will go to the overarching referral page
- navigating directly to unstarted step isn't really supported. Return not found

gigxz:
Looks like there is an accessibility violation being flagged in UI on a related component

martha:
@gigxz , thanks, yes, I added that a11y test to demonstrate the issue I described here:
https://www.figma.com/design/znYqGn21VZO5zF2LjpnUiL?node-id=4991-26298&m=dev#1258931224

Hm, maybe it's preferable to just make a decision on the development side and move forward without waiting for design input. I can update the colors to comply with wcag standards.

gigxz:
I see, sorry hadn't reviewed yet I just noticed the failed build. Thanks for fixing

src/modules/ce/components/ReferralPage.tsx
```diff
@@ -57,34 +53,39 @@ const ReferralPage: React.FC<Props> = ({}) => {
...
-            divider={<Divider orientation='vertical' flexItem />}
-          >
-            <Stack direction='row' alignItems='center' gap={0.5}>
-              <Person sx={{ color: 'grayscale.main' }} />
-              <Typography variant='body1'>
+            <Box>
+              <Typography variant='overline'>
+                Referral to {referral.opportunity.name}
+              </Typography>
+              <Typography variant='h3' component='h1'>
```
martha, line 82: We're waiting for input from design on this question of where to display the client name. Here's how I've implemented it for now:

Screenshot 2025-05-22 at 2 59 26 PM

martha:
@gigxz I think this should be ready for review now, thanks!

src/modules/ce/components/ReferralStepStatusChip.tsx
```diff
@@ -0,0 +1,71 @@
...
+            ...baseChipSx,
+            backgroundColor: 'warning.surface',
+            color: 'warning.darkest',
+          }}
+        />
+      );
+    case CeReferralStepStatus.InProgress:
+      return (
+        <Chip
+          label='Started'
```
gigxz: Designs show label as "In Progress", which feels easier to differentiate from "not started" imo. Did I miss a different iteration of the designs?

src/modules/ce/components/ReferralStepCard.tsx
```diff
@@ -51,22 +54,72 @@ const ReferralStepCard: React.FC<Props> = ({ step }) => {
...
+    if (locked)
+      return {
+        color: 'grayscale.main',
+      };
+  }, [locked]);
+
+  const lastUpdatedString = useMemo(() => {
+    const dateString = parseHmisDateString(updatedAt);
+    const at = dateString ? ` ${formatRelativeDateTime(dateString)}` : '';
+    const by = updatedBy ? ` by ${updatedBy.name}` : '';
```
gigxz: This doesn't feel like something that should live in a referral step component. Maybe we need a param on the existing `lastUpdatedBy` util to specify whether to use a relative or exact date?

src/modules/ce/components/ReferralStepAssignee.tsx
```diff
@@ -1,33 +1,47 @@
...
   step: CeReferralStepSummaryFieldsFragment;
 }
 const ReferralStepAssignee: React.FC<Props> = ({ step }) => {
   const { swimlane, assignees = [] } = step;
 
+  const locked = step.status === CeReferralStepStatus.Unavailable;
+
+  const commonChipSx = useMemo(() => {
+    return {
+      borderRadius: 1,
```
gigxz: These styles are duplicative with styles specified in ReferralStepStatusChip, maybe we need a `CommonRectangularChip` or something? I opened a thread [on figma](https://www.figma.com/design/znYqGn21VZO5zF2LjpnUiL?node-id=5424-153399#1270285775) because I don't quite understand the visual distinction between the rounded vs rectangular chips. So maybe hold off until we can resolve that!

src/modules/ce/components/ReferralStatusChip.tsx
```diff
@@ -8,35 +8,43 @@ interface Props {
...
+          }}
           color='primary'
-          icon={<InProgressIcon />}
+          icon={
+            <InProgressIcon
+              sx={{
+                '&.MuiChip-icon': {
+                  color: 'primary.main',
+                },
+              }}
```
gigxz: Can we move these styles to `theme.ts` so we have a standard style for `<Chip color="primary" />`? thanks!
https://mui.com/material-ui/react-chip/#color-chip

src/modules/ce/components/ReferralStatusChip.tsx
```diff
@@ -8,35 +8,43 @@ interface Props {
 }
 const ReferralStatusChip: React.FC<Props> = ({ status }) => {
   const baseChipSx: SxProps = {
-    borderWeight: '1px',
-    borderColor: 'borders.main',
-    borderStyle: 'solid',
     fontWeight: 600,
+    fontSize: '14px',
+    backgroundColor: 'grayscale.surface',
```
gigxz: likewise lets either move styles to theme, or leave them with the default style. specifying font size in pixels here is not ideal.

src/modules/ce/components/ReferralPage.tsx
gigxz: Sounds good. My thoughts are:
- client name shouldn't be linked. (client may have multiple records; client may not be in hmis; user may have access to see their name but not their profile; user may accidentally click away from current page; etc). we can add links to hmis/warehouse dashboards to the client detail  panel
- I think it adds too much height/padding to the top bar.. I think thats the nature of using overline text though, maybe only solvable by having one horizontal title
- overline text color doesn't match our usual approach in `PageTitle` where overline text is the dark primary

src/modules/ce/components/AssignContactsForm.tsx
```diff
@@ -0,0 +1,86 @@
...
+    },
+    onCompleted: () => onClose?.(),
+  });
+
+  if (error) throw error;
+
+  return (
+    <Stack gap={2} mt={2}>
+      {referral.swimlanes?.map((swimlane) => (
+        <AssignContactFormItem
```
gigxz, line 69: I'm able to fully clear a swimlane. It seems like when I do that, and then a future step with no assignees should become available, it still remains locked  (the backend returns an 
initialized unavailable step).

Are you able to repro that? Maybe it has nothing to do with empty swimlanes, and is some other issue with how my workflow is set up. I'm using the "Enrollment Creator" template from the starter pack.

Repro steps:
- assign contants
- clear contacts
- perform move-in date step
- next step doesn't become available

src/modules/ce/components/ReferralStepCard.tsx
```diff
@@ -51,22 +54,72 @@ const ReferralStepCard: React.FC<Props> = ({ step }) => {
...
+          color='grayscale'
+        >
           View
         </ButtonLink>
       );
     }
-  }, [opportunityId, projectId, referral.id, status, step]);
+  }, [name, opportunityId, projectId, referral.id, status, step]);
+
+  const collapsed = step.status === CeReferralStepStatus.Completed;
```
gigxz: I think its fine to collapse the assignees, but we still need to be able to click in to view the step

gigxz: To simplify maybe remove this `collapsed` state?

src/modules/ce/components/AssignContactsForm.tsx
gigxz: Nevermind, I couldn't reproduce it myself. It was an old referral I had sitting around. 
And I remember we have https://github.com/open-path/Green-River/issues/7615

src/routes/routes.ts
```diff
@@ -116,8 +116,7 @@ const projectDashboardRoutes = {
   // CE
   CE: 'ce',
   OPPORTUNITY: 'ce/opportunities/:opportunityId',
-  REFERRAL: 'ce/opportunities/:opportunityId/referrals',
-  REFERRAL_DETAILS: 'ce/opportunities/:opportunityId/referrals/:referralId',
+  REFERRAL: 'ce/opportunities/:opportunityId/referrals/:referralId',
```
gigxz: @martha what do you think of making this page just be at `projects/:id/ce/referrals/:referralId`?
When navigating there from `/projects/:id/ce#referrals` it injects opportunity into the path and breadcrumbs, which seems unnecessary. I also hit a permission NotFound page on the breadcrumb when testing a user who could see the project and the assigned referral, but not the opportunity page.

As discussed earlier we may also need to expose this page as "floating" outside of the project dashboard, in which case it also won't really make sense to rely on `opportunityId` being in the route. I'm not sure if we rely on that currently anyway, the referralId should be enough to get everything we need.

gigxz (approved):
looks good! left some minor suggestions

src/modules/ce/components/ReferralPage.tsx
martha: Sounds good. I will remove the overline text for now and we can discuss with design next week.

One solution could be to put the name of the opportunity into the breadcrumbs. "Project ABC -> Coordinated Entry -> One Bedroom Unit -> Referral for Kathleen Strock" and then the title of the page could just be "Referral for Kathleen Strock." But this wouldn't be possible if we go with your other comment about rearranging so referrals aren't "under" opportunity.

src/modules/ce/components/ReferralStepAssignee.tsx
martha: Sounds good, I'm fine to hold off on introducing square chips until we discuss more with design.
Screenshot 2025-05-23 at 3 15 09 PM

src/modules/ce/components/ReferralStatusChip.tsx
martha: Thanks @gigxz . I will merge this now, but feel free to add more feedback (especially about https://github.com/greenriver/hmis-frontend/pull/1155/commits/a157a35bc70d495133eb249a0e89a775ad0921e9) and I can address in a later review.

PR #1159, opened 2025-05-22
Add Referrals to User Dashboard

martha:
## Description

Summary of changes:
- Update `PageContainer` to use `PageTitle` internally, for better consistency and shared code
- Rename "my" dashboard to "user" dashboard
  - **Including in the route /my-dashboard -> /dashboard**, with a redirect for backwards compatibility
- Introduce new component `TooltipKeyValue` for new pattern described in Figma
- Update the Dashboard page to include the Referral steps assigned to this user

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5419/files

How to test:
- With backend and frontend code checked out, visit /dashboard route and check out the new assigned tasks widget

## Type of change
New feature


src/modules/userDashboard/components/YourReferrals.tsx
```diff
@@ -0,0 +1,56 @@
...
+  GetUserCeAssignedStepsQuery,
+  GetUserCeAssignedStepsQueryVariables,
+  UserCeReferralStepFieldsFragment,
+} from '@/types/gqlTypes';
+
+const YourReferrals = () => {
+  const { user } = useAuth();
+  if (!user) return <NotFound />;
+
+  // TODO(#7507) - using this "table" to display cards results in some dom validation issues that cause console warning.
```
martha: I think it's acceptable to merge with the console warning, having added a todo and a comment here: https://github.com/open-path/Green-River/issues/7507. But let me know if you agree.

src/modules/userDashboard/components/TooltipKeyValue.tsx
martha, line 1: I haven't put this in our common components dir yet because it's only used here for now, and I have some hesitations about introducing it as a wider pattern. (See figma thread.)

src/modules/userDashboard/components/YourReferrals.tsx
```diff
@@ -0,0 +1,56 @@
...
+  // When we refactor GenericTable to use hooks, we should consider supporting non-table layouts more gracefully.
+  return (
+    <>
+      <Paper>
+        <Stack gap={1} m={2}>
+          <Typography component='h2' variant='h4'>
+            Referrals
+          </Typography>
+          <Typography variant='caption'>
+            Referral steps assigned to you
```
gigxz:
```suggestion
            Referral tasks assigned to you
```
On the referral page they are called "tasks". I think this language also makes it clearer that there is an action required. We should probably revisit other places where we refer to tasks as steps 
(eg "current step" col header)

src/modules/userDashboard/components/YourReferrals.tsx
```diff
@@ -0,0 +1,56 @@
...
+          GetUserCeAssignedStepsQueryVariables,
+          UserCeReferralStepFieldsFragment
+        >
+          queryVariables={{
+            id: user.id,
+          }}
+          queryDocument={GetUserCeAssignedStepsDocument}
+          columns={[]}
+          pagePath='user.ceAssignedSteps'
+          noData='No referral steps assigned to you'
```
gigxz: tasks

src/components/layout/nav/ToolbarMenu.tsx
```diff
@@ -16,10 +16,10 @@ export const TOOLBAR_MENU_ITEMS: (Required<
   })[] = [
   {
     permissions: ['canViewMyDashboard'],
-    path: Routes.MY_DASHBOARD,
-    id: 'navToMyDashboard',
-    activeItemPathIncludes: 'my-dashboard',
-    title: 'My Dashboard',
+    path: Routes.USER_DASHBOARD,
+    id: 'navToUserDashboard',
+    activeItemPathIncludes: 'dashboard',
```
gigxz: I'm not seeing the nav item be in an active state when I'm on the dashboard:
Screenshot 2025-05-23 at 1 16 45 PM

src/components/layout/PageContainer.tsx
```diff
@@ -1,22 +1,26 @@
...
+  overlineText?: string;
   actions?: ReactNode;
+  maxWidth?: ContainerProps['maxWidth'];
 }) => {
   const isTiny = useIsMobile('sm');
   return (
     <Container
       component='main'
-      maxWidth='lg'
+      maxWidth={maxWidth}
```
gigxz: Reviewed with Ash at design standup - suggested just leaving this at `lg` so all the page containers match
```suggestion
      maxWidth='lg'
```

src/components/layout/PageContainer.tsx
```diff
@@ -1,22 +1,26 @@
...
       sx={{ px: { xs: 1, sm: 3, lg: 4 }, pt: 4, pb: 6 }}
```
gigxz: Reviewed at design standup - reduce padding when there is overline text so the main title is on the same heigh as other pages (client search, projects)
```suggestion
      sx={{ px: { xs: 1, sm: 3, lg: 4 }, pt: overlineText ? 1 : 4, pb: 6 }}
```

src/modules/userDashboard/components/YourReferrals.tsx
gigxz: It looks like the approach we took with the Client Search results cards is to wrap each card in td and tr. 

This change looks fine to me:
```jsx
          renderRow={(step) => (
            <tr key={step.id}>
              <td>
                <UserStepCard
                  step={step}
                  currentUserId={user.id}
                />
              </td>
            </tr>
          )}
```
And then on UserStepCard, changing `my: 1` to `mb: 1`

src/modules/userDashboard/components/YourReferrals.tsx
```diff
@@ -0,0 +1,56 @@
...
+    <>
+      <Paper>
+        <Stack gap={1} m={2}>
+          <Typography component='h2' variant='h4'>
+            Referrals
+          </Typography>
+          <Typography variant='caption'>
+            Referral steps assigned to you
+          </Typography>
+        </Stack>
```
gigxz:
```suggestion
      <CommonCard title='Referrals' padContent={false} sx={{ pb: 1 }}>
```

- discussed at design standup: remove caption description
- I would suggest using CommonCard on the User Dashboard. I can't think of a reason we should use custom cards here. Using CommonCard will ensure that all of the cards on the dashboard have the same styles. Here I added extra padding below the cards (Also pointed out by ash)

src/modules/staffAssignment/components/YourClients.tsx
```diff
@@ -101,19 +101,21 @@ const MY_CLIENTS_COLUMNS: ColumnDef<StaffAssignmentWithClientsFragment>[] = [
...
-      </Typography>
       <Paper>
+        <Stack gap={1} m={2}>
+          <Typography component='h2' variant='h4'>
+            Your Clients
+          </Typography>
+          <Typography variant='caption'>
+            Households with active enrollments assigned to you
+          </Typography>
+        </Stack>
```
gigxz: Suggestion:
```jsx
      <CommonCard title='Your Clients' padContent={false}>
        <Typography variant='body2' color='text.secondary' sx={{ px: 2 }}>
          Households with active enrollments assigned to you
        </Typography>
```
- discussed at standup that subtitle was supposed to be body2 (figma did list caption)
- per other comment, use a shared card component
- 'text.secondary' to match mocks

src/modules/userDashboard/components/TooltipKeyValue.tsx
```diff
@@ -0,0 +1,43 @@
...
+  Icon,
+  children,
+  ...typographyProps
+}) => {
+  const isMobile = useIsMobile('sm');
+
+  return (
+    <Tooltip title={title} arrow>
+      <Typography
+        variant='body2'
```
gigxz: discussed in design standup: these are supposed to be `caption`

src/api/operations/ce.queries.graphql
```diff
@@ -146,3 +146,16 @@ query GetAdminCeReferrals(
     }
   }
 }
+
+query GetUserCeAssignedSteps($id: ID!) {
+  user(id: $id) {
+    ceAssignedSteps {
```
gigxz:
```suggestion
    id
    ceAssignedSteps {
```

I'm getting the following warning. I think this should resolve it, untested
Screenshot 2025-05-23 at 3 17 59 PM

src/modules/userDashboard/components/TooltipKeyValue.tsx
martha: Sounds good. "Caption" feels like a misnomer to me, since these are not really captions _to_ anything, if that makes sense. I wonder if we should introduce a new body3 for even tinier text. Relates to thread in design library https://www.figma.com/design/bsw8CX4UOq6SIdXYorZDWV?node-id=2978-20483&m=dev#1152226440

src/api/operations/ce.fragments.graphql
```diff
@@ -180,3 +180,27 @@ fragment CeReferralStepFields on CeReferralStep {
   }
   submittedValues
 }
+
+fragment UserCeReferralStepFields on CeReferralStep {
```
gigxz, line 185:
```suggestion

# used for resolving assigned steps on the User Dashboard
fragment UserCeReferralStepFields on CeReferralStep {
```

src/api/operations/userDashboard.fragments.graphql
```diff
@@ -0,0 +1,24 @@
+fragment UserDashboardFields on UserDashboard {
```
gigxz: can we remove this? appears unused

src/routes/AppRoutes.tsx
```diff
@@ -98,6 +99,12 @@ const ProtectedRoutes: React.FC<{ user: HmisUser }> = ({ user }) => {
 const AppRoutes = () => {
   const { user } = useAuth();
 
+  // Although we don't use it here, fetch the dashboard config for determining
+  // whether to show the Dashboard link in the toolbar. This batches the query
+  // with the other root-permissions query, avoiding the awkwardness of the
+  // Dashboard link flickering into place after the rest of the page has loaded.
+  useGetUserDashboardConfigQuery();
```
gigxz: This is performing the query any time someone loads the login page, resulting in a 401, because the user isn't authenticated yet.

What about fetching the dashboard config in the same place we fetch root permissions (`MainLayout`)? I think it's the same idea for the dashboard config, we want to block page load until both are done fetching.

The MainLayout component also renders `ToolbarMenu`, so we could also pass the dashboard config value directly down to that child, rather than relying on reading it from the apollo cache.

src/routes/protected.tsx
```diff
@@ -165,7 +165,12 @@ export const protectedRoutes: RouteNode[] = [
     element: <App />,
     children: [
       { path: Routes.ALL_PROJECTS, element: <AllProjectsPage /> },
-      { path: Routes.MY_DASHBOARD, element: <MyDashboardPage /> },
+      { path: Routes.USER_DASHBOARD, element: <UserDashboardPage /> },
+      // For backwards compat, old '/my-dashboard' redirects to '/dashboard'
```
gigxz, line 169: nice, thanks!

src/modules/userDashboard/components/YourReferrals.tsx
```diff
@@ -0,0 +1,47 @@
...
+        <GenericTableWithData<
+          GetUserCeAssignedStepsQuery,
+          GetUserCeAssignedStepsQueryVariables,
+          UserCeReferralStepFieldsFragment
+        >
+          queryVariables={{}}
+          queryDocument={GetUserCeAssignedStepsDocument}
+          columns={[]}
+          pagePath='userDashboard.ceReferralSteps'
+          noData='No referral tasks assigned to you'
```
gigxz, line 30: I'm seeing a border under this 🤔  it should probably be fixed in the underlying table component, or as a workaround:
`noData={<>{'No referral tasks assigned to you'}</>}`

src/modules/userDashboard/components/YourReferrals.tsx
```diff
@@ -0,0 +1,47 @@
...
+  UserCeReferralStepFieldsFragment,
+} from '@/types/gqlTypes';
+
+const YourReferrals = () => {
+  const { user } = useAuth();
+  if (!user) return <NotFound />;
+
+  return (
+    <>
+      <CommonCard title='Referrals' sx={{ pb: 1 }}>
```
gigxz: Hm I think this padding was based on my suggestion but maybe the table changed it in a way that looks like too much whitespace at the bottom. I am getting in the weeds here but I think it's nice to have it consistently padded with `2` all around... here's one way to achieve that:


Screenshot 2025-05-28 at 12 10 11 PM


```diff
diff --git a/src/modules/userDashboard/components/UserStepCard.tsx b/src/modules/userDashboard/components/UserStepCard.tsx
index 91aaedc6..ccb4de9c 100644
--- a/src/modules/userDashboard/components/UserStepCard.tsx
+++ b/src/modules/userDashboard/components/UserStepCard.tsx
@@ -37,7 +37,7 @@ const StepCardInfo: React.FC<{
         sx={{
           display: 'inline-flex',
           alignItems: 'center',
-          gap: 1,
+          gap: 0.5,
           mr: isMobile ? 0 : 2,
         }}
       >
@@ -86,7 +86,6 @@ const UserStepCard: React.FC<Props> = ({ step, currentUserId }) => {
   return (
     <Card
       sx={{
-        mb: 1,
         backgroundColor: 'primary.surface',
         borderColor: 'primary.100',
         transition: (theme) =>
diff --git a/src/modules/userDashboard/components/YourReferrals.tsx b/src/modules/userDashboard/components/YourReferrals.tsx
index aa615012..d2485d3e 100644
--- a/src/modules/userDashboard/components/YourReferrals.tsx
+++ b/src/modules/userDashboard/components/YourReferrals.tsx
@@ -17,7 +17,7 @@ const YourReferrals = () => {

   return (
     <>
-      <CommonCard title='Referrals' sx={{ pb: 1 }}>
+      <CommonCard title='Referrals'>
         <GenericTableWithData<
           GetUserCeAssignedStepsQuery,
           GetUserCeAssignedStepsQueryVariables,
@@ -31,6 +31,8 @@ const YourReferrals = () => {
           paginationItemName='task'
           defaultPageSize={10}
           recordType='CeReferralStep'
+          // add padding after each row except the last one
+          tableProps={{ sx: { 'tbody > :not(:last-child) td': { pb: 1 } } }}
           renderRow={(step) => (
             <tr key={step.id}>
               <td>
```

gigxz: Oops that first `gap` change is unrelated I was playing with that, I think it looks better a bit smaller so there is more differentiation between the gap between items and the gap between icon->text. But dont feel strongly

gigxz (approved):
Looks great!!

src/modules/userDashboard/components/YourReferrals.tsx
martha: Hm, does that fix it for you? I'm still seeing a border either way. Thanks for pointing that out!
Screenshot 2025-05-28 at 2 07 37 PM
Something like a `dontWrapNoData` prop seems messy/annoying, but maybe correct to add. I'm worried that modifying `GenericTable` to treat a `ReactNode` differently from a `string` might have unanticipated knock-on effects on other tables.

martha: I think I will propose `noDataCellStyles` as a new prop

src/modules/userDashboard/components/YourReferrals.tsx
martha: Thanks, I like your suggested changes. Here's what it looks like when the tasks are paginated:
Screenshot 2025-05-28 at 2 19 39 PM

gigxz: Oh I didn't think to test pagination, thanks. Looks good!

src/modules/userDashboard/components/YourReferrals.tsx
gigxz: Hmm okay. The "no data" display has gotten zero love. I think it could potentially be fine to just ensure there is no bottom border directly [here](https://github.com/greenriver/hmis-frontend/blob/a766573c0d239115a293a20b05da1e0fcbbdd959/src/components/elements/table/GenericTable.tsx#L421)? I don't see why we would want a border on the "no data" message. If the table is occurring inside a bordered card, it doesn't make a difference anyway.

src/modules/userDashboard/components/YourReferrals.tsx
martha: Regarding the `gap`, I thought about this too, and was a bit torn. Our [design library](https://www.figma.com/design/bsw8CX4UOq6SIdXYorZDWV/OP-HMIS---Design-Library?node-id=2462-16646&p=f&t=PWEYEVMRkZ4RKnhT-0) calls for a pretty radical rewrite of the spacing in our app:
- currently we use MUI's default of spacing(1) = 8px
- design library states that spacing(1) should be 5px
- and on mobile, spacing(1) should be 4px(!)

so in the figma here, the spacing is 5, corresponding to (1) under the new system, but 8px under the old. That's why I went with the wider gap, hoping it would save us from going back and fixing it later when we do https://github.com/open-path/Green-River/issues/7363.

gigxz: ok fair enough!

src/routes/AppRoutes.tsx
martha: Hrm so, the `MobileMenu` also renders a `ToolbarMenu`, and `MobileMenu` is also rendered by the different dashboard contexts (project, client, enrollment) and I think it's not simple to pass down new children to them.

I remember at the time we might not have been too happy with this arrangement but decided to leave it due to time. Maybe it's time to revisit soon.

For now I will revert so that `ToolbarMenu` fetches what it needs

PR #1163, opened 2025-05-29
Add "floating referral" page for non-project-permissioned users

martha:
## Description

Merge notes:
- ~This PR targets https://github.com/greenriver/hmis-frontend/pull/1159, which should be merged first.~

Summary of changes:
- Refactor CE components into organized directories.
- Create a "floating" Referral page outside of the project context. 
- When linking to a Referral from outside of the Project context (on Client Referrals and User Dashboard), check the new access fragment to determine whether the user has access to view this project. If not, render it as "floating" referral.
- Within the Referral world, add attributes to the Referral outlet context so it can return helpers for generating links. The idea is for the child components (ReferralStep, etc.) to just use those helpers, instead of needing to know whether they are being rendered within a Project context or not.

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5435

How to test:
- Find a user who has "can view own referrals" and "can perform own referral tasks," but NOT "can view project". Make sure they have some referral tasks assigned to them.
- Impersonate that user
- Navigate to the user Dashboard (or the Client Referrals page)
- Make sure the user can click the referral link and see the referral, even though they aren't seeing it under the project context.

## Type of change
Non-user-facing bug fix


src/modules/ce/components/referral/ReferralPage.tsx
```diff
@@ -12,29 +12,36 @@ import {
...
+  // Referral is sometimes, but not always, rendered in the context of a Project.
+  project?: ProjectAllFieldsFragment;
+}
 
 /**
  * Provides navigational context for a single Referral.
  * Uses react router outlet to render sub-pages.
  */
-const ReferralPage: React.FC<Props> = ({}) => {
+const ReferralPage: React.FC<Props> = ({ project }) => {
```
martha, line 42: Open question: When a referral is rendered in this new "floating" context, the target project name is no longer visible anywhere. (When rendered in the project context, the target project is in the breadcrumbs.) Is this acceptable?

src/modules/ce/components/referral/ReferralWayfinder.tsx
```diff
@@ -36,10 +27,6 @@ const ReferralWayfinder: React.FC<Props> = ({ open, onClose }) => {
           alertTitle='Referral Completed'
           alertText={`${clientName} has been accepted to ${opportunity.name}`}
           items={[
-            {
```
martha, line 39: I think this Wayfinding dialog requires a bit more thought in general. 
- Where you should be able to go, after finishing a referral, should depend on your permissions in ways that aren't yet reflected here. E.g. if you got here from User Dashboard and don't have any other permissions in the project, you probably should never see "go back to opportunity/project/anything project related" 
- On the other hand, if you do have permission to view/manage units in the project, I think it's appropriate that one of the wayfinding steps after a _rejected_ referral would be, "go back to the opportunity page and see who else is eligible to get referred here." But maybe we just need to rename it from "back to opportunity" to something different.

I'm tempted to leave this out of scope and maybe create a followup ticket so we don't forget to return to this dialog with design later.

src/modules/ce/components/referral/ReferralPage.tsx
gigxz: No, the provider should be able to see the project name. maybe just adjust the page title to `Referral for <client> to <project>`?

src/modules/ce/components/referral/ReferralWayfinder.tsx
gigxz: Agreed I find it a bit confusing. Seems like something we can come back to in a bug bash / design QA pass

gigxz (approved):
looks good!

PR #1166, opened 2025-05-31
Release 164

ttoomey:
## Description

Summary of changes:
Fixes but in interaction between auto fill and calculated values, as seen on income
This change means that display-only fields (where item.type === ItemType.Display) will always be autofilled, even if they're marked as dirty. Assumption is that display fields can't be edited by users so there's no need to preserve their value.

Depends on hmis-warehouse PR:

See GH issue

## Type of change
Bug fix

### DIFF
--- a/graphql.schema.json
+++ b/graphql.schema.json
@@ -4206,6 +4206,22 @@
             "deprecationReason": null
           },
           {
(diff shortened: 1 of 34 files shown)


ttoomey:
got the wrong branch, closing

PR #1168, opened 2025-06-01
Unit group management and unit/opportunity rework

gigxz:
## Description

https://github.com/open-path/Green-River/issues/7753

Summary of changes:
* Add support for creating unit groups
* Add unit group page
* Replace opportunity page with unit page

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5444 and https://github.com/greenriver/hmis-warehouse/pull/5458

~~**WAIT TO DEPLOY UNTIL** https://github.com/greenriver/hmis-warehouse/pull/5458 is in place~~ - this is no longer relevant. Updated to keep "legacy" behavior of managing units without groups for non-CE installations. This will need to be addressed before we enable CE outside of the QA environment.

**How to test**:
- [x] with CE enabled, test all unit and CE referral workflows
- [x] with CE disabled, test unit management workflows
  * test unit assignment during enrollment
- [x] test with and without "can view units" and "can manage units"
- [x] test all unit filters/columns

## Type of change
New feature


gigxz:
@martha I marked this ready for review! merging is still blocked by https://github.com/greenriver/hmis-warehouse/pull/5458 as discussed - I will get started on that
However https://github.com/greenriver/hmis-warehouse/pull/5444 is safe to merge.

src/api/operations/units.queries.graphql
```diff
@@ -12,12 +13,38 @@ query GetUnits(
...
       nodesCount
       nodes {
-        ...UnitFields
+        ...UnitTableRowFields
       }
     }
   }
 }
 
+query GetUnitGroups($id: ID!, $limit: Int = 10, $offset: Int = 0) {
```
martha: Maybe `GetProjectUnitGroups`?

src/api/operations/ce.fragments.graphql
```diff
@@ -1,12 +1,16 @@
...
-  categories
   status
   active
-  expiresAt
   projectId
   projectName
+  unit {
+    id
+    name
+    # unit type / unit group?
```
martha: (Not necessarily in this PR), I do think it would be nice to resolve the unit group name so it can be shown on the Project Available Units table.

src/modules/ce/components/project/ProjectCePage.tsx
```diff
@@ -33,6 +25,14 @@ const ProjectCePage: React.FC = () => {
       });
     }
 
+    if (project.access.canViewUnits) {
+      defs.push({
+        title: 'Available Units',
+        key: 'available-units',
+        // TODO add helper: these units are available for referrals
```
martha: Flagging todo - this is just indicating that we want to add helper text here, right?

src/modules/ce/components/client/ClientOpportunitiesTable.tsx
```diff
@@ -83,9 +83,9 @@ const ClientOpportunitiesTable: React.FC = () => {
           noData='No opportunities'
```
martha, line 89: Maybe add some commentary in this file about why we choose to keep the "opportunities" language in this table/file, even though we've replaced it with "units" in many other tables? I think I agree with the choice -- just to think it through/articulate why...
- From the perspective of looking at a Client, it makes more sense to ask the question "what opportunities is this person eligible for?", rather than "what units is this person eligible for?"
- Whereas from the Project context, it makes more sense to ask "What units are in this project that are accepting referrals?" rather than "what opportunities are there at this project?"

Anyway, if you had additional reasoning for keeping it here it would be cool to add that in a comment too!

src/modules/ce/components/referral/ReferralPage.tsx
```diff
@@ -66,9 +69,9 @@ const ReferralPage: React.FC<Props> = ({ project }) => {
           projectId: project.id,
           referralId: referral.id,
         }),
-        opportunityPath: generateSafePath(ProjectDashboardRoutes.OPPORTUNITY, {
+        unitPath: generateSafePath(ProjectDashboardRoutes.UNIT, {
```
martha: Maybe we can only return `unitPath` at all if `referral.opportunity.unit` is non-null here? (It looks like the child components gracefully handle the case where `unitPath` is null because it will always be null in the "floating" context, but if we try to generate a path when unit is null, we will get an error, I think.)

src/modules/units/components/UnitUtilizationByUnitType.tsx
```diff
@@ -0,0 +1,50 @@
...
+  unitTypes,
+  variant = 'stacked',
+}) => {
+  const hasMultipleUnitTypes = unitTypes.length > 1;
+
+  return (
+    <Stack gap={1}>
+      {unitTypes.map((unitType, idx) =>
+        variant === 'stacked' ? (
+          <Stack gap={0.25}>
```
martha:
```suggestion
          <Stack key={unitType.id} gap={0.25}>
```
to avoid console warning

src/modules/units/components/CreateUnitsDialog.tsx
```diff
@@ -0,0 +1,162 @@
...
+                  setUnitType(option.code);
+                }
+              }}
+              maxWidth={400}
+            />
+          </Grid>
+          <Grid item xs={12}>
+            <NumberInput
+              label={getRequiredLabel('Number of Units to Add', true)}
+              value={numberOfUnits ?? ''}
```
martha, line 140: Let's add a limit (both here and in the backend) to the number of units you are allowed to bulk add? I'm able to get my dev environment to hang for some time by telling it to create 100000 units.

martha: I'm also seeing a silent (swallowed) error by telling it to create 1000000000000000 units (not castable to int). Adding a `<ApolloErrorAlert error={errorState.apolloError} />` solves the swallowing, which I guess we should do anyway so that apollo errors are surfaced; but maybe this should be a validation error that's fixable by the user

src/modules/units/components/ProjectUnitsTable.tsx
```diff
@@ -0,0 +1,174 @@
...
+          ariaLabel: `${label} ${unit.unitGroup.name}`,
+          to: generateSafePath(ProjectDashboardRoutes.UNIT_GROUP, {
+            projectId: project.id,
+            unitGroupId: unit.unitGroup.id,
+          }),
+        });
+      }
+
+      // If unit is occupied, link to hoh Enrollment
+      const occupant =
```
martha: I think there is something funny going on here with `includeClientOccupants`. When the "Client Occupants" column is not selected, this linking logic does not work, because `unit.occupants` is undefined.

It seems like a similar situation to `exitDate`, where we still need to resolve it on Enrollments in the various enrollment tables, even though it is an optional column, because that field is also needed to resolve the status.

src/modules/units/components/Units.tsx
```diff
@@ -94,57 +66,42 @@ const Units = () => {
...
             </Button>
           </ProjectPermissionsFilter>
         }
       />
       <Stack gap={4}>
-        <TitleCard title='Capacity' headerSx={{ p: 2 }}>
-          <UnitCapacityTable projectId={project.id} />
-        </TitleCard>
+        {unitGroups.length > 0 && (
+          <Masonry columns={{ xs: 1, md: 2 }} spacing={2}>
```
martha: New (to us) MUI component unlocked 😎 

Take or leave this -- I noticed the extra right-margin, and googled it. Seems related to this issue: https://github.com/mui/material-ui/issues/34234

That issue is stale, but there are some workarounds suggested. I tried adding
```
MuiMasonry: {
  styleOverrides: {
    root: {
      width: 'auto',
    },
  },
},
```
to our `theme.ts` and it seems to have the intended effect. But I haven't really battle-tested it to see if there are any other unwanted side effects.

src/modules/units/components/UnitGroupCard.tsx
martha, line 1: I noticed something strange going on with the unit capacity display when there are multiple unit groups in a project. I think this is a backend bug, but I'll add my comment here since the backend work is already merged.

When I set up a second unit group with some units in it that share a type with units in the first unit group, the second group's total capacity and vacancies show correctly, but the bar chart displaying the per-unit-type breakdown shows information from the first unit group that's inaccurate for this second unit group. See screenshot:
Screenshot 2025-06-04 at 2 52 38 PM

Poking around in the console, I think it is related to the apollo client caching results by ID. The `UnitTypeCapacity` object's ID corresponds to the ID of the `Hmis::UnitType` model. But in this case we want to differentiate it by unit group. Maybe we can get around this with adding some kind of composite ID on the backend.
Screenshot 2025-06-04 at 3 05 10 PM

src/modules/units/components/UnitGroupCard.tsx
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    <Paper sx={{ px: 2, pt: 1, pb: 2 }}>
+      <Stack
+        justifyContent='space-between'
+        direction='row'
+        alignItems='flex-start'
+      >
+        <div>
+          <Typography variant='overline'>Unit Group</Typography>
+          <Typography variant='h5' sx={{ mb: 1 }}>
```
martha:
```suggestion
          <Typography component='h2' variant='h5' sx={{ mb: 1 }}>
```

martha: Also, I'm not totally sure about all merits/disadvantages of this approach, but we could put the overline text _in_ the heading text. That way it would get read by screen readers as part of this heading, as opposed to part of the previous section's content. We do this in `PageTitle.tsx`:
https://github.com/greenriver/hmis-frontend/blob/b325723d685dac1216bee699f17547703401daae/src/components/layout/PageTitle.tsx#L32-L40

src/modules/units/components/UnitGroupFormDialog.tsx
```diff
@@ -0,0 +1,164 @@
...
+    <Dialog open={open} onClose={handleClose} fullWidth maxWidth='sm'>
+      <DialogTitle>Create Unit Group</DialogTitle>
+      <DialogContent sx={{ my: 2 }}>
+        {errorState.errors.length > 0 && (
+          <Box sx={{ mb: 2 }}>
+            <ErrorAlert key='errors' errors={errorState.errors} fixable />
+          </Box>
+        )}
+        <ApolloErrorAlert error={errorState.apolloError} />
+        <Grid container spacing={2}>
```
martha: Could this be simplified as just a `Stack` as opposed to a `Grid`?

src/modules/units/components/UnitManagementTable.tsx
```diff
@@ -122,6 +114,16 @@ const UnitManagementTable = ({
         rowSecondaryActionConfigs={rowSecondaryActionConfigs}
         loading={loading}
         loadingVariant='linear'
+        // Only link to Unit page if CE is enabled. For now we don't have anything non-CE to show.
```
martha: Commented out purposefully, or do you want to add this back in?

src/modules/units/components/UnitPage.tsx
```diff
@@ -0,0 +1,106 @@
...
+    },
+  });
+
+  // Set the breadcrumb so it says the correct name of this unit and group
+  useEffect(() => {
+    if (!unit) return;
+
+    overrideBreadcrumbTitles({
+      [ProjectDashboardRoutes.UNIT]: unit.name,
+      // fixme unable to inject unit group name
```
martha: Fixme flag - do you want to solve this on this PR? I think it is ok to leave for later. Maybe we should start a catchall ticket for design/bug bash followups?

src/modules/units/components/UnitPage.tsx
```diff
@@ -0,0 +1,106 @@
...
+    const defs = [];
+
+    defs.push({
+      title: 'Overview',
+      key: 'overview',
+      contents: <UnitOverview unit={unit} />,
+    });
+
+    const opportunity = unit.latestOpportunity;
+    if (opportunity && project.access.canViewPrioritizedClientLists) {
```
martha: Maybe we want to also hide the "Eligible Clients" tab for an opportunity that's already closed? I was clicking around and thought it could be confusing to show, especially under that name, since it's not really true anymore that those clients are "eligible" for this opportunity.

On the other hand, I guess I could imagine someone wanting to find out "who else _was_ eligible for this opportunity at the time?"

src/modules/units/components/UnitPage.tsx
```diff
@@ -0,0 +1,106 @@
...
+  if (loading) return <Loading />;
+  if (error) throw error;
+  if (!unit) return <NotFound />;
+
+  return (
+    <>
+      <PageTitle title={unit.name} />
+      <Box sx={{ mb: 2 }}>
+        <UnitReferralStatus unit={unit} />
+        {/* TODO button to mark as available/unavailable? */}
```
martha, line 98: This seems useful, especially once we have "mark available as of future date" functionality

martha (approved):
All looks great. A couple of my comments relate to minor bugs, they could be fixed here or on a future PR, no strong preference either way from me.

Thank you!!

src/modules/units/components/UnitPage.tsx
martha: I started collecting a few more of these, so I created https://github.com/open-path/Green-River/issues/7799 for a place to put my notes somewhere shareable.

Feel free to remove the note about this one on that ticket, if you decide to address it in this PR!

src/modules/units/components/CreateUnitsDialog.tsx
gigxz: thanks, added!

src/modules/ce/components/client/ClientOpportunitiesTable.tsx
gigxz: Good point. I added a comment, but I'm not tied to "opportunity". This may be clearer as "Vacancies" or "Available Units" etc.

src/modules/units/components/Units.tsx
gigxz: Thanks I didn't even notice that. Added on the component instead of the theme since I'm not sure its always applicable / we don't use it elsewhere yet

src/modules/units/components/UnitPage.tsx
gigxz: Its a nice to have, removed the TODO for now

src/modules/units/components/UnitPage.tsx
gigxz: Agreed I think it makes sense to hide in this case! updated

src/modules/units/components/ProjectUnitsTable.tsx
gigxz: The "View Enrollment" secondary menu action only appears if the "Client Occupants" column is selected. I think that's probably OK, leaving as is

src/modules/units/components/UnitGroupCard.tsx
gigxz: Good find, thanks!
Fixing on the backend by adding `object.id` to the id string here to make it unique:
https://github.com/greenriver/hmis-warehouse/blob/cfc17f17060b70d1cd7801f0fe60268252aa4d09/drivers/hmis/app/graphql/types/hmis_schema/unit_group.rb#L44

src/modules/units/components/UnitManagementTable.tsx
gigxz: Removed because table has bulk actions

src/modules/ce/components/referral/ReferralPage.tsx
gigxz: Good call, updated!

PR #1172, opened 2025-06-06
Enable bulk marking units available/unavailable

martha:
## Description

Summary of changes:
- Add new shared component `GenericMutationButton` with confirmation dialog
- Add bulk actions to Unit Management table
- Add `awaitRefetch` logic to the existing bulk-delete button, so the loading state is clearer to the user
- Update the existing `useUnitCeActions` to use the new logical helpers `canBeMarkedAvailable` etc.

Issue: https://github.com/open-path/Green-River/issues/7773
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5467

Screenshot 2025-06-06 at 5 21 06 PM

## Type of change
New feature


src/components/elements/table/EnhancedTableToolbar.tsx
```diff
@@ -42,9 +42,11 @@ const EnhancedTableToolbar = <T extends { id: string }>({
       }}
     >
       <Stack
-        justifyContent='space-between'
-        alignItems={'center'}
-        direction={isTiny ? 'column' : 'row'}
+        justifyContent={stackVertically ? undefined : 'space-between'}
```
martha, line 45: These are unrelated improvements to the EnhancedTableToolbar on mobile. Before (current QA)


After (local)

src/modules/dataFetching/components/GenericMutationButton.tsx
```diff
@@ -0,0 +1,93 @@
...
+    Pick<ConfirmationDialogProps, 'title' | 'confirmText'>;
+  children?: ReactNode;
+  dialogContent: ReactNode;
+  queryDocument: TypedDocumentNode<Mutation, MutationVariables>;
+  variables: MutationVariables;
+  idPath: string;
+  onSuccess?: VoidFunction;
+};
+
+const GenericMutationButton = <Mutation, MutationVariables>({
```
martha, line 28: We could definitely consider combining this with `DeleteMutationConfirmationDialog`/`Button`. I haven't done that here due to time constraint.

Or we could eliminate it and move this code into `UnitBulkActionButton`, which is its only usage at the moment.

src/modules/units/hooks/useUnitCeActions.tsx
```diff
@@ -51,32 +50,28 @@ export const useUnitCeActions = ({
...
           actions.push({
             title: 'Mark as Unavailable for Referrals',
             key: 'markUnavailable',
             ariaLabel: `Mark Unit ${unit.id} as Unavailable for Referrals`,
             onClick: () => {
               markUnitsUnavailable({ variables: { unitIds: [unit.id] } });
             },
-            disabled:
-              unit.latestOpportunity.status === CeOpportunityStatus.Locked,
-            disabledReason:
```
martha, line 69: I want to call out that, with this update to using the new helpers from the backend, we _are_ losing something: some Mark Unavailable actions that would have formerly been disabled with a `disabledReason` will instead no longer appear.

My feeling is that the previous logic was already hard to follow (and therefore hard to maintain), and it's a worthwhile tradeoff to keep it in the backend. Hopefully the columns "Occupancy" and "Referral Status" help with the user's understanding of why the action may be missing. But let me know what you think!

src/components/elements/table/EnhancedTableToolbar.tsx
gigxz: great, thanks!

src/modules/dataFetching/components/GenericMutationButton.tsx
gigxz: this looks useful, thanks!

src/modules/units/hooks/useUnitCeActions.tsx
gigxz: agreed!

gigxz (approved):
🥳

PR #1182, opened 2025-06-18
Add referral history

martha:
## Description

Summary of changes:
- Resolve new Events fields on the Referral
- Add a new `ReferralTimeline` component for rendering the events
- Update the `CommonButtonDrawer` so that its children are responsible for padding

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5498

How to test:
- Visit a Referral and open the Activity panel
- Activity events should appear for referral start, step completion, and referral completion (accept/reject)

## Type of change
New feature

### DIFF
--- a/graphql.schema.json
+++ b/graphql.schema.json
@@ -5125,6 +5125,47 @@
             "deprecationReason": null
           },
           {
+            "name": "events",
+            "description": null,
+            "args": [
+              {
+                "name": "limit",
+                "description": null,
+                "type": {
+                  "kind": "SCALAR",
+                  "name": "Int",
+                  "ofType": null
+                },
+                "defaultValue": null,
+                "isDeprecated": false,
+                "deprecationReason": null
+              },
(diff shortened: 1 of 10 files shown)


martha:
Screenshot:
Screenshot 2025-06-18 at 11 21 43 AM

A few questions to confirm with design:
- [x] Excluding sort/filter for now. Should we still include “Last active x days ago” bar? - Leave out the bar.
- [x] Icons - I've skipped them; should we nail down their semantic meaning and map them to event types? Or just skip for now and include them later? - Skip for now is fine
- [x] Drop shadow - do we want to introduce this as a pattern? (MUI library only includes box shadow around the whole box, so this is custom styling for us. I would probably want to include it in our design library and guidelines on when to use it) - Exclude shadow for now and use border stroke

gigxz (approved):
looks good!

PR #1186, opened 2025-06-24
Clear cache when household is fully joined

martha:
## Description

Summary of changes: This PR clears the Household entry from the cache after a household ID no longer exists post-join. This prevents the user from seeing a 500 error as described on the linked ticket.

Depends on hmis-warehouse PR: Relates to https://github.com/greenriver/hmis-warehouse/pull/5523, but doesn't depend on it

How to test: See test instructions on the ticket

## Type of change
Bug fix

### DIFF
--- a/src/modules/household/hooks/usePerformJoinHousehold.tsx
+++ b/src/modules/household/hooks/usePerformJoinHousehold.tsx
@@ -1,5 +1,6 @@
 import { useCallback, useState } from 'react';
 import { clientBriefName } from '@/modules/hmis/hmisUtil';
+import { cache } from '@/providers/apolloClient';
 import {
   HouseholdClientFieldsFragment,
   HouseholdFieldsFragment,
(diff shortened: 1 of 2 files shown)


src/modules/household/hooks/usePerformJoinHousehold.tsx
```diff
@@ -25,6 +28,12 @@ export function usePerformJoinHousehold({
       if (data.joinHousehold) {
         setJoinedHousehold(data.joinHousehold.receivingHousehold);
         setRemainingHousehold(data.joinHousehold.donorHousehold || undefined);
+
+        if (donorHouseholdId && !data.joinHousehold.donorHousehold) {
+          // There were no remaining members in the donor household, so clear it from the cache
+          cache.evict({ id: `Household:${donorHouseholdId}` });
```
gigxz, line 36: Maybe add a comment explaining: "if there _are_ remaining household members in the donor household, the donor household gets updated in the cache by the donorHousehold resolved on the mutation return value. Thats why we only need to evict from the cache explicitly if there is no donorHousehold resolved"

That wasn't immediately obvious to me (why we have this `if` condition for cache eviction)

gigxz (approved):

src/modules/household/hooks/usePerformJoinHousehold.tsx
martha: good point, thanks!

PR #1190, opened 2025-06-27
Unit waitlist ui updates

gigxz:
## Description

* display time when waitlist was last updated (time when someone was added/removed, doesn't mean that the whole list was refreshed at that time)
* display message if candidate pool doesn't exist yet
Screenshot 2025-06-27 at 3 12 07 PM

## Type of change
new feature


src/modules/ce/components/unit/PrioritizedClientsTable.tsx
```diff
@@ -81,18 +86,49 @@ const PrioritizedClientsTable: React.FC<Props> = ({ opportunity }) => {
...
+  // If CandidatePool has not been generated yet (due to change in eligibility or prioritization requirements), show a message
+  if (!opportunity.candidatesGeneratedAt) {
+    return (
+      <Alert severity='info'>
+        The waitlist for this unit has not been generated yet. Please check back
+        later.
+      </Alert>
+    );
+  }
+  const candidatesGeneratedAt = parseHmisDateString(
```
martha, line 98: Could use `RelativeDateDisplay`'s `getFormattedDates` helper? Maybe move that out into the `hmisUtil`? Seems like getting the relative and absolute date from the hmis date string is going to become an increasingly common pattern in the code (since our design library recommends it almost everywhere dates are shown), so I like the idea of having a common helper that does both

src/modules/ce/components/unit/PrioritizedClientsTable.tsx
```diff
@@ -81,18 +86,49 @@ const PrioritizedClientsTable: React.FC<Props> = ({ opportunity }) => {
...
-      GetCeOpportunityCandidatesQueryVariables,
-      CeCandidateFieldsFragment
-    >
-      columns={columns}
-      queryVariables={{ opportunityId: opportunity.id }}
-      queryDocument={GetCeOpportunityCandidatesDocument}
-      pagePath='ceOpportunity.candidates'
-      paginationItemName='candidates'
-    />
+    <Stack rowGap={2}>
```
martha, line 103: Is there any reason to specifically use `rowGap` here and not just `gap`, which I think of as our normal pattern? (but I'm grepping and seeing that `rowGap` is also used too, so not sure)

martha (approved):
All looks good. Thanks for the schema cleanup too. 🚂

src/modules/ce/components/unit/PrioritizedClientsTable.tsx
gigxz: Yes that sounds like a good future improvement, going to merge as is for now

src/modules/ce/components/unit/PrioritizedClientsTable.tsx
gigxz: no reason in particular, I think it maybe doesnt wrap so the col gap might not be necessary

PR #1202, opened 2025-07-11
Add AHA score mutation and form item

martha:
## Description

Summary of changes:
- Add `clientId` as a new local constant
- Create `AhaScore`, a new `DynamicGroup` type

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5563

To do:
- [x] Disable re-fetch

## Type of change
new feature (custom client build)


src/modules/form/components/DynamicFormFields.tsx
```diff
@@ -29,7 +29,7 @@ const DynamicFormFields: React.FC<Props> = ({ handlers, ...props }) => {
       let value = baseValue;
       // Convert string values to appropriate number types
       if (item) {
-        if (item.type === ItemType.Integer) value = parseInt(value) || 0;
+        if (item.type === ItemType.Integer) value = parseInt(value) || null;
```
martha: @gigxz let's definitely discuss this change before merging. I am a bit hesitant about what existing behavior might break with this change, but in this specific case, it seems like we would not want to save a 0 when the fetched value is null and/or unparseable. (0 being an allowable value for this score.)

src/modules/external/aha/components/AhaScore.tsx
```diff
@@ -0,0 +1,134 @@
...
+const EXPECTED_LINK_IDS = [
+  SCORE_LINK_ID,
+  ALT_AHA_FLAG_LINK_ID,
+  DW_CLIENT_LINK_ID,
+  GENERATOR_LINK_ID,
+];
+
+// Custom component for rendering AHA score. Renders a button to fetch the score, as well as sub-items including:
+// score (value), alt-AHA flag, and hidden items for values that should be submitted but not shown to the user.
+// Similar to DisabilityGroup, this component is tightly coupled to its expected form structure.
```
gigxz, line 25: Clarify expected structure of item
```suggestion
// Similar to DisabilityGroup, this component is tightly coupled to its expected form structure.
//
// `item` is expected to be of type 'GROUP', and have child items with certain link_ids (EXPECTED_LINK_IDS).
```

src/modules/external/aha/components/AhaScore.tsx
```diff
@@ -0,0 +1,134 @@
...
+        setHasScore(false);
+        return;
+      }
+
+      setErrorState(emptyErrorState);
+
+      if (data.fetchAhaScore?.score) {
+        setHasScore(true);
+
+        if (severalItemsChanged) {
```
gigxz, line 65: I feel like this prop shouldn't be optional (doesn't need to be) but that's a preexisting issue. Hopefully this doesn't arise

src/modules/external/aha/components/AhaScore.tsx
```diff
@@ -0,0 +1,134 @@
...
+              [SCORE_LINK_ID]: data.fetchAhaScore.score,
+              [ALT_AHA_FLAG_LINK_ID]: data.fetchAhaScore.altAhaFlag,
+              [DW_CLIENT_LINK_ID]: data.fetchAhaScore.dwClientId,
+              [GENERATOR_LINK_ID]: data.fetchAhaScore.generator,
+            },
+            type: ChangeType.User,
+          });
+        }
+      } else {
+        // TODO(#7812) If score is not available (or bad quality?), calculate alt-AHA
```
gigxz: I don't think this is a TODO for this code. The logic for whether to prompt for Alt-AHA will likely be defined in the Form Definition, based on whether the `alt_aha_flag` is 0 or 1 (or null?)

src/modules/external/aha/components/AhaScore.tsx
```diff
@@ -0,0 +1,134 @@
...
+        setHasScore(false);
+      }
+    },
+    onError: (apolloError) => {
+      setErrorState({ ...emptyErrorState, apolloError });
+      setHasScore(false);
+    },
+  });
+
+  // console.error instead of throwing an error, so we can still preview how the form looks in non-client contexts
```
gigxz, line 86: Can we pass a dummy clientId value in from the `FormPreview` component instead? we aren't getting any kind of alerts on console errors

src/modules/external/aha/components/AhaScore.tsx
```diff
@@ -0,0 +1,134 @@
...
+    return EXPECTED_LINK_IDS.every((expectedLinkId) => {
+      return item.item?.find((i) => i.linkId === expectedLinkId);
+    });
+  }, [item]);
+
+  const [errorState, setErrorState] = useState<ErrorState>(emptyErrorState);
+
+  // hasScore is a boolean that can be undefined when the user hasn't clicked the fetch button yet.
+  // `false` indicates the user did click the button but the score wasn't available.
+  const [hasScore, setHasScore] = useState<boolean | undefined>(undefined);
```
gigxz: Just to make sure I'm understanding this, is it a combination of "has score" and "has fetched" eg
undefined= haven't fetched yet
true = fetched and found score
false = fetched and didn't find score
?

I think using a boolean state with 3 states is making the render logic kind of hard to follow. For future readability/maintainability, I'd suggest splitting it out into 2 states (has fetched and has score). Or you can rely on the [called](https://www.apollographql.com/docs/react/data/mutations#result) result value on the mutation to know whether the mutation has been called.

src/modules/form/components/DynamicFormFields.tsx
gigxz: Yes lets extract this change, it feels risky. I don't disagree-- forcing it to zero seems odd-- but I don't feel comfortable with making such a broad change as part of a very specific custom integration work. For example I think this code wouldve broken saving anything as zero, because `0 || null` = `null`. 
Can you find a workaround for AHA? Maybe we can save as -1 to indicate that there was no score?

gigxz (changes requested):
Looks good! only requested change is to back out the DynamicFormFields change

src/modules/form/components/DynamicFormFields.tsx
martha: Sure, I've reverted it. That workaround seems reasonable. This is in the `itemChanged` callback, so it's not called anymore unless a score is returned. https://github.com/greenriver/hmis-frontend/pull/1202/commits/ab5985993761937478ad94b522bff28aaa531cc3

src/modules/external/aha/components/AhaScore.tsx
martha: Yes - I briefly looked at making this required, but it's optional because we reuse the same type on view-only groups, I think. Let's leave it as is for now.

src/modules/external/aha/components/AhaScore.tsx
martha: Thanks! I'll use a second piece of state. `called` is set to true as soon as the mutation is called, when it hasn't returned yet, at which point we don't yet want to show the "score unavailable for this client" message

PR #1204, opened 2025-07-15
Rename Coordinated Entry => Referrals Project dash, don't allow both legacy and CE referrals

gigxz:
## Description
https://github.com/open-path/Green-River/issues/7915
Summary of changes:
* Hide Legacy referral page in the project if Coordinated Entry is enabled there
* Some TODOs for https://github.com/open-path/Green-River/issues/7321 once we have additional configuration in place
* Allow Unit page to be rendered under the ce tab, so you're not jumping to another page from "availble units" table

## Type of change
New feature

### DIFF
--- a/src/components/layout/PageTitle.tsx
+++ b/src/components/layout/PageTitle.tsx
@@ -7,11 +7,13 @@ const PageTitle = ({
   overlineText,
   actions,
   endElement,
+  tabbedPage,
 }: {
   title: ReactNode;
   overlineText?: string;
   endElement?: ReactNode; // element to appear directly to the right of the title
   actions?: ReactNode; // element to float to the right (e.g. action button)
+  tabbedPage?: boolean; // whether this is a tabbed page, which will reduce the margin below the page title
 }) => {
   const isTiny = useIsMobile('sm');
 
@@ -22,7 +24,7 @@ const PageTitle = ({
       justifyContent={'space-between'}
       width={isTiny ? 'fit-content' : '100%'}
       sx={{
-        mb: isTiny ? 1 : 3,
+        mb: isTiny || tabbedPage ? 1 : 3,
         alignItems: isTiny ? 'left' : 'center',
         // fixed min height (if not mobile), so height is the same whether there are actions or not
         minHeight: isTiny ? '' : '40px',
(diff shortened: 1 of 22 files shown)


src/components/layout/PageTitle.tsx
```diff
@@ -7,11 +7,13 @@ const PageTitle = ({
   overlineText,
   actions,
   endElement,
+  tabbedPage,
 }: {
   title: ReactNode;
   overlineText?: string;
   endElement?: ReactNode; // element to appear directly to the right of the title
   actions?: ReactNode; // element to float to the right (e.g. action button)
+  tabbedPage?: boolean; // whether this is a tabbed page, which will reduce the margin below the page title
```
gigxz: this is not good because the CommonTabs have the collapse feature, so we dont know if there will be tabs or not. look at another appraoch

src/components/layout/dashboard/contextHeader/useDashboardBreadcrumbs.tsx
```diff
@@ -164,6 +164,10 @@ export const useProjectBreadcrumbConfig = (
         title: 'Unit',
         parent: ProjectDashboardRoutes.UNITS,
       },
+      [ProjectDashboardRoutes.CE_UNIT]: {
```
martha, line 167: I'm trying to brainstorm, what are the other examples of times where we render the same page under different routes, with different breadcrumbs?

So far the only example I can think of is the Referral route, which can be rendered both within and outside of the Project context. That situation might have a stronger justification than this one, I think, because the same user wouldn't see both pages. If you have permission to see the Project, you would only ever view the referral in the Project context.

Is there a similar justification here -- where a user who can manage Units would only see the Unit page from within the Unit context, and other users would see the Unit page from within the CE context?

martha (approved):
Looks good! Thanks for the todos, those are helpful 🙏

src/components/layout/dashboard/contextHeader/useDashboardBreadcrumbs.tsx
gigxz: Good question, nope it's really just a context thing! I discussed this with Ash: Without this change, following the typical workflow of Project>Referrals>Available Units>click a unit> would bounce you out of the left-hand `Referrals` tab (fka Coordinated Entry) and into the `Units` tab under Configuration. We agreed that this was a jarring experience, it's confusing how to go back to the previous page. Imo it makes sense to be able to view those Units directly under the Referral context since it's part of that workflow.

PR #1207, opened 2025-07-17
Remove default filtering from admin referral table

gigxz:
## Description

* Remove default filters from the Admin referrals table. In-progress referrals are already floated to the top by default.  A few reasons:
  * The use-case for this table is not strongly known. For example you might be going there to review pending denials, vs review stalled referrals, etc. A default filter is annoying if it's not what you need.
  * Now that we have Custom Status support, Initialized/InProgress might not even be the true representation of an in progress referral. In that case, the table could even show up empty by default.
* Make some more columns hideable, turns out https://github.com/open-path/Green-River/issues/7832 may not be a bug (or at least not, for everyone..)
* 
## Type of change
bug/feature

### DIFF
--- a/src/modules/ce/components/admin/AdminReferralsTable.tsx
+++ b/src/modules/ce/components/admin/AdminReferralsTable.tsx
@@ -15,7 +15,6 @@ import {
 } from '@/routes/routes';
 import {
   CeReferralAdminFieldsFragment,
-  CeReferralStatus,
   GetAdminCeReferralsDocument,
   GetAdminCeReferralsQuery,
   GetAdminCeReferralsQueryVariables,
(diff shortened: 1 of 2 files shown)


src/modules/ce/components/admin/AdminReferralsTable.tsx
```diff
@@ -136,12 +134,6 @@ const AdminReferralsTable: React.FC<Props> = ({}) => {
       >
         columns={COLUMNS}
         queryVariables={{}}
-        defaultFilterValues={{
-          referralStatus: [
-            CeReferralStatus.Initialized,
-            CeReferralStatus.InProgress,
-          ],
-        }}
```
gigxz, line 144: We also have a default filter like this on `ProjectReferralsTable`. I think we need to remove it from there too.

If an installation has a custom status called `Assigned` that represents an intermediate in-progress status, it should not be excluded from the referral table by default. Am I correct in understanding that that's what would happen @martha?

(I have no issue with removing the default filter imo the default filter can be confusing)

martha: Yes, that's right. https://github.com/greenriver/hmis-warehouse/pull/5556#discussion_r2203803672

I agree with removing the default filter in both places!

martha (approved):

src/modules/ce/components/admin/AdminReferralsTable.tsx
gigxz: great, thanks! sorry I didn't respond to that comment thread!

PR #1211, opened 2025-07-18
Add form for sending Direct CE Referrals

martha:
## Description

Summary of changes:
- Add `SendReferralForm` which presents the enrollment, project, and unit group picklists
- Add `SendReferralSubForm` which fetches and presents the "delegated handoff" step form
- Support disabled picklist options

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5585

How to test:
- 

## Type of change
New feature


src/api/operations/ce.operations.graphql
```diff
@@ -62,6 +62,31 @@ mutation SubmitCeReferralStep(
   }
 }
 
+mutation CreateDirectCeReferral(
```
gigxz: The ce.operations and ce.queries files are getting big, maybe we could move these to `directReferrals.` files?

src/modules/ce/components/directReferral/SendReferralForm.tsx
```diff
@@ -0,0 +1,117 @@
...
+          itemChanged={({ value }) =>
+            setFormState((old) => ({
+              ...old,
+              selectedProject: value,
+              selectedUnitGroup: undefined,
+            }))
+          }
+        />
+        {formState.selectedProject && (
+          <SentryErrorBoundary>
```
gigxz: do we expect this pick list fetch to raise for some reason?

src/modules/ce/components/directReferral/SendReferralForm.tsx
```diff
@@ -0,0 +1,117 @@
...
+  project: ProjectAllFieldsFragment;
+}
+
+type FormState = {
+  selectedEnrollment?: PickListOption;
+  selectedProject?: PickListOption;
+  selectedUnitGroup?: PickListOption;
+};
+
+const SendReferralForm: React.FC<Props> = ({ project }) => {
```
gigxz, line 27: Screenshot 2025-07-23 at 11 11 38 AM

This is a bit of a funky state to end up in. It lets you select projects that don't actually have unit groups that can accept referrals, so you can't proceed but there's also no error message.

Maybe we can hide projects from the list if they aren't referrable? It looks like we do hide most un-referrable projects, but not all of them. In this case my problem is the workflow doesn't have the new inbound task marker

src/modules/ce/components/project/ProjectCeReferralsPage.tsx
```diff
@@ -69,31 +69,30 @@ const ProjectCeReferralsPage: React.FC = () => {
...
-  //         Send Referral
-  //       </ButtonLink>
-  //     );
-  //   }
-  // }, [project.id, showOutgoingReferrals]);
+  const actions = useMemo(() => {
+    if (showOutgoingReferrals) {
+      return (
+        <ButtonLink
+          variant='contained'
```
gigxz: Other Project pages with buttons in the top-right use a different style for buttons. Suggest updating style to match (outlined, possibly with icon if we have one?). See pages: Overview, Enrollments, Units, Inventory etc all follow this pattern.

gigxz (approved):
looks good!

src/modules/ce/components/directReferral/SendReferralForm.tsx
martha: Thanks, I agree, we should filter out projects that aren't referrable! Do you think we should also filter out projects that have no availability? Or would something like this potentially be useful to communicate to the user that the project has no availability (vs. just hiding the project from the list entirely)?
Screenshot 2025-07-23 at 5 31 20 PM

gigxz: Hmm yeah I think that seems smart to keep for now, since it communicates something dynamic amount the project (capacity) as opposed to communicating something about configuration, in a way. Something to get user feedback on perhaps. but I think that looks good

src/modules/ce/components/directReferral/SendReferralForm.tsx
martha: No I don't think so, I'll remove

src/modules/ce/components/project/ProjectCeReferralsPage.tsx
martha: Sure! Maybe,
Screenshot 2025-07-23 at 5 37 21 PM
but I'll check in with design tomorrow to confirm

PR #1212, opened 2025-07-22
Support printing all case notes

gigxz:
## Description

https://github.com/open-path/Green-River/issues/7866

- Add "card view" for case notes tables
- Add print page for case notes table, and button to access it from client case notes page.
- Print function is permissions by a new permission. The permission doesnt grant any actual access, this is more to give admins the ability to hide print functionality since it may not be desired to allow downloading data

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5598

How to test:
* enable permission
* client profile => case notes => card toggle / print => print
* enrollment => case notes => card toggle
* storybook case note card 

## Type of change
new feature


graphql.schema.json
```diff
@@ -12456,6 +12472,12 @@
             "isDeprecated": false,
             "deprecationReason": null
           },
+          {
```
martha: It looks like there are some AHA-related changes that don't really belong with this diff(?) I think they may be already on release-173, so maybe all this needs is to have the base branch merged in?

src/modules/hmis/hmisUtil.ts
```diff
@@ -312,17 +312,22 @@ export const lastUpdatedBy = ({
   dateUpdated,
   user,
   relativeDate = false,
+  dateFormat = 'date',
 }: {
   dateUpdated?: string | null;
   user?: { name: string } | null;
   relativeDate?: boolean;
+  // adding this, but I feel like the better default behavior would be to use whichever granularity the `dateUpdated` argument is in. why are we defaulting to showing timestamps as YY/MM/DDDD elsewhere? not our usual pattern for audit timestamps
```
martha: I agree that what you have added is fine for now. I think I also agree with you about the desired pattern, but I feel like I remember running into use cases where we wanted to override a date's granularity. For example, a date is returned from the backend as a timestamp, but in the ui we only want to show the date. However, maybe that actually goes against our styleguide/would indicate the need for returning a different type from the backend?

martha (approved):
Everything looks great! thank you!

src/modules/hmis/hmisUtil.ts
gigxz: Ah thanks for pointing out, I forgot to come back to this comment. I think it's ok if there are some places we want to do that, but I think the _default_ should be to display the date in the same granularity as the field type (eg datetime => date with time, date => date). Since this method is still only used in a few places, I will take a stab at updating that, and I can add an override method in the place you added it for.

It looks like this pr https://github.com/greenriver/hmis-frontend/pull/1155/files changed several places to display timestamps as Dates instead of Date Times (client profile card, client search results card, scan card). I'll assume those were unintentional, and the only intended one was ReferralStepCard, since that's the file that seems related to the PR.

martha: Yes - thanks! Sorry for missing that, many moving parts 😬

gigxz: Updated as such with https://github.com/greenriver/hmis-frontend/pull/1212/commits/4be0362c3821e4bc5ad4aeb59cbda852b973bb06
It looks like the ReferralStepCard usage that it was originally introduced for is no longer present, but I added a comment where we intentionally format relative timestamps as relative dates.

PR #1196, opened 2025-07-02
Display relevant-to-matching Client details on Referral page

gigxz:
## Description

https://github.com/open-path/Green-River/issues/7591

Summary of changes:

[//]: # 'remove if not applicable'
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5540

[//]: # 'remove if not applicable'
How to test:

to add:
- [x] mci id
- [x] link to client profile only if viewable
- [x] opportuinty eligibility requirements
- [x] make cards collapsible
- [x] dynamic eligibility values

Screenshot 2025-07-07 at 10 31 43 AM

## Type of change
[//]: # (e.g., Bug fix, New feature, Documentation, Code clean-up, Dependency update)

## Checklist before requesting review
- [ ] I have performed a self-review of my code
- [ ] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [ ] My code includes comments and/or descriptive variable names to help other engineers understand the intent (or not applicable)
- [ ] My code follows the style guidelines of this project (eslint)
- [ ] I have updated the documentation (or not applicable)
- [ ] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

PR #1216, opened 2025-07-23
Minor CE bug fixes

gigxz:
## Description
Fix some small bugs that were surfaced in qa:
* Fix bug where user would see "Start" button for a task even if they didn't have access to perform the task
* Fix bug where user would see "Client 123" in place of client name for referral (places I missed with https://github.com/greenriver/hmis-frontend/pull/1196)

## Type of change
bug fix

### DIFF
--- a/src/modules/ce/components/referral/ReferralStepAction.tsx
+++ b/src/modules/ce/components/referral/ReferralStepAction.tsx
@@ -52,8 +52,10 @@ const ReferralStepAction: React.FC<Props> = ({ step, referralId, path }) => {
   if (error) throw error;
 
   const buttonSx = { width: '116px' };
+  const { canPerformStep } = step.access;
 
-  if (status === CeReferralStepStatus.Available && step.access.canPerformStep) {
+  // If the step is "Available" (meaning nobody has clicked "Start" yet), we need to perform a mutation to initialize it
+  if (status === CeReferralStepStatus.Available && canPerformStep) {
     return (
       <LoadingButton
         sx={buttonSx}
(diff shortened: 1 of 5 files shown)


src/modules/ce/components/referral/ReferralStepAction.tsx
```diff
@@ -66,9 +68,10 @@ const ReferralStepAction: React.FC<Props> = ({ step, referralId, path }) => {
     );
   }
 
-  if (status === CeReferralStepStatus.InProgress) {
-    // Someone has already kicked off the "Start Step" mutation, but we don't save in-progress form values,
-    // so from the user's perspective, whether or not the step is "started" isn't very meaningful.
+  // If the step is "In Progress", just link to the step page.
+  // Someone has already kicked off the "Start Step" mutation, but we don't save in-progress form values,
+  // so from the user's perspective, whether or not the step is "started" isn't very meaningful.
+  if (status === CeReferralStepStatus.InProgress && canPerformStep) {
```
martha, line 74: I think this was intentional -- the `ReferralStepForm` component shows a `DynamicView` if the current user can't perform the step. So we are losing something by making this change: the ability of a user who can view, but not perform, a (started) step to preview what info/questions that step will collect. I think that loss seems ok to me, just surfacing it.

Another alternative -- given that the whole "start step" concept is abstracted from the user, could we just get rid of the `StartCeReferralStep` mutation altogether, and have the `SubmitCeReferralStep` mutation both start and submit the step? Then this button could say "Start" for a user with permission and "View" for a user without.

(That solution also has a nice parity with the current CreateReferral mutation, which both creates the referral and starts the workflow engine, never actually leaving a referral in the "initialized" state.)

martha (approved):
Thank you!

src/modules/ce/components/referral/ReferralStepAction.tsx
gigxz: Hmm okay, I  didn't realized this was on purpose. My concern with "view" is that the user can't tell... did this task filled out and everything was left blank? vs nobody has filled it out yet. I think lets start with hiding it (since there is no _data_ to show anyway) and we can see if we get feedback.

PR #1220, opened 2025-07-29
Follow up on QA items for AHA score

martha:
## Description

- Targets release-173 to follow up on QA that's on that branch. We can retarget to 174 if you prefer since this component isn't in use yet.

Summary of changes:
- Disable re-fetching the score on a previously saved, unlocked assessment.
- _Actually save to the db_ a value of -1 when the score was fetched and couldn't be found. This circumvents the generic required logic on the field, enabling users to save the form after having tried to fetch the score. I'm not totally sold on the incomplete (ts failing) solution I have drafted; I'll add a comment thread to discuss this inline
- Custom override the label to show the score as required.
- Fix bug where autofill values were overwriting previously saved values on readonly fields. I'll add a comment thread to discuss this inline

How to test: Using the form definition below,
- Test that you can fetch and save AHA score for a client with an MCI uniq id
- Test fetching, submitting, and reopening/unlocking the assessment. The score should still appear and the button to refetch should be disabled.
- Test fetching, "save & submit later", and reopening the assessment. Same as above.
- Test submitting without fetching the score. You should see a validation error.
- Test fetching score on a client that doesn't have an MCI uniq id, or doesn't have one that's available in the external system. You should see "score is not available" message and still be able to submit

<Details>
<Summary>new form JSON to use</Summary>

```json
{
  "item": [
    {
      "item": [
        {
          "text": "Assessment Date",
          "type": "DATE",
          "bounds": [
            {
              "id": "min-assmt-date",
              "type": "MIN",
              "severity": "error",
              "value_local_constant": "$entryDate"
            },
            {
              "id": "max-assmt-date",
              "type": "MAX",
              "severity": "error",
              "value_local_constant": "$today"
            },
            {
              "id": "max-assmt-date-exit",
              "type": "MAX",
              "severity": "error",
              "value_local_constant": "$exitDate"
            }
          ],
          "link_id": "assessment_date",
          "mapping": {
          },
          "required": true,
          "assessment_date": true,
          "disabled_display": "HIDDEN"
        },
        {
          "item": [
            {
              "text": "AHA Score",
              "type": "INTEGER",
              "link_id": "aha_score",
              "mapping": {
                "custom_field_key": "housing_needs_for_ce_aha_score_2"
              },
              "required": true,
              "read_only": true,
              "warn_if_empty": false,
              "disabled_display": "HIDDEN"
            },
            {
              "text": "Alt AHA Flag",
              "type": "BOOLEAN",
              "link_id": "aha_alt_aha_flag",
              "mapping": {
                "custom_field_key": "housing_needs_for_ce_aha_alt_aha_flag"
              },
              "required": false,
              "read_only": true,
              "warn_if_empty": false,
              "disabled_display": "HIDDEN"
            },
            {
              "text": "Data Warehouse Client ID",
              "type": "STRING",
              "hidden": true,
              "link_id": "aha_dw_client_id",
              "mapping": {
                "custom_field_key": "housing_needs_for_ce_aha_dw_client_id"
              },
              "required": false,
              "read_only": true,
              "warn_if_empty": false,
              "disabled_display": "HIDDEN"
            },
            {
              "text": "Generator",
              "type": "STRING",
              "hidden": true,
              "link_id": "aha_generator",
              "mapping": {
                "custom_field_key": "housing_needs_for_ce_aha_generator"
              },
              "required": false,
              "read_only": true,
              "warn_if_empty": false,
              "disabled_display": "HIDDEN"
            }
          ],
          "text": "Allegheny Housing Assessment (AHA) Score",
          "type": "GROUP",
          "link_id": "aha_group",
          "component": "AHA",
          "disabled_display": "HIDDEN"
        }
      ],
      "text": "Allegheny Housing Assessment (AHA) Score",
      "type": "GROUP",
      "link_id": "section_1",
      "disabled_display": "HIDDEN"
    }
  ]
}
```
</Details>

## Type of change

- [x] My code follows the style guidelines of this project (eslint) - typescript fails, please see inline comment thread.

### DIFF
--- a/src/modules/external/aha/components/AhaScore.tsx
+++ b/src/modules/external/aha/components/AhaScore.tsx
@@ -45,11 +45,12 @@ const AhaScore = ({
   // Use RHF to watch the current value of the score field.
   // This is to disable the fetch button after unlocking an assessment that's already fetched the score.
   const values = useDynamicFieldWatchValues([SCORE_LINK_ID]);
-  const existingValue = values[SCORE_LINK_ID];
+  const scoreValue = values[SCORE_LINK_ID];
 
-  const [hasScore, setHasScore] = useState<boolean>(
-    !!existingValue && existingValue >= 0
-  );
+  // If client has no score (-1 is returned), hasScore is false and the button stays enabled
+  const hasScore =
+    scoreValue !== null && scoreValue !== undefined && scoreValue >= 0;
+
   const [hasFetched, setHasFetched] = useState<boolean>(false);
 
   const clientId = handlers?.localConstants.clientId;
@@ -62,36 +63,26 @@ const AhaScore = ({
       const errors = data.fetchAhaScore?.errors || [];
       if (errors.length > 0) {
         setErrorState({ ...emptyErrorState, errors });
-        setHasScore(false);
         return;
       }
 
       setHasFetched(true);
       setErrorState(emptyErrorState);
 
-      if (data.fetchAhaScore?.score) {
-        if (data.fetchAhaScore.score >= 0) {
-          setHasScore(true);
-        }
-
-        if (severalItemsChanged) {
-          severalItemsChanged({
-            values: {
-              [SCORE_LINK_ID]: data.fetchAhaScore.score,
-              [ALT_AHA_FLAG_LINK_ID]: data.fetchAhaScore.altAhaFlag,
-              [DW_CLIENT_LINK_ID]: data.fetchAhaScore.dwClientId,
-              [GENERATOR_LINK_ID]: data.fetchAhaScore.generator,
-            },
-            type: ChangeType.User,
-          });
-        }
-      } else {
-        setHasScore(false);
+      if (data.fetchAhaScore && severalItemsChanged) {
+        severalItemsChanged({
+          values: {
+            [SCORE_LINK_ID]: data.fetchAhaScore.score,
+            [ALT_AHA_FLAG_LINK_ID]: data.fetchAhaScore.altAhaFlag,
+            [DW_CLIENT_LINK_ID]: data.fetchAhaScore.dwClientId,
+            [GENERATOR_LINK_ID]: data.fetchAhaScore.generator,
+          },
+          type: ChangeType.User,
+        });
       }
     },
     onError: (apolloError) => {
       setErrorState({ ...emptyErrorState, apolloError });
-      setHasScore(false);
     },
   });
 
@@ -99,13 +90,7 @@ const AhaScore = ({
   if (!isComponentValid) throw new Error('Invalid Aha form component');
 
   if (viewOnly) {
-    return item.item?.map((i) => {
-      if (i.linkId === SCORE_LINK_ID && !hasScore) {
-        return renderChildItem(i, { value: null });
-      } else {
-        return renderChildItem(i);
-      }
-    });
+    return item.item?.map((i) => renderChildItem(i));
   }
 
   if (!clientId) {
--- a/src/modules/external/aha/components/AhaScore.tsx
+++ b/src/modules/external/aha/components/AhaScore.tsx
@@ -42,7 +42,7 @@ const AhaScore = ({
 
   const [errorState, setErrorState] = useState<ErrorState>(emptyErrorState);
 
-  // Use dynamic field watch to get the current values of the expected link IDs.
+  // Use RHF to watch the current value of the score field.
   // This is to disable the fetch button after unlocking an assessment that's already fetched the score.
   const values = useDynamicFieldWatchValues([SCORE_LINK_ID]);
   const existingValue = values[SCORE_LINK_ID];
--- a/src/modules/external/aha/components/AhaScore.tsx
+++ b/src/modules/external/aha/components/AhaScore.tsx
@@ -47,7 +47,9 @@ const AhaScore = ({
   const values = useDynamicFieldWatchValues([SCORE_LINK_ID]);
   const existingValue = values[SCORE_LINK_ID];
 
-  const [hasScore, setHasScore] = useState<boolean>(!!existingValue);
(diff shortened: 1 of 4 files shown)


src/modules/external/aha/components/AhaScore.tsx
```diff
@@ -83,16 +95,25 @@ const AhaScore = ({
...
   });
 
-  // console.error instead of throwing an error, so we can still preview how the form looks in non-client contexts
+  // Throw an error if the children don't match the expected structure
+  if (!isComponentValid) throw new Error('Invalid Aha form component');
+
+  if (viewOnly) {
+    return item.item?.map((i) => {
+      if (i.linkId === SCORE_LINK_ID && !hasScore) {
+        return renderChildItem(i, { value: null });
```
martha: Actually saving the value as -1 in the db requires manually overwriting the display value so we show "Data not collected" instead of -1 when the assessment is reopened. However, this change doesn't satisfy typescript, and I'm not sure it is the right way forwards.

Would it be ok to just display -1?

src/modules/form/hooks/rhf/useDynamicFieldAutofillValue.tsx
```diff
@@ -32,6 +32,7 @@ export const useDynamicFieldAutofillValue = (
     **/
     if (
       !result &&
+      !!item.autofillValues &&
```
martha: This code was the culprit for the bug described on ticket comments:
> I see that it shows the current score in readonly mode but then when I unlock it I can no longer see the score.

Here's my understanding of the bug:
- the existing code assumes that readonly items being displayed on editable forms always have some autofill value. And therefore that, if the `autofillValues` call on line 20 returns null, it means the conditions are no longer met, and the field should be nullified.
- however, in this use case, the `autofillValues` call on line 20 is returning null because the fields don't have any autofill values specified. But they should not be nullified.

I'm proposing a solution where this logic should only apply to items that do have an `autofillValues` specified. Let me know if you have any concerns about that, or pointers on existing forms to confirm/QA after this change.

gigxz: Nice job tracking this down! This looks correct to me. I would suggest moving this further up, since there is really no need for this to invoke `autofillValues(...)` for an item that does not have any `autofillValues` specified. (It will [always return null](https://github.com/greenriver/hmis-frontend/blob/799df7d1cbf46ace15f81cb766ab69f4e5aaa742/src/modules/form/util/formUtil.ts#L557-L564) in this case anyway). I think that will make it easier to understand what's going on.

So, I can't comment up there in the diff, but something like:
```tsx
  return useMemo(() => {
    // If the item does not have any autofill values defined, return null as there is nothing to evaluate.
    if (!item.autofillValues) return null;

    // Evaluate the autofill conditions for the item.
    // Returns an object { value: <computed value> } if there is an autofill value to apply,
    // or null if no conditions are satisfied.
    const result = autofillValues({
      item,
      values,
      itemMap,
      localConstants,
      viewOnly,
    });


    // ... no changes to rest of hook
```

src/modules/external/aha/components/AhaScore.tsx
gigxz: Yes I think it's fine to display -1, lets go with that

src/modules/external/aha/components/AhaScore.tsx
```diff
@@ -102,12 +123,27 @@ const AhaScore = ({
...
+                fontWeight: 600,
+              }}
+            />
+          }
+          helperText={item.helperText}
+        >
           <LoadingButton
             loading={loading}
-            disabled={hasScore} // If value has already been fetched, disable the button
+            disabled={hasScore} // If value already exists, disable the button
```
gigxz, line 128: It looks like `hasScore` will be false if the value is `<0`, so does this mean the button will continually be enabled if the fetched score is -1? I think thats probably fine, just checking.

src/modules/external/aha/components/AhaScore.tsx
```diff
@@ -39,7 +42,14 @@ const AhaScore = ({
...
 
-  const [hasScore, setHasScore] = useState<boolean>(false);
+  // Use RHF to watch the current value of the score field.
+  // This is to disable the fetch button after unlocking an assessment that's already fetched the score.
+  const values = useDynamicFieldWatchValues([SCORE_LINK_ID]);
+  const existingValue = values[SCORE_LINK_ID];
+
+  const [hasScore, setHasScore] = useState<boolean>(
+    !!existingValue && existingValue >= 0
+  );
```
gigxz: Since we're _watching_ the SCORE_LINK_ID field now, do we still need to maintain an internal state field for whether the score exists or not?
I haven't tested this out since I don't have the full setup locally, but I wonder if it behaves the same way if `hasScore` was populated from the watched form field:

```suggestion
  // Use RHF to watch the current value of the score field.
  // This is to disable the fetch button after unlocking an assessment that's already fetched the score.
  const values = useDynamicFieldWatchValues([SCORE_LINK_ID]);
  const existingValue = values[SCORE_LINK_ID];

  const hasScore = !!existingValue && existingValue >= 0; // this should update whenever existingValue changes, without needing to call setHasScore all over
```
If that works the same way, it's less stuff to manage in this component, which is nice. I might be missing some reason for why this doesn't work though.

src/modules/external/aha/components/AhaScore.tsx
martha: Yes, this was my intention, to not prevent the user from re-fetching if there is no score. I'll add comments

src/modules/external/aha/components/AhaScore.tsx
martha: Thanks, yeah this seems like a nice improvement! It seems to work well with a minor modification to accommodate 0 as a valid value for the score

gigxz (approved):
👍

PR #1222, opened 2025-07-31
Add alt-AHA calculator form component

martha:
## Description

Summary of changes:
- Add new `AltAhaScore` form component
- Add a dummy local constant for `enrollmentId` in assessment contexts, similar to `clientId`
- Pass form definition identifier to `DynamicForm`, so it can be provided in form context and used by the alt-aha form component.
  - I haven't updated all DynamicForms to do this, and not all of them use forms with identifiers anyway.

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5614

## Type of change
New feature, custom build


src/modules/assessments/components/AssessmentForm.tsx
```diff
@@ -180,6 +180,7 @@ const AssessmentForm: React.FC<Props> = ({
       clientAge: age(client),
       clientRaceEthnicity: raceEthnicityDisplayString(client.race),
       clientId: client.id,
+      enrollmentId: enrollment.id,
```
gigxz: This seems fine, but what is it used for? I think there's maybe a form content dependency here that I'm not following

src/modules/form/components/DynamicForm.tsx
```diff
@@ -41,6 +41,7 @@ export interface DynamicFormProps
   extends Omit<DynamicFormBaseProps, 'handlers' | 'onSaveDraft'> {
   clientId?: string;
   definition: FormDefinitionJson;
+  identifier?: string;
```
gigxz: I think this needs some explanation as to why `identifier` is passed separately from `definition`. If I grep for usages of `<DynamicForm\b`, the vast majority are passing a definition but _not_ an identifier, and it feels very unclear what the implications of that are. If later on we decided to use this `identifier` prop to do something else, it could have unexpected behavior. Just feels like something that could be a future-bug, and should have comments explaining whats going on.

gigxz: Alternatively, another route would be for the [IndividualAssessment component](https://github.com/greenriver/hmis-frontend/blob/aeb621232137b6b7e1b5a2e98521775d04f22ded/src/modules/assessments/components/IndividualAssessment.tsx#L114) to provide an "assessment context" that the Alt-AHA component can pull from. That might be a clearer way to go. We could also use that to provide the Enrollment ID. (It seems a bit arbitrary to me that Enrollment ID is being provided via Local Constants whereas identifier is being provided via the form context.) I think that points to a need for having some additional context about the assessment available.

src/modules/assessments/components/AssessmentForm.tsx
gigxz: Nevermind- I see its being used to pass state down to the AltAhaScore component

martha: This is for the AltAhaScore component to pass an enrollment ID to the backend that can be saved as the "owner" of the calculation log, because the assessment hasn't been persisted yet.

(Peripherally relates to this backend convo about calculation log owner https://github.com/greenriver/hmis-warehouse/pull/5614#discussion_r2292048976)

src/modules/form/components/DynamicForm.tsx
martha: Sure, that makes sense to me! I was thinking about these pretty surface-level, something like, "enrollment ID is kind of like client ID, so I'll add it with the local constants" and "form identifier comes from the form definition, so I'll pass it in from the same place where we pass the json definition." I like your suggestion of adding IndividualAssessment context instead.

src/modules/external/aha/components/AltAhaScore.tsx
```diff
@@ -0,0 +1,103 @@
...
+    formDefinitionIdentifier,
+    handlers,
+  ]);
+
+  return (
+    <Stack direction='column' gap={1} alignItems='flex-start'>
+      {label}
+      {errorState && hasErrors(errorState) && (
+        <Stack gap={1} sx={{ mb: 1 }}>
+          <ApolloErrorAlert error={errorState.apolloError} />
```
gigxz:
```suggestion
          <ApolloErrorAlert error={errorState.apolloError} inline />
```

This is appearing as a snackbar at the top of the screen but its out of view. (Ideally the snackbar should be sticky, but that wasn't easy for me to figure out quickly). Suggest presenting the error inline instead so the user can see it

src/modules/external/aha/components/AltAhaScore.tsx
```diff
@@ -0,0 +1,103 @@
...
+  });
+
+  const { formDefinitionIdentifier, enrollmentId } =
+    useContext(AssessmentContext) || {};
+
+  const handleFetch = useCallback(() => {
+    if (!enrollmentId || !formDefinitionIdentifier) {
+      // In the preview environment, mock the response instead of calling the mutation
+      onChange?.(10);
+    } else {
```
gigxz: This worries me because if there is a render issue with the context, or this is rendered somewhere in the app without a context, the user will see "10" as the score (and perhaps even get that value stored for the client?). Suggest raising instead:

```suggestion
    if (!enrollmentId || !formDefinitionIdentifier) {
      throw new Error(
        'unable to calculate alt-aha without assessment context values'
      );
    } else {
```

gigxz (approved):
looks good just a couple minor suggestions!

src/modules/external/aha/components/AltAhaScore.tsx
martha: Or maybe the button should be disabled? (to avoid sentry noise, since users are likely to click the button in preview)

gigxz: Sure that seems OK. I'm not too concerned about previews, I guess, at some point we should dedicate some time to making it more robust. (For example, provide an easy way for the form component to know that its being rendered in preview)

PR #1223, opened 2025-08-04
Add support for rounding to form expressions

martha:
## Description

Summary of changes:
- Add support for rounding functions to the form expression evaluator

Test against the json form here: https://qa-hmis.openpath.host/admin/forms/housing_needs_assessment

Screenshot 2025-08-04 at 3 57 20 PM

## Type of change
New feature

### DIFF
--- a/src/modules/form/util/expressions/expressions.evaluateFormula.test.ts
+++ b/src/modules/form/util/expressions/expressions.evaluateFormula.test.ts
@@ -22,4 +22,30 @@ describe('evaluateFormula', () => {
       undefined
     );
   });
+
+  describe('rounding functions', () => {
+    it('rounds division to nearest integer', () => {
+      expect(evaluateFormula('ROUND(a / b)', context({ a: 7, b: 3 }))).toBe(2);
+      expect(evaluateFormula('ROUND(a / b)', context({ a: 8, b: 3 }))).toBe(3);
+    });
+
+    it('rounds division to specific decimal places', () => {
+      expect(
+        evaluateFormula('ROUNDTO(a / b, 2)', context({ a: 10, b: 3 }))
+      ).toBe(3.33);
+      expect(
+        evaluateFormula('ROUNDTO(a / b, 1)', context({ a: 22, b: 7 }))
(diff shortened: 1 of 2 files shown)


src/modules/form/util/expressions/expressions.evaluateFormula.test.ts
```diff
@@ -22,4 +22,30 @@ describe('evaluateFormula', () => {
       undefined
     );
   });
+
+  describe('rounding functions', () => {
```
gigxz, line 57: suggest adding tests where denominator value is nil and 0

gigxz (approved):
Looks good. I couldn't tell how all the edge cases were handled (missing and 0 denominators) so I suggested adding test cases for those.

src/modules/form/util/expressions/expressions.evaluateFormula.test.ts
martha: Thanks! Sure, I added some tests for edge cases. I think these edge cases predate this PR, and I wasn't sure what the existing behavior was, so I investigated and found that our expression evaluator just does whatever the default javascript behavior is for edge cases like dividing by 0. With dividing by undefined, it will return undefined due to the try/catch here https://github.com/greenriver/hmis-frontend/blob/965c49be4cabae7fd5374385a6a0a40cf5e7925c/src/modules/form/util/expressions/formula.ts#L70-L79.

my understanding is that formula autofill is an advanced-user feature, and we expect users to ensure that the form behaves the way they want in terms of these edge cases. In this example for AMI/FPL,
- we are always dividing by an ami or fpl, which are not set to 0, so we should never divide by 0
- if ami or fpl is undefined, the autofill_when condition is not met, so we should never divide by undefined

Does that match your understanding?

gigxz: this seems ok, lets go ahead and merge @martha

martha:
Sorry, I merged without checking, this should have been retargeted to release-176. I'll revert and retarget

PR #1235, opened 2025-08-18
Add global eligible client list

gigxz:
## Description

Summary of changes:
* Add a tab to the CE admin page for viewing a list of all Ce Clients that belong to any candidate pool.
  * Also adds the ability to click into each client and see which unit groups they are eligible for
  * With dynamic columns and dynamic filters
* Add dynamic columns to unit-level eligible client lists. For example to support having a column per priority score value.

Issue: https://github.com/open-path/Green-River/issues/8043
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5659 

How to test:
* Admin => coordinated entry => eligible clients. Search and filter table, set up hmis table configs for more columns (see backend PR), click in to see eligible projects
* To test configurable columns on unit-level waitlist, set up db configuration and view waitlist

Screenshot 2025-09-02 at 8 50 07 PM

Screenshot 2025-09-02 at 8 50 25 PM

Screenshot 2025-09-02 at 9 01 25 PM

## Type of change
New feature


src/modules/ce/components/admin/ConsolidatedWaitlistDialog.tsx
```diff
@@ -0,0 +1,26 @@
...
+interface Props {
+  id: string;
+  open: boolean;
+  onClose: () => void;
+}
+
+const ConsolidatedWaitlistDialog: React.FC<Props> = ({ id, ...props }) => {
+  return (
+    <CommonDialog maxWidth='lg' fullWidth enableBackdropClick {...props}>
+      <DialogTitle>what waitlists is this client on?</DialogTitle>
```
gigxz: clean up

gigxz:
I'm working on a bug with table filtering, will push a fix

src/components/elements/tableFilters/filters/FilterMenu.tsx
```diff
@@ -15,34 +15,38 @@ export interface TableFilterMenuProps<T> {
   setFilterValues: (value: Partial<T>) => any;
 }
 
-const TableFilterMenu = <T,>(props: TableFilterMenuProps<T>): JSX.Element => {
+const TableFilterMenu = <T,>({
+  filters,
+  filterValues,
+  setFilterValues,
+}: TableFilterMenuProps<T>): JSX.Element => {
```
gigxz, line 22: The diff in this file is no longer necessary since I moved the transformation to the separate util. But I think it makes it more readable to destructure, so keeping the change

gigxz:
filter bug is resolved now

src/components/elements/tableFilters/filters/FilterItem.tsx
```diff
@@ -73,12 +73,9 @@ const TableFilterItem = <T,>({
             filter.pickListOptions ||
             localResolvePickList(filter.enumType, true) ||
             [];
-          const variant =
-            filter.variant ||
-            (options.length < 3 && filter.multi ? 'checkboxes' : 'select');
```
gigxz, line 78: removed this dynamic functionality, imo not helpful and is more confusing to have a panel with lots of different input types. callers can explicitly provide variant if desired. otherwise it always will use a dropdown

src/modules/dataFetching/types.ts
```diff
@@ -26,32 +27,32 @@ export interface DateFilter<I> extends BaseFilter<I> {
...
-  type: 'picklist';
+export interface RemotePickListFilter<I> extends BaseFilter<I> {
+  type: 'remote_picklist';
   pickListReference: PickListType; // name of remote pick list
   variant?: SelectElementVariant;
   pickListArgs?: PickListArgs;
 }
 
+export interface LocalPickListFilter<I> extends BaseFilter<I> {
+  type: 'local_picklist';
```
martha, line 45: Am I understanding this correctly:
- `RemotePickListFilter` is when the picklist is one of our `PicklistType`s such as `PickListType.Project`. "remote" because the pick list options are fetched by `pickListReference`.
- `LocalPickListFilter` is when the pick list options are provided to the filter "locally," via `pickListOptions`. But "local" doesn't necessarily mean "locally defined here in the frontend source code;" for example the usage you've implemented here is to fetch the dynamically defined pick list options separately, and then pass those options to a `LocalPickListFilter`.

Is that right? I found "local" a bit confusing in this context, if so.

martha (approved):
all looks good!

src/modules/dataFetching/types.ts
gigxz: Yep I think you're understanding that correctly. The pick list is "local" because the `pickListOptions` are passed directly to the filter object (line 46 just below this comment), as opposed to the RemotePickListFilter  which receives a `pickListReference` that it must resolve from the backend. 

You're correct that for dynamic filtering, we do fetch the options from the backend first, and _then_ we define the LocalPickListFilter object based on that. So there is a backend fetch involved - I think this is what you're referring to? I would still say that from the perspective of the FilterMenu, the list is local, because it doesn't have to do any fetching to retrieve it.

PR #1242, opened 2025-08-27
Updated Unit Group management ux

gigxz:
## Description

Issue: https://github.com/open-path/Green-River/issues/8124
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5686

How to test:
- [ ] Test Unit management behavior for project that does NOT accept direct referrals nor waitlist referrals ("supportsReferrals"= false in the api). It should remain unchanged, units are still "floating" i.e. not tied to groups.
- [ ] Test Unit management behavior for a project that DOES have direct or waitlist referrals enabled:
  - [ ] Creating a new project, setting up the first unit group, adding the first units to it (allows selecting unit type)
  - [ ] Adding a new unit group to a project
  - [ ] Adding subsequent units to an existing group (disallows choosing Unit Type, Unit Type is inferred by backend)
  - [ ] Bulk managing start/stop accepting referrals on the main Units page
    - [ ] Those actions are unavailable for units that belong to groups that do not have a Workflow configured

## Type of change
New feature


src/modules/units/components/Units.tsx
```diff
@@ -24,28 +22,33 @@ import { useGetProjectUnitGroupsQuery } from '@/types/gqlTypes';
...
+  }, [data]);
+
+  // Enable Unit Groups management UI if:
+  // - the project supports CE referrals (direct or waitlist), OR
+  // - the project already has at least one unit group. (This will allow UI to switch over when Units are migrated into Unit Groups)
+  // TODO(#7814) remove this flag once Unit Group migration is complete. At that point all projects will have Unit Groups Enabled.
+  const unitGroupsEnabled = useMemo(
+    () => !!projectSupportsReferrals || unitGroups.length > 0,
+    [projectSupportsReferrals, unitGroups.length]
+  );
```
gigxz, line 51: Tagging @martha regarding this approach in relation to https://github.com/open-path/Green-River/issues/7814. This should allow us to deploy this frontend change independently, and then once we make the migration, the interfaces will change for users. There are a few links to `#7814` peppered in this PR for things that can be cleaned up after the migration.

src/modules/units/components/UnitGroupCard.tsx
```diff
@@ -76,5 +71,21 @@ const UnitGroupCard: React.FC<Props> = ({
...
+
+  if (linkToUnitGroup) {
+    return (
+      <CardActionArea
+        component={RouterLink}
+        to={generateSafePath(ProjectDashboardRoutes.UNIT_GROUP, {
+          projectId,
+          unitGroupId: unitGroup.id,
+        })}
+        aria-label='Manage Unit Group'
```
martha:
```suggestion
        aria-label={`Manage Unit Group ${unitGroup.name}`}
```
also something to note, all of this card's interior content is hidden to screen readers in favor of the aria label, including the "x/y units vacant".

martha (approved):
Looks good! I noticed long unit group names can cause some style issues even on wide screens, but I think it's fine if we decide to defer fixing edge cases like that until later.
Screenshot 2025-08-28 at 11 18 52 AM

src/modules/units/components/UnitGroupCard.tsx
gigxz: Thanks, added. I think it's probably OK for that content to be hidden since it's shown elsewhere on the page. 

I'm facing issues with <a> tag being nested inside <a> tag because of the settings button inside the clickable card. I think I'm going to remove the button for now since the accessibility of the clickable card appears OK to me. Let me know if you have a different suggestion though, we can come back to it

gigxz:
Thanks Martha I updated to fix the wrapping issue

PR #1243, opened 2025-08-28
Support Legacy and CE Referrals side-by-side

gigxz:
## Description

Issue: https://github.com/open-path/Green-River/issues/8119

Add support for using CE referrals and Legacy referrals side-by-side during transition period. See issue for testing instructions.

Also renamed existing module `referrals`=>`legacyReferrals`.

## Type of change
New feature


src/modules/legacyReferrals/components/ProjectLegacyReferrals.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { generateSafePath } from '@/utils/pathEncoding';
+
+/**
+ * This component is used to show the legacy referrals page.
+ *
+ * If the project has CE referrals enabled, it disables the ability to create
+ * new legacy referrals and referral requests.
+ * TODO(#8142) fully sunset legacy referrals, remove this tab
+ */
+const ProjectLegacyReferrals = () => {
```
gigxz, line 22: UI unchanged aside from disabling action buttons in "legacy mode". Copied from the deleted `ProjectReferralsPage`.

src/routes/protected.tsx
```diff
@@ -290,15 +289,19 @@ export const protectedRoutes: RouteNode[] = [
...
-                  'canManageOutgoingReferrals',
-                ]}
-                redirectRoute={Routes.PROJECT}
-              >
-                <ProjectReferralsPage />
-              </ProjectEditRoute>
+              // No ProjectRoute wrapper due to complexity: some parts of page are visible for Legacy referrals,
+              // other parts visible if CE is enabled, etc. ProjectCeReferralsPage component handles permissions internally.
+              // See useProjectCeVisibility hook for details.
+              <ProjectCeReferralsPage />
```
martha: If this component now renders both CE and Legacy referrals, maybe to reduce confusion could we rename it `ProjectReferralsPage`?

src/modules/legacyReferrals/components/ProjectLegacyReferrals.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const { globalFeatureFlags: { externalReferralsEnabled } = {} } =
+    useGlobalFeatureFlags();
+  const { project } = useProjectDashboardContext();
+
+  // If this project has CE Referrals enabled (either Direct or Waitlist),
+  // show this page "legacy mode" which turns off the ability to create new
+  // legacy referrals and referral requests.
+  const { supportsReferrals, sendsDirectReferrals } =
+    project.coordinatedEntryFeatures || {};
+  const legacyMode = supportsReferrals || sendsDirectReferrals;
```
martha: This is minor but, here we are overloading the term "legacy" to mean two related but different things:
- Legacy Referrals = the old referrals (as opposed to CE referrals)
- Legacy Mode = a mode that the Legacy Referral UI can be in, only when CE is enabled

I think this overload might get confusing -- for example, for a client that is still using Legacy Referrals but does _not_ have CE enabled, we would say "they are using Legacy Referrals, but not in Legacy Mode."

Maybe we could rename `legacyMode` to something descriptive of what it does, like "disableNewLegacyReferrals"?

martha (approved):
Looks great! I just made a couple of naming suggestions, feel free to take or leave them.

src/modules/legacyReferrals/components/ProjectLegacyReferrals.tsx
gigxz: good point, updated!

src/routes/protected.tsx
gigxz: updated!

PR #1245, opened 2025-08-28
Auto-expand menu after it is auto-collapsed

gigxz:
## Description

* Add support for using react-router-dom Location State to expand desktop nav.
* Set the flag for auto-expanded when leaving "focus mode" (Back to X button in the context header)
* Set the flag for auto-expanded when leaving assessment page after a submission

Issue: https://github.com/open-path/Green-River/issues/7757, test instructions on issue

## Type of change
bug fix

### DIFF
--- a/src/components/layout/dashboard/contextHeader/ContextHeader.tsx
+++ b/src/components/layout/dashboard/contextHeader/ContextHeader.tsx
@@ -3,17 +3,21 @@ import MenuIcon from '@mui/icons-material/Menu';
 import { Box, Button } from '@mui/material';
 import React, { ReactNode, useCallback, useMemo } from 'react';
 
-import { useLocation, useNavigate } from 'react-router-dom';
+import { NavigateOptions, useLocation, useNavigate } from 'react-router-dom';
 import CommonStickyBar from '../../CommonStickyBar';
 import {
   CONTEXT_HEADER_HEIGHT,
   STICKY_BAR_HEIGHT,
 } from '../../layoutConstants';
+
 import { useIsMobile } from '@/hooks/useIsMobile';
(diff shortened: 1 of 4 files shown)


martha (approved):
Maybe I'm overthinking it, but I'm having a hard time totally wrapping my head around this change. At first look, it seemed to me like this is a case where ["you might not need an effect"](https://react.dev/learn/you-might-not-need-an-effect), since `useDashboardState` already returns a callback for opening/closing the menu. 

On looking more closely, I realized we actually are already doing that (line 98 of ContextHeader). So this PR introduces a second mechanism for closing the menu. Now we have both:
- Mechanism 1 (existing):
  - When: On clicking "Back to..." in the context header bar
  - How: `handleOpenMenu();` line 98 of ContextHeader.
  - Problems: 
    - it's causing the menu to flicker open on mobile too (but that seems easy to fix)
    - it's not called on Submit - would need to be refactored/moved to somewhere else
- Mechanism 2 (introduced in this PR):
  - When: on clicking "Submit"
  - How: the effect in `useDashboardState` listens for changes on the new EXPAND_DESKTOP_NAV_KEY
  - Problems: 
    - it's an effect, which always makes me look sideways
    - it introduces a second way the open menu state is controlled, which could be confusing in the future

tldr I'm happy to merge this now, it does seem to fix the problem, but I'd like to at least add more comments explaining the need for both mechanisms, and maybe go so far as to open a followup ticket.

gigxz:
Thank you for pointing that out @martha, I think I wasn't totally following the existing mechanism. Closing this in favor of a cleaner approach, where we just directly close the nav menu when desired: https://github.com/greenriver/hmis-frontend/pull/1247

PR #1247, opened 2025-08-29
Auto-expand menu when exiting Assessment pages

gigxz:
## Description

Issue: https://github.com/open-path/Green-River/issues/7757

1. When navigating away from Assessment pages after submission, re-expand the nav menu on desktop.
2. Fix bug where mobile menu would flicker when leaving assessment page from existing back button mechanism

Note: the left nav still does not auto-expand when using the browser Back functionality. (That could be a future enhancement, but was too much code churn to introduce referer state, though I did explore this approach a bit. There are a _lot_ of places where we link to an assessment page).

## Type of change
Bug fix

### DIFF
--- a/src/components/layout/dashboard/contextHeader/ContextHeader.tsx
+++ b/src/components/layout/dashboard/contextHeader/ContextHeader.tsx
@@ -86,13 +86,14 @@ const ContextHeader: React.FC<Props> = ({
       navigate(-1);
     }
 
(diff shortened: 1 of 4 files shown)


martha (approved):
Cool looks good, thanks Gig! I wish we could keep the `<a>` markup on the back-button, but I think it's probably not worth the time right now.

PR #1253, opened 2025-09-03
Always use unit group UI

martha:
## Description

Summary of changes:
- Remove `unitGroupsEnabled` flag
- Default to always using the unit group UI 

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5700

How to test:
- Navigate to a project with no units, that is not already using unit groups
- You should be prompted to add a unit group (not a floating unit)

## Type of change
New feature

### DIFF
--- a/src/modules/units/components/UnitManagementTable.tsx
+++ b/src/modules/units/components/UnitManagementTable.tsx
@@ -50,14 +50,14 @@ const UnitManagementTable: React.FC<Props> = ({
     return [
       UNIT_COLUMNS.unitType,
       UNIT_COLUMNS.unitId,
-      UNIT_COLUMNS.unitGroup,
+      ...(unitGroupId ? [] : [UNIT_COLUMNS.unitGroup]), // if looking at units in one group, no need to show the unit group col
       UNIT_COLUMNS.unitOccupancyStatus,
       UNIT_COLUMNS.clientOccupants,
       ...(coordinatedEntryFeatures.supportsReferrals
         ? [UNIT_COLUMNS.ceReferralStatus]
         : []),
     ];
-  }, [coordinatedEntryFeatures.supportsReferrals]);
+  }, [coordinatedEntryFeatures.supportsReferrals, unitGroupId]);
 
   const filters = useFilters({
     type: 'UnitFilterOptions',
--- a/src/modules/units/components/UnitManagementTable.tsx
+++ b/src/modules/units/components/UnitManagementTable.tsx
@@ -26,7 +26,6 @@ import { generateSafePath } from '@/utils/pathEncoding';
 interface Props {
   projectId: string;
   unitGroupId?: string; // if this table is for a specific unit group
-  unitGroupsEnabled?: boolean; // TEMP(#7814), remove when all projects moved to unit groups
   projectSupportsReferrals?: boolean; // whether to show CE details
   noUnitsMessage?: string; // custom message to show when there are no units
 }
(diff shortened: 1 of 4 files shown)


src/modules/units/components/Units.tsx
```diff
@@ -143,8 +133,8 @@ const Units = () => {
         projectId={project.id}
         open={addUnitsDialogOpen}
         onClose={() => setAddUnitsDialogOpen(false)}
-        allowSelectUnitType={!unitGroupsEnabled}
-        allowSelectUnitGroup={unitGroupsEnabled}
+        allowSelectUnitType={false} //todo @martha - discuss this ux. How should this dialog behave when unit group is new and doesn't yet have an established type?
```
martha: Flagging for discussion

src/modules/units/components/Units.tsx
```diff
@@ -143,8 +133,8 @@ const Units = () => {
         projectId={project.id}
         open={addUnitsDialogOpen}
         onClose={() => setAddUnitsDialogOpen(false)}
-        allowSelectUnitType={!unitGroupsEnabled}
-        allowSelectUnitGroup={unitGroupsEnabled}
+        allowSelectUnitType={true} //todo @martha - discuss this ux. How should this dialog behave when unit group is new and doesn't yet have an established type?
```
martha: Flagging for discussion: Currently, the UI does allow you to create units on the project Unit page (outside the Unit Group page). You have to select the unit group when you do so. However, that Unit Group might not have any existing units, in which case the backend doesn't know what Unit Type to use.

I think the most straightforward possible solutions are:
a) what's implemented here: pass `true` here; Unit Type dropdown is always shown, but if you try to select an invalid unit type for an existing unit group, you will get an error.
b) alternative: pass `false` here; Unit Type dropdown is not shown, and if you try to create a unit in a Unit Group that doesn't already have an established type, you get an error saying "unit type is required." We could make this error more descriptive/instructive, e.g. "to add units to this unit group, go to the unit group page." but that seems unnecessarily confusing.

Alternatively, we could do a little more refactoring to make this dynamic, but my understanding was that we want to avoid additional churn on the UI for this feature.

What do you think?

gigxz: Ah good find, I missed that in qa. So this would come up if the user creates a unit group but leaves it empty, and then goes back to the Units page and tries to add units to it from there. 

Another idea as a simple fix we can just exclude empty unit groups from this feature?
`unitGroups={unitGroups.filter((ug) => ug.unitTypes.length > 0)}` or something. It might be a bit more complex than that, but basically disallow adding units to the unit group from this modal if the unit  group is empty. 

(b) also seems fine. I would rather one of those than (a) because (a) makes the experience worse for the majority of cases, whereas (b) is only annoying in the specific edge case described.

martha: Sure, seems reasonable to me. https://github.com/greenriver/hmis-frontend/pull/1253/commits/689eceeeac75c04507d95ffd48cff8ae508d8022

src/modules/units/components/UnitManagementTable.tsx
```diff
@@ -51,12 +49,12 @@ const UnitManagementTable: React.FC<Props> = ({
     return [
       UNIT_COLUMNS.unitType,
       UNIT_COLUMNS.unitId,
-      ...(unitGroupsEnabled ? [UNIT_COLUMNS.unitGroup] : []),
+      UNIT_COLUMNS.unitGroup,
```
gigxz: Suggest hiding this column when you're looking at 1 group already? That table are a bit squished and it is duplicative with unit type. I'm not sure about getting rid of unit type altogether, since they could technically be named differently. But it feels like we shouldn't need to show the Unit Group column when you're looking at the Unit Group page.

`...(unitGroupId ? [] : [UNIT_COLUMNS.unitGroup]),`

gigxz (approved):

PR #1254, opened 2025-09-03
Simplify project CE interface for direct referral projects

gigxz:
## Description

* Hide Eligibility card from Referral Details if the project has no eligibility requirements
* Show text in MatchRuleCard if there are no cards (unit group and unit page)
* **Removes support for Unit Page** for projects that do not support waitlist referrals. This is an unnecessary page for those projects, it doesn't show anything useful, and adds navigational confusion. I updated the related components to pass around `coordinatedEntryFeatures` directly so it will be easier to flip this back later if/when we need to. (For example, if we add more attributes or functionality to the Unit page).
* **Hide Eligibility/Prioritization cards** for projects that do not support waitlist referrals. These values are only relevant for waitlist-based projects.

## Type of change
feature update

### DIFF
--- a/src/modules/ce/components/referral/ReferralDetailContent.tsx
+++ b/src/modules/ce/components/referral/ReferralDetailContent.tsx
@@ -207,15 +207,17 @@ const ReferralDetailContent: React.FC<Props> = ({ referral }) => {
           <CommonDetailGrid rows={sourceEnrollmentDetails} />
         </CommonCollapsibleCard>
       )}
-      {eligibilityRequirementsDetails && (
-        <CommonCollapsibleCard
-          title='Eligibility Requirements'
-          TitleComponent='h3'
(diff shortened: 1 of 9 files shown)


src/modules/ce/components/referral/ReferralDetailContent.tsx
```diff
@@ -207,15 +207,17 @@ const ReferralDetailContent: React.FC<Props> = ({ referral }) => {
           <CommonDetailGrid rows={sourceEnrollmentDetails} />
         </CommonCollapsibleCard>
       )}
-      {eligibilityRequirementsDetails && (
```
gigxz, line 210: No need to show this if the project has no requirements. For example for a direct referral, this is just an empty collapse header.

martha (approved):
great, thanks Gig!

PR #1257, opened 2025-09-05
Add Client Eligibility frontend form component

martha:
## Description

Summary of changes: Add new `ClientEligibility` form component that calls the `CalculateClientEligibility` mutation and displays eligible project types.

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5721

How to test:
- Use a form fragment such as:
```json
{
  "text": "Score?",
  "type": "INTEGER",
  "link_id": "score",
  "mapping": {
    "custom_field_key": "marthas_custom_assessment_score"
  },
  "disabled_display": "HIDDEN"
},
{
  "text": "Eligible Project Types",
  "type": "CHOICE",
  "link_id": "eligible_project_types",
  "mapping": {
    "custom_field_key": "marthas_custom_assessment_eligible_project_types"
  },
  "repeats": true,
  "component": "CLIENT_CE_ELIGIBILITY",
  "disabled_display": "HIDDEN",
  "pick_list_reference": "ProjectType"
}
```
- in combination with a rule similar to:
```ruby
Hmis::Ce::Match::Rule.create!(
  name: "Must have score of at least 5",
  rule_type: "eligibility_requirement",
  applicability_config: {},
  owner: Hmis::Hud::Project.find(703), # pick a project with a project type that the client isn't otherwise eligible for, for clearer testing
  expression: "`cde.custom_assessment.marthas_custom_assessment_score` != NULL AND `cde.custom_assessment.marthas_custom_assessment_score` >= 5",
)
```

Screenshot 2025-09-05 at 11 31 25 AM

## Type of change


src/modules/ce/components/ClientEligibility.tsx
```diff
@@ -0,0 +1,134 @@
...
+const projectTypePickList = localResolvePickList('ProjectType');
+
+interface ClientEligibilityProps extends DynamicInputCommonProps {
+  value?: PickListOption[] | null;
+  onChange?: (value: PickListOption[] | null) => void;
+  handlers?: FormDefinitionHandlers;
+}
+
+// Shows a button to calculate client eligibility based on current form values, and displays the result.
+const ClientEligibility = ({
```
martha: I played around with a refactor to reduce shared code with `AltAhaScore` by creating a new generic `MutationButtonField`.

I didn't really _love_ the result -- just based on vibes, it seems less graceful and maintainable to me -- so I reverted it. That work is here if you are curious to check it out: https://github.com/greenriver/hmis-frontend/commit/e95b45e4b8e3fe4100233a7f51ea5ddbc6550c25 

I do still think we might want to introduce a pattern like this if we expect to start adding more and more form components that fit this mold, so let's discuss.

src/modules/ce/components/ClientCeEligibility.tsx
```diff
@@ -0,0 +1,134 @@
...
+              projectTypePickList?.find(
+                (pickListOption) => pickListOption.code === projectTypeCode
+              )
+            )
+            .filter((option): option is PickListOption => option !== undefined);
+
+          onChange(matchingPickListOptions);
+        }
+      },
+      onError: (apolloError: any) => {
```
gigxz:
```suggestion
      onError: (apolloError) => {
```

src/modules/ce/components/ClientCeEligibility.tsx
```diff
@@ -0,0 +1,134 @@
...
+import {
+  CalculateClientCeEligibilityMutation,
+  PickListOption,
+  useCalculateClientCeEligibilityMutation,
+} from '@/types/gqlTypes';
+
+const projectTypePickList = localResolvePickList('ProjectType');
+
+interface ClientEligibilityProps extends DynamicInputCommonProps {
+  value?: PickListOption[] | null;
```
gigxz:
```suggestion
  value?: PickListOption[] | null; // array of PickListOptions representing Project Types that this client is eligible for
```

src/modules/ce/components/ClientCeEligibility.tsx
```diff
@@ -0,0 +1,134 @@
...
+          // Filter out undefined values to keep typescript happy
+          const matchingPickListOptions = eligibleProjectTypes
+            .map((projectTypeCode) =>
+              projectTypePickList?.find(
+                (pickListOption) => pickListOption.code === projectTypeCode
+              )
+            )
+            .filter((option): option is PickListOption => option !== undefined);
+
+          onChange(matchingPickListOptions);
```
gigxz: Suggestion to make this a bit more concise, without changing the logic. It would be nice to have a comment explaining why this needs to be represented as an array of PickListOptions (as opposed to an array of project types) but I wasn't totally clear on that. I think it's probably because we have some other code expecting that array values are always pick list shaped?

```suggestion
          // Convert the returned project type codes to pick list options
          // (because..? this is this expected format for form items that collect multi-value arrays?)
          const eligibleProjectTypesAsPickList = eligibleProjectTypes.map(
            (code) => ({ code, label: HmisEnums.ProjectType[code] })
          );
          onChange(eligibleProjectTypesAsPickList);
```

src/modules/ce/components/ClientCeEligibility.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  const hasRequiredContext = !!enrollmentId && !!formDefinitionIdentifier;
+
+  const handleFetch = useCallback(() => {
+    if (hasRequiredContext) {
+      calculateClientEligibility({
+        variables: {
+          enrollmentId,
+          formDefinitionIdentifier: formDefinitionIdentifier,
+          valuesByLinkId: handlers?.getValuesForSubmit().valuesByLinkId || {},
```
gigxz: Can we raise instead of silently passing `{}` if handlers is undefined at this point?
Another option is to make `handlers` a required prop and take [this approach](https://github.com/greenriver/hmis-frontend/blob/3dc0c8139484a08ea649a5430a4cd1cab0a4e3bf/src/modules/form/components/DynamicField.tsx#L173).

src/modules/ce/components/ClientCeEligibility.tsx
```diff
@@ -0,0 +1,134 @@
...
+        },
+      });
+    }
+  }, [
+    hasRequiredContext,
+    calculateClientEligibility,
+    enrollmentId,
+    formDefinitionIdentifier,
+    handlers,
```
gigxz: I'm not sure about depending on `handlers` here because I'm uncertain how often that changes. It might be safer to just depend on the specific function we're using:
```jsx
  const { getValuesForSubmit } = handlers; // assuming handlers is required
  const handleFetch = useCallback(() => {
    if (hasRequiredContext) {
      calculateClientEligibility({
        variables: {
          enrollmentId,
          formDefinitionIdentifier: formDefinitionIdentifier,
          valuesByLinkId: getValuesForSubmit().valuesByLinkId,
        },
      });
    }
  }, [
    hasRequiredContext,
    calculateClientEligibility,
    enrollmentId,
    formDefinitionIdentifier,
    getValuesForSubmit,
  ]);
```

src/modules/ce/components/ClientCeEligibility.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+      {value && value.length > 0 && (
+        <Stack direction='column' gap={0.5}>
+          {value.map((projectType) => (
+            <Typography key={projectType.code} variant='body2'>
+              {projectType.label}
+            </Typography>
+          ))}
+        </Stack>
+      )}
```
gigxz:
```suggestion
```

The values themselves should not be displayed, they should just be stored as the items value. If the form author wants them to be displayed, they could use an autofill (I think). In the immediate case, we will have a subsequent Item that displays "Eligible" if the list contains a certain set of project types, and "Ineligible" if the list does not contain any of those project types.

src/modules/ce/components/ClientCeEligibility.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+      {value && value.length === 0 && (
+        <Typography
+          variant='body2'
+          sx={{ fontStyle: 'italic', color: 'text.secondary' }}
+        >
+          No eligible project types found
+        </Typography>
+      )}
```
gigxz:
```suggestion
```

Also not needed, leaving this up to the form author as far as what to display.

src/modules/ce/components/ClientCeEligibility.tsx
martha: Yes that's right, that code is here: https://github.com/greenriver/hmis-frontend/blob/6c27c774eb1ed1b78f09e79f0fb7a888a2757322/src/modules/form/util/formUtil.ts#L799

I've added a comment referring to that method

src/modules/ce/components/ClientCeEligibility.tsx
martha: Sounds good.

Should this still display the values in the read-only view? It does by default since we've implemented it as a Choice element and haven't overridden the `DynamicView` component for this form type:
Screenshot 2025-09-10 at 3 50 43 PM

I'm finding that behavior helpful while testing, but we don't need to keep it if it isn't a client need.

src/modules/ce/components/ClientCeEligibility.tsx
gigxz: I see, thanks!

src/modules/ce/components/ClientCeEligibility.tsx
gigxz: Yes to just always hiding this component from read-only view, and see snippet on https://github.com/open-path/Green-River/issues/8129#issuecomment-3276528930 for how to add an additional item to display the result for testing/qa

gigxz (approved):
Github is wonky again with the diff (At least for me) but this looks ready to merge!

With the additional suggestion to hide result when rendering as `DynamicViewField`, but that doesn't need to block merge (or need re-review).

PR #1263, opened 2025-09-09
Support redirect from Unit page to All Units page

gigxz:
## Description

https://github.com/open-path/Green-River/issues/8124#issuecomment-3271697068

Bug: the CE Admin Available Units table links to the Unit Page, which displays "Not Found" if the project does not support waitlist-based referrals. This PR patches that by adding a redirect to the All Units Page if Unit page is not available.

We will likely want to revisit this-- it may be better for Available Units table to conditionally link (and display information to differentiate), but it currently doesn't have knowledge as to which Units are for "waitlist-based" vs "direct", so that will require some additional graphql work. We also may add back the Unit page at some point for direct referrals if needed. 

This PR solves the immediate bug so the user is not presented with an action that links to "Not Found".

## Type of change
bug fix

### DIFF
--- a/src/components/accessWrappers/ProjectRoute.tsx
+++ b/src/components/accessWrappers/ProjectRoute.tsx
@@ -1,3 +1,4 @@
+import { useMemo } from 'react';
 import { Navigate } from 'react-router-dom';
 
 import Loading from '../elements/Loading';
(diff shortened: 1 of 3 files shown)


martha (approved):
Great! Thanks.

PR #1268, opened 2025-09-10
Add Assessment Details column

martha:
## Description

Summary of changes:
- Add new Data Column for displaying the CDEs on an Assessment that have `show_in_summary: true`
- Display this col on the Client, Project, Enrollment, and Household assessment screens
- Resolve CDEs on Assessment conditionally, only if the column is shown
- Add new style util `textTruncationSx` for shared styles between this new column and the similar, existing Service Details column. I've just parked this in `theme.ts`, as it's a bit similar to `customVisuallyHidden`, but let me know if you think there's a better place for it.

How to test:
- Set some CDEDs to "show_in_summary: true." (See issue comment for ruby script) 
- Fill out an assessment that collects those CDEDs
- View the assessment in the different screens; you should be able to enable the assessment details column and see the summary details
- (unchanged but helpful: Once you enable the column on an enrollment, your preference persists across enrollments)

## Type of change
New feature

### DIFF
--- a/src/modules/assessments/util.tsx
+++ b/src/modules/assessments/util.tsx
@@ -1,13 +1,29 @@
+import { Stack, Typography } from '@mui/material';
 import { AlwaysPresentLocalConstants } from '../form/util/formUtil';
 import { ClientAssessmentType } from './assessmentTypes';
 import RelativeDateDisplay from '@/components/elements/RelativeDateDisplay';
 import { ColumnDef } from '@/components/elements/table/types';
+import { textTruncationSx } from '@/config/theme';
+import { DataColumnDef } from '@/modules/dataFetching/types';
 import { HhmAssessmentType } from '@/modules/enrollment/components/HouseholdAssessmentsTable';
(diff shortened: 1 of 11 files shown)


src/modules/assessments/util.tsx
```diff
@@ -99,6 +115,42 @@ export const ASSESSMENT_COLUMNS: {
...
+      <Stack>
+        {(assessment.customDataElements || [])
+          .filter((cde) => cde.displayHooks.includes(DisplayHook.TableSummary))
+          .map((cde) => {
+            const val = customDataElementValueAsString(cde);
+            return (
+              <Typography
+                key={cde.key}
+                variant='body2'
+                sx={textTruncationSx('300px')}
```
gigxz:
```suggestion
                sx={textTruncationSx()}
```

Is this necessary? I'm seeing that truncation still works appropriately when we don't pass a max width.  With 300px, it was cutting off my text even when there was plenty of space on the table to show it. It's probably not a big deal, but would err on the side of trying to show as much text as is possible based on the screen size.

gigxz (approved):
👍

PR #1270, opened 2025-09-11
Fix unit "referral in progress" status permission bug

gigxz:
## Description

Patch UI bugs for users who do not have permission to view all Referrals in the project, but do have permission to view units:
1. User should always see "Referral in Progress" Status on a unit, even if they don't have permission to resolve that referral.
2. User should never see "top candidates" on a unit that has a referral in progress, even if they don't have permission to see that referral. Also adds an extra check to ensure we don't expose "Start referral" button on a locked opportunity, even though the other topCandidate change already fixes that, just for good measure.

**How to test**: can mimic scenario by stubbing `opportunity.referral` query on the backend to return nil.

## Type of change
Bug fix

### DIFF
--- a/src/modules/units/components/UnitOverview.tsx
+++ b/src/modules/units/components/UnitOverview.tsx
@@ -8,6 +8,7 @@ import OpportunityBanner from '@/modules/ce/components/unit/OpportunityBanner';
 import { useProjectDashboardContext } from '@/modules/projects/components/ProjectDashboard';
 import { cache } from '@/providers/apolloClient';
 import {
+  CeOpportunityStatus,
   UnitDetailFieldsFragment,
   useMarkUnitsAvailableMutation,
 } from '@/types/gqlTypes';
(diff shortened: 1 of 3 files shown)


src/modules/units/components/UnitOverview.tsx
```diff
@@ -34,11 +35,16 @@ const UnitOverview: React.FC<Props> = ({ unit }) => {
       {opportunity?.active && (
         <Grid item xs={12}>
           <OpportunityBanner
-            topCandidate={opportunity.candidates.nodes[0]}
+            topCandidate={
+              opportunity.status === CeOpportunityStatus.Open
```
martha, line 39: Maybe the OpportunityBanner should be responsible for this? We already pass the whole `opportunity` as a prop to `OpportunityBanner` so, maybe instead of passing `topCandidate` as a separate prop, it should just be determined internally (and not determined if the opportunity is locked)

Relates to commented-out: https://github.com/greenriver/hmis-frontend/pull/1270/files#diff-91e910d49ad227c21a3eb39bd0b51adbcc133e0b685ddb9013638cdcfb12a426R125-R129

martha (approved):
Looks great, nice fix

src/modules/units/components/UnitOverview.tsx
gigxz: I started working on that but got muddled in the OpportunityBanner. I'm going to merge this as an immediate fix to the UI bug, and open another PR to refactor that component

gigxz: Opened https://github.com/greenriver/hmis-frontend/pull/1271... it has quite a bit of churn because OpportunityBanner was originally set up to support showing Closed/Accepted referral status too, which we no longer show. I think it might need more of a redesign to think through all these scenarios. (e.g. is accepting referrals but has no candidates; is accepting referrals and I cant see candidates; has an in progress referral I can't see; is not accepting referrals; etc.)

PR #1275, opened 2025-09-17
8123 increase proxy buffer in nginx to support long redirect urls

ttoomey:
nginx: remove Active Storage proxy; raise proxy header buffer to 32k

- Remove unused /rails/active_storage/blob location from default.conf.template
- Set proxy_buffer_size 32k in proxy_headers.conf.template to accommodate large upstream headers (e.g., Set-Cookie) and avoid “upstream sent too big header” errors

### DIFF
--- a/docker/templates/default.conf.template
+++ b/docker/templates/default.conf.template
@@ -45,17 +45,6 @@ server {
       proxy_set_header Host $backend_server;
     }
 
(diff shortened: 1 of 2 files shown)


gigxz (approved):

ttoomey:
deployed to QA, seems to address the 502 on redirect

PR #1284, opened 2025-09-22
Support flag to lock edits on a form

martha:
## Description

Summary of changes:
- Resolve new `access` and `adminEditableOnly` fields on the form definition object
- Display a flag if the form is locked for non-admins to edit
- Hide edit buttons if the current user doesn't have permission to edit

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/5796

How to test:
- After running the migration, manually update a form to have the 'admin_editable_only' flag set to true.
- Visit that form in the UI. It should display a flag stating that it is locked, and you should not be able to edit.
Screenshot 2025-09-22 at 3 52 44 PM
- As a super-admin, you should still be able to edit.
Screenshot 2025-09-22 at 3 53 18 PM

## Type of change
New feature


src/modules/admin/components/forms/FormDefinitionActionsCard.tsx
```diff
@@ -58,41 +57,46 @@ const FormDefinitionActionsCard: React.FC<Props> = ({ formIdentifier }) => {
             {publishedBy && `by ${publishedBy.name}`}
           </Typography>
         )}
-        <RootPermissionsFilter permissions='canManageForms'>
```
martha, line 61: The diff here is
`<RootPermissionsFilter permissions='canManageForms'>`
=> 
`{formIdentifier.access.canManageForm && (`
; the rest is unchanged.

src/modules/admin/components/forms/FormDefinitionDetailPage.tsx
```diff
@@ -51,6 +51,17 @@ const FormDefinitionDetailPage = () => {
                     </Typography>
                   </Alert>
                 )}
+                {formIdentifier.adminEditableOnly && (
+                  <Alert severity='info' icon={<LockIcon />}>
+                    <Typography variant='body2'>
+                      This form is locked
+                      {formIdentifier.access.canManageForm && (
+                        <> to non-admins</>
```
martha, line 59: We can simplify this and remove the extra flag if you have a preference :)

martha:
@gigxz the linked backend PR is approved, but I'll hold off on merging either until both frontend and backend are approved.

src/modules/admin/components/forms/FormDefinitionActionsCard.tsx
```diff
@@ -58,41 +57,46 @@ const FormDefinitionActionsCard: React.FC<Props> = ({ formIdentifier }) => {
...
-              )}
-            </>
-          )}
-        </RootPermissionsFilter>
+        {formIdentifier.access.canManageForm && (
+          <>
+            <Divider />
+            {formIdentifier.managedInVersionControl ? (
+              <DuplicateFormButton formIdentifier={formIdentifier} />
+            ) : (
```
gigxz: I didn't test this, but just want to confirm. Can users still duplicate forms that are managed in version control? I see `return false if resource.managed_in_version_control?` in the policy, so I just want to make sure we didn't lose that functionality.

If so, We may need another policy check to see if you can *duplicate- to go along with this UI and the `CreateDuplicateFormDefinition` mutation

martha: You are right, this is a regression. Good catch, I'll fix

martha: Is there any reason we don't allow duplicating a form that _isn't_ managed in version control? It seems like the UX is to provide an Edit button if it's editable and a Duplicate button otherwise, but I can't think of anything inherent in the policy that should _prevent_ users from duplicating a form that isn't managed in version control.

gigxz: I don't think we need to prevent it in the policy, but UI-wise imo we shouldn't expose that until/unless we have a plan for dealing with the `mapping`s. If you duplicate a custom assessment with a bunch of custom data elements, then those become shared across 2 forms, it would probably not be desired

src/modules/admin/components/forms/FormDefinitionActionsCard.tsx
```diff
@@ -58,41 +60,36 @@ const FormDefinitionActionsCard: React.FC<Props> = ({ formIdentifier }) => {
...
-                      ? formatRelativeDateTime(draftUpdatedOn)
-                      : ''}{' '}
-                    {draftUpdatedBy && `by ${draftUpdatedBy.name}`}
-                  </Typography>
-                </>
-              )}
-            </>
-          )}
-        </RootPermissionsFilter>
+        {canManageForm || canDuplicateForm || (canPublishForm && <Divider />)}
```
gigxz: Can you explain the logic here? I'm not really following. Wouldn't this just show `true` if `canManageForm` was true? Perhaps this was supposed to be `{(canManageForm || canDuplicateForm || canPublishForm) && <Divider />}`?

src/modules/admin/components/forms/FormDefinitionActionsCard.tsx
```diff
@@ -58,41 +60,36 @@ const FormDefinitionActionsCard: React.FC<Props> = ({ formIdentifier }) => {
...
+        {managedInVersionControl && canDuplicateForm && (
+          <DuplicateFormButton formIdentifier={formIdentifier} />
+        )}
+        {canManageForm && (
```
gigxz, line 70: Is the "can manage form" flag just used for gating the ability to edit form content and title, or is used for anything else? (eg adding/removing form rules)

gigxz (dismissed):

src/modules/admin/components/forms/FormDefinitionActionsCard.tsx
gigxz: If it's for editing, it might be clearer to say its for editing, but I think "manage" is also probably fine. It's just a bit more ambiguous.

src/modules/admin/components/forms/FormDefinitionActionsCard.tsx
martha: yes that sounds right, and I think matches what I have implemented -- the policy doesn't prevent duplication, but the frontend does prevent it unless the form is managed in version control.

src/modules/admin/components/forms/FormDefinitionActionsCard.tsx
martha: Oh yep you're right, this is a typo. It should be what you suggested. I'll fix!

martha: I updated it to be a bit more precise. I think the behavior we want is to show the divider if any of the buttons will be shown.

src/modules/admin/components/forms/FormDefinitionActionsCard.tsx
martha: Hm good question. I named it this only to match the name of the "can manage form" permission on the backend. On the frontend, it's currently only used for editing form content and title, as you said. 

It looks like we already use a different backend permission for managing form rules, "can configure data collection." I think in this case it might be clearer to leave as-is so the backend and frontend match?

gigxz (approved):

src/modules/admin/components/forms/FormDefinitionActionsCard.tsx
gigxz: Yep gotcha. The idea with the policies going forward is that the names don't _have to_ match the underlying permissions, I think we want to make them pretty specific so it's clear what actions the user can take. For example a policy might have "can edit thing" and "can publish thing" and "can create thing" even if they're governed by the same underlying perms for now. But no changes needed at this point I think we are good!

PR #1295, opened 2025-10-10
Pre-warm logo in cache before showing Login page

martha:
## Description

Summary of changes:
- Pre-warm the logo image in the backend cache before showing the Login screen.
- Motivation/backstory: 
  - What changed: In release-184, we updated asset compilation, impacting how logo images are served. 
    - The old version was reading a file off the file system, going through rails and nginx, but with no db query or s3 fetch involved.
    - The new version goes through rails and nginx, then makes a db query to determine if the image is stale. If it's stale, _or_ if the rails memory cache is old (>24h), it makes a fetch call to s3
  - Result: With these recent changes, if you don't have the logo cached, there is an awkward pause when the page has loaded but the logo has not yet. (This video is in my local dev environment, so it exaggerates the slowness. It should be faster on hosted envs.)
https://github.com/user-attachments/assets/af1b26b1-bd3d-4a92-af6b-d4f79b48fb65
    - Because of the 24h expiry of the rails memory cache, we expect users to see this effect about once per day.

## Type of change
UI cleanup

### DIFF
--- a/src/modules/hmisAppSettings/Provider.tsx
+++ b/src/modules/hmisAppSettings/Provider.tsx
@@ -104,9 +104,22 @@ export const HmisAppSettingsProvider: React.FC<Props> = ({ children }) => {
     const cachedUser = getValidCachedUser();
     const promises: Array<Promise<any>> = [];
 
-    const saveSettings = (value: HmisAppSettings) => {
+    // Pre-warm the backend rails cache with the logo image
+    const prefetchLogo = (logoPath?: string) => {
+      if (!logoPath) return Promise.resolve();
+      const src = `${window.origin}${logoPath}`;
+      return new Promise<void>((resolve) => {
+        const img = new Image();
+        img.onload = () => resolve();
+        img.onerror = () => resolve();
+        img.src = src;
+      });
+    };
+
+    const saveSettings = (value: HmisAppSettings): Promise<void> => {
       setAppSettings(value);
       storage.setAppSettings(value);
+      return prefetchLogo(value.logoPath);
     };
 
     if (cachedUser) {
@@ -114,6 +127,7 @@ export const HmisAppSettingsProvider: React.FC<Props> = ({ children }) => {
       const cachedAppSettings = storage.getAppSettings();
       if (cachedAppSettings) {
         setAppSettings(cachedAppSettings);
+        promises.push(prefetchLogo(cachedAppSettings.logoPath));
       } else {
         promises.push(fetchHmisAppSettings().then(saveSettings));
       }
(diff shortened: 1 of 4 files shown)


martha:
Tagging @eanders for review of my PR description, to make sure I've conveyed the situation correctly.

I also thought of a question while writing this up. Could we extend the rails memory cache lifespan? If we know when the image is stale and refresh the cache when that happens, what is the reason to explicitly invalidate the rails cache once every 24 hours?

gigxz:
This doesn't seem ideal to me still because we can't ensure that the input elements and buttons won't shift around when the logo loads, because the skeleton height may be different from the logo height. I think it would be better to prevent the whole login form from loading until we can display it fully.

Can we warm the cache by fetching the logo URL in the HmisAppSettingsProvider, so the app settings aren't considered fully fetched until the image has been loaded? I _think_ we could do that by prefetching the logo url somewhere in [this use effect](https://github.com/greenriver/hmis-frontend/blob/3eba5c326b5bf74a90b83e69562a14554ea17eb4/src/modules/hmisAppSettings/Provider.tsx#L102-L103).  That way, the user sees the usual Loading spinner until that image is prefetched as well. I think that would mimic the old experience, and allow us to keep the Login form as- is.

martha:
Thanks @gigxz for the good suggestion. This looks like it works pretty well. How I tested locally:

- Since Rack mini profiler disables the rails cache, set `RACK_MINI_PROFILER=off` in `.env.local` and then restart the backend web server.
- in a Rails console, run `Rails.cache.clear`
- in the browser (chrome), on the login page, right-click the refresh button, "Empty Cache and Hard Reload"
- the spinner spins until the image is loaded (no flicker) ✅  (at least without any browser throttling in place)

For comparison, following the same steps on release-184 without this change, there is a noticeable flicker.

src/modules/hmisAppSettings/Provider.tsx
```diff
@@ -104,16 +104,30 @@ export const HmisAppSettingsProvider: React.FC<Props> = ({ children }) => {
...
+        img.onload = () => resolve();
+        img.onerror = () => resolve();
+        img.src = src;
+      });
+    };
+
+    const saveSettings = (value: HmisAppSettings): Promise<void> => {
       setAppSettings(value);
       storage.setAppSettings(value);
+      return prefetchLogo(value.logoPath);
```
gigxz, line 122: Do we do something with the return value of `saveSettings` now? I'm not seeing that, but maybe I'm missing something

gigxz (approved):
looks good!

src/modules/hmisAppSettings/Provider.tsx
martha: Good question, I had the same confusion, but I believe (if I'm understanding chatgpt and [this corroborating article](https://web.dev/articles/promises#chaining) correctly) that thanks to promise chaining, when we do
```javascript
promises.push(fetchHmisAppSettings().then(saveSettings));
```
then if `saveSettings` also returns a promise, that last promise is the one that's appended to the `promises` array.

gigxz: Ah I see, thanks!

PR #1308, opened 2025-10-29
Fix visibility/UX for user who can view but not edit enrollments

martha:
## Description

Summary of changes:
For a user who can view, but not edit, an enrollment:
- If there is an in-progress household assessment
  - Prevent the user from navigating to the last (summary) page since there is nothing for them to do here.
  - Fix a logical bug that was showing unstarted forms as unlocked.
- On the Enrollment Overview page
  - Hide the "Go to Exit" button if no exit assessment exists.
  - Unstyle household tasks as links
  - ~Hide the Delete Enrollment button~ reverted, see inline comment thread

Depends on hmis-warehouse PR: No dependency, but I'll work on a backend PR for system tests in parallel.

## Type of change
Bug fix


src/modules/enrollment/components/pages/EnrollmentOverview.tsx
```diff
@@ -73,8 +73,16 @@ const EnrollmentOverview = () => {
...
+            {
+              // User sees the Delete Enrollment button here under two circumstances:
+              // 1. The enrollment is incomplete, and the user has permission to edit the enrollment
+              ((enrollment.access.canEditEnrollments &&
+                enrollment.inProgress) ||
+                // 2. The enrollment is complete but missing an intake (bad data), and the user has permission to delete the enrollment.
+                // (Normally completed enrollments can only be deleted by deleting their intake assessment, so this workaround enables cleaning up bad data)
+                (enrollment.access.canDeleteEnrollments &&
+                  !enrollment.inProgress &&
+                  !enrollment.intakeAssessment)) && (
```
martha: Dug this logic up referring back to https://github.com/greenriver/hmis-frontend/pull/614 and  https://github.com/greenriver/hmis-warehouse/pull/3926.

src/modules/enrollment/components/EnrollmentReminders.tsx
```diff
@@ -80,16 +80,21 @@ const reminderDesciption = (
...
       return (
         <Stack gap={0.4}>
-          {/* style reminder as a link, even though it's not (the row is clickable) */}
-          <Link component='span'>{reminderTitle(reminder)}</Link>
+          {/* if user can edit this enrollment, style the reminder as a link (the row is clickable) */}
+          {canEditEnrollment ? (
+            <Link component='span'>{reminderTitle(reminder)}</Link>
+          ) : (
+            reminderTitle(reminder)
+          )}
```
martha, line 98: I was a bit torn about whether to keep the link styling, since we now usually avoid styling links in table rows, and instead add a 3-dots menu at the end of the row for action items. But I don't think that pattern works very well here, since this is meant to be like a quick-action link, and it also seemed like too much churn to introduce. Instead, I've just made the link styling conditional on user edit access.

src/modules/assessments/components/household/HouseholdAssessments.tsx
```diff
@@ -257,6 +257,8 @@ const HouseholdAssessments: React.FC<Props> = ({
     px: 4,
   };
 
+  const showSummaryTab = enrollment.access.canEditEnrollments;
```
martha: I took the approach of simply hiding the whole summary tab, instead of making that tab view-only, since I don't think anything on there is relevant to a user who can view the enrollment but not edit any assessment data.

src/modules/enrollment/components/pages/EnrollmentOverview.tsx
gigxz: Ah, thanks for digging that up. Looks like the mutation is a bit more permissive than what we expose in the frontend. Was there a bug here related to this issue, or is this change just bringing the frontend in line with the mutation?


If I'm understanding this correctly, providers (who can edit enrollment but not delete enrollments) previously didn't see the "Delete Enrollment" button on the enrollment overview page. The only way they could perform the "delete" was to remove WIP enrollments from an existing household on the household management screen. So, this change is exposing a new way for them to delete WIP enrollments, and it's also allowing them to delete the HoH's WIP enrollment which wasn't previously exposed.

Am I getting that right? This change seems like it warrants some more discussion, and a release note if we choose to implement it. Lets check in on it next standup!

src/modules/enrollment/components/EntryExitDatesWithAssessmentLinks.tsx
```diff
@@ -34,6 +34,8 @@ const EntryExitDatesWithAssessmentLinks: React.FC<Props> = ({ enrollment }) => {
 
   const canLinkToIntake =
     enrollment.access.canEditEnrollments || !enrollment.inProgress;
+  const canLinkToExit =
+    enrollment.access.canEditEnrollments || enrollment.exitDate;
```
gigxz: We probably shouldn't link to starting an exit for an enrollment that is WIP, since it can't be exited yet.
```suggestion
    !enrollment.inProgress && (enrollment.access.canEditEnrollments || enrollment.exitDate);
```

src/modules/assessments/components/AssessmentForm.tsx
```diff
@@ -385,10 +385,14 @@ const AssessmentForm: React.FC<Props> = ({
...
-        {locked && assessment ? (
+        {locked ? (
+          // Locked scenarios:
+          // - Locked assessment that was previously submitted
+          // - Locked assessment that was previously saved, not submitted
+          // - Locked because the user doesn't have access to edit
           <FormContainer
             actions={
-              canEdit && !isPrintView ? (
+              assessment && canEdit && !isPrintView ? (
```
gigxz: I'm not sure about the approach of introducing a read-only view of a form where the underlying assessment doesn't exist. It doesn't really seem useful to display an empty form of all Data-Not-Collected values for an assessment that doesn't exist, and may even be misleading if the viewer thinks the assessment exists / has been started. It's also inconsistent with the individual-household experience, where read-only viewers can only navigate to assessments that _already exist_.

I'd like to propose 2 alternative approaches:
1. Adjust the `HouseholdAssessments` tabbed view to only display tabs for assessments that exist. (diff below). or,
2. Adjust `HouseholdAssessmentTabPanel` to just render `<AssessmentAlert locked />` instead of the whole IndividualAssessment component, for enrollments where there is no assessment and the viewer can't create one.




```diff
+++ b/src/modules/assessments/components/household/HouseholdAssessments.tsx
@@ -115,12 +115,18 @@ const HouseholdAssessments: React.FC<Props> = ({

   const hasRefetched = useHasRefetched(fetchAssessmentsStatus.networkStatus);

+  const readOnlyView = !enrollment.access.canEditEnrollments;
+
   useEffect(() => {
     if (!householdMembers) return;
     if (!assessmentByEnrollmentId) return;
     setTabs((oldTabs) => {
-      const newTabs: TabDefinition[] = householdMembers.map(
-        ({ client, enrollment, relationshipToHoH }, index) => {
+      const newTabs: TabDefinition[] = householdMembers
+        .filter(({ enrollment }) =>
+          // If user has read-only access, only show tabs for which there is an assessment
+          readOnlyView ? !!assessmentByEnrollmentId[enrollment.id] : true
+        )
+        .map(({ client, enrollment, relationshipToHoH }, index) => {
           const assessment = assessmentByEnrollmentId[enrollment.id];
           const assessmentId = assessment?.id;
           const assessmentInProgress = assessment?.inProgress;
@@ -165,11 +171,10 @@ const HouseholdAssessments: React.FC<Props> = ({
           }

           return tabData;
-        }
-      );
+        });
       return newTabs;
     });
-  }, [householdMembers, assessmentByEnrollmentId, role]);
+  }, [householdMembers, assessmentByEnrollmentId, role, readOnlyView]);
```

src/modules/enrollment/components/pages/EnrollmentOverview.tsx
martha: Makes sense, let's discuss more!

The existing bug(?) is that with the current frontend logic, a user who has "can delete" and _not_ "can edit" sees the button and gets an Access Denied error if they click it. That sounds more like a permission requirement issue: "can delete" should depend on "can edit". (We have never run into this in prod; I just noticed it while I was clicking around the enrollment overview UI with "can edit" permission removed.)

Sounds like we could revert this change, since that's a permission configuration that never arises in real life. 

I think this also relates to https://github.com/open-path/Green-River/issues/8067 so we could add a note there.

src/modules/assessments/components/AssessmentForm.tsx
martha: Yes that sounds good, I applied your diff.

I am on the fence about reverting these changes in AssessmentForm. It's this part that is making me nervous:
```jsx
{locked && assessment ? <DynamicView /> : <DynamicForm />
```

It seems bug-prone that we are ignoring `locked` when there is no assessment. Even if this wouldn't arise with the current code, it feels like a place where it would be easy to introduce regressions in the future. What do you think? Could we just keep the change I've made, and add a comment like this https://github.com/greenriver/hmis-frontend/pull/1308/commits/a1d2b472fd9666f3364e664aa6651954eae23502 to clarify we don't expect this to arise?

src/modules/enrollment/components/pages/EnrollmentOverview.tsx
martha: Discussed 11/3. Agreed on reverting the logical change I've made here. Instead made the minimal change to fix the bug - https://github.com/greenriver/hmis-frontend/pull/1308/commits/5bd15630b057e4f312ae4bc06bb053c4783a35f9

src/modules/assessments/components/AssessmentForm.tsx
gigxz: @martha good point, we weren't accounting for that before. I guess I would be in favor of throwing an error from the component if `locked` is true but `assessment` is null/undefined, since it's not something we can render. 
Looking at the component logic, we only [set locked to true](https://github.com/greenriver/hmis-frontend/blob/9341258f2f01eb855092cad98c2ad735e4cb5be9/src/modules/assessments/components/AssessmentForm.tsx#L121) if there is an assessment, so the problem seems to arise from the [initial state](https://github.com/greenriver/hmis-frontend/blob/9341258f2f01eb855092cad98c2ad735e4cb5be9/src/modules/assessments/components/AssessmentForm.tsx#L114-L116). The IndividualAssessment renders `AssessmentForm` passing it a null assessment, even if `enrollment.access.canEditEnrollments` is false - which is probably something we shouldn't do. It doesn't logically make sense to render a _form_ if the user doesn't actually have permission to submit that form. So maybe we can pull it up to the IndividualAssessment level, and have it throw or render a `<alert> assessment does not exist </alert>` in that case. What do you think of that?

martha: Thanks! I took a stab at this, but not sure I am totally following your suggestion. The `AssessmentForm` does still sometimes need to render a form if no assessment exists, when it's a user who _does_ have `canEdit` and they are filling out a new assessment. We can lift the `locked` state up into `IndividualAssessment`, which is somewhat neater (https://github.com/greenriver/hmis-frontend/pull/1308/commits/1151a62f9e6daa2bebcaf97ae74538ea98d89979). But the `assessment` prop can still be nil in `AssessmentForm`, so to assuage my concern from my comment above, we would still need something like this https://github.com/greenriver/hmis-frontend/pull/1308/commits/2dba3a961f858ef15f165e09b85a77a8d50d67ed handling the situation within that component, even though it should never arise with the current code.

gigxz: Ah I see. Sorry, I was having a hard time following what was happening because of how much is going on in the AssessmentForm component, and the amount of logic included inside the jsx. It could definitely use a refactor. I think you're right what you had before is clearer!

martha: - AssessmentForm should raise instead of displaying anything if the user can't edit but there is no assessment.
- typescript doesn't know the error has already been thrown, so revert back to the previous `locked && assessment && ...` but add a comment that this will not happen

gigxz (approved):
👍  looks good!

PR #1313, opened 2025-11-06
Support managing editor_user_ids in form builder

martha:
## Description

Summary of changes:
- Update the `ControlledSelect` RHF select element to support multi-select
- Add a multi-select for editor users to the Form Item editor

How to test: See linked ticket

## Type of change
New feature


src/modules/form/components/rhf/ControlledSelect.tsx
```diff
@@ -54,11 +56,38 @@ const ControlledSelect: React.FC<ControlledSelectProps> = ({
...
+      },
+      {}
+    );
+
+    if (multiple) {
+      if (!field.value) return [];
+
+      if (!Array.isArray(field.value))
+        throw new Error(
+          `ControlledSelect expects an array when multiple is true, but received ${field.value}`
```
martha: Let's discuss this. I think adding these new raises may make the component more brittle, so I'm a bit hesitant. But, we do want to surface developer errors.

I spent some time trying to avoid this by getting the typing to be a bit more sophisticated (https://github.com/greenriver/hmis-frontend/pull/1313/commits/51b424249d2ffc8dd6a4941c6c09013d8d2dc781), similar to how MUI implements Autocomplete for example. But, although it made a much nicer api, I wasn't happy with how much type casting this involved. I still think there might be a graceful way to do it, but started to feel I was spending more than reasonable time. Let me know if you want me to bring this back and pursue it more though.

gigxz: Hmm, good question. What if we split this logic into another component, so we retain `ControlledSelect` as-is and add `ControlledMultiSelect`? That should make the typing easier, and the component logic would be easier to reason about hopefully since we don't need to handle both cases.

martha: Thanks, that seems sensible! I also tried out another alternative of branching logic within the same component, similar to [TableFilterItemSelect](https://github.com/greenriver/hmis-frontend/blob/f2a1a7b74c097c3aad36cdcfbec0576b2153db08/src/components/elements/tableFilters/filters/items/TableFilterItemSelect.tsx#L6) (the component referenced in the comment at the top of ControlledSelect) which would let us keep the api with passing `multiple` as a prop, more similar to GenericSelect and mui's Autocomplete. https://github.com/greenriver/hmis-frontend/pull/1313/commits/1b54009a9e12465daafc865e6e63d27e41cb94bf. This try still involved some casting though.

Let me know if you have a preference between the two. I think your suggestion is probably more maintainable, so that's where I've left it.

gigxz (approved):
This approach looks good to me, thanks!

src/modules/form/components/rhf/ControlledSelect.tsx
gigxz: Nice. I agree separate component is more maintainable and better type safety!

PR #1314, opened 2025-11-07
Fix referral filter types

gigxz:
## Description

Issue: https://github.com/open-path/Green-River/issues/8473

Fix bug where tables were using the incorrect Filter type for the query. It would be better if we could have type checking here (e.g. ensure the filter type matches the filter type on the query variable type like `GetProjectOutgoingDirectCeReferralsQueryVariables`) but I wasn't able to figure that out in a reasonable timeframe.

## Type of change
Bug fix


martha (approved):
Thanks!

PR #1316, opened 2025-11-10
Fix CE a11y issues

martha:
## Description

I made a quick pass through the CE screens with the Axe devtools auto scanner, and it turned up a few issues that all had pretty straightforward fixes. It was quicker to just propose fixes than split out tickets, so I've done that here. I'll comment inline about the warnings fixed if it's not obvious.

## Type of change
a11y fix

### DIFF
--- a/src/components/elements/CommonButtonDrawer.tsx
+++ b/src/components/elements/CommonButtonDrawer.tsx
@@ -54,23 +54,25 @@ const CommonButtonDrawer: React.FC<Props> = ({
         }}
         {...DrawerProps}
       >
-        <Typography p={2} component='h2' variant='h5'>
-          {title}
-        </Typography>
-        <IconButton
-          aria-label='close'
-          onClick={() => setOpen(false)}
-          sx={{
-            position: 'absolute',
-            right: 8,
-            top: 8,
-            color: 'grayscale.light',
-          }}
-        >
-          <CloseIcon />
-        </IconButton>
-        <Divider orientation='horizontal' flexItem />
-        {children && React.cloneElement(children, { onClose: handleClose })}
+        <aside>
+          <Typography p={2} component='h2' variant='h5'>
+            {title}
+          </Typography>
+          <IconButton
+            aria-label='close'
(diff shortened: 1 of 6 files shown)


src/components/elements/table/GenericTable.tsx
```diff
@@ -646,7 +646,9 @@ const GenericTable = <T extends { id: string }>({
                   rowsPerPageOptions={[10, 25, 50, 100]}
                   slotProps={{
                     select: {
-                      'aria-label': 'rows per page',
+                      inputProps: {
```
martha, line 649: This was applying the aria label on the select component's parent div, which isn't valid a11y markup. The fix applies the aria label on the input itself.

src/components/elements/CommonButtonDrawer.tsx
```diff
@@ -54,23 +54,25 @@ const CommonButtonDrawer: React.FC<Props> = ({
...
-            right: 8,
-            top: 8,
-            color: 'grayscale.light',
-          }}
-        >
-          <CloseIcon />
-        </IconButton>
-        <Divider orientation='horizontal' flexItem />
-        {children && React.cloneElement(children, { onClose: handleClose })}
+        <aside>
```
martha, line 57: This fixes a warning "all page content should be contained by landmarks" because the MUI Drawer component renders content outside of `main`.

src/components/elements/CommonDetailGrid.tsx
```diff
@@ -97,7 +97,6 @@ export const CommonDetailGridItem: React.FC<{
         item
         {...valueBreakPoints}
         sx={itemSx}
-        aria-labelledby={labelId}
```
martha, line 100: I'm proposing to remove this because the common detail grid is usually used for displaying key/value text pairs, so I think it doesn't need to aria label the values, since aria labels should be used on interactive elements. (https://www.w3.org/TR/using-aria/) This fixes an axe warning related to wcag 4.1.2. Happy to revert this though if we think the change is too far-ranging

gigxz (approved):

PR #1320, opened 2025-11-12
Propose UI notice for stale opportunities

martha:
## Description

Following up from this [thread](https://github.com/greenriver/hmis-warehouse/pull/5997#discussion_r2519725367), this PR adds an Alert on the Unit page if the opportunity is active, but the rules are out of date.

- For now, this won't come up frequently because we are managing rules on the backend, so when we make updates to rules we should remember to refresh open opportunities.
- But, I was thinking that in the future, when we make rules manageable in the UI, we likely will provide the user with some control over whether opportunities get refreshed or not. If they choose not to refresh opportunities, then I think this kind of alert could be helpful for reminding them of why the ruleset listed here may differ from the ruleset listed on the parent Unit Group page.
- We can definitely workshop the language. 
  - "stop and re-start" is a little clunky
  - I also thought about drafting some language along the lines of "the same might also be true for other units in the unit group, so you can also go back up one level to the unit page and do this operation in bulk" (obviously phrased more gracefully than that!), but I think it might be better to just keep it simple.
  - edited to add: perhaps we should also make it clearer that "the requirements may be out of date" = that means the client list might also be outdated

Screenshot 2025-11-12 at 5 17 02 PM

## Type of change
New feature

### DIFF
--- a/src/modules/units/components/UnitOverview.tsx
+++ b/src/modules/units/components/UnitOverview.tsx
@@ -94,9 +94,8 @@ const UnitOverview: React.FC<Props> = ({ unit }) => {
               </LoadingButton>
             }
           >
-            This unit is stale, so the requirements below may appear out of
-            date. To bring them up to date, stop and then re-start accepting
-            referrals.
+            The requirements below may be outdated. To refresh them, stop and
+            re-start accepting referrals for this unit.
           </Alert>
         </Grid>
       )}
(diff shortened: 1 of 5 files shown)


src/modules/units/components/UnitOverview.tsx
```diff
@@ -69,6 +79,27 @@ const UnitOverview: React.FC<Props> = ({ unit }) => {
...
+                variant={'outlined'}
+                onClick={() => markUnitUnavailable()}
+              >
+                Stop Accepting Referrals
+              </LoadingButton>
+            }
+          >
+            This unit is stale, so the requirements below may appear out of
+            date. To bring them up to date, stop and then re-start accepting
+            referrals.
```
gigxz: How about:
> The requirements below may be outdated. To refresh them, stop and re-start accepting referrals for this unit.

I think "stale" language might not be clear, and also want to clarify the action that must be taken _on the unit_.

gigxz (approved):
Looks good! Tested stopping and re-starting from the unit page

PR #114, opened 2022-12-23
E2E test for Organization, Project, Funder, ProjectCoC, and Inventory

gigxz:
https://github.com/greenriver/hmis-warehouse/pull/2715

PR #1324, opened 2025-11-19
Fix misc keyboard accessibility issues

martha:
## Description

Summary of changes:
- Remove the override of 'enter' on TextInputs. [#117](https://github.com/open-path/hmis-accessibility/issues/117)
  - This was blocking the default MUI behavior that hitting 'enter' on a DatePicker's calendar adornment should open the datepicker popup.
  - This override was not working anyway, as of the RHF refactor. As of release-190, our current behavior is that hitting 'enter' while on a form text input submits the form. [QA example](https://qa-hmis.openpath.host/client/==SmJGR2FsS09FNmdIclVWOC9mRm1XUT09/enrollments/==SENHYjh5NFZaOHNIclVWOC9mRm1XUT09/assessments/==SDBwT3hXYTV5Y05WZnNEZjhHckN1dz09)
  - We can separately discuss whether that behavior should be prevented. We had overridden it in the past, pre-RHF. But fwiw, I think we should keep it, since it is the html standard. [source](https://www.w3.org/TR/2011/WD-html5-20110525/association-of-controls-and-forms.html#implicit-submission), [discussion](https://stackoverflow.com/questions/57064691/is-implicit-submission-submit-on-enter-key-useful-to-people-using-readers-and)
- Maintain focus on the Menu/Close buttons when the left-hand menu is opened/closed [#65](https://github.com/open-path/hmis-accessibility/issues/65)
- In Assessment navigation, as well as scrolling to the right spot on the page, update keyboard focus to the selected element [#114](https://github.com/open-path/hmis-accessibility/issues/114)
- Prevent the Edit Client form from submit-on-enter when radio button is focused. [#116](https://github.com/open-path/hmis-accessibility/issues/116)
  - The update matches our existing override for [Radio Inputs on Dynamic Forms](https://github.com/greenriver/hmis-frontend/blob/release-190/src/components/elements/input/RadioGroupInputOption.tsx#L41-L49).
  - in contrast to the above, this proposed change (and our existing pattern in dynamic forms) is counter to the html standard. I think it's justifiable because having the form submit in this situation is too unintuitive, especially because the form is long and the submit button is below the fold.

How to test: Each of the 4 sub-issues linked above has repro instructions.

## Type of change

### DIFF
--- a/src/components/elements/input/RadioGroupInputOption.tsx
+++ b/src/components/elements/input/RadioGroupInputOption.tsx
@@ -1,8 +1,9 @@
 import { Checkbox, FormControlLabel, Radio } from '@mui/material';
-import { KeyboardEventHandler, useCallback, useId } from 'react';
+import { useCallback, useId } from 'react';
 
 import CommonHtmlContent from '@/components/elements/CommonHtmlContent';
 import { PickListOption } from '@/types/gqlTypes';
+import { preventImplicitSubmission } from '@/utils/forms';
 
 interface Props {
   option: PickListOption;
@@ -38,16 +39,6 @@ const RadioGroupInputOption: React.FC<Props> = ({
     [onChange, disabled, option]
   );
 
-  // Prevent form submission on Enter. Enter should toggle the state.
-  const onKeyDown: KeyboardEventHandler<HTMLButtonElement> = useCallback(
-    (e) => {
-      if (e.key === 'Enter' || e.key === ' ' || e.code === 'Space') {
-        handleClick(e);
-      }
-    },
-    [handleClick]
-  );
-
   return (
     <>
       <FormControlLabel
@@ -58,7 +49,7 @@ const RadioGroupInputOption: React.FC<Props> = ({
         control={
           <ControlComponent
             disabled={disabled}
-            onKeyDown={onKeyDown}
+            onKeyDown={preventImplicitSubmission}
             data-checked={checked}
             inputProps={{
               'aria-labelledby': helperText
--- a/src/components/elements/input/RadioGroupInputOption.tsx
+++ b/src/components/elements/input/RadioGroupInputOption.tsx
@@ -1,5 +1,5 @@
 import { Checkbox, FormControlLabel, Radio } from '@mui/material';
-import { useCallback, useId } from 'react';
+import { KeyboardEventHandler, useCallback, useId } from 'react';
 
 import CommonHtmlContent from '@/components/elements/CommonHtmlContent';
 import { PickListOption } from '@/types/gqlTypes';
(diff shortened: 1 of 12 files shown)


src/components/elements/input/TextInput.tsx
```diff
@@ -62,9 +62,6 @@ const TextInput = ({
       id={htmlId}
       fullWidth={fullWidth}
       label={horizontal ? undefined : label} // hide the label if this text input is part of a horizontal group (informally deprecated)
-      onKeyDown={(e) =>
-        !props.multiline && e.key === 'Enter' && e.preventDefault()
-      }
```
gigxz, line 67: I'm not sure about this change, continuing conversation on the ticket https://github.com/open-path/hmis-accessibility/issues/116

src/modules/form/components/client/names/MultiNameInput.tsx
```diff
@@ -51,19 +62,26 @@ const MultiNameInput = ({ id, value, onChange, label }: Props) => {
...
+                control={
+                  <Radio
+                    onKeyDown={(e) => {
+                      // Prevent form submission on Enter. Enter should select the focused option instead.
+                      // (Following the same pattern from RadioGroupInputOption, used in Dynamic Forms)
+                      if (
+                        e.key === 'Enter' ||
+                        e.key === ' ' ||
+                        e.code === 'Space'
+                      ) {
```
gigxz: Should we pull this into a shared util or hook rather than repeating it? We have this type of code key-checking code (and preventing default etc) scattered around the input components, so it might be nice to clean that up a bit. Also recognizing that is scope creep and the proposed code here matches our current convention of inlining it.

src/components/elements/input/RadioGroupInputOption.tsx
```diff
@@ -38,16 +39,6 @@ const RadioGroupInputOption: React.FC<Props> = ({
     [onChange, disabled, option]
   );
 
-  // Prevent form submission on Enter. Enter should toggle the state.
-  const onKeyDown: KeyboardEventHandler<HTMLButtonElement> = useCallback(
-    (e) => {
-      if (e.key === 'Enter' || e.key === ' ' || e.code === 'Space') {
-        handleClick(e);
```
gigxz, line 45: It seems like now the space bar action sometimes toggles the state and sometimes doesn't. If Ii tab into the radio group directly and then hit space bar, it wont toggle the state (but it does on qa). However if I hit an arrow key to go up it down and _then_ hit the space bar, it does toggle the state. A bit of a weird regression, I think? Assuming that space bar still should support toggling.

gigxz (approved):
Looks great! The focus moving when you scroll to another form section is really handy. I just encountered one oddity with the radio button behavior.

src/components/elements/input/RadioGroupInputOption.tsx
martha: Thanks for the QA, you're right, this is a regression. I had been hoping we could just rely more on mui/html's defaults for tab and click behaviors, but because we want radio groups to be clearable, we need to handle it directly.

I think it should be fixed now - solved by rolling back the change to keep the previous `onKeyDown` handler.

gigxz: Ah I see, makes sense that clearable radio buttons is not the norm. The change looks good to me, thanks!

gigxz (approved):

PR #1326, opened 2025-11-24
Propose UI updates to referral forms

martha:
## Description

Summary of changes:
- Add the `without_top_level_cards` variant to the `DynamicForm` rendered on a referral step.
- Propose style updates on the Direct Referral sender form so it visually better matches assessments

See inline comment for screenshots of both changes.

## Type of change
UI update


src/modules/ce/components/referral/ReferralStepForm.tsx
```diff
@@ -85,6 +85,7 @@ const ReferralStepForm: React.FC<Props> = ({ step }) => {
     <>
       {editable ? (
         <DynamicForm
+          variant='without_top_level_cards'
```
martha, line 88: The actual form screenshotted below won't ever appear as editable in this context (it is filled out in the sending project for direct referrals.) I actually only noticed this and snagged the screenshots because of a [bug](https://github.com/greenriver/hmis-warehouse/pull/6036), but I thought that in general it seemed like something we should improve visually.

**before**:
Image

**after**:
Image


Also note, we don't have a similar variant `without_top_level_cards` for `DynamicView`, so the readonly step still looks like this:

Image

I haven't gone ahead with adding that yet because I started wondering if we should discuss further or bring this up with design

src/modules/ce/components/directReferral/SendReferralForm.tsx
```diff
@@ -32,84 +32,86 @@ const SendReferralForm: React.FC<Props> = ({ project }) => {
...
-          itemChanged={({ value }) =>
-            setFormState((old) => ({
-              ...old,
-              selectedProject: value,
-              selectedUnitGroup: undefined,
-            }))
-          }
-        />
-        {formState.selectedProject && (
+    <Stack gap={2}>
```
martha: I think github is making this diff look a bit more complicated than it is; it's just:
- wrap the whole component in a Stack
- remove the maxWidth on the Paper
- move the SendReferralSubForm out of the Paper

**before**:

Image

**after**:
Image
...
Image

For comparison, here's what it looks like if we keep the original styling (revert all changes here) but just add `variant='without_top_level_cards'` in `SendReferralSubForm`:

Image

martha: Discussed, we should use the variant instead of rendering in a separate card to support referral forms that don't have a top level group

src/modules/ce/components/referral/ReferralStepForm.tsx
martha: Discussed and decided we should add support by this for using the `DynamicFormLayout` in both `DynamicForm` and `DynamicView`

src/modules/form/components/viewable/DynamicView.tsx
```diff
@@ -53,19 +58,13 @@ const DynamicView: React.FC<
...
         }}
       >
-        <Grid
-          container
-          direction='column'
-          spacing={2}
-          data-testid='dynamicView'
-          {...GridProps}
-        >
+        <DynamicFormLayout variant={variant} GridProps={GridProps}>
```
martha, line 61: This removes `data-testid='dynamicView'`. Based on grepping both frontend and backend codebases, I _think_ that is currently unused (e.g. by system tests), but let me know if there's something I'm not thinking of here.

src/modules/form/components/DynamicFormLayout.tsx
```diff
@@ -1,18 +1,20 @@
...
   variant?: 'standard' | 'without_top_level_cards';
+  GridProps?: GridProps;
 }
 
 const DynamicFormLayout = ({
   errors,
   children,
   saveButtons,
   variant = 'standard',
+  GridProps,
```
martha, line 17: Support passing `GridProps` because it's currently used on the `DynamicView` in two places, both legacy-referrals-world:
- https://github.com/greenriver/hmis-frontend/blob/46530d98b3ab4a1091bec0a79d575c3d30f9aa33/src/modules/legacyReferrals/components/admin/AdminReferralPostingForm.tsx#L81-L85
- https://github.com/greenriver/hmis-frontend/blob/46530d98b3ab4a1091bec0a79d575c3d30f9aa33/src/modules/legacyReferrals/components/ProjectReferralPostingForm.tsx#L88-L95

we could also support this as another variant? Or, we could just remove support for this. If I'm understanding correctly, while we still support some views on legacy referrals, neither of these forms would be shown to users anymore (once #8631 is in place, soon)

martha:
thanks @gigxz for talking this over with me this morning, I made the updates we discussed and think it's ready for re-review.

gigxz (approved):
looks good, thanks!

src/modules/form/components/DynamicFormLayout.tsx
gigxz: I think fine to leave it as-is to avoid churn on these legacy components

PR #1328, opened 2025-11-25
Disable configuring custom form to collect HUD service type

martha:
## Description

Summary of changes:
- Update the New Form Rule dialog to use the Custom Services (type/category) picklists for custom forms, and the HUD picklists for the HUD service form. 
- ~Super-admins still see all options in the list, with a warning reminder to help prevent introducing bugs.~

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/6042

## Type of change
Bug fix


src/modules/admin/components/formRules/NewFormRuleDialog.tsx
```diff
@@ -194,19 +201,43 @@ const NewFormRuleDialog: React.FC<Props> = ({
     },
   });
 
+  // Logic for Service Type / Service Category picklists:
+  // - If the user has super-admin permissions (can administrate config), show all services including HUD and custom.
```
gigxz: I'm not sure we need this, as it can also cause confusion among super-admins (to be able to record a custom service using the HUD form and vice versa). Did you have a particular use-case in mind?

martha: Thanks for helping me think it through, I'll remove it!

I was mixed up about the use case where the standard service form, with a customized fragment, can collect a custom data element. That is allowed/supported, it just requires a manual update to the CDED so that the owner is `Hmis::Hud::Service` not `Hmis::Hud::CustomService`.

Now I realize, that's totally separate from the logic here. I can't think of a use case for actually recording a custom service using the HUD form. Sounds like we should remove support for that, for the same reason we're removing support for "mixed" service categories -- it would raise an error at the moment of data collection.

gigxz: Okay I see. Sounds good. Thanks!

gigxz (approved):

PR #1330, opened 2025-12-02
Remove empty group items when all child items are filtered out from Data Collected About rule

gigxz:
## Description

Issue: https://github.com/open-path/Green-River/issues/8472

* Update `applyDataCollectedAbout` such that it filters out `GROUP` items  when all child items were filtered out.
  * This mirrors logic on the backend which filters out groups when all child items are filtered out due to project applicability: https://github.com/greenriver/hmis-warehouse/blob/ef351980650f5ffb2583abd669b87193abd3e617/drivers/hmis/app/models/hmis/form/definition_item_filter.rb#L58-L59
  * Client applicability form filtering is currently handled on the frontend so that clients in the same household can use the same form definition (as opposed to each household member needing to fetch the FD separately to get their own client filtering).

**How to test**: render a form where there is a group and all children are not applicable due to `data_collected_about` value. The bug only presented when the child item _also_ had enable_when, due to [this unsafe accessor](https://github.com/greenriver/hmis-frontend/blob/1b528db5bb52587b41cbdef491cc864704df3e8d/src/modules/form/components/group/QuestionGroup.tsx#L24).

## Type of change
Bug fix

### DIFF
--- a/src/modules/form/components/group/QuestionGroup.tsx
+++ b/src/modules/form/components/group/QuestionGroup.tsx
@@ -21,7 +21,7 @@ const QuestionGroup = ({
   const hasTitle = !!item.text;
   const isConditionalGroup =
     !!item.item &&
-    (!!item.enableWhen || !!item.item[0].enableWhen) &&
+    (!!item.enableWhen || !!item.item[0]?.enableWhen) &&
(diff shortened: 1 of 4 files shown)


martha (approved):
Looks great to me. Thanks for the unit test too, cool!

PR #1332, opened 2025-12-03
Make form rules fetch more standard

martha:
## Description

[low priority] We're currently fetching Form Rules with a global Form Rules query, that is filtered by form definition ID, which I think is a non-standard pattern for us; see `FormProjectMatchTable` for comparison, for example. (I think this might be a historical relic since there used to just be one table of all form rules across the app?) And this seems to have required some caching acrobatics that I think can be simplified a lot if we take an approach that looks more like our usual.

This is a code cleanup, and isn't actually the fix to the bug in the linked ticket (that's [here](https://github.com/greenriver/hmis-warehouse/pull/6058)) -- so if we consider it too much churn, I'm fine with closing this without merge.

Summary of changes: 
- Update the `GetFormRules` query to accept a form definition ID and query for that definition's rules
- Update both `FormRuleTable` and `NewFormRuleDialog` to simplify the cache clearing
- No need to continue passing `formCacheKey` around

How to test:
- Open form definition detail page; confirm rules are fetched and displayed successfully
- Play with adding/removing form rules. Confirm that both the rules table and the project match table update automatically when changes are made.

## Type of change
Code cleanup

### DIFF
--- a/src/api/operations/config.queries.graphql
+++ b/src/api/operations/config.queries.graphql
@@ -1,20 +1,25 @@
 query GetFormRules(
+  $id: ID!
   $limit: Int = 25
   $offset: Int = 0
   $filters: FormRuleFilterOptions
   $sortOrder: FormRuleSortOption
 ) {
-  formRules(
-    limit: $limit
-    offset: $offset
-    filters: $filters
-    sortOrder: $sortOrder
-  ) {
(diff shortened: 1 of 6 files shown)


gigxz (approved):
much neater, thanks for the improvement!

PR #1344, opened 2025-12-30
Normalize initial value for boolean checkbox display

gigxz:
## Description

Issue: https://github.com/open-path/Green-River/issues/7239

#### Buggy behavior:
* A boolean checkbox field would have value 'undefined' if untouched, and would be saved as null (or not saved at all, in the case of custom fields). When form reopened, it displayed as "data not collected".
* If user checked and then unchecked it, it would save as false, and displays as "No" when reopened.

#### Updated behavior
* Boolean checkbox fields are always initialized to `false` if empty. When form is reopened for viewing, the value always shows as "No" if false or null.
* This also results in a response being recorded where there previously was none, specifically for CDEs. No CDE would be created if a checkbox was untouched, whereas now it will be created with value "false".

#### How to test:
1. **Testing backwards compatibility:** save or submit a form using the old code with a checkbox left untouched. Reopen, you should see "data not collected". Now switch to this branch and reopen the same assessment. It should now show "No"
2. **Testing new behavior:** save or submit a form with checkbox unchecked
3. **Testing default booleans**: boolean items that do NOT specify `"component": "CHECKBOX"` should be unchanged. Confirm they still support tri-state if the "yes no radio" component is left empty. 

### Screenshots
for reference:

Item `"type": "BOOLEAN"` (is actually tri-state)
Image

Item `"type": "BOOLEAN", "component": "CHECKBOX"` SHOULD be bi-state, but was tri-state. (untouched = null, check and uncheck = false, check = true)

Image

## Type of change
Bug fix

### DIFF
--- a/src/modules/form/hooks/rhf/useEnrichedFormData.tsx
+++ b/src/modules/form/hooks/rhf/useEnrichedFormData.tsx
@@ -10,6 +10,7 @@ import {
   getEnrichedValueForChoiceItem,
   getInitialValues,
   getItemMap,
+  normalizeBooleanCheckboxValues,
 } from '@/modules/form/util/formUtil';
 import { FormDefinitionJson } from '@/types/gqlTypes';
 
@@ -50,10 +51,12 @@ export const useEnrichedFormData = <T extends FieldValues>({
   });
 
   const [defaultValues] = useState<DefaultValues<T>>(() => {
+    // 1. Start with baseline values: form definition defaults (from "initial" attribute) + previously submitted data (from initialValues prop)
     const newValues: Record<string, any> = {
       ...getInitialValues(definition, localConstants),
       ...cloneDeep(initialValues),
     };
+    // 2. Layer on values that are calculated by "autofill_values" attribute on form items
     Object.keys(itemMap).forEach((linkId) => {
       const change = autofillValues({
         item: itemMap[linkId],
@@ -65,6 +68,9 @@ export const useEnrichedFormData = <T extends FieldValues>({
       //console.info('form init autofill', linkId, change)
       if (change) newValues[linkId] = change.value;
     });
+    // 3. Normalize boolean checkbox fields to be bi-state (true/false only)
+    normalizeBooleanCheckboxValues(newValues, itemMap);
+
     return newValues as DefaultValues<T>;
   });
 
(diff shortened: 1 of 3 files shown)


src/modules/form/hooks/rhf/useEnrichedFormData.tsx
```diff
@@ -65,6 +68,9 @@ export const useEnrichedFormData = <T extends FieldValues>({
       //console.info('form init autofill', linkId, change)
       if (change) newValues[linkId] = change.value;
     });
+    // 3. Normalize boolean checkbox fields to be bi-state (true/false only)
```
martha:
```suggestion
    // 3. Normalize boolean checkbox fields to be bi-state (true/false only) as opposed to tri-state (true/false/null)
```
?

src/modules/form/util/formUtil.normalizeBooleanCheckboxValues.test.ts
```diff
@@ -0,0 +1,179 @@
...
+      },
+    };
+
+    const result = normalizeBooleanCheckboxValues(values, itemMap);
+
+    expect(result.checkbox1).toBe(false);
+    expect(Object.keys(result)).toHaveLength(1);
+  });
+
+  it('handles empty itemMap', () => {
```
martha: Is this expected?

src/modules/form/util/formUtil.normalizeBooleanCheckboxValues.test.ts
```diff
@@ -0,0 +1,179 @@
...
+    expect(result.checkbox1).toBe(false);
+    expect(result.checkbox2).toBe(false);
+    expect(result.textField).toBe('some value');
+  });
+
+  it('leaves existing boolean checkbox values unchanged', () => {
+    const values = {
+      checkbox1: true,
+      checkbox2: false,
+      checkbox3: undefined,
```
martha, line 59: Reading the code, it took me a minute to parse why this test is different from the above, I think because of line 45 which creates overlap between the two tests. Maybe the two tests could be combined, or checkbox3 could be removed from the second test.

martha: Or maybe these both can be removed in favor of the test on line 103.

martha (approved):
Looks good! thank you!

src/modules/form/util/formUtil.normalizeBooleanCheckboxValues.test.ts
gigxz: no it's not

src/modules/form/util/formUtil.normalizeBooleanCheckboxValues.test.ts
gigxz: yeah those were redundant, I'll clean it up

PR #1351, opened 2026-01-07
Add UI for CE default contacts

martha:
## Description

Summary of changes:
- Add new Admin Default Contacts page for viewing all default contacts in the data source
- Add a new card on the Project Overview page for viewing this specific project's default contacts
- Add a modal for editing default contacts, which supports editing in both Project and Global modes
- Introduce a new 'warning' variant on `GenericSelect` to display when default contacts are missing or inactive

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/6145

How to test: See ticket comment

## Type of change


src/modules/ce/components/defaultContacts/AdminDefaultContactsTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+    key: 'organization',
+    render: (row: ProjectWithCeDefaultContactsFragment) =>
+      row.organization.organizationName,
+    optional: { defaultHidden: false },
+  },
+];
+
+interface Props {}
+const AdminDefaultContactsTable: React.FC<Props> = ({}) => {
+  const { data: { ceSwimlanes } = {}, loading, error } = useGetSwimlanesQuery();
```
martha, line 47: This component queries for all template swimlanes in order to know what to render in the column headers. I explored the alternative of using `getColumnDefs` to deduce the swimlanes/columns from the project query results, but the challenge is that, with pagination, the first page of results might not include all the swimlane columns that should be shown.

src/components/elements/input/GenericSelect.stories.tsx
martha, line 1: This PR introduces new 'warning' and (now unused) 'error' treatments on GenericSelect. 
Image

If this feels out of scope I'm happy to revert and just use our existing 'warnIfEmpty' treatment. See thread: https://www.figma.com/board/axGI9vwmuBRC4uZ11ZJ3RP?node-id=3902-14009#1511398207

src/modules/ce/components/defaultContacts/AdminDefaultContactsTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+          GetDefaultSwimlaneAssignmentsQuery,
+          GetDefaultSwimlaneAssignmentsQueryVariables,
+          ProjectWithCeDefaultContactsFragment
+        >
+          columns={columns}
+          queryVariables={{}}
+          filters={filters}
+          queryDocument={GetDefaultSwimlaneAssignmentsDocument}
+          pagePath='projectsWithCeDefaultContacts'
+          paginationItemName='coordinated entry project'
```
martha: User-facing record name is "coordinated entry project" to try to clarify why some projects may not appear in this list. ~It may be confusing that all projects still appear in the table's Project Filter dropdown. We could make a new customized filter/dropdown to avoid that confusion, but I am not necessarily sure the complexity is worth it?~ this no longer uses a custom Project Filter

src/modules/projects/components/ProjectOverviewPage.tsx
```diff
@@ -96,6 +97,9 @@ export const ProjectFormTitle = ({
 
 const ProjectOverviewPage = () => {
   const { project } = useProjectDashboardContext();
+  const showDefaultContactCard =
```
martha, line 100: Is there any specific permission governing whether you can _view_ (not edit) the project's Default Contacts card? I didn't see anything that seemed like a good fit, so I figured you can always view this if you can view the project?

src/components/elements/input/GenericSelect.tsx
```diff
@@ -80,14 +103,24 @@ const GenericSelect = <
         tagValue.map((option: T, index: number) => {
           // Avoid console warning: A props object containing a "key" prop is being spread into JSX
           const { key, ...rest } = getTagProps({ index });
+
+          // If the selected value is an object with a 'disabled' property set to true, disable the chip from being unselected
```
gigxz: Looks like the Autocomplete component provides a prop for this: https://mui.com/material-ui/react-autocomplete/#disabled-options

Here's one way we could implement it:

```tsx
    // default method to determine if an option is disabled
    const getOptionDisabled = (option: T) => {
      if (getOptionDisabledProp) { // if method was provided as a prop, use that
        return getOptionDisabledProp(option);
      }
      if (isPickListOption(option)) { // convenience: define default behavior for PickListOption type
        return !!option.disabled;
      }
      return false;
    }


  return (
    <Autocomplete
      ...
      getOptionDisabled={getOptionDisabled}
```

src/modules/admin/hooks/useAdminDashboardNavItems.tsx
```diff
@@ -23,6 +23,13 @@ export const useAdminDashboardNavItems = () => {
             permissions: ['canAdministrateCoordinatedEntry'],
             hide: !globalFeatureFlags?.coordinatedEntryEnabled,
           },
+          {
+            id: 'default-contacts',
+            title: 'Default Contacts',
```
gigxz: I would put this at the bottom of the CE section, since it's not part of day-to-day workflow operations

src/modules/ce/components/defaultContacts/EditCeDefaultContactsModal.tsx
```diff
@@ -0,0 +1,322 @@
...
+            helperText={`Tasks: ${swimlane.taskNames.join(', ')}`}
+            color={
+              showMissingWarning || anyUserInactive ? 'warning' : undefined
+            }
+          />
+        </Box>
+      );
+    },
+    [formState, handleChangeUsers, projectMode, usersPickList]
+  );
```
gigxz: This looks like a good candidate for pulling out into a separate component:
````tsx
<SwimlaneUserSelect
     value={formState[swimlane.id] || []}  
     onChange={(value) => handleChangeUsers(swimlane.id, value)}
     warnIfEmpty={} // whether it should use warning treatment when empty
     pickListArgs={} // args to pass to picklist fetching query
/>
```

src/modules/ce/components/defaultContacts/EditCeDefaultContactsModal.tsx
```diff
@@ -0,0 +1,322 @@
...
+            Cancel
+          </Button>
+          {(loading || hasSwimlanes) && (
+            <LoadingButton
+              onClick={handleSubmit}
+              type='submit'
+              loading={submitLoading}
+              disabled={loading || submitLoading}
+            >
+              Save
```
gigxz, line 213: I'm not seeing the cache update after saving. Can you reproduce this?
1. Click "Edit global contacts"
2. Go to a select box that's empty. Select a user
3. Hit save, wait for modal to close
4. Click "Edit global contacts" again
5. Unexpected behavior: the select box that I previously selected a user for is still empty
6. Refresh the page, reopen the modal and the value is now selected

src/components/elements/input/GenericSelect.tsx
gigxz: I'm also not seeing the disabled values handled correctly (Eg I can still remove them from the dropdown or by hitting backspace), but hopefully using the Autocomplete native functionality will fix that.

I also noticed that when I have the same user applied twice to a swimlane (once globally and once at project-level), I get a console error about the code not being unique. That should be resolvable by using a unique id for the code, for example the database id of the ce_default_swimlane_assignment

src/modules/ce/components/defaultContacts/AdminDefaultContactsTable.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+interface Props {}
+const AdminDefaultContactsTable: React.FC<Props> = ({}) => {
+  const { data: { ceSwimlanes } = {}, loading, error } = useGetSwimlanesQuery();
+  const [editingProject, setEditingProject] =
+    useState<ProjectWithCeDefaultContactsFragment | null>(null);
+
+  const filters = useFilters({
+    type: 'ProjectFilterOptions',
+    omit: ['status', 'funder', 'organization', 'projectType', 'ceEnabled'],
```
gigxz: I'm seeing 'searchTerm' as the only filter - search term we always pull out into a search bar at the top rather than keeping it in the filter. Maybe we can add that pattern here and remove the filter altogether? (or keep projectType/organization those seem potentially useful.

src/modules/ce/components/defaultContacts/AdminDefaultContactsTable.tsx
```diff
@@ -0,0 +1,141 @@
...
+        key: swimlane.id,
+        optional: { defaultHidden: false },
+        render: (row: ProjectWithCeDefaultContactsFragment) => {
+          if (!row.ceSwimlanes.find((s) => s.id === swimlane.id)) {
+            // The project's applicable swimlanes list doesn't include this column, so show n/a
+            return (
+              <Typography variant='body2' color='text.secondary'>
+                Not applicable
+              </Typography>
+            );
```
gigxz: Suggest using shared component:
```suggestion
            return <NotCollectedText>Not applicable</NotCollectedText>;
```

src/modules/ce/components/defaultContacts/EditCeDefaultContactsModal.tsx
gigxz: This should be resolved with https://github.com/greenriver/hmis-frontend/pull/1351/commits/0f434a964c25712e01743c870a5a3ac5c7eb3fe9, but definitely  should test this again to confirm

src/modules/ce/components/defaultContacts/EditCeDefaultContactsModal.tsx
gigxz: @martha take a look at https://github.com/greenriver/hmis-frontend/pull/1351/commits/0f434a964c25712e01743c870a5a3ac5c7eb3fe9 which has the changes we discussed! I left out the swimlane selection refactor with some todos, which are referring to this comment.

src/modules/projects/components/ProjectOverviewPage.tsx
gigxz: Yes I think that's fine!

src/modules/ce/components/defaultContacts/AdminDefaultContactsTable.tsx
gigxz: This seems reasonable to me!

src/components/elements/input/GenericSelect.tsx
martha: `getOptionDisabled` is MUI's api for disabling certain options in the dropdown, but what we want here is slightly different: to lock certain tags that are selected. I was basing my code on [this example](https://v5.mui.com/material-ui/react-autocomplete/#fixed-options) which uses `renderTags`. It looks like there is still more to debug there, so thanks for the catch!

martha: I think the behavior should be improved now. I'm not necessarily delighted with the churn ([here](https://github.com/greenriver/hmis-frontend/pull/1351/changes#diff-1ac9a3e40d32a9f1204667cfa64341814c210321d1a7d47e72a0f8bae4a18ab4R60-R89)) but curious what you think.

src/modules/ce/components/defaultContacts/EditCeDefaultContactsModal.tsx
martha: This looks resolved, thanks!

src/modules/ce/components/defaultContacts/EditCeDefaultContactsModal.tsx
martha: Thanks, your changes look good! I think we should return some kind of loading state for the global modal, otherwise it feels like the button is just really slow. I can take a shot at that.

martha: Let me know what you think of https://github.com/greenriver/hmis-frontend/pull/1351/commits/dbc51c7149af9f303e026102dbdb2d33df8003f1. It's not perfect because there is still kind of a flicker, see video. But maybe acceptable.


https://github.com/user-attachments/assets/e55777f9-fcdf-4c93-9a9c-aed239a5d8ec

src/modules/ce/components/defaultContacts/EditProjectCeDefaultContactsModal.tsx
```diff
@@ -0,0 +1,40 @@
...
+  project,
+  open,
+  onClose,
+}) => {
+  const dialogContentHeader = (
+    <Box>
+      <Stack direction='row' spacing={4}>
+        <CommonLabeledTextBlock title='Project'>
+          {project.projectName}
+        </CommonLabeledTextBlock>
```
gigxz, line 22: Whoops I accidentally removed this! I tested the below suggestion
```suggestion
        <CommonLabeledTextBlock title='Project'>
          {project.projectName}
        </CommonLabeledTextBlock>
        <CommonLabeledTextBlock title='Organization'>
          {project.organization.organizationName}
        </CommonLabeledTextBlock>
```

src/components/elements/input/GenericSelect.tsx
gigxz: Nice, thanks for that update. I tested it out and it seems functional now. I think it's a bit weird for the parent of the component to need to pass `fixedOptions` AND make sure that each fixed option has `disabled: true` in the picklist option, and if those 2 are mismatched then the disabled treatment doesn't match the fixed option behavior. Instead, I'd suggest doing what the MUI example does and disabling based on presence in fixedOptions list. Here is what I tested that seemed to work:

```diff
@@ -144,9 +136,8 @@ const GenericSelect = <
           // If the selected value is an object with a 'disabled' property set to true, visually disable the chip.
           // See https://v5.mui.com/material-ui/react-autocomplete/#fixed-options
           // Note that this *only* visually disables the chip. The parent component should modify onChange to prevent the option from being deselected.
-          const disabled = isPickListOption(option)
-            ? !!option.disabled
-            : !!rest.disabled;
+          const disabled = fixedOptions?.includes(option) || rest.disabled;
+
``

Comment (gigxz) [reply src/components/elements/input/GenericSelect.tsx]: Re:code churn, I think it's ok, it could be made more brief by not destructuring reason/details – also makes it clearer to the reader that the method is not messing with those.

```diff
--- a/src/components/elements/input/GenericSelect.tsx
+++ b/src/components/elements/input/GenericSelect.tsx
@@ -61,28 +61,20 @@ const GenericSelect = <
     (
       event: React.SyntheticEvent,
       newValue: AutocompleteValue<T, Multiple, boolean, Creatable>,
-      reason: AutocompleteChangeReason,
-      details?: AutocompleteChangeDetails<T>
+      ...args: [AutocompleteChangeReason, AutocompleteChangeDetails<T>?]
     ) => {
       // Only handle fixedOptions when multiple is true and fixedOptions are provided
       if (fixedOptions && rest.multiple && Array.isArray(newValue)) {
         // Ensure fixedOptions are always included, then add any additional selected options
+        // Cast is needed because TypeScript can't infer that Multiple=true means AutocompleteValue is T[],
+        // but we've verified the runtime conditions so it is safe
         const mergedValue = [
           ...fixedOptions,
           ...newValue.filter((option: T) => !fixedOptions.includes(option)),
-        ];
-
-        // TypeScript can't infer that Multiple=true means AutocompleteValue is T[],
-        // but we've verified the runtime conditions, so this is safe
-        return onChange?.(
-          event,
-          mergedValue as AutocompleteValue<T, Multiple, boolean, Creatable>,
-          reason,
-          details
-        );
+        ] as AutocompleteValue<T, Multiple, boolean, Creatable>;
+        return onChange?.(event, mergedValue, ...args);
       }
-
-      return onChange?.(event, newValue, reason, details);
+      return onChange?.(event, newValue, ...args);
     },
     [onChange, fixedOptions, rest.multiple]
   );
```

src/modules/ce/components/defaultContacts/EditCeDefaultContactsModal.tsx
gigxz: This works 👍

src/modules/ce/components/defaultContacts/AdminDefaultContactsTable.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+  if (loading) return <Loading />;
+  if (error) throw error;
+
+  return (
+    <>
+      <Stack gap={2}>
+        <CommonSearchInput
+          label='Search project default contacts'
+          name='search project default contacts'
```
gigxz: I think "Search projects" is clearer? since you're not able to search by contact name.

src/components/elements/input/GenericSelect.tsx
martha: Thanks, yeah that was my idea when introducing `fixedOptions` but I didn't carry it through 🤦‍♀️ ! Thanks for the improvements.

gigxz (approved):

PR #1376, opened 2026-02-11
Occurrence point bug fix

gigxz:
## Description

Fix bug where changing the value of an "occurrence point field" (Enrollment details card) would cause other values to appear as missing.

Ticket: https://github.com/open-path/Green-River/issues/8816

## Type of change
Bug fix


src/modules/form/components/DependentFormItemWrapper.tsx
```diff
@@ -35,7 +35,14 @@ const DependentFormItemWrapper: React.FC<Props> = ({ item, renderChild }) => {
...
       unregister(item.linkId);
     };
-  }, [disabled, visible, unregister, register, item]);
+  }, [
+    disabled,
+    visible,
+    unregister,
+    register,
+    item.disabledDisplay,
+    item.linkId,
```
gigxz, line 44: Depend on specific needed fields rather than rerunning if the `item` object ref changes. This was happening because the mutated definition was not memoized so it would change on re-render, and this hook would cause the field to unmount when the `item` object ref changed.

This change alone would fix the presenting bug. However the "real" fix is memoizing the form, which I did as well, but I chose to keep this change because it makes sense to depend on fewer variables.

src/modules/form/components/OccurrencePointForm.tsx
```diff
@@ -73,6 +73,16 @@ const OccurrencePointForm: React.FC<OccurrencePointFormProps> = ({
     [localConstantsProp]
   );
 
+  // Keep a stable reference to definitionForDisplay to avoid unnecessary re-renders of DynamicView.
+  // This could be avoided if the parent component memoized the definitionForDisplay.
```
gigxz, line 77: I avoided a larger refactor to have the parent pass a memoized prop, because I was spending more time than allotted on this bug fix

martha (approved):
Good find 🥳

PR #1393, opened 2026-03-02
Minimatch resolutions for vulnerability

gigxz:
## Description

Fix the minimatch vulnerability CVE-2026-27903 by patching each affected major version in place using Yarn resolutions, instead of forcing the whole tree onto a single minimatch version.

In tree | Patched version | Notes
-- | -- | --
3.1.2 | 3.1.4 | Hoisted; used by glob, eslint, eslint-plugins (older stack).
8.0.4 | 8.0.6 | Only under @sentry/bundler-plugin-core → glob.
9.0.5 | 9.0.7 | typescript-estree, Storybook, graphql-config, etc.

**Why not pin everything to one version?**
Forcing the entire dependency tree to a single minimatch version (e.g. 9.0.7) would have meant bumping some dependencies from 3.x to 9.x. minimatch has had breaking API and behavior changes across major versions, so that kind of bump felt too risky without dedicated testing of every consumer (ESLint, glob, Storybook, etc.). Patching each major line keeps behavior compatible while addressing the CVE.

**Follow-up**
Will open a ticket to upgrade eslint to a supported, modern version (incl migrating to flat config).

## Type of change
Dependency updates

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,12 @@
     ]
   },
   "resolutions": {
-    "giget": "^3.1.2"
+    "giget": "^3.1.2",
+    "glob/minimatch": "8.0.6",
+    "eslint/minimatch": "3.1.5",
+    "eslint-plugin-import/minimatch": "3.1.5",
+    "@storybook/cli/minimatch": "9.0.9",
+    "graphql-config/minimatch": "9.0.9"
   },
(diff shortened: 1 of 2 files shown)


martha (approved):
Looks good. Thanks for the explanation!

PR #1405, opened 2026-03-16
Support SearchQuery object in URL params for client search

martha:
## Description

Summary of changes:
- Add a new hook `useResolvedSearchQueryId` which resolves a search query ID to the params that it corresponds to
- Update the `ClientSearch` component:
  - Remove all previous search-params behavior
  - Implement new search params using the new `useResolvedSearchQueryId` hook as well as the existing `useSearchParamsState` hook

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/6300

## Type of change


martha:
@gigxz requesting your preliminary review to see if this seems like an acceptable approach. I have some concerns about brittleness of the new effects in `ClientSearch`, but I think to some extent, this kind of code is necessary when dealing with url search params, because we need to both control/set the params and listen/respond to changes.

src/modules/search/components/ClientSearch.tsx
```diff
@@ -115,26 +117,80 @@ export const MOBILE_SEARCH_RESULT_COLUMNS: ColumnDef<ClientSearchResultFieldsFra
...
+  // but we can interact with them as if they are React state
+  const [{ searchQueryId, searchType, displayType }, setSearchParams] =
+    useSearchParamsState({
+      paramsDefinition: {
+        // the ID of the ClientSearchQuery from the backend
+        searchQueryId: { type: 'string', default: null },
+        // Whether the search type is broad or specific (advanced)
+        searchType: { type: 'string', default: 'broad' },
+        // Whether the results are displayed as a table or cards
+        displayType: { type: 'string', default: 'table' },
```
gigxz: @martha this looks like a great start!
I have the following suggestions, which I think will simplify the change:
1. Pull in https://github.com/greenriver/hmis-frontend/pull/1407, which moves "search type" state into the routing layer. With that change, it no longer needs to be managed internally in this component.
2. Descope moving `displayType` into the search parameters. This doesn't feel critical to this refactor, and I think it would be simpler to reason about if we are only dealing with searchQueryId for this branch. (Also, `displayType` is purely client-side UI state, it is not relevant to the server. So It might make more sense to thrown in a URL hash (eg. `#cards`) as opposed to using search params. But that's maybe a style thing. Doesn't feel super relevant to the search query ID work).

src/modules/search/components/ClientSearch.tsx
```diff
@@ -265,7 +291,34 @@ const ClientSearch = () => {
           >
             queryVariables={{ input: searchInput }}
             queryDocument={SearchClientsDocument}
-            onCompleted={() => setHasSearched(true)}
+            onCompleted={() => setHasSearched(true)} // set hasSearched when the GenericTable returns any results (whether from network or cache)
+            onCompleteNetworkQuery={(data) => {
```
gigxz: Yikes, that seems like pretty unexpected behavior for an `onCompleted` callback. I can look into that and see if we really need the extra invocations. It looks like it was added for the initial bed night UI.

src/modules/search/hooks/useResolvedSearchQueryId.ts
```diff
@@ -0,0 +1,33 @@
+import { useGetSearchQueryQuery } from '@/types/gqlTypes';
+
+interface UseResolvedSearchQueryIdProps {
+  searchQueryId?: string;
+  user?: any;
+  onCompleted?: (resolvedParams: Record<string, string>) => void;
```
gigxz: The pattern of this hook is a bit unusual to me. Typically we just return the data itself, usually memoized using useMemo. What's the reasoning for exposing an onCompleted callback?

For example
```tsx
  // Resolve the search query ID into usable search params (memoized)
  const { resolvedSearchInput, loading: searchQueryLoading } = useResolvedSearchQueryId({
    searchQueryId
  });
  // If the search query ID was resolved to a new set of search params, update the search input to perform the search
  useEffect(() => {
    if (resolvedSearchInput) setSearchInput(resolvedSearchInput);
  }, [resolvedSearchInput]);

```

src/modules/search/components/ClientSearch.tsx
gigxz: This had a very narrow use, I opened https://github.com/greenriver/hmis-frontend/pull/1408 to rename it

src/modules/search/hooks/useResolvedSearchQueryId.ts
```diff
@@ -0,0 +1,33 @@
...
+    // cache-first is Apollo's default; Setting it explicitly to call out that having this uuid in the cache already is likely
+    // if the user is navigating with the back-button, so we don't want to hit the network again in that case
+    fetchPolicy: 'cache-first',
+    onCompleted: (data) => onCompleted?.(data?.searchQuery?.params),
+  });
+
+  if (error) throw error;
+
+  return {
+    resolvedParams: data?.searchQuery?.params,
```
gigxz: Outstanding work lists `validate/normalize incoming params`. But I am wondering if this query can just return a typed object instead?

I know we intend to use this elsewhere, but all other usages should only use a single "free text" field. So I think we could expose it as a type that closely matches `HmisSchema::ClientSearchInput`?


text_search (can be reused across other use-cases that support free-text search which may always contain PII)
personal_id (only used for advanced client search)
warehouse_id (only used for advanced client search)
first_name (only used for advanced client search)
last_name (only used for advanced client search)
ssn_serial (only used for advanced client search)
dob (only used for advanced client search)

martha: Thanks, I think that works and I've pushed a preliminary approach. I think some of the code could probably be neatened up a bit, so I'll work on that next after addressing other comments.

src/modules/search/hooks/useResolvedSearchQueryId.ts
martha: Hm, my reasoning was just that in general I try to avoid `useEffect`s when possible because I find they can be hard to follow and lead to weird bugs (https://react.dev/learn/you-might-not-need-an-effect)

In this case, I previously thought I was seeing a bug with the `useEffect` that wasn't present with the callback approach. But I can't repro or even remember it now. I think maybe it got resolved by memoizing the return value, as you said. 

So I'm okay either way -- if we prefer to keep our usual approach to maintain a consistent pattern I'll go with that.

src/modules/search/searchUtil.ts
```diff
@@ -1,46 +1,60 @@
...
-export const searchParamsToState = (
-  searchParams: URLSearchParams
-): ClientSearchInput => {
-  const variables: Record<string, any> = {};
-  const fieldNames: string[] = SearchFormDefinition.item.map(
-    (item) => item.mapping?.fieldName as string
-  );
+/**
+ * Maps a loaded `SearchQuery` into `ClientSearchInput`.
+ * Returns `null` when there is no record or no mapped search criteria
```
gigxz: Seems to return `null` when there is no data, but `{}` when there are no mapped search criteria

src/modules/search/hooks/useSearchQuery.ts
```diff
@@ -0,0 +1,37 @@
...
+  const { data, loading, error } = useGetSearchQueryQuery({
+    variables: { id: searchQueryId || '' },
+    skip,
+    // cache-first is Apollo's default; Setting it explicitly to call out that having this uuid in the cache already is likely
+    // if the user is navigating with the back-button, so we don't want to hit the network again in that case
+    fetchPolicy: 'cache-first',
+  });
+
+  if (error) throw error;
+
```
gigxz: Maybe the write-to-cache logic could be encapsulated in this hook too, since they will likely often be used together? Like:

```diff
--- a/src/modules/search/hooks/useSearchQuery.ts
+++ b/src/modules/search/hooks/useSearchQuery.ts
@@ -1,4 +1,7 @@
+import { useCallback } from 'react';
+import apolloClient from '@/providers/apolloClient';
 import {
+  GetSearchQueryDocument,
   SearchQueryFieldsFragment,
   useGetSearchQueryQuery,
 } from '@/types/gqlTypes';
@@ -6,10 +9,12 @@ import {
 interface Props {
   searchQueryId?: string | null;
 }
+type SearchQueryFields = Omit<SearchQueryFieldsFragment, 'id' | '__typeName'>;

 type Result = {
   searchQuery: SearchQueryFieldsFragment | null;
   loading: boolean;
+  writeSearchQueryToCache: (id: string, fields: SearchQueryFields) => void;
 };

 /**
@@ -28,9 +33,23 @@ const useSearchQuery = ({ searchQueryId }: Props): Result => {

   if (error) throw error;

+  // Helper to write a search query to the cache. This can be used to
+  // populate the cache with a search query when it is received from the server with a searchQueryId.
+  const writeSearchQueryToCache = useCallback(
+    (id: string, params: SearchQueryFields) => {
+      apolloClient.writeQuery({
+        query: GetSearchQueryDocument,
+        variables: { id },
+        data: { searchQuery: { __typename: 'SearchQuery', id, ...params } },
+      });
+    },
+    []
+  );
+
   return {
     searchQuery: data?.searchQuery || null,
     loading: skip ? false : loading,
+    writeSearchQueryToCache,
   };
 };
```

src/modules/search/hooks/useSearchQuery.ts
```diff
@@ -0,0 +1,37 @@
...
+  loading: boolean;
+};
+
+/**
+ * Loads `searchQuery` by id
+ */
+const useSearchQuery = ({ searchQueryId }: Props): Result => {
+  const skip = !searchQueryId;
+
+  const { data, loading, error } = useGetSearchQueryQuery({
```
gigxz: The naming here feels really generic, I wonder if we can come up with something more descriptive @martha. It's not bad but I can definitely see it being confusing in the future to be like, "oh I want to do a search, here's the search query hook".

* "persisted search query" (`usePersistedSearchQuery`)
* "persisted [client] search params" (`usePersistedSearchParamsQuery`) (=> generates `useGetPersistedSearchParamsQuery`, no QueryQuery)

What do you think of those?
(This could apply to the graphql query too. If it feels like too much churn to change there at this point, that's OK)

src/modules/search/components/OmniSearch.tsx
```diff
@@ -137,9 +143,10 @@ const OmniSearch: React.FC = () => {
         });
       }
       if (option.__typename === 'SeeMore') {
-        const search = debouncedSearch
-          ? new URLSearchParams({ textSearch: debouncedSearch })
-          : undefined;
+        const search =
+          debouncedSearch && typeof option.searchQueryId === 'string'
+            ? new URLSearchParams({ searchQueryId: option.searchQueryId })
+            : undefined;
```
gigxz, line 149: nice!

src/modules/search/searchUtil.ts
```diff
@@ -1,46 +1,60 @@
...
+  data: SearchQueryFieldsFragment | null
+): ClientSearchInput | null {
+  if (!data) return null;
+  const result: ClientSearchInput = {};
+  for (const key of SEARCH_QUERY_FIELD_NAMES) {
+    const v = data[key];
+    // Omit empty/unset fields.
+    // This prevents the constructed search input object from appearing
+    // to have changed compared to the object ClientSearch created from
+    // user input, which is desirable to avoid double-querying.
```
gigxz: Was this an issue with the fingerprinting, if so I wonder if the backend could drop unset attributes during normalization?

gigxz (commented):
Looks great, nice work! 👏 
I tested and everything worked as expected. Just a few minor comments about code organization.

src/modules/search/searchUtil.ts
martha: It's not an issue with fingerprinting, but with GraphQL caching and prop equality, if I'm describing that correctly. Here's my understanding. I'll try to update comments to get this more clear.

1. User searches. `searchInput` gets set to `{textSearch: "example"}`
2. `handleSearchCompleted` does 2 things:
    - pre-warms the cache with a SearchQuery object. This object must have all fields (lastName, dob, etc.) defined, or else at step 3 GraphQL thinks it needs to make a network request.
    - sets `searchQueryId` equal to the value returned from the query.
3. the `useSearchQuery` hook runs, because `searchQueryId` changed (from null to non-null). It hits the cache and returns the pre-warmed query inserted at step 2.
4. `resolvedSearchInput` (memoized) is re-computed, because `searchQuery` changed (from null to non-null). `resolvedSearchInput` is now set to `{textSearch: "example"}`
5. `setSearchInput` runs in a `useEffect`, because `resolvedSearchInput` has changed (from null to non-null). `searchInput` was already set to `{textSearch: "example"}` before (at step 1). now it's being re-set to `{textSearch: "example"}`, and since that's unchanged, `GenericTableWithData` does *not* re-query.

If we change line 37 of `searchQueryToClientSearchInput` to `result[key] = v` (instead of `if (v) result[key] = v`), then:
- at step 4, `resolvedSearchInput` would be set to `{textSearch: "example", firstName: null, dob: null, ...}`     
- at step 5, `searchInput` would now appear to be a changed prop and `GenericTableWithData` would perform a second network query.

martha: Comments updated here, feedback welcome https://github.com/greenriver/hmis-frontend/pull/1405/commits/dc863a2e0d84b03f58216bddb697cc91b395c784

src/modules/search/searchUtil.ts
martha: Hm, true. Just thinking it through... I think "when data is returned but there are no mapped search criteria" is kinda undefined territory right now. That would mean the backend returned a resolved `SearchQuery` object, but none of its params map to `ClientSearchInput` fields. I'll just update the comment for clarity so it's accurate.

gigxz (approved):
💯

src/modules/search/searchUtil.ts
gigxz: Thank you for the explanation and the updated comment! That makes sense and I think will be helpful to have in the future next time we touch this code.

PR #1406, opened 2026-03-16
Persistent table filters in URL

gigxz:
## Description

Issue: https://github.com/open-path/Green-River/issues/149

Table filter state can be synced to the URL so that filter selections persist across forward/back navigation and can be shared or bookmarked. Filter and sort utilities are refactored into shared hooks/utils/type files (per guidelines for the repo), and a single `useTableFilters` hook.

### Backwards compatibility
Old `useFilters` hook is exported for backwards compatibility. I did change the location of it, so there is a diff for updating those imports.

### What changed
- **useTableFilters** — New hook similar to old `useFilters`. It builds filter config and optionally syncs filter values to the URL. Tables choose `syncToUrl: true` (persist in URL) or `syncToUrl: false` (local only).
- **useSearchParamsState** — Setter now supports an updater function (like `setState`) and clearing all params with `{}`. This could be pulled into its own PR if desired. https://github.com/greenriver/hmis-frontend/pull/1409
- **Filter/sort refactor** — Logic moved out of `filterUtil` and data-fetching types into `tableFilterUtil`, `tableFilterTypes`, and `tableSortUtil`. 
- **Tables** — Some tables now persist filters in the URL. To reduce the size of the diff, I didn't update all tables. These tables are URL-synced now: **ProjectConfigTable, ServiceTypeTable, ClientAccessSummaryTable, EnrollmentAccessSummaryTable, UserAuditHistory, AllProjectsPage, ProjectHouseholdsTable, ProjectClientEnrollmentsTable, AdminOpportunitiesTable, AdminReferralsTable, AdminCeClientsTable (contains dynamic filters).**

### QA Instructions
**How to QA (URL-synced tables)**

- Set a filter; confirm URL updates and results match.
- Set additional filters; confirm all are reflected in the URL and results.
- Change optional columns (show/hide); confirm column choices persist.
- Hit Reset; confirm filter settings clear but optional column settings do not.
- Navigate forward (e.g. into a row), then back; confirm filters and columns are restored.
- Reload the page; confirm filters and columns are restored from the URL.
- Generally try to break it or produce unexpected behavior

**Other QA**
- Bed night workflow should be QAd for regressions


gigxz:
Marked as draft because this is not a priority for review

src/hooks/useTableFilters.ts
```diff
@@ -0,0 +1,115 @@
...
+  buildTableFilterType,
+  buildUrlParamsDefinition,
+} from '@/utils/tableFilterUtil';
+
+export interface FilterParams<T = Record<string, unknown>> {
+  type?: string | null; // filter input type for inferring filters if not provided
+  pickListArgs?: PickListArgs; // optional: pick list args to be applied to all PickList filter items
+  omit?: Array<string>; // optional: skip some filters
+  dynamicFilters?: TableFilterConfigFieldsFragment[]; // optional: dynamic filters to include
+  /** Whether to sync filter values to URL. If true, filterValues will be set from URL. */
```
martha:
```suggestion
  /** Whether to sync filter values to URL. If true, filterValues will be set as URL search params. */
```
?

src/hooks/useTableFilters.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  // Build URL param definition from filter configuration; if syncing filter state to URL.
+  const urlParamsDefinition: SearchParamsStateType = useMemo(() => {
+    if (!syncToUrl || !filters || Object.keys(filters).length === 0) {
+      return {};
+    }
+    return buildUrlParamsDefinition(filters, omitFromUrl);
+  }, [syncToUrl, filters, omitFromUrl]);
+
+  // state/setState for filter values that are synced to URL.
```
martha:
```suggestion
  // state/setState for filter values that are synced to URL.
  // Has no effect if urlParamsDefinition is empty (when syncToUrl is false)
```

src/hooks/useTableFilters.ts
```diff
@@ -0,0 +1,115 @@
...
+  omitFromUrl?: Array<string>;
+  /** Initial filter values (e.g. defaults when URL has no params). Typed by the same T as useTableFilters<T>. */
+  initialFilterValues?: Partial<T>;
+}
+
+/** useTableFilters export is intentionally overloaded so that when syncToUrl is false,
+ * the return type is only { filters }; when true it includes filterValues and setFilterValues.
+ * Callers then get accurate types without optional/undefined.
+ *
+ * If using useTableFilters with 'syncToUrl: false', filter state is managed by GenericTableWithData.
```
martha:
```suggestion
 * If using useTableFilters with 'syncToUrl: false', filter state is managed by GenericTableWithData.
 * If using 'syncToUrl: true', the calling component owns filter state, 
 * and typically passes `filterValues` and `setFilterValues` directly to GenericTableWithData
 * so they can be read and set by the filter widget internal to GenericTableWithData.
```
Maybe add something like this? (At first when I read this comment I was confused because the filter widget itself, which modifies the filter state, is internal to GenericTableWithData in both cases. But filter state is "owned" by the parent component only when `syncToUrl` is true)

src/modules/admin/components/projectConfig/ProjectConfigTable.tsx
```diff
@@ -86,9 +87,10 @@ const ProjectConfigTable = ({
     []
   );
 
-  const filters = useFilters({
-    type: 'ProjectConfigFilterOptions',
-  });
+  const { filters, filterValues, setFilterValues } =
+    useTableFilters<ProjectConfigFilterOptions>({
+      type: 'ProjectConfigFilterOptions',
+    });
```
martha, line 93: One thing about the new approach is that the call signature is always repetitive, passing both `ProjectConfigFilterOptions` to the hook's generic type and `'ProjectConfigFilterOptions'` as the type prop.

I think that is fine, and justifiable, but was kinda curious if ai had any suggestions for improving to reduce repetition. Below is what Composer came up with. I don't think this is necessarily preferable, since the `FilterTypesMap` idea also feels ugly to me.. but it does constrain the ugliness to 1 place instead of each time we call the hook.

## **Central map + inferred `type`**

Define a map from the GraphQL input name (string literal) to the TS type, then overload `useTableFilters` so `K extends keyof FilterTypesMap` implies `T = FilterTypesMap[K]`:

```ts
type FilterTypesMap = {
  EnrollmentAccessSummaryFilterOptions: EnrollmentAccessSummaryFilterOptions;
  ClientFilterOptions: ClientFilterOptions;
  // …
};

// Signature idea:
declare function useTableFilters<K extends keyof FilterTypesMap>(
  params: FilterParams<FilterTypesMap[K]> & { type: K; syncToUrl?: true }
): { filters: TableFilterType<FilterTypesMap[K]>; filterValues: Partial<FilterTypesMap[K]>; ... };
```

Call site:

```ts
const { filters, filterValues, setFilterValues } = useTableFilters({
  type: 'EnrollmentAccessSummaryFilterOptions',
});
```

No generic; the string literal narrows `K` and picks the right `T`.

**Tradeoff:** You maintain `FilterTypesMap` when new filter inputs appear. That can be hand-written, or generated (e.g. from the same GraphQL/codegen pipeline that already produces the interfaces).

src/hooks/useTableFilters.ts
```diff
@@ -0,0 +1,115 @@
...
+
+export interface FilterParams<T = Record<string, unknown>> {
+  type?: string | null; // filter input type for inferring filters if not provided
+  pickListArgs?: PickListArgs; // optional: pick list args to be applied to all PickList filter items
+  omit?: Array<string>; // optional: skip some filters
+  dynamicFilters?: TableFilterConfigFieldsFragment[]; // optional: dynamic filters to include
+  /** Whether to sync filter values to URL. If true, filterValues will be set from URL. */
+  syncToUrl?: boolean;
+  /** Keys to exclude from URL (e.g. PII). type 'text' is always excluded. */
+  omitFromUrl?: Array<string>;
```
martha: Something about this comment that took me awhile to understand: the excluding of free-text filters is not actually happening at the "omit from URL" step (even though omitting them from the URL is the reason why we do it - ?). It's happening earlier, when creating the filter definitions, at the `getFilterForType` step.

So in other words, `text` type filters aren't just unsupported in the URL, they are unsupported in the table filters widget in general. If I got that right, maybe we can update the comments around here, something like
```javascript
/* Graphql FilterOptions input type for inferring filters, such as 'ClientFilterOptions'. Text filters are unsupported and will be skipped.  */
type?: string | null;
...
/* Keys to exclude from URL (e.g. PII). Text filters are already excluded, see above */
omitFromUrl?: Array<string>;
```

src/utils/tableFilterUtil.ts
```diff
@@ -0,0 +1,206 @@
...
+  if (inputType in HmisEnums) {
+    filter = {
+      ...baseFields,
+      enumType: inputType as keyof typeof HmisEnums,
+      type: 'enum',
+    };
+  }
+
+  if (!filter && inputType === 'String') {
+    // console.log('Skipping free-text filter:', fieldName);
```
martha:
```suggestion
```

src/hooks/useTableFilters.ts
```diff
@@ -0,0 +1,115 @@
...
+): {
+  filters: TableFilterType<T>;
+  filterValues: Partial<T>;
+  setFilterValues: (values: Partial<T>) => void;
+};
+export default function useTableFilters<T = Record<string, unknown>>(
+  params: FilterParams<T> & { syncToUrl: false }
+): { filters: TableFilterType<T> };
+
+export default function useTableFilters<T = Record<string, unknown>>({
```
martha, line 58: I'm remembering that when we similarly implemented URL search params for optional columns, we also decided to store the user's recent selections in local storage in the browser. ([relevant code](https://github.com/greenriver/hmis-frontend/blob/ba02f81e70d786b94108156c35d52fb8c852a240/src/modules/dataFetching/hooks/useOptionalColumns.tsx#L48-L53))

I think the use case is different and we shouldn't bring in that complexity here, but maybe it would be good to document that decision somewhere.

martha (approved):
🙌 beautiful, looks great. I left some comments mostly related to clarifying code comments and documentation.

If you haven’t already, it would probably be good to split out a separate ticket for upgrading the remaining tables. Also because it would be a good place to add a note to help us remember that `searchTerm` can now be removed from the `omit` list when we upgrade future tables.

src/hooks/useTableFilters.ts
gigxz: These are actually not returned at all if syncToUrl is false. I can add another comment about that here

src/hooks/useTableFilters.ts
gigxz: Added, thanks!

src/hooks/useTableFilters.ts
gigxz: I got rid of this `omitFromUrl` entirely, it wasn't needed. The `omit` prop was sufficient. I updated the comment:
```
  /** (Optional) Keys to skip when inferring filters. Filters that are text-based are always skipped
   * because they are not supported in the filter widget (due to limiting PII in URL). */
  omit?: Array<string>;
  ```

File: src/hooks/useTableFilters.ts (hunk 7 above)
Comment (gigxz) [reply src/hooks/useTableFilters.ts:58]: Added a comment to the hook
```
 * Design notes:
 * -  The useTableFilters export is intentionally overloaded so that when syncToUrl is false,
 * the return type is only { filters }; when true it includes filterValues and setFilterValues.
 * Callers then get accurate types without optional/undefined.
 * - Unlike Optional Table Columns (which similarly have logic to sync to URL),
 * filter values do not persist to localStorage. This was an intentional design choice.
 ```

martha (approved):

src/modules/admin/components/projectConfig/ProjectConfigTable.tsx
gigxz: Thanks @martha for this. The typesafety could definitely be improved, I think generating them instead of manually maintaining a map of all Filter types would probably be easier. I will go ahead and merge this but play around with those options (briefly) on another branch.

gigxz: Opened https://github.com/greenriver/hmis-frontend/pull/1422

PR #1407, opened 2026-03-17
Move broad and specific client search to separate routes

gigxz:
## Description

This change moves the client search type (broad vs. specific) from React state inside ClientSearch into the router by giving each mode its own route (`/` for broad, `/advanced` for specific).

This matches the pattern we already use for the toggle between "Clients" and "Households" on the Project Enrollments page.

This change is intended to reduce local state and branching in ClientSearch, to reduce state management changes in https://github.com/greenriver/hmis-frontend/pull/1405. Each route renders a separate instance of the search UI so the two modes don’t share in-memory state and there’s no “bleed” (e.g. advanced form values or results appearing in broad mode or vice versa).

In addition, the URL now encodes the selected search mode, so users can bookmark or share a link that opens client search in the correct mode (broad or specific/advanced).

## Type of change
Refactor

### DIFF
--- a/src/modules/search/components/ClientSearchTypeToggle.tsx
+++ b/src/modules/search/components/ClientSearchTypeToggle.tsx
@@ -1,50 +1,70 @@
 import LocationSearchingIcon from '@mui/icons-material/LocationSearching';
 import SearchIcon from '@mui/icons-material/Search';
 
-import CommonToggle, {
-  CommonToggleProps,
-} from '@/components/elements/CommonToggle';
+import { useCallback } from 'react';
+import { useNavigate } from 'react-router-dom';
+import CommonToggle from '@/components/elements/CommonToggle';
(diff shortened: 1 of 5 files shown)


src/modules/search/components/ClientSearchTypeToggle.tsx
```diff
@@ -1,50 +1,70 @@
...
 
+/**
+ * Toggle between broad (text) and specific (advanced) client search modes.
+ * Uses React Router navigation to switch between "broad" and "specific" client search routes.
+ */
 const ClientSearchTypeToggle: React.FC<ClientSearchTypeToggleProps> = ({
   value,
-  onChange,
-}) => (
-  <LabelWithContent
```
gigxz, line 20: Most of the diff here is whitespace, I didn't change anything below `<LabelWithContent ...` except `onChange={handleChange}`

martha (approved):
Awesome, thank you!

PR #1409, opened 2026-03-18
Update behavior of useSearchParamState setter to match SetState

gigxz:
## Description

Update useSearchParamState `setFilterParams` behavior to match behavior of React `SetStateAction`. This is more useful and predictable. Previously, it wasn't possible to clear out the state by passing `{}` for example.

Pulled in for use in https://github.com/greenriver/hmis-frontend/pull/1406 which uses this hook to store Filter state in search params.

**How to test**: test bed night workflow and other workflows that use useSearchParamState. No changes expected.

## Type of change
New feature

### DIFF
--- a/src/hooks/useSearchParamState.ts
+++ b/src/hooks/useSearchParamState.ts
@@ -164,7 +164,6 @@ const useSearchParamsState = ({
     const currentParams = getAllCurrentParams(searchParams);
     populateParams(currentParams, accumulator);
 
-    // `navigate` instead of `useSearchParams` as workaround for https://github.com/remix-run/react-router/issues/8393
     navigate(
       {
         pathname,
@@ -233,6 +232,7 @@ const useSearchParamsState = ({
 
       const accumulator = new URLSearchParams();
       populateParams(currentParams, accumulator);
+      // `navigate` instead of `useSearchParams` as workaround for https://github.com/remix-run/react-router/issues/8393
       navigate({
         pathname,
         hash,
--- a/src/hooks/useSearchParamState.ts
+++ b/src/hooks/useSearchParamState.ts
@@ -164,6 +164,7 @@ const useSearchParamsState = ({
     const currentParams = getAllCurrentParams(searchParams);
     populateParams(currentParams, accumulator);
 
+    // `navigate` instead of `useSearchParams` as workaround for https://github.com/remix-run/react-router/issues/8393
     navigate(
       {
         pathname,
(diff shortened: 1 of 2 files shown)


src/hooks/useSearchParamState.ts
```diff
@@ -164,6 +164,7 @@ const useSearchParamsState = ({
     const currentParams = getAllCurrentParams(searchParams);
     populateParams(currentParams, accumulator);
 
+    // `navigate` instead of `useSearchParams` as workaround for https://github.com/remix-run/react-router/issues/8393
```
martha: I _think_ this comment belongs with the `navigate` call on line 236. (?) This navigate call is just setting the initial search params on page load, but the linked bug is about changing search params removing the url hash.

src/hooks/useSearchParamState.ts
```diff
@@ -186,31 +187,52 @@ const useSearchParamsState = ({
     return getValues(paramsDefinition, searchParams);
   }, [initial, mounted, paramsDefinition, searchParams]);
 
+  /**
+   * Updates URL search params (same keys as paramsDefinition). Behaves like React setState:
+   * - Pass an object to set/replace: keys present are updated, keys in paramsDefinition but
+   * not in the object are cleared. setValues({}) clears all defined params.
+   * - Pass an updater function to merge with current values: setValues(prev => ({ ...prev, page: 2 })).
+   *
+   * Any keys not present in paramsDefinition are ignored.
```
martha, line 195: Thanks for adding this comment! Maybe to be even more explicit, we could add something like,
```suggestion
   * Any keys not present in paramsDefinition are ignored, enabling multiple usages of `useSearchParamState` within the same page
```

src/hooks/useSearchParamState.ts
```diff
@@ -186,31 +187,52 @@ const useSearchParamsState = ({
...
+      for (const key of Object.keys(paramsDefinition)) {
+        const hasKey = Object.prototype.hasOwnProperty.call(nextValues, key);
+        let value = hasKey ? nextValues[key] : undefined;
+
+        if (hasKey && paramsDefinition[key]?.type === 'date' && isDate(value)) {
+          // Try to serialize date. If it is invalid, it will be set to null.
+          value = formatDateForGql(value);
+        }
+
+        const isEmpty =
```
martha, line 221: This all looks great, I think your code with the more explicit variable names makes everything a lot clearer. Thanks for that!

martha (approved):
🎉

PR #1422, opened 2026-03-30
Make table filtering type-safe with generated types

gigxz:
## Description

Follow-up to https://github.com/greenriver/hmis-frontend/pull/1406 to make table-filtering typesafe.
Related issue: https://github.com/open-path/Green-River/issues/149

## What changed

- **Codegen / scripts**
  - Added `scripts/generateGqlFilterOptionsTypes.js`, run after GraphQL codegen
  - Emits `src/types/gqlFilterOptionsTypes.generated.ts`: imports each `*FilterOptions` type from `gqlTypes`, exports **`FilterOptionsByName`**, **`FilterOptionsGraphqlTypeName`** (`keyof` that map), and **`FILTER_OPTIONS_GRAPHQL_TYPE_NAMES`** (`as const` list).
- **`useTableFilters` / `useFilters`**
  - `Args` uses **`type: K`** where **`K extends keyof FilterOptionsByName`**, so invalid `type` strings are rejected.
  - **`omit`**, **`initialFilterValues`**, and return types use **`FilterOptionsByName[K]`** instead of a free generic or plain `string` for `type`.
  - Call sites **no longer pass** `useTableFilters<SomeFilterOptions>` / `useFilters<SomeFilterOptions>` (the old generic was the **model** type, not the GraphQL type name); behavior is unchanged, with **inference from `type: '…FilterOptions'`**.
  - Type-checking will fail if callers pass values to `omit` that are not part of the generated filters. There were a few places where we did this, which I removed.

## How to verify

- `yarn graphql:codegen` — regenerates `gqlFilterOptionsTypes.generated.ts` when the schema changes.
- `yarn tsc` — should pass; new filter inputs in the schema require a codegen run so the registry stays aligned.
- No expected change in behavior for any tables, including dynamic filters

## Type of change
Code cleanup

### DIFF
--- a/scripts/generateGqlFilterOptionsTypes.js
+++ b/scripts/generateGqlFilterOptionsTypes.js
@@ -1,0 +1,53 @@
+'use strict';
+
+/**
+ * Reads graphql.schema.json (from graphql-codegen introspection) and emits
+ * src/types/gqlFilterOptionsTypes.generated.ts with all input object types
+ * whose name ends with `FilterOptions`.
+ *
+ * This is used to make table filtering type-safe (see useTableFilters.ts).
+ */
+import fs from 'fs';
+import path from 'path';
+import { fileURLToPath } from 'url';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const repoRoot = path.join(__dirname, '..');
+
+const generatedFileHeader =
+  '// **** THIS FILE IS GENERATED, DO NOT EDIT DIRECTLY ****\n\n';
+
+const schemaPath = path.join(repoRoot, 'graphql.schema.json');
+const outPath = path.join(
+  repoRoot,
+  'src/types/gqlFilterOptionsTypes.generated.ts'
+);
+
+const raw = fs.readFileSync(schemaPath, 'utf8');
+const schema = JSON.parse(raw);
+
+const types = schema.__schema?.types ?? [];
+const names = types
+  .filter(
+    (t) =>
+      t.kind === 'INPUT_OBJECT' &&
+      typeof t.name === 'string' &&
+      t.name.endsWith('FilterOptions')
+  )
+  .map((t) => t.name)
+  .sort((a, b) => a.localeCompare(b));
+
+if (names.length === 0) {
+  console.error(
+    'generateGqlFilterOptionsTypes: no *FilterOptions INPUT_OBJECT types found'
+  );
+  process.exit(1);
+}
+
+const importLine = `import type {\n${names.map((n) => `  ${n}`).join(',\n')},\n} from './gqlTypes';\n\n`;
+
+const mapEntries = names.map((n) => `  ${n}: ${n};`).join('\n');
+
+const output = `${generatedFileHeader}${importLine}export type FilterOptionsByName = {\n${mapEntries}\n};\n\nexport type FilterOptionsGraphqlTypeName = keyof FilterOptionsByName;\n\nexport const FILTER_OPTIONS_GRAPHQL_TYPE_NAMES = [\n${names.map((n) => `  '${n}'`).join(',\n')},\n] as const;\n`;
+
+fs.writeFileSync(outPath, output, 'utf8');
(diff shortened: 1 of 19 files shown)


src/modules/ce/components/client/ClientReferralsTable.tsx
```diff
@@ -32,10 +32,7 @@ const ClientReferralsTable: React.FC = () => {
     clientId: string;
   };
 
-  const filters = useFilters({
-    type: 'ClientCeReferralFilterOptions',
-    omit: ['workflowTemplate'],
```
gigxz, line 37: Removed because this is not part of ClientCeReferralFilterOptions, which is now surfaced via type error

src/modules/search/components/ClientSearch.tsx
```diff
@@ -216,7 +216,6 @@ const ClientSearch: React.FC<ClientSearchProps> = ({ searchType }) => {
 
   const filters = useFilters({
     type: 'ClientFilterOptions',
-    omit: ['searchTerm'],
```
gigxz, line 278: Removed because this is not part of ClientFilterOptions, which is now surfaced via type error

scripts/generateGqlFilterOptionsTypes.js
```diff
@@ -0,0 +1,53 @@
...
+  process.exit(1);
+}
+
+const importLine = `import type {\n${names.map((n) => `  ${n}`).join(',\n')},\n} from './gqlTypes';\n\n`;
+
+const mapEntries = names.map((n) => `  ${n}: ${n};`).join('\n');
+
+const output = `${generatedFileHeader}${importLine}export type FilterOptionsByName = {\n${mapEntries}\n};\n\nexport type FilterOptionsGraphqlTypeName = keyof FilterOptionsByName;\n\nexport const FILTER_OPTIONS_GRAPHQL_TYPE_NAMES = [\n${names.map((n) => `  '${n}'`).join(',\n')},\n] as const;\n`;
+
+fs.writeFileSync(outPath, output, 'utf8');
```
martha, line 53: Is there a reason to use `writeFileSync` here instead of `writeFile`, which `generateEnumDescriptions.js` uses?

src/modules/projects/components/tables/ProjectClientEnrollmentsTable.tsx
```diff
@@ -53,15 +53,14 @@ const ProjectClientEnrollmentsTable = ({
...
-        staffAssignmentsEnabled ? '' : 'assignedStaff',
-      ],
-      pickListArgs: { projectId: projectId },
-    });
+  const { filters, filterValues, setFilterValues } = useTableFilters({
+    type: 'EnrollmentsForProjectFilterOptions',
+    omit: [
+      'bedNightOnDate', // Not applicable, this filter is available via bed night workflow
+      ...(staffAssignmentsEnabled ? [] : ['assignedStaff']),
+    ] as (keyof EnrollmentsForProjectFilterOptions)[],
```
martha: if we want, I think the type cast can be avoided like this:
```suggestion
    // omit has two branches, instead of spread+ternary, so TS does not widen omit keys to string[].
    // This allows type safety without casting
    omit: staffAssignmentsEnabled
      // bedNightOnDate: not applicable, bed night filter is available via bed night workflow
      ? ['bedNightOnDate']
      // assignedStaff: only show this filter if staff assignments are enabled for this project
      : ['bedNightOnDate', 'assignedStaff'],
```

Otherwise I always prefer to add a comment justifying the cast/why we are confident it will be cast-able at runtime. I would apply the same comment to the other typescript casts in this PR, but don't have neat suggestions for those - it's up to you how worthwhile you think it is to try to avoid them.

martha (approved):
Neat 😎 !

src/modules/projects/components/tables/ProjectClientEnrollmentsTable.tsx
gigxz: Ah that's a neat approach, thanks! Added https://github.com/greenriver/hmis-frontend/pull/1422/commits/7f9e0dfcbb986d0b580949be04072ff9ffb00377

scripts/generateGqlFilterOptionsTypes.js
gigxz: No reason, I don't think it really matters since this script is executed using `yarn graphql:codegen`

PR #1434, opened 2026-04-09
Patch pagination bug with persistent table filters

gigxz:
## Description

Fixes an issue where tables that sync filters to the URL could not move to the next page: the footer looked like it updated, but the list stayed on the first page and flickered.

The cause was twofold: optional hook arguments were using fresh empty objects and arrays on every render, which made URL-backed filter state look “new” every time and broke memoization. Separately, the table reset the current page whenever the filter prop’s identity changed, even when the actual filters were unchanged.

The fix stabilizes those default empty values so they stay the same across renders, and only resets pagination when the meaning of the filters changes, not when the same values are carried in a new object. Together, that restores reliable pagination for URL-backed filters without changing how filters behave for users.

### How to test

Exercise a table that uses URL-backed filters and run through navigation **in mixed orders** (don’t follow a single linear script). Goal: confirm nothing gets stuck, flashes wrongly, or shows the wrong slice of data.

Try combinations such as:

- Open the page, paginate forward and back, then jump with the page control if available.
- Change **rows per page**, then paginate again.
- Apply one or more **filters**, confirm results, then change **page** and **page size**.
- Change **sort** (if available), paginate, then change sort again.
- **Clear / remove filters** partially and fully; paginate after each change.
- Apply filters, go to a later page, then **change a filter** (or apply a new one)—confirm you land on **page 1** with updated results.
- Use **browser back / forward** after changing filters or page. Currently only _filters_ are expected to be maintained (not sort or page number)

**Expected Behavior**

- Pagination, page size, sort, and filters all stay consistent with what’s on screen and in the URL.
- **Whenever filters actually change** (including clearing them in a way that changes the query), the table should **reset to the first page** for the new result set.

## Type of change
Bug fix

### DIFF
--- a/src/hooks/useTableFilters.ts
+++ b/src/hooks/useTableFilters.ts
@@ -27,6 +27,14 @@ type Args<K extends keyof FilterOptionsByName> = {
   initialFilterValues?: Partial<FilterOptionsByName[K]>;
 };
 
+/** Stable references for defaults (to avoid new instances each render, impacting useMemo/useEffect deps) */
+const EMPTY_INITIAL_FILTER_VALUES: Partial<
+  FilterOptionsByName[keyof FilterOptionsByName]
+> = Object.freeze({});
+const EMPTY_PICK_LIST_ARGS: PickListArgs = Object.freeze({});
+const EMPTY_OMIT: Array<keyof FilterOptionsByName[keyof FilterOptionsByName]> =
+  [];
+
 /**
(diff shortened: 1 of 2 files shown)


martha (approved):
Thanks, looks good to me! 

I played around per the QA instructions and didn't notice anything unusual. It looks like changing the sort when on page 2+ of results doesn't reset the results to page 1, which is unintuitive, but that looks expected at this time. Sounds like we will fix it when we make the update to put pagination in the URL.

PR #1435, opened 2026-04-10
Update Bulk Services to use client search params

martha:
## Description

Summary of changes: Update the Bulk Services UI to use the new database-backed client search params, instead of putting client search information in the URL.

## Type of change
New feature
PII handling improvement

### DIFF
--- a/src/api/operations/bulkService.queries.graphql
+++ b/src/api/operations/bulkService.queries.graphql
@@ -17,6 +17,7 @@ query BulkServicesClientSearch(
   ) {
     offset
     limit
+    searchQueryId
     nodesCount
     nodes {
       id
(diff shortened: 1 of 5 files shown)


gigxz (commented):
Code looks good, and the basic path and add-new-client path work for me.

There is one regression though, the existing workflow injects the household id for search when you are adding a new household as part of the bed night flow.
On QA:
- bed night flow
- click add new client > add new household
- add household member(s)
- click "back to bed nights/bulk services" button
- the household id should auto-populate the search bar so the newly added household members appear in the search results for recording bed nights

The relevant code is at: https://github.com/greenriver/hmis-frontend/blob/0f3a1df9c985d468f1e49965fce31e192a5b93fe/src/modules/household/components/CreateHouseholdPage.tsx#L79-L83

martha:
Thanks @gigxz good catch. Since we no longer support `searchTerm` in the url directly, but `householdId` is guaranteed to not contain pii, I added the approach of using a "one-shot" url param for `householdId` that's consumed and cleared after the user navigates back. Let me know what you think of this approach.

gigxz (approved):

PR #1437, opened 2026-04-13
Remove unnecessary cache clearing

martha:
## Description

Summary of changes: Followup to https://github.com/greenriver/hmis-warehouse/pull/6377#discussion_r3072693466

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/6377

How to test: Edit a project config. After saving, the project configs table should get updated right away, without another network request to reload the whole table.

## Type of change
Code clean-up

### DIFF
--- a/src/modules/admin/components/projectConfig/ProjectConfigDialog.tsx
+++ b/src/modules/admin/components/projectConfig/ProjectConfigDialog.tsx
@@ -55,7 +55,6 @@ const ProjectConfigDialog: React.FC<ProjectDialogProps> = ({
               })}
               getErrors={(data) => data.updateProjectConfig?.errors || []}
               onCompleted={(data) => {
-                evictProjectConfigs();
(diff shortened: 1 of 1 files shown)


gigxz (approved):
is there a reason its removed on line 58 but not 82?

martha:
> is there a reason its removed on line 58 but not 82?

Yes, the evict call on line 82 is repopulating the results for the root field `projectConfigs` after a new project config has been created. Since that list result is sorted and paginated, the frontend doesn't strictly "know" where in the list the newly created config will go, or even that it will be on the currently displayed page of results, unless it re-queries the backend. The evict pattern is normal for our codebase, I think, although sometimes we do something more elaborate like manually inserting the newly created record in the cache.

Whereas the evict call removed from line 58 is now strictly unnecessary, because after an update mutation, apollo updates the cached record's fields, and it's then reflected in the cached `projectConfigs` list field's existing reference to that record.

gigxz (approved):

PR #1443, opened 2026-04-14
Bump alpine image to 3.23

gigxz:
## Description
Issue: https://github.com/open-path/Green-River/issues/8987

We were at 3.19 which is EOL. Bump to 3.23 and pin the version.

* NODE_VERSION is currently on 24.14.1 (set in `.nvmrc). Verified this tag exists on docker hub
* NGINX version had to be bumped from 1.25 to 1.28 because `1.25-alpine3.23` is not published

Testing: I built the image locally, haven't done testing beyond that

## Type of change
Dependency update

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,7 +2,7 @@
 # build container
 #####################################################################
 ARG NODE_VERSION
-FROM node:${NODE_VERSION}-alpine AS builder
+FROM node:${NODE_VERSION}-alpine3.23 AS builder
 
 ARG GITHASH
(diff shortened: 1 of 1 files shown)


dtgreiner:
I think this looks good, and it may take a staging deploy to fully verify. I may take a stab at updating to debian instead of alpine to bring it in line with the rest of the OP images.

dtgreiner (approved):

gigxz:
Deployed and tested on QA, I'm going to go ahead and merge into 210

PR #1447, opened 2026-04-16
Prevent console warnings when changing referral contacts

gigxz:
## Description

Prevent these console warnings from appearing when changing/removing contacts on a referral:

> Cache data may be lost when replacing the participants field of a CeReferralSwimlane object.
This could cause additional (usually avoidable) network requests to fetch data that were otherwise cached.
To address this problem (which is not a bug in Apollo Client), define a custom merge function for the CeReferralSwimlane.participants field, so InMemoryCache can safely merge these objects:

> Cache data may be lost when replacing the assignees field of a CeReferralStep object.
This could cause additional (usually avoidable) network requests to fetch data that were otherwise cached.
To address this problem (which is not a bug in Apollo Client), define a custom merge function for the CeReferralStep.assignees field, so InMemoryCache can safely merge these objects:

https://www.apollographql.com/docs/react/caching/cache-field-behavior#opting-in-to-the-default-behavior-for-non-normalized-fields contains an explanation of the problem and fix

## Type of change
Code clean-up

### DIFF
--- a/src/providers/apolloClient.tsx
+++ b/src/providers/apolloClient.tsx
@@ -113,6 +113,16 @@ export const cache = new InMemoryCache({
     // CeReferralSwimlane uses 'cacheKey' because 'id' was non-unique and changing it would be backwards-incompatible, so new cacheKey was added.
     CeReferralSwimlane: {
       keyFields: ['cacheKey'],
+      fields: {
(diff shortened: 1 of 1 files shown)


martha (approved):
thanks!

PR #1456, opened 2026-05-01
Refactor Referrals out of Admin

martha:
## Description

Summary of changes:
- Add a new top-level nav link for Referrals. Resolve the new `canIndexReferrals` root permission for determining visibility.
- Refactor `AdminReferralsPage` to simply `ReferralsPage`. Move the content that was previously on the Referrals, Available Units, and Eligible Clients admin pages so that they are now rendered as 3 tabs on the Referrals page.
- Set up redirects so the old urls are not broken.

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/6428

## Type of change
New feature


src/components/layout/nav/ToolbarMenu.tsx
```diff
@@ -76,6 +81,13 @@ const ToolbarMenu: React.FC<ToolbarMenuProps> = ({ mobile }) => {
         activeItemPathIncludes: 'project',
         title: 'Projects',
       },
+      {
+        path: Routes.REFERRALS,
+        id: 'navToReferrals',
+        activeItemPathIncludes: 'referrals',
```
gigxz, line 87: This doesn't seem to be working, I'm seeing that the "Referrals" nav item does not appear highlighted even when I am on page the referrals page (nor when I click into a referral)

martha: Follow-up question, should the route indeed be `/ce/referrals`? (I think I saw that in a figma comment from you somewhere, but can't find it at the moment.) Maybe we should update these to be just `/referrals` instead? I think this somewhat relates to your question about whether to keep "Coordinated Entry" in the overline text, if CE-language might be confusing/irrelevant to some users.

gigxz: Yeah I think we can drop the `/ce` prefixes, that also lets us get rid of that initial rerender when its routing from referrals=>ce/referrals

martha:
I renamed some graphql types for clarity now that this Referrals table is no longer admin-only:
- `CeReferralTableFields` is used on the Project referrals table, but it’s also used as a “base” type that's spread on all the other referral table types, and it’s used for typescript in the referral column definitions. I’m proposing to rename it to `CeReferralBaseTableFieldsFragment`
- That frees us up to rename `CeReferralAdminFields` to `CeReferralTableFields`
- Just for parity/consistency, I'm proposing renaming `ClientCeReferralTableFields` to `CeReferralClientTableFields`
- Lastly, the query `GetAdminCeReferrals` can be renamed to `GetCeReferrals`

martha:
I'm working on breadcrumbs now, but plan to open that work in a separate PR to make review more manageable. They should go out together in the same release. Edit: Opened https://github.com/greenriver/hmis-frontend/pull/1458, keeping as draft pending design feedback

src/routes/protected.tsx
```diff
@@ -176,15 +177,46 @@ export const protectedRoutes: RouteNode[] = [
...
+            permissions='canAdministrateCoordinatedEntry'
+            otherwise={<NotFound />}
+          >
+            <ReferralsPage currentTab='eligible-clients' />
+          </RootPermissionsFilter>
         ),
+      },
+      {
+        path: ReferralRoutes.REFERRAL,
+        // todo @martha - is this going to have /referrals/referral/...?
```
gigxz: todo can be removed? seems like its just /referrals

src/modules/ce/components/admin/AdminCeClientsTable.tsx
```diff
@@ -135,7 +135,7 @@ const AdminCeClientsTable: React.FC = () => {
         value={search}
         onChange={setSearch}
         fullWidth
-        size='small'
+        size='medium'
```
gigxz: unsure of this, it was mentioned by design but is not part of workspace designs https://greenriver.slack.com/archives/C033YNSN0Q1/p1778185621194159?thread_ts=1778097842.089159&cid=C033YNSN0Q1
(personally I think small looks better..)

src/modules/ce/components/ReferralsTable.tsx
```diff
@@ -146,29 +146,28 @@ const AdminReferralsTable: React.FC<Props> = ({}) => {
         value={search}
         onChange={setSearch}
         fullWidth
-        size='small'
+        size='medium'
```
gigxz: *same here
if anything, only make it big on provider view>?

gigxz (approved):
Looks good to me, I think we should address a few outstanding design things before merging

src/routes/protected.tsx
martha: Thanks! Ha, I thought I removed this, but I did recently give cursor an instruction to stop removing my todo comments unless I say so, so maybe the bot added it back >:|

PR #1464, opened 2026-05-08
Add Workspace switcher to Referrals page

gigxz:
## Description

Issue: https://github.com/open-path/Green-River/issues/9204
Depends on Backend PR: https://github.com/greenriver/hmis-warehouse/pull/6455
Opened against https://github.com/greenriver/hmis-frontend/pull/1456 because it depends on referral refactor

* If workspaces are configured for referrals, render a "workspace switcher" that supports filtering all tabs by a project group.
* If no workspaces are configured, there is no change to the UI
* Components `WorkspaceSelect` and `useWorkspaceSelector` are implemented generically (not CE-specific) so they can be re-used if we introduce workspace switchers elsewhere, such as the User Dashboard.

See issue for how to test

Screenshot 2026-05-08 at 3 24 59 PM

## Type of change
New feature


src/modules/ce/components/ReferralsPage.tsx
```diff
@@ -26,56 +46,73 @@ const ReferralsPage: React.FC<Props> = ({ currentTab }) => {
...
-        currentTab={currentTab}
-        onChangeTab={handleChangeTab}
-      />
+    <PageContainer
+      title='Referrals'
+      actions={hasWorkspaces ? workspaceSelector : undefined}
+    >
+      {workspaceLoading ? (
+        <Loading />
+      ) : (
```
martha: Could also do:
`if (workspaceLoading) return <Loading/>;`
above on line 99. To me it looks a little neater and more in line with our usual pattern. But we definitely do both throughout the codebase, so up to your preference :)

src/modules/ce/components/ReferralsPage.tsx
```diff
@@ -26,56 +46,73 @@ const ReferralsPage: React.FC<Props> = ({ currentTab }) => {
...
-      <CommonTabs
-        tabDefinitions={tabDefinitions}
-        ariaLabel='Referrals'
-        collapseSingleTab
-        currentTab={currentTab}
-        onChangeTab={handleChangeTab}
-      />
+    <PageContainer
+      title='Referrals'
+      actions={hasWorkspaces ? workspaceSelector : undefined}
```
martha: The `hasWorkspaces` guard is somewhat redundant, since `workspaceSelector` is already `null` when there are no workspaces, so we could simplify by just passing it directly here, and then removing `hasWorkspaces` entirely from the api of the `useWorkspaceSelector` hook.

src/components/elements/input/WorkspaceSelect.tsx
```diff
@@ -0,0 +1,79 @@
...
+      size='small'
+      label='Workspace'
+      ariaLabel='Workspace'
+      options={workspaceOptions}
+      value={selectedWorkspaceOption}
+      getOptionLabel={(option) => option.label || option.code}
+      isOptionEqualToValue={(option, value) => option.code === value.code}
+      blurOnSelect
+      disableClearable
+      onChange={(_event, value) => onChange(value?.code || null)}
```
martha: When "All Projects" is selected, is it intentional that we are passing `all` to onChange, instead of `null`? The result is that `workspace=all` ends up in the URL, but has the same meaning as no workspace search param. Line 35 will be null anyway, assuming that no `workspaceOption` matches a slug of `all`.

This was flagged by my AI review as something to clean up. But I think it's optional; the current functionality you have is fine.

src/modules/ce/components/ReferralsPage.tsx
```diff
@@ -26,56 +46,73 @@ const ReferralsPage: React.FC<Props> = ({ currentTab }) => {
...
-        navigate(Routes.REFERRALS);
+        navigate(`${Routes.REFERRALS}${search}`);
         break;
       case 'available-units':
-        navigate(ReferralRoutes.AVAILABLE_UNITS);
+        navigate(`${ReferralRoutes.AVAILABLE_UNITS}${search}`);
         break;
       case 'eligible-clients':
-        navigate(ReferralRoutes.ELIGIBLE_CLIENTS);
+        navigate(`${ReferralRoutes.ELIGIBLE_CLIENTS}${search}`);
```
martha: It looks like `react-router-dom` exposes a nicer api for this, using `navigate`'s `search` prop instead of manually concatenating the search param with the url.
```diff
diff --git a/src/modules/ce/components/ReferralsPage.tsx b/src/modules/ce/components/ReferralsPage.tsx
index ce6b77df..23efa55a 100644
--- a/src/modules/ce/components/ReferralsPage.tsx
+++ b/src/modules/ce/components/ReferralsPage.tsx
@@ -1,5 +1,5 @@
 import React, { useMemo } from 'react';
-import { useLocation, useNavigate } from 'react-router-dom';
+import { useNavigate, useSearchParams } from 'react-router-dom';
 
 import CommonTabs from '@/components/elements/CommonTabs';
 import Loading from '@/components/elements/Loading';
@@ -22,7 +22,7 @@ interface Props {
 const ReferralsPage: React.FC<Props> = ({ currentTab }) => {
   const [access] = useRootPermissions();
   const navigate = useNavigate();
-  const location = useLocation();
+  const [searchParams] = useSearchParams();
 
   // Load workspace selector (if configured) for selecting project group
   const {
@@ -78,21 +78,17 @@ const ReferralsPage: React.FC<Props> = ({ currentTab }) => {
   }, [showUnitsAndEligibleClients, selectedProjectGroupId]);
 
   const handleChangeTab = (key: string) => {
-    const workspace = new URLSearchParams(location.search).get('workspace');
-    const search = workspace
-      ? `?workspace=${encodeURIComponent(workspace)}`
+    const workspaceParam = searchParams.get('workspace');
+    const search = workspaceParam
+      ? new URLSearchParams({ workspace: workspaceParam }).toString()
       : '';
-    switch (key) {
-      case 'referrals':
-        navigate(`${Routes.REFERRALS}${search}`);
-        break;
-      case 'available-units':
-        navigate(`${ReferralRoutes.AVAILABLE_UNITS}${search}`);
-        break;
-      case 'eligible-clients':
-        navigate(`${ReferralRoutes.ELIGIBLE_CLIENTS}${search}`);
-        break;
-    }
+    const routes: Partial<Record<string, string>> = {
+      'referrals': Routes.REFERRALS,
+      'available-units': ReferralRoutes.AVAILABLE_UNITS,
+      'eligible-clients': ReferralRoutes.ELIGIBLE_CLIENTS,
+    };
+    const pathname = routes[key];
+    if (pathname) navigate({ pathname, search });
   };
 
   if (workspaceError) throw workspaceError;
```

src/components/elements/input/WorkspaceSelect.tsx
```diff
@@ -0,0 +1,79 @@
...
+              <Box
+                component='span'
+                sx={{
+                  width: 12,
+                  height: 12,
+                  borderRadius: '50%',
+                  bgcolor: 'primary.main',
+                }}
+              />
+            </InputAdornment>
```
martha: I have a couple concerns about the dot + main color on the select component. I started typing up a comment here, but then decided I'll put my design comments in a thread over in Figma, for better separation of concerns :) [Figma thread](https://www.figma.com/design/kqZiUPjXvqSvJVslJCnlTB?node-id=11095-3344&m=dev#1761036057)

Code-wise, if we do end up choosing to establish this pattern as part of our design language, I would prefer to put the custom styling in a thin presentational wrapper around `GenericSelect`. We could call it something like `AccentedSelect`, `PrimaryColorSelect`, `BrandedSelect`, etc. The dot and custom color could live there, and the component would pass the rest of its props down directly to the `GenericSelect` it renders.

martha (approved):
Looks good to me. I don't think any of my comments are blocking. Thanks @gigxz ! Speedy work! 🚄

src/components/elements/input/WorkspaceSelect.tsx
gigxz: I believe the purpose is probably to bring some visual attention to the context switcher on the page and differentiate it from other "input"-type selectors.
Looks like your question is still open in Figma. For now, I just moved the adornment into its own component (within the same file) and added a note saying we can pull it out to another component or variant for re-use when needed.

Even if designs change, I think it'll be non-blocking for merge because nobody will see this functionality yet (not until we add any Workspace records)

src/modules/ce/components/ReferralsPage.tsx
gigxz: thanks, updated!

src/components/elements/input/WorkspaceSelect.tsx
gigxz: Updated to save "all" selection as null!

PR #1474, opened 2026-05-21
Pull lockfile fix from staging to main

gigxz:


martha (approved):

gigxz:
need to chnage rules to allow merge into main for this

PR #1488, opened 2026-06-10
Add Admin UI for managing global CE Match Rules

martha:
## Description

Summary of changes:
- Add a new Admin page for creating global (only, for now) CE Match Rules. Supports client fields and custom assessment fields, with comparator/value inputs driven by field metadata from the backend.
- Add GraphQL operations for creating CE Match Rules and fetching CE match field metadata.
- Update shared RHF input wrappers where needed for the CE match builder:
  - Type `name` and `control` with RHF generics so nested CE match clause paths are checked by TypeScript.
  - Pass the `loading` prop down.

Not included in this PR; will split out into future PRs to make the diff more manageable:
- A warning confirmation UI that shows affected unit groups before saving rules that would remove too many candidates. (For this PR, the impact warning is bypassed by always passing `confirm=true`.)
- An advanced free-text editor for expressions the structured builder does not support.
- Correct CE permissions. For this PR, only GR super-admins can view the new UI. (AI review will probably flag a mismatch between permissions on the frontend and backend)

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/6590

How to test: See issue QA instructions

Screenshot 2026-06-12 at 9 17 20 AM

## Type of change
New feature


src/modules/admin/hooks/useAdminDashboardNavItems.tsx
```diff
@@ -29,6 +29,13 @@ export const useAdminDashboardNavItems = () => {
             permissions: ['canManageCeDefaultContacts'],
             hide: !globalFeatureFlags?.coordinatedEntryEnabled,
           },
+          {
+            id: 'ce-match-rules',
+            title: 'CE Match Rules',
```
gigxz: Not sure if I am conflicting with design advice here, but I would suggest renaming this to "Eligibility Rules" to use consistent language with other places where we surface eligibility rules. The route itself should be updated too

src/modules/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx
```diff
@@ -0,0 +1,151 @@
...
+        confirmed,
+      },
+    });
+  };
+
+  return (
+    <Stack gap={2}>
+      <TitleCard title='Rule Details' headerComponent='h2' padded>
+        <Stack gap={2}>
+          {saved && <Alert severity='success'>Match rule saved.</Alert>}
```
gigxz: Same comment here regarding not introducing "match rule" language (since we already have "eligibility rule" language established and they mean the same thing)

src/modules/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx
```diff
@@ -0,0 +1,151 @@
...
+    },
+  };
+};
+
+/**
+ * The top-level CE match rule form component.
+ * Owns the RHF state to collect rule details and structured clauses,
+ * and the mutation call to submit to the backend.
+ */
+const CeMatchRuleForm = () => {
```
gigxz, line 63: There's no client side validation of the rules here so I can submit something empty and get an error. Since we're using RHF, can we mark those fields as required so we get the client-side validation?

Image

gigxz (approved):
exciting!

PR #1490, opened 2026-06-10
[#9289] Sort CE Assessment tab by Assessment Date

meborn:
## Description

Summary of changes:

Resolves https://github.com/open-path/Green-River/issues/9289

This updates the Enrollment Dashboard CE Assessments tab so records are sorted by Assessment Date, newest to oldest.

The issue suggested changing the backend resolver default in `HasCeAssessments`, but that did not produce the result we needed in the UI. The table uses the generated `GetEnrollmentCeAssessmentsDocument` from `hmis-frontend`, so the frontend still needed its GraphQL operation to request the sort order explicitly. A backend Ruby keyword default does not update the generated query document that Apollo runs in the browser.

This PR follows the existing `GenericTableWithData` pattern used elsewhere in the frontend: the GraphQL operation declares a `$sortOrder` variable, passes it through to the paginated field, and the table sets its default sort option.

Changes included:

- Add `$sortOrder: AssessmentSortOption = ASSESSMENT_DATE` to `GetEnrollmentCeAssessments`
- Pass `sortOrder: $sortOrder` to the `ceAssessments` field
- Set the CE Assessments table default sort option to `AssessmentSortOption.AssessmentDate`
- Regenerate GraphQL types so the generated query document includes the sort argument
- Add `.cursor/` to `.gitignore` so local Cursor rules are not committed

How to test:

1. Visit an enrollment’s CE Assessments tab:

`/client/:clientId/enrollments/:enrollmentId/ce-assessments`

2. Confirm CE Assessments are sorted by Assessment Date from newest to oldest.

## Type of change

Bug fix


meborn:
@martha There are other graphql changes here that I am not sure about. I did pull main for the warehouse and run some migrations. Is this fine? Is there a process there? Or did I just do something weird here?

martha:
@meborn ah yeah, the other graphql changes are from a recent change of mine that was merged to the backend but didn't have a corresponding frontend PR yet. Thanks for the catch! I opened one here: https://github.com/greenriver/hmis-frontend/pull/1491

martha:
> The issue suggested changing the backend resolver default in `HasCeAssessments`, but that did not produce the result we needed in the UI. 

Can you provide more detail about why this didn't produce the result needed?

The change I had in mind was in the `hmis-warehouse` codebase, around [here](https://github.com/greenriver/hmis-warehouse/blob/9b26502f93331fe87a2daa1abc59931f20703176/drivers/hmis/app/graphql/types/hmis_schema/has_ce_assessments.rb#L31-L35). The advantage of adding a default sort order in the concern, as opposed to sorting from the frontend only, is that it's a bit more generic, so if we add this concern on more entities, it would apply there too. (To spell it out a bit more concretely: Right now we only ever resolve CeAssessments on Enrollments, but if we ever wanted to resolve all the CeAssessments in a Project, we would use this concern, so having a default sort order in the concern would ensure consistency.)

But if you ran into some problems with that approach during implementation, let me know what they were!

(We could _also_ make some changes to the frontend codebase as well, if we want to get the sort "widget" working. 
Screenshot 2026-06-11 at 8 49 04 AM

But I think that's optional for the current ticket's scope.)

martha:
I think this can be closed given your backend change fixes the issue, does that sound right @meborn ?

meborn:
@martha Sure! I am currently working on configuring my commits to be signed and then I will take a look at the rspec thing and the sort component on the frontend. Not sure of process here. If closing the issue before the merge is fine that works for me. 

Edit: I thought this was a comment on the issue. I am seeing now it is the frontend PR. If you don't think I should take a look at the sorting that is fine too. Up to you!

martha:
Hm, I guess I'd prioritize adding an rspec test over updating the sort widget, since that sort widget is pretty solidly out of the scope of the ticket, as written.

PR #1494, opened 2026-06-12
Add CE rule free-text editor

martha:
## Description

Summary of changes:
- Add free-text expression editor
- Add confirmation dialog when switching, if there are unsaved changes

Screenshot 2026-06-12 at 9 52 44 AM

## Type of change
New feature


src/modules/admin/components/ceMatchRules/editor/CeMatchExpressionModeSwitch.tsx
```diff
@@ -0,0 +1,92 @@
...
+    structuredExpression.operator !== CeMatchRuleBooleanOperator.And ||
+    structuredExpression.clauses.some(
+      ({ source, customAssessmentFormIdentifier, field, value }) =>
+        hasValue(source) ||
+        hasValue(customAssessmentFormIdentifier) ||
+        hasValue(field) ||
+        hasValue(value)
+    )
+  );
+};
```
gigxz: This feels like something where we could make use of RHF tooling, for example
```tsx
const { isDirty } = useFormState({
    control,
    name: 'structuredExpression',
  });
```

That said, though, probably makes sense to wait to rework this (if at all) until this UI supports "edit" mode for an existing rule. I imagine the form/state management might have to change a bit when we introduce that.

gigxz (approved):

src/modules/admin/components/ceMatchRules/editor/CeMatchExpressionModeSwitch.tsx
martha: That's true and it's a pretty quick change, so I went ahead with it, although I think you're right that it will get reworked when the UI supports edit mode.

PR #1495, opened 2026-06-15
Add Global match rules landing page; scaffold other tabs

martha:
## Description

Summary of changes: Add the Global match rules landing page, supported by reusable components that will also be used for Organization, Project, and Unit Group-level rules.

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/6606/changes

Screenshot 2026-06-22 at 11 58 15 AM

## Type of change
New feature


src/modules/admin/components/ceMatchRules/ruleGroups/CeMatchRuleGroupSection.tsx
```diff
@@ -0,0 +1,68 @@
...
+  ownerLevel: CeMatchRuleOwnerLevel;
+  icon: ReactNode;
+  rules: CeMatchRuleAdminSummaryFieldsFragment[];
+  count: number;
+  title?: string;
+  variant?: 'current' | 'inherited';
+}
+
+/**
+ * A collapsible section of rules, representing a single owner level,
```
gigxz: This doesn't seem to be collapsible, but maybe it will be later?

src/modules/admin/components/ceMatchRules/ruleGroups/CeMatchRuleGroupSection.tsx
```diff
@@ -0,0 +1,68 @@
...
+          variant='outlined'
+          startIcon={<AddIcon />}
+        >
+          Add {ownerLevelLabel} Rule
+        </Button>
+      </Stack>
+      <CeMatchRuleGroupTable
+        ownerLevel={ownerLevel}
+        rules={rules}
+        variant='current'
```
gigxz: Should variant prop be passed through here?

src/api/operations/ceMatchRules.queries.graphql
```diff
@@ -22,3 +22,12 @@ query GetCeMatchCustomAssessmentFields($formDefinitionIdentifier: String!) {
...
 }
+
+query GetCeMatchGlobalRules {
+  ceMatchRules(filters: { ownerType: DATA_SOURCE }) {
+    nodesCount
+    nodes {
+      ...CeMatchRuleAdminSummaryFields
+    }
+  }
+}
```
gigxz, line 33: If this will be reused for the other tabs, is could be adjusted to:
```
query GetCeMatchRules($ownerType: CeMatchRuleOwnerType!) {
  ceMatchRules(filters: { ownerType: $ownerType }) {
    nodesCount
    nodes {
      ...CeMatchRuleAdminSummaryFields
    }
  }
}

```

src/components/elements/CommonTabs.tsx
```diff
@@ -70,11 +72,19 @@ const CommonTabs: React.FC<CommonTabsProps> = ({
           value={currentIndex}
           onChange={handleChange}
           aria-label={ariaLabel}
+          sx={{
+            '& .MuiTab-labelIcon': {
+              // override MUI default so tabs with icons aren't abnormally tall
+              minHeight: 48,
```
gigxz, line 77: I think the icons also look a bit goofy because they are bigger than the text (compared to the designs, where the tab text is larger).
Maybe add in here `svg: { width: '0.8em', height: '0.8em' },` to reduce the size of the icons ?

gigxz: Or the more appropriate way to do it is probably pass fontSize here:
```diff
      {
        title: 'Global',
        key: 'global',
        icon: <GlobalIcon fontSize='small' />,
        iconPosition: 'start' as const,
        contents: <CeMatchGlobalRules />,
      },
```

src/modules/admin/components/ceMatchRules/ruleGroups/CeMatchEffectiveRulesCard.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <Typography variant='cardTitle' component='h2' fontWeight='600'>
+        Effective Rules for {ownerName} ({effectiveRulesCount})
+      </Typography>
+      <Stack direction='row' gap={4} flexWrap='wrap'>
+        {ruleCountSummaries.map(({ label, count }) => (
+          <Typography key={label} variant='body2'>
+            {label}: {count} rules
+          </Typography>
+        ))}
+      </Stack>
```
gigxz, line 38: Maybe this will make more sense once additional support is added for other types.. right now the "summary" and the card title display the same information.

For global rules, can't this just say "Global Rules (8)" (instead of `"Effective Rules for Global (8); Global: 8 rules"`?

src/modules/admin/components/ceMatchRules/ruleGroups/CeMatchRuleGroupSection.tsx
martha: Yes - I ended up reworking this in the next PR, so I'll bring some of those changes in! Thanks!

src/modules/admin/components/ceMatchRules/ruleGroups/CeMatchRuleGroupSection.tsx
martha: Followup to the comment thread above - the new `CeMatchRuleGroup` component is only for rendering the current-level group, so I removed the variant prop. (`CeMatchRuleAccordion`, the collapsible inherited variant, is referenced in a comment here but not added til the next PR)

src/api/operations/ceMatchRules.queries.graphql
martha: Per the thread [here](https://github.com/greenriver/hmis-warehouse/pull/6606#discussion_r3469801019) on the backend PR, I'm leaning towards not reusing this on the other tabs, but I'm open to feedback about that!

gigxz: I think I misunderstood the other tabs. I was thinking they were pages of rules, but really they will be pages of organizations, page of projects, etc.

martha:
I'm converting this to draft for now while working on cleanly bringing in changes from the other branch

src/components/elements/CommonTabs.tsx
martha: Thanks, yeah that's a good point, they look better this way. I figured if we want more tabs with icons in the future they should probably have consistent styling, so I reworked this to have CommonTabs accept an Icon type instead of a react node, so that the CommonTabs component can handle how the icons should be styled and the callers don't have to worry about it.

src/modules/admin/components/ceMatchRules/ruleGroups/CeMatchEffectiveRulesCard.tsx
martha: Yes... I agree, it's duplicative for Global rules. I was pretty much blindly following the mocks here:

Image

I'd like to keep it as-is for now and confirm with design, since I don't feel totally clear on what should be changed/removed. If we change this to "Global Rules (8)" then we end up with:

Image

which to me somehow feels even more duplicative.

gigxz (approved):

PR #1504, opened 2026-06-22
Add VisionLink eligibility flags form component

gigxz:
## Description

Issue: https://github.com/open-path/Green-River/issues/9248
Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/6626

Adds frontend support for Eviction Prevention assessments to fetch VisionLink eligibility flags via the new fetchVisionLinkFlags GraphQL mutation. Introduces a dedicated visionlink module with a form group component wired into dynamic form rendering.

**How to test**: See issue description for minimal FormDefinition and how to mock responses from FetchVisionLink mutation

[See Screenshots](https://github.com/open-path/Green-River/issues/9248#issuecomment-4772701477)

## Type of change
New feature


src/modules/external/visionlink/components/VisionLinkFlags.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+  const { loading, errorState, fetchVisionLinkFlags } = useFetchVisionLinkFlags(
+    { clientId, onFlagsFetched }
+  );
+
+  if (!isComponentValid)
+    throw new Error(
+      'Invalid VisionLink form component. Expected link IDs: ' +
+        VisionLinkLinkIds.join(', ')
+    );
```
martha: This mirrors the behavior of the AHA form component, but I think we might want to consider another strategy, such as not raising in read-only mode, or generally avoid raising but alert to Sentry instead.

In theory we thought we'd only hit this raise in the case of a developer error. In practice, for AHA, it's come up a few times on old form versions. Raising here has 2 main problems, I think: 
- When rendering an old, readonly form in an assessment-collection context, it makes it impossible to view/unlock/edit the assessment at all. Maybe in readonly mode, it would be preferable to just show whatever we have/can map to a link ID. And in edit mode, we update to the latest form version anyway.
- When rendering a form in the admin builder, it makes it impossible to edit the form, meaning the only possible fix is a backend edit directly to the json definition in the database. I think it would be preferable to either not raise, or change the error boundary so that we can still enter the form builder/json editor and fix the problem.

martha (approved):
looks great! 🥳

src/modules/external/visionlink/components/VisionLinkFlags.tsx
gigxz: Updated as discussed to send a Sentry message without raising, and only in edit-mode (to avoid noise from retired malformed forms)

PR #1510, opened 2026-06-24
Lock and visually flag HUD-linked fields

martha:
## Description

Summary of changes:
- Add helper text to common menu items. This looks similar to the existing disabled-reason, but is available when the field is enabled.
Screenshot 2026-06-24 at 2 48 35 PM

- Visually flag HUD-linked fields on the form builder tree, and disable deletion of these fields for non-super-admins.
Screenshot 2026-06-24 at 2 44 52 PM

- In the form item editor, lock certain fields so they can only be edited by super-admins. We can discuss these choices more -- my guess was that for properties like label and helper text, it's not harmful to allow broader editing. But if it would be preferable to just fully lock the item and disallow editing, I can make that change.
Screenshot 2026-06-24 at 2 45 41 PM

Depends on hmis-warehouse PR: N/A. In plan mode, AI also suggested backend changes to prevent super-admins from changing the locked fields, but my intuition was that we can descope the backend changes. My reasoning: Since this is already an admin-only UI backed by admin-only endpoints, we are just trying to add better UX affordances that will help the user not trip up or get into a weird state.

## Type of change
Bug fix

### DIFF
--- a/src/components/elements/CommonMenuButton.tsx
+++ b/src/components/elements/CommonMenuButton.tsx
@@ -165,7 +165,7 @@ const CommonMenuButton = ({
                       variant={'caption'}
                       sx={{ fontStyle: 'italic' }}
                     >
-                      {helperText }
+                      {helperText}
                     </Typography>
                   )}
                 </Stack>
--- a/src/components/elements/CommonMenuButton.tsx
+++ b/src/components/elements/CommonMenuButton.tsx
@@ -28,6 +28,7 @@ export type CommonMenuItem = {
   divider?: boolean;
   disabled?: boolean;
   disabledReason?: string;
+  helperText?: string;
   ariaLabel?: string;
   linkState?: LocationState;
   openInNew?: boolean;
@@ -126,6 +127,7 @@ const CommonMenuButton = ({
             onClick,
             disabled,
             disabledReason,
+            helperText,
             ariaLabel,
             openInNew,
             linkState,
@@ -156,6 +158,14 @@ const CommonMenuButton = ({
                       sx={{ fontStyle: 'italic' }}
                     >
                       {disabledReason}
+                    </Typography>
+                  )}
+                  {helperText && (
+                    <Typography
+                      variant={'caption'}
+                      sx={{ fontStyle: 'italic' }}
+                    >
+                      {helperText || 'hello'}
                     </Typography>
                   )}
                 </Stack>
--- a/src/components/elements/CommonMenuButton.tsx
+++ b/src/components/elements/CommonMenuButton.tsx
@@ -165,7 +165,7 @@ const CommonMenuButton = ({
                       variant={'caption'}
                       sx={{ fontStyle: 'italic' }}
                     >
-                      {helperText || 'hello'}
+                      {helperText }
                     </Typography>
                   )}
                 </Stack>
--- a/src/modules/formBuilder/components/HudChip.tsx
+++ b/src/modules/formBuilder/components/HudChip.tsx
@@ -1,0 +1,16 @@
+import { Chip } from '@mui/material';
+
+const HudChip: React.FC = ({}) => {
+  return (
+    <Chip
+      label='HUD'
+      size='small'
+      variant='outlined'
+      color='primary'
+      aria-label='HUD field'
+      sx={{ fontWeight: 600, height: '20px' }}
+    />
+  );
+};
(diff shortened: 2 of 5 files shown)


src/components/elements/CommonMenuButton.tsx
```diff
@@ -158,6 +160,14 @@ const CommonMenuButton = ({
                       {disabledReason}
                     </Typography>
                   )}
+                  {helperText && (
+                    <Typography
+                      variant={'caption'}
+                      sx={{ fontStyle: 'italic' }}
+                    >
+                      {helperText || 'hello'}
```
gigxz:
```suggestion
                      {helperText }
```

src/modules/formBuilder/components/formTree/FormTreeLabel.tsx
```diff
@@ -97,8 +105,30 @@ const FormTreeLabel: React.FC<FormTreeLabelProps> = ({
...
-    () => [
+  const menuItems = useMemo(() => {
+    // Disable deletion for groups that have children.
+    const groupHasChildren =
+      item?.type === ItemType.Group && !!item?.item && item.item.length > 0;
+
+    // Disable deletion (for non-super-admins) if this is a HUD field that is required,
+    // to prevent the user from deleting the field and breaking the form's HUD data collection.
+    // This relies on the form being configured to correctly require HUD fields when the underlying HUD record requires them.
+    const disableDeleteHudItem = isRequiredHudItem && !canAdministrateConfig;
```
gigxz: I think we should disable deletion of `fieldName`-mapped fields for all non-super-admins, regardless if the field is marked as required on the form or not. That feels a bit safer.. Often the HUD-required fields are marked "Warn if empty" for example for 99s but they should still be present.

The "required" value on the item doesn't necessarily map to being required _by HUD_, so it also feels like it could be confusing to say "This item maps to a required HUD field" as the reason

src/modules/formBuilder/components/HudChip.tsx
```diff
@@ -0,0 +1,16 @@
+import { Chip } from '@mui/material';
+
+const HudChip: React.FC = ({}) => {
```
gigxz: So these aren't exactly HUD fields.. grep for `"field_name":` in `./hmis-warehouse/drivers/hmis/lib`  you'll see all the different fields that are "standard". Most of them are "HUD" but not all. This might get confusing to administrators, we certainly don't want to mislead them about what's a HUD-mandated field and what's not. I'm thinking about the client and enrollment forms in particular.

Maybe we could use something more generic to identify these fields as non-custom fields. For example:
* Built-in field
* Standard field
* Core field
* System field
* Fixed mapping


What do you think? Out of those I think I like the first 2 the best..

gigxz: And of course, it _would_ be great if we could identify HUD-required fields in the builder eventually, but the presence of `field_name` vs `custom_field_key` isn't sufficient for that

gigxz:
> In the form item editor, lock certain fields so they can only be edited by super-admins. We can discuss these choices more -- my guess was that for properties like label and helper text, it's not harmful to allow broader editing. But if it would be preferable to just fully lock the item and disallow editing, I can make that change.

This seems fine to me to allow some editing. The editor component is definitely pretty complex with the amount of conditional rendering that's going on, and this is adding to that pre-existing issue, but I think it's manageable and something we can refactor on a v2 pass.

src/components/elements/CommonMenuButton.tsx
martha: Thanks haha!

src/modules/formBuilder/components/formTree/FormTreeLabel.tsx
martha: Thanks, that makes sense to me!

src/modules/formBuilder/components/HudChip.tsx
martha: Makes sense. "Standard field" sounds good. Among your other suggestions I also like "System field", since it mirrors the language of "System forms", which feels like a related concept: there are certain forms, AND certain form fields (even on custom forms), that are required for the system to function correctly and can't be freely modified by users.

I'll go with "System field" for that reason, unless you have hesitations!

gigxz (approved):

PR #1514, opened 2026-06-26
Fix bug where client with masked DOB is considered adult

martha:
## Description

Summary of changes:
- Add `age` to the `ClientNameDobVet` fragment.
  - Alternatives considered: add to the `AssessedClientFields` fragment. 
    - Why: This change would have had a smaller footprint, scoped only to assessments. 
    - Why not: Adding it to the more broadly-used `ClientNameDobVet` fragment helps guard against similar/future bugs in other contexts. The backend cost of querying this field is low, since `age` is an in-memory calculation from the already-loaded DOB.
- Update the `age(client)` helper used by `evaluateEnableWhen` to use the explicit `age` field first, and fallback to `dob` if unavailable (unexpected). 
  - This also fixes the other usage of the `age` helper, which is the client card at the top of the left-hand nav on the client dashboard. Previously, if DOB was masked, it would not display the client age; now it does.

Depends on hmis-warehouse PR: N/A

## Type of change
Bug fix

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@
 .env.development.local
 .env.test.local
 .env.production.local
+.cursor/
 bundle_analysis*
 untracked
 docker-compose.yml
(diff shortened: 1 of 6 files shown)


gigxz (approved):
nice!

PR #1516, opened 2026-06-26
Display Referral History on Unit page

meborn:
## Description

Summary of changes:

Adds a dedicated `Referral History` tab to the Unit page. The tab shows a paginated referral table for the selected unit with client name, status, referred by, referral date, last updated date, and origin.

The table links each row to the CE referral detail page and preserves Unit breadcrumb context when navigating from Unit Referral History, including both the standard Unit route and CE Unit route. The Unit page also stays reachable for historical referral review even when the current unit group no longer has a waitlist workflow; workflow-dependent Eligible Clients UI remains hidden in that case.

Accessibility improvements:
- Updated Unit tab aria label to reflect the broader tab set.
- Added meaningful row action names for referral history rows.

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/6662

How to test:

1. Open a waitlist-enabled project with a unit that has historical referrals across multiple opportunities.
2. Open the Unit page and confirm the `Referral History` tab is visible for users with referral visibility.
3. Confirm the table shows client, status, referred by, referral date, last updated date, and origin.
4. Click a referral row and confirm the referral detail page opens.
5. Confirm breadcrumbs return to the Unit page when the referral was opened from Unit Referral History.
6. Repeat from the CE available-units route and confirm breadcrumbs return to the CE Unit route without errors.
7. Confirm users without referral visibility do not see the referral history tab/data.

## Type of change
New feature, accessibility improvement

### DIFF
--- a/src/components/elements/Breadcrumbs.tsx
+++ b/src/components/elements/Breadcrumbs.tsx
@@ -17,7 +17,6 @@ export interface Breadcrumb {
   label: string;
   to: string;
   icon?: SvgIconComponent;
-  routeParams?: Record<string, string>;
 }
 interface Props extends BreadcrumbsProps {
   crumbs: Breadcrumb[];
@@ -44,7 +43,7 @@ const Breadcrumbs = ({ crumbs, variant = 'body2', sx, ...rest }: Props) => {
       }}
       {...rest}
     >
-      {crumbs.map(({ label, to, icon: Icon, routeParams }, index) => {
+      {crumbs.map(({ label, to, icon: Icon }, index) => {
         if (index === crumbs.length - 1) {
           return (
             <Typography variant={variant} key={to} sx={{ display: 'inline' }}>
@@ -54,10 +53,7 @@ const Breadcrumbs = ({ crumbs, variant = 'body2', sx, ...rest }: Props) => {
         }
 
         // Fill in path with params we already have (e.g. replace :clientId with id)
-        const filledInPath = generateSafePath(to, {
-          ...params,
-          ...routeParams,
-        });
+        const filledInPath = generateSafePath(to, params);
         return (
           <RouterLink
             aria-label={label}
--- a/src/components/elements/Breadcrumbs.tsx
+++ b/src/components/elements/Breadcrumbs.tsx
@@ -17,6 +17,7 @@ export interface Breadcrumb {
   label: string;
   to: string;
   icon?: SvgIconComponent;
+  routeParams?: Record<string, string>;
 }
 interface Props extends BreadcrumbsProps {
   crumbs: Breadcrumb[];
@@ -43,7 +44,7 @@ const Breadcrumbs = ({ crumbs, variant = 'body2', sx, ...rest }: Props) => {
       }}
       {...rest}
     >
-      {crumbs.map(({ label, to, icon: Icon }, index) => {
+      {crumbs.map(({ label, to, icon: Icon, routeParams }, index) => {
         if (index === crumbs.length - 1) {
           return (
             <Typography variant={variant} key={to} sx={{ display: 'inline' }}>
@@ -53,7 +54,10 @@ const Breadcrumbs = ({ crumbs, variant = 'body2', sx, ...rest }: Props) => {
         }
 
         // Fill in path with params we already have (e.g. replace :clientId with id)
-        const filledInPath = generateSafePath(to, params);
+        const filledInPath = generateSafePath(to, {
+          ...params,
+          ...routeParams,
+        });
         return (
           <RouterLink
             aria-label={label}
--- a/src/modules/ce/referralColumns.tsx
+++ b/src/modules/ce/referralColumns.tsx
@@ -33,12 +33,6 @@ export const REFERRAL_COLUMNS: Record<
     render: (referral: CeReferralBaseTableFieldsFragment) =>
       parseAndFormatDate(referral.createdAt),
   },
-  updatedDate: {
-    header: 'Last Updated',
-    key: 'updatedDate',
-    render: (referral: CeReferralBaseTableFieldsFragment) =>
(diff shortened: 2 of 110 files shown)


src/modules/units/components/UnitPage.tsx
```diff
@@ -102,24 +111,11 @@ const UnitPage: React.FC<Props> = ({}) => {
   if (!project.coordinatedEntryFeatures?.supportsWaitlistReferrals)
     return <NotFound />;
 
-  // If the unit group doesn't have a workflow template identifier, creating client-list-based referrals will not work.
```
meborn, line 105: Allow users to still view history and don't redirect.

vite.config.js
```diff
@@ -140,6 +141,7 @@ export default defineConfig(({ command, mode }) => {
         host: devHmisHost,
         https:
           env.SERVER_HTTPS === undefined ? true : env.SERVER_HTTPS === 'true',
+        watch: usePolling ? { usePolling } : undefined,
```
meborn, line 144: This fixed my reload issue where the frontend wasn't picking up my changes.

src/components/elements/Breadcrumbs.tsx
```diff
@@ -17,6 +17,7 @@ export interface Breadcrumb {
   label: string;
   to: string;
   icon?: SvgIconComponent;
+  routeParams?: Record<string, string>;
```
martha: I think the breadcrumb change here is a lot of churn that we should consider keeping out of scope for the current ticket.

My understanding is that this change was introduced in order to have the referral breadcrumbs be more flexible. So if you click a referral from the Unit History page, the breadcrumbs keep the unit context/path, but if you got there from the Project Referrals page, the breadcrumbs keep the project context/path.

However, I think that flexibility has a few downsides.
- It's a new pattern, and it complexifies the mental model for breadcrumbs.
  - Currently: breadcrumbs are (pretty much) static/deterministic based on the route. (They can get dynamic labels using `overrideBreadcrumbTitles`, but the links themselves are not dynamic.)
  - With this change: breadcrumbs can be different for the same route depending on context/history
- That complexity introduces more confusing edge cases to deal with. One that I noticed so far is: If you are looking at the Unit History page and click into a referral, the breadcrumbs are showing that you're still in the context of a Unit, but the left-hand nav now auto-highlights the "Referrals" nav link. I think that inconsistency is confusing: are you now in the Project Units context, or the Project Referrals context?
- This is sorta a separate thought, but just adding it to the brain dump: We already support flexibility with different contexts where referrals can appear -- they can be in the target project context, the source project context, (`/projects/xxx/ce/referrals/xxx`) or the Referrals context (go to referrals in the top nav, or `/referrals/xxx`). Given that we already have a fair amount of code/complexity to support this, I'm hesitant to add a whole _separate_ pattern of referral context flexibility.

So, overall my feeling is that this breadcrumb change would need more product discussion. For this PR, I'd prefer to revert these changes and accept the previous breadcrumb behavior, which is: Once you click from Unit History into a Referral, you're now in the Project Referrals context and the breadcrumbs reflect that.

What do you think?

src/modules/ce/referralColumns.tsx
```diff
@@ -32,6 +33,12 @@ export const REFERRAL_COLUMNS: Record<
     render: (referral: CeReferralBaseTableFieldsFragment) =>
       parseAndFormatDate(referral.createdAt),
   },
+  updatedDate: {
```
martha: Following up on my [backend PR comment](https://github.com/greenriver/hmis-warehouse/pull/6662/changes#r3493556759), I'd suggest removing this column for now.

src/modules/units/components/UnitPage.tsx
```diff
@@ -53,9 +55,10 @@ const UnitPage: React.FC<Props> = ({}) => {
       key: 'overview',
       contents: <UnitOverview unit={unit} />,
     });
-
     const opportunity = unit.latestOpportunity;
+    const hasWaitlistWorkflow = !!unit.unitGroup?.workflowTemplateIdentifier;
```
martha: Per [my comment below](https://github.com/greenriver/hmis-frontend/pull/1516/changes#r3494267944), if we revert that change, we should revert this too to avoid confusion

src/modules/units/components/UnitPage.tsx
martha: Hm, I want to think through this edge case before making changes to this code. The edge case is where:
- the project does have `supportsWaitlistReferrals`, but
- this particular unit group has no `workflowTemplateIdentifier`. That would be either because:
  - It hasn't been set up yet - in which case there wouldn't be any referral history
  - It used to have one but it has been removed - which is unexpected, because we disable changing the workflow template identifier once it's been set from the UI. So this would only happen resulting from a direct change in the db.

I _think_ therefore that we can revert this change, as it's something we expect to never come up. Especially because the ticket description says,
> The Unit page is currently only exposed in projects that have waitlists enabled. Projects that only receive direct referrals do not use the Unit page, which means that this historical visibility won't be available for them. Changing that is out-of-scope for this ticket.

Alternatively, if you see strong arguments for keeping the change in place, we should add some code comments here explaining why this is unexpected.

vite.config.js
martha: good find!

martha (commented):
Everything looks great to me with the unit history table itself! 🚀  My comments are related to code that I think we can remove/churn we can avoid.

martha (approved):
Nice, this is ready from my perspective! Before merging, I'd say to:
- wait until the backend PR is ready and merged so they go in the same release -- still a couple outstanding comments there
- add an imperative-voice title to the PR such as "Add CE referral history to units", which will become the squashed commit's title when you merge

martha (approved):
LGTM!

PR #1517, opened 2026-06-30
Add ability to edit existing CE match rules

martha:
## Description

Summary of changes:
- Add CE match rule detail routes from the global, organization, project, and unit group rule tables. They all render the same component, but adding distinct routes for each owner enables more useful breadcrumbs.
- Add read-only view, unlock-to-edit, update, and delete flows for existing CE match rules.
- Convert returned structured expression metadata into the editor's field source and custom assessment selections.
- Add reusable code-styled text input/display components for free-text expressions.

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/6664

## Type of change
New feature


src/components/elements/CodeTextBlock.tsx
```diff
@@ -0,0 +1,25 @@
+import { Box, BoxProps } from '@mui/material';
```
martha, line 1: Image

src/modules/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx
```diff
@@ -1,212 +1,228 @@
...
-  onSaved: (rule: CeMatchRuleDetailsFragment) => void;
-  onCancel: VoidFunction;
+  ruleId?: string;
+  initialValues?: CeMatchRuleFormValues;
+  onSaved?: (rule: CeMatchRuleDetailsFragment) => void;
+  onCancel?: VoidFunction;
+  onDelete?: VoidFunction;
 }
 
-const defaultCeMatchRuleFormValues = (): CeMatchRuleFormValues => ({
```
martha, line 41: Moved to `ceMatchRuleFormUtil` now that it is also used by `CeMatchExpressionModeSwitch`

src/modules/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx
```diff
@@ -1,212 +1,228 @@
...
-  name: '',
-  mode: 'structured',
-  structuredExpression: {
-    operator: CeMatchRuleBooleanOperator.And,
-    clauses: [newDraftClause()],
-  },
-  freeTextExpression: '',
-});
-
-const buildCeMatchRuleInput = (
```
martha, line 51: Moved to `useCeMatchRuleFormSubmission` hook (as `buildCreate...` and `buildUpdate...`)

src/modules/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx
```diff
@@ -1,212 +1,228 @@
...
+    // If this is an existing rule, reset the form and exit edit mode
+    clearErrors();
+    reset(defaultValues);
+    setEditing(false);
+  }, [clearErrors, defaultValues, onCancel, reset, ruleId]);
+
+  const handleUnlock = useCallback(() => {
+    clearErrors();
+    // Editable fields unregister when they unmount on cancel, so rehydrate
+    // RHF state from the saved values before mounting them again.
```
martha, line 132: This `handleUnlock` solves a bug where: if you unlock a rule, make no changes, cancel, then unlock again, the form would be blank.

This is because our RHF fields unregister on unmount by default (which is behavior we define in `ControlledTextInput` et.al., not RHF default behavior, but I think changing it and testing its knock-on effects is out-of-scope here). And then, since this same component handles both edit and readonly modes, the whole component/form doesn't get remounted, so the default values don't get rehydrated.

We could alternatively solve this by moving edit/readonly to their own components. Then when the user unlocks, `useForm` would be called again when the whole component remounts, and the default values would hydrate. I opted away from that route because I think the sole component is a bit easier to maintain (e.g. when we want to add the Applicability card, we just add it once here, instead of tracking down 2 components to add it to). But I could go either way, so would be happy for feedback.

src/modules/admin/components/ceMatchRules/editor/clause/CeMatchClause.tsx
```diff
@@ -65,15 +66,18 @@ const CeMatchClause: React.FC<Props> = ({
...
   const customAssessmentFields =
     customAssessmentFieldsData?.ceMatchCustomAssessmentFields ||
     emptyCustomAssessmentFields;
 
   const fields = useMemo(() => {
-    if (source === 'client') return clientItems;
-    if (source === 'custom') return customAssessmentFields;
+    if (source === CeMatchRuleFieldSource.Client) return clientItems;
+    if (source === CeMatchRuleFieldSource.CustomDataElement)
+      return customAssessmentFields;
```
gigxz, line 81: This comment really isn't about _this_ diff, but something I'm noticing. Seems like there will be a lot of churn on this file to add new sources. I think we can refactor that later on when we get to adding them.

I think the "Custom data element" source is unique in that it has the prerequisite (must select assessment type before selecting the field). I don't anticipate other sources having that (eg PSDEs, households, they will each just be namespaced hard-coded lists of fields. Similar to Client).

Given that, I wonder if we might want to adjust this lightly into a UI where the user _first_ selects one of "Standard Field" or "Custom Field", and then all standard fields can share a single grouped, searchable field dropdown (each group maps to a backend "source"). If user selects "Custom Field" then they get the form picker instead.

So, nothing concrete to change here -- just something to keep in mind next time we come back to this file to introduce the new field sources (https://github.com/open-path/Green-River/issues/9139 / https://github.com/open-path/Green-River/issues/9358 / https://github.com/open-path/Green-River/issues/7898)


I'm not sure if that approach would make https://github.com/greenriver/hmis-warehouse/pull/6664/changes#r3521224739 less relevant, but I still suggest making it in case we don't do a larger refactor.

src/modules/admin/components/ceMatchRules/pages/CeMatchRuleDetailPage.tsx
```diff
@@ -0,0 +1,95 @@
...
+  return (
+    <>
+      <PageTitle overlineText='Coordinated Entry Rules' title={rule.name} />
+      <CeMatchRuleForm
+        ruleId={rule.id}
+        initialValues={initialValues}
+        ownerType={rule.ownerType}
+        ownerId={rule.ownerId}
+        ownerName={rule.ownerName}
+        onSaved={returnToOwnerRules}
```
gigxz: This feels a little jarring to me on save because I'm dumped back to a long list of rules and lose track of the one I was changing. I think I would feel more confident in the change as a user if it kept me on the "View rule" page so I can visually confirm that the change saved.

src/modules/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx
```diff
@@ -1,212 +1,228 @@
...
+    if (!ruleId) {
+      onCancel?.();
+      return;
     }
-    return `This eligibility rule will apply to all units${substring}.`;
-  }, [ownerName, ownerType]);
+
+    // If this is an existing rule, reset the form and exit edit mode
+    clearErrors();
+    reset(defaultValues);
```
gigxz: I don't think this should be necessary on cancel since the whole form is unmounting. but doesn't hurt.

src/modules/admin/components/ceMatchRules/pages/CeMatchRuleDetailPage.tsx
gigxz: From what I can tell, it would work to just lock the rule again on save:

```diff
--- a/src/modules/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx
+++ b/src/modules/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx
@@ -22,10 +22,7 @@ import ErrorAlert from '@/modules/errors/components/ErrorAlert';
 import { hasErrors, hasOnlyWarnings } from '@/modules/errors/util';
 import { getRequiredLabel } from '@/modules/form/components/RequiredLabel';
 import ControlledTextInput from '@/modules/form/components/rhf/ControlledTextInput';
-import {
-  CeMatchRuleDetailsFragment,
-  CeMatchRuleOwnerType,
-} from '@/types/gqlTypes';
+import { CeMatchRuleOwnerType } from '@/types/gqlTypes';

 interface Props {
   ownerType: CeMatchRuleOwnerType;
@@ -33,7 +30,6 @@ interface Props {
   ownerName?: string;
   ruleId?: string;
   initialValues?: CeMatchRuleFormValues;
-  onSaved?: (rule: CeMatchRuleDetailsFragment) => void;
   onCancel?: VoidFunction;
   onDelete?: VoidFunction;
 }
@@ -50,7 +46,6 @@ const CeMatchRuleForm: React.FC<Props> = ({
   ownerName,
   ruleId,
   initialValues,
-  onSaved = () => undefined,
   onCancel,
   onDelete,
 }) => {
@@ -77,12 +72,14 @@ const CeMatchRuleForm: React.FC<Props> = ({
   // If the component remounts with different default values, reset the form
   useEffect(() => reset(defaultValues), [defaultValues, reset]);

+  const handleSaved = useCallback(() => setEditing(false), []);
+
   const { errorState, loading, submit, clearErrors } =
     useCeMatchRuleFormSubmission({
       ownerType,
       ownerId,
       ruleId,
-      onSaved,
+      onSaved: handleSaved,
     });

   const mode = watch('mode');
```

gigxz (approved):
Looks good! Some minor comments, I think standardizing the "source" labels from the backend would be nice.

src/modules/admin/components/ceMatchRules/pages/CeMatchRuleDetailPage.tsx
martha: That makes sense. For parity, when editing a new rule, I added a `navigate` call so that it redirects to the rule detail page after the rule is saved.

src/modules/admin/components/ceMatchRules/editor/clause/CeMatchClause.tsx
martha: Thanks, that makes sense to me! I agree with saving this refactor for later.

PR #1524, opened 2026-07-06
Persist page numbers on forward/back navigation

meborn:
## Description

Summary of changes:

Adds URL-backed pagination for the high-priority HMIS tables from open-path/Green-River#8954 so browser Back/Forward and refresh restore the table page position.

- Adds `useTablePagination` to read/write 1-indexed `page` and `pageSize` URL params while exposing the 0-indexed pagination state expected by table components.
- Extends `GenericTableWithData` with optional controlled pagination while preserving existing internal pagination for tables that do not opt in.
- Wires URL-backed pagination into:
  - Referrals
  - Available Units
  - Eligible Clients
  - Project Households
- Keeps pagination reset behavior when filters or parent query variables change.
- Uses active `rowsPerPage` when deciding whether to hide pagination, so a refreshed URL with a smaller `pageSize` still shows controls when additional pages exist.

How to test:

Manual QA:

- On `/referrals`, navigate to page 2 or 3, open a referral, use browser Back, and confirm the same table page is restored.
- Repeat for `/referrals/available-units` and `/referrals/eligible-clients`.
- On a project households table, navigate to another page, open an enrollment/client row, use browser Back, and confirm the same table page is restored.
- Refresh a URL with pagination params and confirm the table opens on the expected page.
- Change filters or workspace context and confirm pagination resets to the first page.
- Test a URL with a smaller `pageSize`, such as `pageSize=10`, and confirm pagination controls still appear when there are more than 10 results.
- Regression test existing URL-backed filters: on the Projects page, load a direct URL with a non-default filter such as status=CLOSED, refresh/open it in a new tab, and confirm the table shows the requested filter without briefly reverting to the default status=OPEN behavior.

## Type of change
New feature


src/modules/dataFetching/util.ts
```diff
@@ -0,0 +1 @@
+export const DEFAULT_TABLE_PAGE_SIZE = 25;
```
martha, line 1: Maybe `util.ts` or `tableUtil.ts` instead of `constants.ts` to match our usual patterns more closely?

src/modules/ce/components/admin/AdminCeClientsTable.tsx
```diff
@@ -106,6 +107,10 @@ const AdminCeClientsTable: React.FC<Props> = ({ projectGroupId }) => {
     dynamicFilters: tableConfigLookup?.ceClientsGlobalConfig?.filters,
     omit: ['projectGroupId'], // only exposed via Workspaces
   });
+  const pagination = useTablePagination({
+    pageParam: 'eligibleClientsPage',
```
martha: I think my preference is towards keeping the pagination params simple and consistent, rather than overriding, when possible. (prefer `page=1&pageSize=10` instead of `eligibleClientsPage=1&eligibleClientsPageSize=10`)

Did you add an override here because of concern about multiple tables in the same page? `CommonTabs` only renders the active tab content, so I think we should be okay on that front. Locally, I tested out removing the overrides on all 3 CE admin tables and didn't notice any ill effects, even with CPU throttling.

We _would_ want to override the param names if we ever render multiple paginated tables on the same page, so I think it's good to keep `useTablePagination`'s `pageParam` and `pageSizeParam` args around, even if we aren't using them.

src/modules/dataFetching/components/GenericTableWithData.tsx
```diff
@@ -76,6 +76,8 @@ export interface Props<
   noData?: ReactNode | ((filters: Partial<FilterOptionsType>) => ReactNode);
   defaultPageSize?: number;
   rowsPerPageOptions?: number[];
+  /** Controlled Pagination: current page/page size and setters. Use when pagination state should live outside the table, e.g. URL-backed via useTablePagination. */
```
martha, line 79: Per our offline discussion, did you ever experiment with just going ahead and making url-backed pagination the default behavior of `GenericTableWithData`? I'm pretty sure I prefer the approach you implemented here, and it's certainly closer to the spec of the actual ticket, but just wanted to get your thoughts in case you did go down this road.

Thoughts from me:

**Pros of making it the default:**
- Every `GenericTableWithData` gets shareable/back-button-friendly pagination with no caller work.
- Removes repetitive caller code in every table `const pagination = useTablePagination(); pagination={pagination}`

**Cons:**
- It's a departure from the `filters` pattern
- Rollout risk is higher because it would be a behavior change for many tables at once

src/hooks/useSearchParamState.ts
```diff
@@ -179,9 +179,16 @@ const useSearchParamsState = ({
 
   const values = useMemo(() => {
     if (initial && !mounted) {
-      // If the above useEffect hasn't run yet, return the initial values right away
-      // to avoid flickering
-      return initial;
+      // Before the initial-param effect runs, use URL values that already exist
```
martha: I puzzled over this change for awhile, because I'm hesitant about churn to the existing hook (gives us more responsibility for regression testing). And I also remembered that this was tricky to get right originally (dug up history from last year: https://github.com/greenriver/hmis-frontend/pull/1076#discussion_r1982641986)

Ultimately though, I concluded that the change you have here is correct, and we should merge it. While poking around with AI, I learned that your change actually fixes a subtle existing bug, which I'll document here for posterity.
- Without this change: visit the Projects page. It automatically filters to `status=OPEN`
- Change the filter to `status=CLOSED`. Results correctly update
- Reload the page, or navigate directly to the current url in a new tab. The data appears to load correctly BUT, if you inspect on the Network tab, there's a double-query happening: one for `status=OPEN` and one for `status=CLOSED`.
- That double-query goes away on your branch, with these changes. It correctly only queries once for `status=CLOSED`.

I believe that's the equivalent of the bug you probably saw: navigating directly to a url with `page=3` wouldn't work, and you'd get incorrectly redirected to page 1.

So! Good catch, good change, and I would just recommend adding a bit more inline commentary so that when we return to this code in the future, we know what problem it was solving. We should also add to the ticket QA instructions that the QA-er should regression test that table filters are still working as expected.

Also for posterity, I'll add my chatty-generated summary of the issue below. Your change is Option B, which I think is the one we should choose.

<Details><Summary>Summary of options</Summary>

# URL State Semantics Plan

## Goal

Decide a clean, reusable philosophy for `useSearchParamsState` before expanding URL-backed state to pagination, sorting, and similar table concerns.

## Current Understanding

`useSearchParamsState` has two concepts:

- `paramsDefinition.default`: the value returned by `getValues` when a param is absent or invalid.
- `initial`: a caller-provided value that the hook writes into the URL on mount when that key is missing.

The older #1076 discussion ([link](https://github.com/greenriver/hmis-frontend/pull/1076#discussion_r1982641986)) landed on `initial` as value-shaped state, not `URLSearchParamsInit`, and simplified first render to return `initial` directly before the URL-writing effect runs. That worked well for optional columns, where `initial` comes from table config or local storage and may differ from the ordinary default.

## Key Tension

Pagination is different from most filters because every paginated table has two meaningful default values: page and page size. There is a reasonable argument that those should appear in the URL immediately, just like default filters such as `status=OPEN` on Projects table.

But using `initial` for pagination exposes a bug in the old first-render behavior: if the user loads `?page=3&pageSize=10`, returning `initial` directly ignores the explicit URL values and renders as page 1/default size.

## Options To Evaluate

### Option A: Keep `initial` only for exceptional seeded state

Use `paramsDefinition.default` for pagination defaults and omit `initial` from `useTablePagination`.

Pros:

- Preserves the simpler #1076 semantics.
- Avoids changing shared hook behavior for pagination.
- URL params win immediately on direct load.
- Keeps routine fallback defaults separate from URL-populating initial state.

Cons:

- Pagination defaults do not appear in the URL until the user changes pagination.
- Behavior differs from default filters that are intentionally populated into the URL.

### Option B: Define `initial` as missing-param URL seed

Keep using `initial` for pagination, but update first-render behavior so existing URL params win and `initial` fills only missing keys.

Pros:

- Matches the mount effect’s behavior: populate only missing params, preserve existing URL params.
- Allows default pagination params to appear in the URL immediately.
- Better supports future URL-backed state that wants explicit URL canonicalization.

Cons:

- Changes the mental model from “first render returns `initial`” to “first render returns URL values plus initial fallbacks.”
- Reopens the complexity that #1076 intentionally simplified.
- Needs regression coverage because this is shared hook behavior.

### Option C: Add an explicit mode/API for URL seeding

Keep `initial` semantics narrow and introduce a clearer option for “write missing defaults into the URL,” for example `populateMissingParams`, `urlDefaults`, or a wrapper hook for table state.

Pros:

- Avoids overloading `initial` with two meanings.
- Makes pagination’s URL-canonicalization intent explicit.
- Gives future sort/pagination/filter behavior a cleaner API surface.

Cons:

- Larger API change.
- More design work now.
- May be overkill if Option B can be clearly documented and tested.

## Recommended Next Step

Agree on the intended contract:

- `default` means the fallback value when the URL does not provide a usable param.
- `initial` means a value-shaped state seed that should be written into the URL for missing keys.
- Explicit URL params should always take precedence over both `initial` and `default`, including on first render.

If that contract feels right, Option B is the most coherent direction: keep pagination’s `initial` so defaults appear in the URL, but fix first render to merge URL values with initial fallbacks and add regression tests.

If that contract feels too broad for `initial`, choose Option C and design a more explicit API before pagination/sort reuse grows.

</Details>

src/hooks/useTablePagination.ts
```diff
@@ -0,0 +1,106 @@
...
+  return value;
+};
+
+const useTablePagination = ({
+  defaultPageSize = DEFAULT_TABLE_PAGE_SIZE,
+  pageParam = DEFAULT_PAGE_PARAM,
+  pageSizeParam = DEFAULT_PAGE_SIZE_PARAM,
+}: Args = {}): TablePaginationState => {
+  // Define the URL-backed keys that this hook owns. Other query params are preserved by useSearchParamsState.
+  const paramsDefinition = useMemo(
```
martha: Minor style thing: I usually prefer to avoid casting in typescript unless it's unavoidable, and here I think it is avoidable if we explicitly type `paramsDefinition` as a `SearchParamsStateType`. That is,

```diff
diff --git a/src/hooks/useTablePagination.ts b/src/hooks/useTablePagination.ts
index 67981f4e..558b4dc9 100644
--- a/src/hooks/useTablePagination.ts
+++ b/src/hooks/useTablePagination.ts
@@ -1,6 +1,6 @@
 import { useCallback, useMemo } from 'react';
 
-import useSearchParamsState from '@/hooks/useSearchParamState';
+import useSearchParamsState, { SearchParamsStateType } from '@/hooks/useSearchParamState';
 import { DEFAULT_TABLE_PAGE_SIZE } from '@/modules/dataFetching/constants';
 
 // Shared defaults keep table pagination URLs consistent unless a page needs custom param names.
@@ -36,14 +36,14 @@ const useTablePagination = ({
   pageSizeParam = DEFAULT_PAGE_SIZE_PARAM,
 }: Args = {}): TablePaginationState => {
   // Define the URL-backed keys that this hook owns. Other query params are preserved by useSearchParamsState.
-  const paramsDefinition = useMemo(
+  const paramsDefinition: SearchParamsStateType = useMemo(
     () => ({
       [pageParam]: {
-        type: 'number' as const,
+        type: 'number',
         default: 1,
       },
       [pageSizeParam]: {
-        type: 'number' as const,
+        type: 'number',
         default: defaultPageSize,
       },
     }),
```

src/hooks/useTablePagination.ts
```diff
@@ -0,0 +1,106 @@
...
+    },
+  });
+
+  // Convert the human-readable 1-indexed URL page into the 0-indexed value used by table components.
+  const urlPage = positiveIntegerOrDefault(paginationParams[pageParam], 1);
+  const rowsPerPage = positiveIntegerOrDefault(
+    paginationParams[pageSizeParam],
+    defaultPageSize
+  );
+  const page = urlPage - 1;
```
martha, line 70: Great attention to detail ✨ 😮

src/modules/dataFetching/components/GenericTableWithData.tsx
```diff
@@ -152,13 +155,20 @@ const GenericTableWithData = <
   FilterOptionsType,
   SortOptionsType
 >) => {
-  const [page, setPage] = useState(0);
-  const [rowsPerPage, setRowsPerPage] = useState(defaultPageSize);
+  const [internalPage, setInternalPage] = useState(0);
```
martha, line 158: Minor style thing: maybe keep `page`-related state definitions all together in one block (so move lines 166-170 up here under line 160)?

src/modules/dataFetching/components/GenericTableWithData.tsx
```diff
@@ -283,7 +306,7 @@ const GenericTableWithData = <
       ) => {
         const newRowsPerPage = parseInt(event.target.value) || defaultPageSize;
         setRowsPerPage(newRowsPerPage);
-        setPage(0);
+        if (!hasControlledPagination) setPage(0);
```
martha, line 311: This tripped me up, so maybe add a comment. Does this reflect the right understanding of your intention with this change?

```suggestion
        // If pagination is uncontrolled (managed by this component), reset page number to 0 when rows per page changes.
        // If pagination is controlled by the parent, let the parent determine this behavior.
        if (!hasControlledPagination) setPage(0);
```

martha (approved):
Looking good overall! I have a few minor style-related comments and some longer philosophical discussion threads

src/modules/ce/components/admin/AdminCeClientsTable.tsx
meborn: Good call. The custom params came from an AI suggestion that I agreed with because keeping useTablePagination flexible for possible multiple-table pages seemed useful. I didn’t dig far enough into whether these CE admin tabs actually needed the override.

I removed the overrides from the CE admin tables and left the pageParam / pageSizeParam hook args in place for any future page that does render multiple paginated tables together.

src/modules/dataFetching/components/GenericTableWithData.tsx
meborn: I didn’t experiment with making URL-backed pagination the default in GenericTableWithData. Since our offline discussion called out specific tables where we wanted this behavior, I kept the implementation opt-in and focused on those call sites.

I agree with your tradeoff read. If we find ourselves adding useTablePagination() to a lot of tables, it would be worth reconsidering whether URL-backed pagination should become the default. For now, the opt-in approach feels safer and more flexible: it matches the ticket scope, avoids a broader behavior change across every GenericTableWithData, and lets us roll this out incrementally.

src/hooks/useSearchParamState.ts
meborn: Thanks for digging into this and for confirming the behavior. I agree with the contract you outlined: default is the fallback when the URL has no usable value, while initial seeds missing URL params, and explicit URL params should always take precedence.

I added a couple inline comments to make that clearer, especially around the first-render behavior that prevents direct-link state like ?page=3&pageSize=10 from being overwritten by seeded values. I’ll also add a QA note to regression test existing URL-backed filters along with the new pagination behavior.

PR #1525, opened 2026-07-08
8458 Support JWT/IdP authentication behind a feature flag

ttoomey:
[Warehouse PR](https://github.com/greenriver/hmis-warehouse/pull/6681)

> These changes must be tested against the `9357-idp-seam-integration` warehouse branch. If backend is run WITHOUT `AUTH_METHOD=jwt`, and the frontend is run WITHOUT `HMIS_BEHIND_OAUTH_PROXY=true` it should work as-is in devise mode

### Description

Adds an optional JWT/SSO authentication path (via `oauth2-proxy`), selected at runtime from the `authMethod` field returned by `GET /hmis/app_settings`, while leaving the existing Devise/Okta session auth unchanged.

- **Auth method resolution**: Adds `resolveAuthMethod`/`useAuthMethod`, both defaulting to `'devise'`. Only the exact value `'jwt'` enables SSO; any other value collapses to `'devise'` so routing, keepalive, and logout can't interpret it inconsistently. Fetched at runtime so one asset bundle serves either auth method.
- **Public landing page**: Adds `<PublicLanding />`, shown to unauthenticated users under `'jwt'` in place of the Devise/Okta form. Sign-in redirects to `/oauth2/sign_in`, adding a stored `connector_id` to skip the IdP picker when a prior IdP is remembered.
- **Remembered IdP**: `fetchCurrentUser` stores the returned `primaryIdp` as the last-used connector id. Cleared only on explicit logout, not on session expiry.
- **App settings bootstrap**: `HmisAppSettingsProvider` loads the public `app_settings` first, then interprets `currentUser`. Under `'jwt'` a `401`/tokenless response leaves the user unset (rendering `<PublicLanding />`); other errors keep the existing reload-once-on-401 recovery. Non-JSON error bodies from `oauth2-proxy` no longer surface as a `SyntaxError`.
- **Terminal account states**: Handles `account_deactivated` and `no_warehouse_account` returned on a `200` under `'jwt'` by showing a terminal page with a sign-out action, rather than looping the user back to sign-in.
- **Logout transport**: Under `'jwt'`, logout is a CSRF-bearing `DELETE` that follows the `redirect_url` the backend returns. Keepalive and Devise/Okta transport are unchanged.
- **Logout/impersonation failure handling**: Adds explicit "still signed in" and "still impersonating" dialogs with retry/dismiss, preserves the session-tracking record on a failed sign-out (so the app doesn't falsely report the session ended), and keeps impersonation exiting through `stopImpersonating`.
- **Local dev**: Adds an optional `HMIS_BEHIND_OAUTH_PROXY` flag to the Vite config to run the dev server behind the warehouse's local SSO stack; default `yarn dev` behavior is unchanged. Documented in the README.
- Adds tests for `resolveAuthMethod`, the terminal-error type guard, and session transport/logout behavior.

### Type of Change

New feature


eanders (commented):
This looks good to me, but I'll defer to @gigxz for approval.

src/modules/auth/api/sessions.ts
```diff
@@ -158,6 +183,12 @@ export function resetLocalSession() {
 }
 
 export async function logout() {
+  // Under JWT/SSO, Hmis::Idp::SessionsController#destroy explicitly skips CSRF
+  // verification (and Hmis::BaseController#valid_request_origin? always passes
```
gigxz: I don't see a method `valid_request_origin?` on the backend, should this reference a different method?

src/modules/hmisAppSettings/Provider.tsx
```diff
@@ -116,33 +143,56 @@ export const HmisAppSettingsProvider: React.FC<Props> = ({ children }) => {
...
+        const fetchedUser = await userPromise;
+        if (userError) {
+          // JWT/SSO: an unauthenticated 401 is expected - leave the user unset so
+          // PublicLanding renders. Any other error (and any error under
+          // Devise/Okta) propagates to the reload-once-on-401 recovery below.
+          if (
+            !(isJwtAuth && (userError as HttpError | undefined)?.status === 401)
+          ) {
+            throw userError;
+          }
```
gigxz: fetchCurrentUser (UsersController#show) doesn't authenticate, so it shouldn't return 401, I think? I believe it just returns an empty object is there is no authenticated user, but maybe I'm missing something.

gigxz (approved):
Looks good!

src/modules/auth/api/sessions.ts
ttoomey: `valid_request_origin?` is from rails
https://api.rubyonrails.org/v7.1/classes/ActionController/RequestForgeryProtection.html#method-i-valid_request_origin-3F

this is probably a case of AI adding too much into the comments, I'll clean it up

src/modules/hmisAppSettings/Provider.tsx
ttoomey: good catch, there's no need to handle 401s. Will simplify

src/modules/hmisAppSettings/Provider.tsx
```diff
@@ -60,18 +89,96 @@ export const HmisAppSettingsProvider: React.FC<Props> = ({ children }) => {
...
+  }, [attemptLogout]);
+
+  const retryLogout = useCallback(() => {
+    attemptLogout(false);
+  }, [attemptLogout]);
+
+  // A sign-out can fail for a transient reason, and the session it failed to end
+  // is still usable, so dismissing and going back to work is a legitimate choice.
+  // Sign out stays in the user menu for a later attempt.
+  const dismissLogoutFailure = useCallback(() => setLogoutFailed(false), []);
```
gigxz, line 176: I didn't test this, but the AI finding sounds correct to me:
> Failed logout() always clearStore()s; a retry via attemptLogout(false) leaves mounted queries empty (Apollo clearStore does not refetch), so “Stay signed in” can show a hollow app.
**Fix**: On failure, skip clearStore (keep clearing user/settings), or after failure/dismissLogoutFailure call apolloClient.resetStore() / reload so active queries refill while the session is still live.

I would advocate a different fix which is to just `reloadWindow()` if failed logout is dismissed. (Because I think the same issue is present with appSettings having been cleared as well)

src/modules/auth/components/LogoutFailedDialog.stories.tsx
```diff
@@ -0,0 +1,74 @@
+import { Meta, StoryObj } from '@storybook/react';
+import { expect, fn, userEvent, within } from '@storybook/test';
+
+import LogoutFailedDialog from './LogoutFailedDialog';
+
+export default {
+  component: LogoutFailedDialog,
+} as Meta<typeof LogoutFailedDialog>;
+
+type Story = StoryObj<typeof LogoutFailedDialog>;
```
gigxz, line 10: I'm not getting the separate stories to actually render, but I think we just don't have a nice solution for how to render Dialog stories. It's probably ok to leave as-is, appreciate the tests. We will likely come back to the wording/treatment for these Dialogs with design

gigxz (approved):

PR #1531, opened 2026-07-10
Add Bulk Voider user interface

martha:
## Description

Summary of changes:
- Add a Coordinated Entry admin page for entering comma-separated warehouse client IDs and confirming a bulk void run.

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/6706

## Type of change
New feature


src/api/operations/bulkVoidCeClients.operations.graphql
```diff
@@ -0,0 +1,8 @@
+mutation BulkVoidCeClients($destinationClientIds: [ID!]!, $projectId: ID!) {
+  bulkVoidCeClients(
+    destinationClientIds: $destinationClientIds
+    projectId: $projectId
+  ) {
+    success
```
ttoomey, line 6: should this select errors?
```suggestion
    success
    errors { ...ValidationErrorFields }
```

ttoomey (approved):
Looks good, passing along these suggestions from Claude that are maybe worth reviewing to see if they are real and worth addressing

`src/modules/ce/components/admin/AdminBulkVoidPage.tsx`
  1. [warning] AdminBulkVoidPage.tsx:99 — On mutation failure (network/GraphQL error), onCompleted never fires, so confirmOpen stays
  true and the dialog remains open. But displayError/ApolloErrorAlert renders in the page Stack behind the modal backdrop, so the user
  sees a stuck-open dialog with no visible error. Render the error inside the DialogContent, or close the dialog on error, so the
  failure is actually visible.
  2. [warning] AdminBulkVoidPage.tsx:79 — bulkVoidCeClients() is called from onSubmit and its returned promise is neither awaited nor
  .catch-ed. With no default errorPolicy set on the client and no onError on the hook, Apollo's mutate promise rejects on error,
  producing an unhandled promise rejection (Sentry/console noise) on every failure. The hook's error field still populates, so display
  works, but add onError (a no-op is fine) or .catch to suppress the rejection.
  3. [suggestion] AdminBulkVoidPage.tsx:160 — The confirm dialog text void {destinationClientIds.length} clients isn't
  pluralization-guarded and will read "void 1 clients". The input's helperText already handles the singular case (:106); mirror that
  here.

martha:
Thank you very much @ttoomey  for all the reviews!

PR #1532, opened 2026-07-14
Support editing null/not-null conditions in CE Match Rule UI

martha:
## Description

Summary of changes:
* Add the two comparators to `CeMatchClauseComparatorSelect` for non-multiple fields, labeled "Has a value" / "Does not have a value".
* Disable the value input and mark it optional when a null-check comparator is selected (`CeMatchClause`, `CeMatchClauseValueInput`).
* Force `value: null` when submitting a null-check clause in `useCeMatchRuleFormSubmission`, regardless of any stale form value.

Depends on hmis-warehouse PR: #6715

## Type of change
New feature

### DIFF
--- a/src/modules/admin/components/ceMatchRules/editor/clause/CeMatchClause.tsx
+++ b/src/modules/admin/components/ceMatchRules/editor/clause/CeMatchClause.tsx
@@ -182,16 +182,15 @@ const CeMatchClause: React.FC<Props> = ({
             onComparatorChange={resetValueSelection}
           />
         </Grid>
-        {/* Hide and unmount the Value input if the comparator indicates we're checking for null/not null. */}
-        {!isNullCheck && (
-          <Grid item xs={12} md={4}>
-            <CeMatchClauseValueInput
-              clausePath={clausePath}
-              control={control}
-              selectedField={selectedField}
-            />
-          </Grid>
-        )}
+        <Grid item xs={12} md={4}>
+          <CeMatchClauseValueInput
+            clausePath={clausePath}
+            control={control}
+            selectedField={selectedField}
+            disabled={isNullCheck}
+            required={!isNullCheck}
+          />
+        </Grid>
       </Grid>
     </Stack>
   );
--- a/src/modules/admin/components/ceMatchRules/editor/clause/CeMatchClause.tsx
+++ b/src/modules/admin/components/ceMatchRules/editor/clause/CeMatchClause.tsx
@@ -182,15 +182,15 @@ const CeMatchClause: React.FC<Props> = ({
             onComparatorChange={resetValueSelection}
           />
         </Grid>
-        <Grid item xs={12} md={4}>
-          <CeMatchClauseValueInput
-            clausePath={clausePath}
-            control={control}
-            selectedField={selectedField}
-            disabled={isNullCheck}
-            required={!isNullCheck}
-          />
-        </Grid>
+        {!isNullCheck && (
+          <Grid item xs={12} md={4}>
+            <CeMatchClauseValueInput
+              clausePath={clausePath}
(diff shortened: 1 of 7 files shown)


src/modules/admin/components/ceMatchRules/editor/clause/CeMatchClauseValueInput.tsx
gigxz, line 62: AI-review caught this:

> `hmis-frontend/src/modules/admin/components/ceMatchRules/editor/clause/CeMatchClauseValueInput.tsx#L107-L110` — Boolean fields still fail client validation under null-check comparators.
> Issue: Choosing “Has a value” / “Does not have a value” clears the value to '' via onComparatorChange, and required={false} is ignored by the boolean branch’s hardcoded validate (value !== ''). Choice/text paths respect required; boolean does not — submit is blocked for ItemType.Boolean fields (e.g. boolean CDEs).
> Fix: Gate the boolean validator on required, e.g. validate: (value) => !required || value !== '' || requiredMessage.


I did confirm that you can't save is-null/is-not-null rules against boolean fields:

Image
Image

src/modules/admin/components/ceMatchRules/editor/clause/CeMatchClause.tsx
```diff
@@ -179,6 +187,8 @@ const CeMatchClause: React.FC<Props> = ({
             clausePath={clausePath}
             control={control}
             selectedField={selectedField}
+            disabled={isNullCheck}
```
gigxz: This is fine, but we could also just not mount this input if a null-comparator comparator is selected - Not sure if that would fix the other bug I pointed out.

src/modules/admin/components/ceMatchRules/editor/clause/CeMatchClauseComparatorSelect.tsx
gigxz: Just curious, and maybe worth documenting in the code:
Is null-comparison not possible for array fields, or is it just something we're hiding from the frontend because we don't think it's needed? (e.g. "my_array_field = NULL" a.k.a. no options selected)

src/modules/admin/components/ceMatchRules/editor/clause/CeMatchClauseValueInput.tsx
martha: Thanks!

src/modules/admin/components/ceMatchRules/editor/clause/CeMatchClause.tsx
martha: Yes, it looks like we could and it does.

src/modules/admin/components/ceMatchRules/editor/clause/CeMatchClauseComparatorSelect.tsx
martha: I _think_ it's the first one, null-comparison is not possible for array fields, because of how `CdeFieldMap` uses the default of `[]` for repeating CDEDs. https://github.com/greenriver/hmis-warehouse/blob/e6150b3b84333b04dc9c38efdc87e4f451f3ac3b/drivers/hmis/app/models/hmis/ce/match/expression/cde_field_map.rb#L38-L44

Spec test: https://github.com/greenriver/hmis-warehouse/blob/ccaa1f710a6268e2830b4f98df3c2e3e45c68bbe/drivers/hmis/spec/models/hmis/ce/match/expression/cde_field_map_spec.rb#L108-L119

I'll add a comment!

gigxz: thanks!

gigxz (approved):

PR #1533, opened 2026-07-14
Enable attaching Funder and Project Type applicability to CE Match Rules

meborn:
## Description

Summary of changes:

Enables project type and funder filtering on global (DataSource-owned) CE match rules. This allows administrators to constrain when a rule is evaluated based on the project's funder or type.

- Adds an "Applies To" card to the CE match rule form with multi-selects for Funders and Project Types (only displayed for global rules)
- Updates the owner text to clarify that global rules can be further constrained by funder or project type
- Tracks `applicabilityDirty` state separately from expression changes to avoid unnecessary updates
- Improves TypeScript generics on `ControlledMultiSelect` for better type safety with react-hook-form paths

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/6717

How to test:
1. Navigate to a global CE match rule (DataSource-owned)
2. Verify the "Applies To" card appears with Funder and Project Type multi-selects
3. Select funders and/or project types and save the rule
4. Verify the selections persist on reload
5. Confirm the "Applies To" card does **not** appear for project-specific rules

## Type of change

New feature

### DIFF
--- a/src/modules/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx
+++ b/src/modules/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx
@@ -29,6 +29,7 @@ import { getRequiredLabel } from '@/modules/form/components/RequiredLabel';
 import ControlledMultiSelect from '@/modules/form/components/rhf/ControlledMultiSelect';
 import ControlledTextInput from '@/modules/form/components/rhf/ControlledTextInput';
 import { localResolvePickList } from '@/modules/form/util/formUtil';
+import { MultiHmisEnum } from '@/modules/hmis/components/HmisEnum';
 import { HmisEnums } from '@/types/gqlEnums';
 import {
   CeMatchRuleDetailsFragment,
@@ -47,16 +48,6 @@ interface Props {
 
 const projectTypeOptions = localResolvePickList('ProjectType') || [];
 const funderOptions = localResolvePickList('FundingSource') || [];
-
-const formatApplicabilityValues = (
-  values: readonly string[],
-  labels: Record<string, string>,
-  emptyLabel: string
-) => {
-  if (!values.length) return emptyLabel;
-
-  return values.map((value) => labels[value] || value).join(', ');
-};
 
 /**
  * The top-level CE match rule form component.
@@ -130,8 +121,6 @@ const CeMatchRuleForm: React.FC<Props> = ({
   const showWarningDialog = hasOnlyWarnings(errorState);
   const expressionDirty =
     !!dirtyFields.freeTextExpression || !!dirtyFields.structuredExpression;
-  const applicabilityDirty =
-    !!dirtyFields.projectTypes || !!dirtyFields.funders;
   const showApplicabilityCard = ownerType === CeMatchRuleOwnerType.DataSource;
 
   const handleCancel = useCallback(() => {
@@ -185,8 +174,16 @@ const CeMatchRuleForm: React.FC<Props> = ({
         </Stack>
       </TitleCard>
       {showApplicabilityCard && (
-        <TitleCard title='Applies To' headerComponent='h2' padded>
-          <Typography variant='body2' mt={-2} mb={1}>
+        <TitleCard
+          title='Applies To'
+          headerComponent='h2'
+          headerSx={{
+            pb: 0,
+            '& > .MuiTypography-root': { pb: 0 },
+          }}
+          padded
+        >
+          <Typography variant='body2' mb={2}>
             This rule will only be evaluated for projects matching the selected
             funders and project types.
           </Typography>
@@ -212,18 +209,18 @@ const CeMatchRuleForm: React.FC<Props> = ({
             {!editing && (
               <>
                 <CommonLabeledTextBlock title='Funders'>
-                  {formatApplicabilityValues(
-                    displayValues.funders,
-                    HmisEnums.FundingSource,
-                    'All funders'
-                  )}
+                  <MultiHmisEnum
+                    values={displayValues.funders}
+                    enumMap={HmisEnums.FundingSource}
+                    noData='All funders'
+                  />
                 </CommonLabeledTextBlock>
                 <CommonLabeledTextBlock title='Project Types'>
-                  {formatApplicabilityValues(
-                    displayValues.projectTypes,
-                    HmisEnums.ProjectType,
-                    'All project types'
-                  )}
+                  <MultiHmisEnum
+                    values={displayValues.projectTypes}
+                    enumMap={HmisEnums.ProjectType}
+                    noData='All project types'
+                  />
                 </CommonLabeledTextBlock>
               </>
             )}
@@ -274,7 +271,7 @@ const CeMatchRuleForm: React.FC<Props> = ({
               loading={loading}
               variant='contained'
               onClick={handleSubmit((values) =>
-                submit(values, { expressionDirty, applicabilityDirty })
+                submit(values, { expressionDirty })
               )}
             >
               Save Rule
@@ -311,7 +308,6 @@ const CeMatchRuleForm: React.FC<Props> = ({
               submit(values, {
                 confirmed: true,
                 expressionDirty,
-                applicabilityDirty,
               })
             )()
           }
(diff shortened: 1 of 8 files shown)


src/modules/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx
```diff
@@ -42,6 +45,19 @@ interface Props {
   onDelete?: VoidFunction;
 }
 
+const projectTypeOptions = localResolvePickList('ProjectType') || [];
+const funderOptions = localResolvePickList('FundingSource') || [];
+
+const formatApplicabilityValues = (
```
martha: We actually have an existing helper that I _think_ we can use as a drop-in replacement here. I haven't tested this out super thoroughly but it looks okay on first glance:

```diff
diff --git a/src/modules/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx b/src/modules/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx
index fa825a53..cc5697c5 100644
--- a/src/modules/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx
+++ b/src/modules/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx
@@ -29,6 +29,7 @@ import { getRequiredLabel } from '@/modules/form/components/RequiredLabel';
 import ControlledMultiSelect from '@/modules/form/components/rhf/ControlledMultiSelect';
 import ControlledTextInput from '@/modules/form/components/rhf/ControlledTextInput';
 import { localResolvePickList } from '@/modules/form/util/formUtil';
+import { MultiHmisEnum } from '@/modules/hmis/components/HmisEnum';
 import { HmisEnums } from '@/types/gqlEnums';
 import {
   CeMatchRuleDetailsFragment,
@@ -48,16 +49,6 @@ interface Props {
 const projectTypeOptions = localResolvePickList('ProjectType') || [];
 const funderOptions = localResolvePickList('FundingSource') || [];
 
-const formatApplicabilityValues = (
-  values: readonly string[],
-  labels: Record<string, string>,
-  emptyLabel: string
-) => {
-  if (!values.length) return emptyLabel;
-
-  return values.map((value) => labels[value] || value).join(', ');
-};
-
 /**
  * The top-level CE match rule form component.
  * Renders the rule details in either editable or readonly mode.
@@ -212,18 +203,18 @@ const CeMatchRuleForm: React.FC<Props> = ({
             {!editing && (
               <>
                 <CommonLabeledTextBlock title='Funders'>
-                  {formatApplicabilityValues(
-                    displayValues.funders,
-                    HmisEnums.FundingSource,
-                    'All funders'
-                  )}
+                  <MultiHmisEnum
+                    values={displayValues.funders}
+                    enumMap={HmisEnums.FundingSource}
+                    noData='All funders'
+                  />
                 </CommonLabeledTextBlock>
                 <CommonLabeledTextBlock title='Project Types'>
-                  {formatApplicabilityValues(
-                    displayValues.projectTypes,
-                    HmisEnums.ProjectType,
-                    'All project types'
-                  )}
+                  <MultiHmisEnum
+                    values={displayValues.projectTypes}
+                    enumMap={HmisEnums.ProjectType}
+                    noData='All project types'
+                  />
                 </CommonLabeledTextBlock>
               </>
             )}

```

src/modules/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx
```diff
@@ -165,6 +184,52 @@ const CeMatchRuleForm: React.FC<Props> = ({
           )}
         </Stack>
       </TitleCard>
+      {showApplicabilityCard && (
+        <TitleCard title='Applies To' headerComponent='h2' padded>
+          <Typography variant='body2' mt={-2} mb={1}>
```
martha: Nit: maybe it's a bit more graceful to move this inside the Stack rather than applying a negative margin?

```diff
s/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx
index fa825a53..2264a388 100644
--- a/src/modules/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx
+++ b/src/modules/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx
@@ -1,4 +1,4 @@
-import UnlockIcon from '@mui/icons-material/Lock';
+ import UnlockIcon from '@mui/icons-material/Lock';
 import { Button, Stack, Typography } from '@mui/material';
 import { useCallback, useEffect, useMemo, useState } from 'react';
 import { useForm } from 'react-hook-form';
@@ -186,11 +186,11 @@ const CeMatchRuleForm: React.FC<Props> = ({
       </TitleCard>
       {showApplicabilityCard && (
         <TitleCard title='Applies To' headerComponent='h2' padded>
-          <Typography variant='body2' mt={-2} mb={1}>
-            This rule will only be evaluated for projects matching the selected
-            funders and project types.
-          </Typography>
           <Stack gap={2}>
+            <Typography variant='body2'>
+              This rule will only be evaluated for projects matching the selected
+              funders and project types.
+            </Typography>
             {editing && (
               <>
                 <ControlledMultiSelect

```

src/modules/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx
```diff
@@ -114,6 +130,9 @@ const CeMatchRuleForm: React.FC<Props> = ({
   const showWarningDialog = hasOnlyWarnings(errorState);
   const expressionDirty =
     !!dirtyFields.freeTextExpression || !!dirtyFields.structuredExpression;
+  const applicabilityDirty =
```
martha: Finding turned up from my AI review:

> `CeMatchRuleForm` decides whether to send `projectTypes` and `funders` solely from `dirtyFields.projectTypes` / `dirtyFields.funders` (`src/modules/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx`, lines 132-134). However, `CeMatchExpressionModeSwitch.switchEditorMode` calls `reset(...)`, which preserves the current applicability values but clears their dirty flags, and then marks only `mode` dirty (`src/modules/admin/components/ceMatchRules/editor/CeMatchExpressionModeSwitch.tsx`, lines 45-59). Because `buildUpdateCeMatchRuleInput` omits applicability unless `applicabilityDirty` is true, a user can change a funder or project type, switch between structured/free-text mode, and save; the applicability selections remain visible but are not sent, so the old server configuration is retained. Preserve those dirty flags across the reset or determine applicability changes by comparing current values with the saved defaults.

I have two thoughts/suggestions:

1. Is it worth the complexity to track `applicabilityDirty`? This is an uncommon pattern in our app - usually we just submit the whole form, and the backend is responsible for translating that into changes that should be made to the model. I think this change also relates to my comment here about expecting all values on every submit: https://github.com/greenriver/hmis-warehouse/pull/6717#discussion_r3595914738
2. Now that there are more form fields besides just `expression`, that `mode` switcher probably should only reset dirty flags on the `expression` fields (using `resetField` on `freeTextExpression` and `structuredExpression` instead of a general `reset`). This would also let us update the Save button to be disabled when nothing is dirty, which _is_ a common pattern in our app. I'd say that button update is out of scope for this ticket, so it's up to you whether or not to pull that in. But I would suggest making the `reset` change.

martha (changes requested):
Looking good so far! I have a couple of related questions/changes on the frontend and backend. Let me know if you'd like to look at those together.

src/modules/admin/components/ceMatchRules/editor/CeMatchRuleForm.tsx
meborn: The negative margin was intended to bring the subtitle closer to the card title to match the mocks. After looking more closely at TitleCard, I found a cleaner solution using its header styling support to adjust the padding directly.

martha: Oh, I see. I think it looks fine either way, and in general I think our mocks can be taken as suggestions (although that might change if we prioritize updating the code to match the design library!) I would probably remove this custom styling, but I'd say it is up to you!

martha (approved):

PR #1538, opened 2026-07-22
Fix timing problems with household assessment tabs

martha:
## Description

Summary of changes:
- In `HouseholdAssessments`, set the initial tab based on the window.location.hash using initial state, rather than an effect. Keep the effect solely to deal with out-of-range hashes, since `tabs` isn't available yet on first render.
- In `useAssessmentFormDefinition`, if the query has already run, return the `previousData`'s form definition to avoid entirely unmounting the `HouseholdAssessments` component.

My observation based on local testing was that either of these was independently enough to solve the bug. I've included both, since they both seem like best-practice improvements to me.

Corresponding hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/6730 (does not depend)

How to test: See ticket comment

## Type of change
Bug fix

### DIFF
--- a/src/modules/assessments/components/household/HouseholdAssessments.tsx
+++ b/src/modules/assessments/components/household/HouseholdAssessments.tsx
@@ -75,7 +75,7 @@ const HouseholdAssessments: React.FC<Props> = ({
 }) => {
   // prove remount
   useEffect(() => {
-    console.warn('[HouseholdAssessments] mounted', window.location.hash);
+    console.log('[HouseholdAssessments] mounted', window.location.hash);
   }, []);
 
   // track if there are any dirty assessment forms
@@ -184,7 +184,7 @@ const HouseholdAssessments: React.FC<Props> = ({
   const { hash } = useLocation();
 
   useEffect(() => {
-    console.warn('[HouseholdAssessments] state', {
+    console.log('[HouseholdAssessments] state', {
       currentTab,
       hash,
       windowHash: window.location.hash,
@@ -203,7 +203,7 @@ const HouseholdAssessments: React.FC<Props> = ({
   ]);
 
   useEffect(() => {
-    console.warn('[HouseholdAssessments] hash sync effect', {
+    console.log('[HouseholdAssessments] hash sync effect', {
       hasRefetched,
       hash,
       windowHash: window.location.hash,
@@ -211,7 +211,7 @@ const HouseholdAssessments: React.FC<Props> = ({
     });
     // todo @martha
     if (hasRefetched) {
-      console.warn('[HouseholdAssessments] hasRefetched', window.location.hash);
+      console.log('[HouseholdAssessments] hasRefetched', window.location.hash);
       return;
     }
     const hashString = hash.replace('#', '');
@@ -221,13 +221,13 @@ const HouseholdAssessments: React.FC<Props> = ({
       (isFinite(hashNum) && hashNum >= 0 && hashNum <= tabs.length);
 
     if (isValid) {
-      console.warn(
+      console.log(
         '[HouseholdAssessments] setting current tab to hashString',
         hashString
       );
       setCurrentTab(hashString);
     } else {
-      console.warn('[HouseholdAssessments] setting current tab to 1');
+      console.log('[HouseholdAssessments] setting current tab to 1');
       setCurrentTab('1');
     }
   }, [hasRefetched, hash, tabs.length]);
--- a/src/modules/assessments/components/household/HouseholdAssessments.tsx
+++ b/src/modules/assessments/components/household/HouseholdAssessments.tsx
@@ -73,6 +73,11 @@ const HouseholdAssessments: React.FC<Props> = ({
   enrollment,
   formDefinition,
 }) => {
+  // prove remount
+  useEffect(() => {
+    console.log('[HouseholdAssessments] mounted', window.location.hash);
+  }, []);
+
   // track if there are any dirty assessment forms
   const [formStates, setFormStates] = useState<
     Record<string, HouseholdAssessmentFormState>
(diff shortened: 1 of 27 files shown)


src/modules/assessments/hooks/useAssessmentFormDefinition.tsx
```diff
@@ -33,26 +33,30 @@ const useAssessmentFormDefinition = ({
...
+    });
 
   const { formDefinition, itemMap } = useMemo(() => {
-    // Find the definition that we actually have
-    const formDefinition = data?.assessmentFormDefinition;
+    // Find the definition from the fetched data OR from the previous data, if this query has already run.
+    // This guards against totally remounting consumers when there's a refetch (e.g. assessmentDate updates after submit).
+    const formDefinition = data
+      ? data.assessmentFormDefinition
+      : previousData?.assessmentFormDefinition;
```
gigxz: I'm not quite understanding why `data.assessmentFormDefinition` is missing, is it a caching issue? (Maybe: Previously we were looking the form up by `role` and now we're looking it up by `id` after initial form submit, so `data` is null during that fetch? Looking at the comment on line 30)

I don't quite follow how the assessment date plays into it either, but I'm guessing it's because the date is part of the cacheKey.
This solution seems workable, but I'm mainly wondering if (1) there is an underlying cache key issue we should fix, or (2) the comment could be expanded to give more detail on why this happens

src/modules/assessments/components/household/HouseholdAssessments.tsx
```diff
@@ -176,22 +184,14 @@ const HouseholdAssessments: React.FC<Props> = ({
     });
   }, [householdMembers, assessmentByEnrollmentId, role, readOnlyView]);
 
-  const { hash } = useLocation();
-
+  // Once tabs are loaded, clamp an out-of-range tab (e.g. from a stale hash) back
+  // to the first tab.
```
gigxz: Make sense to me. Maybe more accurately we should say "When tabs or currentTab changes, clamp an out-of-range.." since this now triggers every time the currentTab changes.

src/modules/assessments/hooks/useAssessmentFormDefinition.tsx
martha: The `assessmentDate` changing causes the query `useGetAssessmentFormDefinitionQuery` to refetch with changed parameters. And since there is no existing definition in the cache for that `assessmentDate`, it hits the network. At that point, yes, as you said, `data` is then `undefined` during that fetch, `loading` becomes `true`, the parent `IntakeAssessmentPage` entirely unmounts the `HouseholdAssessments` component, and the bug manifests.

Looking on the backend, the reason to pass the assessment date (and to include it in the cache key) is that we apply date-based funder rules to the form definition. https://github.com/greenriver/hmis-warehouse/blob/ccaa1f710a6268e2830b4f98df3c2e3e45c68bbe/drivers/hmis/app/graphql/types/forms/form_definition.rb#L140-L145

If the user changing the assessment date actually produces a different form definition, hypothetically we probably should've refetched when the assessment date was first selected, so the user could've seen the right questions while filling out the form -- rather than only refetching after save. So maybe that is the real underlying "bug" here. But we don't have an existing pattern for "refetch the whole form when this one value changes," and it doesn't seem worth implementing for this rare edge case. (?)

As far as the underlying caching, I think it's working correctly. If in general we want to refetch the form definition when the assessment date changes, in case applicable funders changes what questions are available, then we _should_ keep storing the assessment date in the cache key. The suggestion on line 30, about augmenting the cache manually so it can perform whether we originally fetched by `id` or by `role`, is also interesting but I think it is unrelated.

src/modules/assessments/components/household/HouseholdAssessments.tsx
```diff
@@ -108,13 +107,22 @@ const HouseholdAssessments: React.FC<Props> = ({
...
+
+  // Initialize the tab from the URL hash. Read window.location.hash
+  // directly (rather than useLocation().hash) because tab navigation writes the
+  // hash via the separate data router (router.navigate), which does not notify
+  // the BrowserRouter's location context — leaving useLocation().hash stale.
+  const [currentTab, setCurrentTab] = useState<string | undefined>(() => {
+    const hashString = window.location.hash.replace('#', '');
+    const hashNum = hashString ? parseInt(hashString) : -1;
+    const isValid =
+      hashString === 'summary' || (isFinite(hashNum) && hashNum >= 0);
```
gigxz:
```suggestion
      hashString === SUMMARY_TAB_ID || (isFinite(hashNum) && hashNum >= 0);
```

src/modules/assessments/hooks/useAssessmentFormDefinition.tsx
```diff
@@ -33,26 +33,30 @@ const useAssessmentFormDefinition = ({
...
+        role,
+        assessmentDate,
+      },
+    });
 
   const { formDefinition, itemMap } = useMemo(() => {
-    // Find the definition that we actually have
-    const formDefinition = data?.assessmentFormDefinition;
+    // Find the definition from the fetched data OR from the previous data, if this query has already run.
+    // This guards against totally remounting consumers when there's a refetch (e.g. assessmentDate updates after submit).
```
gigxz:
```suggestion
    // Prefer previousData while refetching so consumers (e.g. IntakeAssessmentPage) keep
    // formDefinition and do not unmount HouseholdAssessments. Apollo clears `data` when
    // query variables change — e.g. assessmentDate when exitDate appears after submit.
    // assessmentDate is part of the query (and Apollo cache key) because the backend applies
    // date-based funder rules, so the same form can return different questions by date.
```

gigxz (approved):
non-blocking suggestions to clarify for future readers!

src/modules/assessments/hooks/useAssessmentFormDefinition.tsx
gigxz: Thanks for that explanation, I was missing the part where `data` was cleared because the variables changed. That makes sense. I proposed a suggestion to expand the comment. nice work tracking this down!

PR #1557, opened 2026-08-06
Sync CE top-level table text search to URL via searchQueryId

meborn:
## Description

Summary of changes:

Resolves [open-path/Green-River#9285](https://github.com/open-path/Green-River/issues/9285)

Sync top-level CE Referrals (`/referrals`) and Eligible Clients (`/referrals/eligible-clients`) text search to the URL via `searchQueryId`, never the raw term.

- New `usePersistedTableSimpleTextSearch` hook: restore from `persistedClientSearchParams`, write URL + Apollo cache on `onNetworkDataReady`, clear URL when the input is emptied (avoids debounce wipe during restore)
- Wired `ReferralsTable` and `AdminCeClientsTable` to that hook
- Request `searchQueryId` on `GetCeReferrals` / `GetCeClients` (+ codegen)

Nested CE tables are unchanged.

[Depends on hmis-warehouse PR: Branch `9285-search-query-id`](https://github.com/greenriver/hmis-warehouse/pull/6822)

How to test:

1. On **Referrals**, search by client name → URL gets `searchQueryId` (not the raw term); reload / open in a new tab restores input + results
2. Navigate into a referral and use browser Back → searched table restores
3. Clear the search → `searchQueryId` removed; full list returns
4. Repeat on **Eligible Clients**
5. Confirm table filters / pagination / optional columns still work with an active search
6. Confirm nested project CE referral tables do **not** put search in the URL

## Type of change
Bug fix / enhancement (URL-restorable CE table search)


src/modules/ce/components/admin/AdminCeClientsTable.tsx
```diff
@@ -83,8 +86,15 @@ const AdminCeClientsTable: React.FC<Props> = ({ projectGroupId }) => {
...
-  // Internal state for search and dialog
-  const [search, setSearch, debouncedSearch] = useDebouncedState<string>('');
+  const {
+    search,
+    setSearch,
+    activeSearchTerm,
+    searchQueryIdLoading,
+    onNetworkDataReady,
+  } = usePersistedTableSimpleTextSearch<GetCeClientsQuery>({
+    getSearchQueryId: getCeClientsSearchQueryId,
```
martha, line 96: Nit, seems like this would be clearer inline rather than defining the helper on line 62?

```suggestion
    getSearchQueryId: (data) => data.ceClients.searchQueryId,
```

src/modules/search/hooks/usePersistedTableSimpleTextSearch.ts
```diff
@@ -0,0 +1,105 @@
...
+      : (debouncedSearch ?? '');
+
+  const onNetworkDataReady = useCallback(
+    (data: TQuery) => {
+      const returnedSearchQueryId = getSearchQueryId(data);
+      if (returnedSearchQueryId && searchQueryId !== returnedSearchQueryId) {
+        writeClientSearchParamsToCache(
+          returnedSearchQueryId,
+          clientSearchInputToSearchParamsCacheFields({
+            textSearch: activeSearchTerm || null,
```
martha, line 81: This is a finding from my AI review that I thought was interesting enough to share:

> `[warning]` Stale-term cache write: `onNetworkDataReady` writes `textSearch: activeSearchTerm || null` using the hook's *current* `activeSearchTerm`, not the term the completed request was actually issued with. If responses arrive out of order (user typed further while an earlier request was in flight), the earlier id can be cached against a later term and pushed to the URL.

After having AI sketch a couple fix ideas,
> - Drop the client-derived cache write and trust the server as source of truth, suppressing the loading flash for self-created ids via a ref — fully correct, ~10 lines, but diverges from the existing convention and costs one extra fetch per new search.
> - Shared/server fix: Have the backend echo the resolved term on the paginated payload so all three consumers can pair term-to-response correctly — most correct and consistent, but the most wiring across both repos.

I think either fix would probably be bug-prone and not worthwhile, so I'm not suggesting any code changes here. The part that I think is interesting is that the existing users of `Hmis::ClientSearchQuery` -- that is, global client search and Bulk Services -- don't have the same problem because they don't use debounced search; they only make a network request after an explicit action, like a submit-button-click or enter-key. 

So my vague thought is just: maybe that type of weird complexity should push us in the direction of avoiding debounced search in general, especially in the future if we want to standardize our search patterns.

src/modules/ce/components/admin/AdminCeClientsTable.tsx
martha: And same in the `ReferralsTable` component below too

src/modules/search/hooks/usePersistedTableSimpleTextSearch.ts
martha, line 1: While clicking around, I noticed a weird bug that I think has to do with URL replacement, or potentially the new code conflicting with existing search param code for page numbers and optional columns. Here's an illustration (possibly not the only one, just one I've been able to repro consistently):
- open the HMIS in a fresh tab
- navigate to Referrals
- navigate to Eligible Clients
- search a client name -> search request returns, search query ID is successfully populated in the URL
- click into a client profile from the 3-dots menu of their search row
- click Back -> returns you to the search results, as expected
- click Back again -> returns you to the Eligible Clients list, no search, as expected
- click Back again -> this returns me to my fresh-tab homepage (google), which is what looks weird/buggy to me. I would expect clicking Back would take me back to Referrals, and then the HMIS homepage. It seems to point to url history getting clobbered.

I was curious if this also manifests from regular client search, but I couldn't reproduce it there, so I think it might be specific to the newly-added code.

I'm still exploring the bug and some possible fixes with AI, but wanted to post this in case you have additional context or have come across something similar before? (_possibly_ relates to https://github.com/open-path/Green-River/issues/9406? seems like it may also relate to https://github.com/open-path/Green-River/issues/9121#issuecomment-5049584852?)

martha: Here is an (ai generated) diff that _seems_ to resolve the bug. Before committing this, there is something still a little weird to me about this proposed fix in general, where the fix comes from checking equality and avoiding a `setParams` call when 2 things are equal... but wouldn't it be better if we just weren't making that unnecessary call in the first place? I would want to qa it more thoroughly and also understand the performance implications better too before committing this change.

```diff
diff --git a/src/hooks/useSearchParamState.ts b/src/hooks/useSearchParamState.ts
index be015d16..54c1e1ad 100644
--- a/src/hooks/useSearchParamState.ts
+++ b/src/hooks/useSearchParamState.ts
@@ -118,6 +118,26 @@ const populateParams = (
   });
 };
 
+// Read the params currently in the address bar rather than a render-time snapshot.
+// Multiple `useSearchParamsState` instances (searchQueryId, pagination, filters,
+// optional columns) each navigate independently; merging against a stale snapshot
+// lets a later writer drop params a sibling just wrote and navigate from a stale
+// history position. Always merging against the live URL keeps writes atomic.
+// const liveSearchParams = () => new URLSearchParams(window.location.search);
+
+// Order-insensitive comparison of two query strings, so we can skip navigations
+// that wouldn't actually change the URL. Skipping is load-bearing: a redundant
+// navigation to the same URL can still push a duplicate entry and, when it fires
+// from behind the history tip, truncate forward history.
+const searchStringsEqual = (a: string, b: string) => {
+  const normalize = (s: string) => {
+    const params = new URLSearchParams(s);
+    params.sort();
+    return params.toString();
+  };
+  return normalize(a) === normalize(b);
+};
+
 // Type for internal state representation of search params
 type ValueType = Record<string, any>; // Future improvement: constrain value type
 
@@ -241,11 +261,17 @@ const useSearchParamsState = ({
 
       const accumulator = new URLSearchParams();
       populateParams(currentParams, accumulator);
+
+      const nextSearch = accumulator.toString();
+      // Skip a no-op navigation. Without this, e.g. resetting to page 0 when already
+      // on page 0 pushes a duplicate entry (and can truncate forward history).
+      if (searchStringsEqual(nextSearch, window.location.search)) return;
+
       // `navigate` instead of `useSearchParams` as workaround for https://github.com/remix-run/react-router/issues/8393
       navigate({
         pathname,
         hash,
-        search: accumulator.toString(),
+        search: nextSearch,
       });
     },
     [hash, navigate, paramsDefinition, pathname, searchParams]
diff --git a/src/hooks/useTablePagination.ts b/src/hooks/useTablePagination.ts
index c35c0221..6fab076d 100644
--- a/src/hooks/useTablePagination.ts
+++ b/src/hooks/useTablePagination.ts
@@ -69,13 +69,18 @@ const useTablePagination = ({
 
   const setPage = useCallback(
     (newPage: number) => {
+      // No-op when the page isn't actually changing (e.g. GenericTableWithData resets
+      // to page 0 on every filter/search change). Writing it anyway pushes a redundant
+      // history entry that competes with other URL writers on the same interaction.
+      if (newPage === page) return;
+
       // Table controls provide 0-indexed pages; store 1-indexed pages in the URL.
       setPaginationParams((currentParams) => ({
         ...currentParams,
         [pageParam]: Math.max(newPage + 1, 1),
       }));
     },
-    [pageParam, setPaginationParams]
+    [page, pageParam, setPaginationParams]
   );
 
   const setRowsPerPage = useCallback(
```

meborn: See this PR where I added these suggestions https://github.com/greenriver/hmis-frontend/pull/1569 to be tested independently.

gigxz (approved):
Looks good to me, tested locally, suggest merging with the useTablePagination fix https://github.com/greenriver/hmis-frontend/pull/1569#pullrequestreview-5042425592 

This ended up being much gnarlier than I originally thought, thanks for digging through it and suggesting future improvements!

PR #1558, opened 2026-08-07
Support mark/unmark client record as restricted

gigxz:
## Description
Issue: https://github.com/open-path/Green-River/issues/9292 (does not close)

Depends on hmis-warehouse PR: https://github.com/greenriver/hmis-warehouse/pull/6823

Adds a Client Profile chip to mark or unmark a client as restricted, using the warehouse `setClientRestricted` mutation.

- Resolve `restricted` and `canMarkRestricted` on the client dashboard
- Warning / grayscale status chip with confirm dialogs for restrict and remove-restriction
- Hover styling for clickable `status` chips

**How to test:**
1. Manually grant `can_mark_clients_as_restricted` on an HMIS role in the db
2. As a user with `can_mark_clients_as_restricted` at a project where the client is enrolled, open the Client Dashboard. See “Unrestricted Record”, restrict, then remove restriction.
3. As a user without that permission, confirm the unrestricted chip is hidden; if the client is already restricted, the warning chip appears but is not clickable. (Note: the "can view restricted clients" permission is NOT applied; marking a client as restricted does nothing to visibility)

**Impact of deploying this:**
- On its own with https://github.com/greenriver/hmis-warehouse/pull/6823, this has **no customer impact**. No users will have permission to mark clients restricted, and no client records are restricted, so nothing additional will appear on the client profile.

## Type of change
New feature


src/modules/client/components/profileCard/ClientRestrictedChip.tsx
```diff
@@ -0,0 +1,110 @@
+import LockIcon from '@mui/icons-material/Lock';
```
martha, line 1: This is design feedback more than code feedback, but I have to say that I don't love the chip-as-action-item pattern being introduced here. 
- It's not in our design library, and I don't think we have any other clickable/actionable Chips in the codebase. A button (or an icon button) is the standard click-to-do-action that is familiar to users across the app's design language.
- The chip label communicates state, not action, so the chip isn't clearly actionable from text alone. You have to hover over it to understand that it is actually a click target, and even once you do, the distinction is a pretty subtle background color difference. I think this is too subtle, especially since Chips throughout the rest of our app are usually not clickable. Maybe this would be improved with a Tooltip if we don't want to rework the design to use a Button.

Overall, I'm okay with introducing this, given that MUI offers `clickable` Chips out of the box, with good a11y affordances (it is registered in the dom with `role="button"` and is keyboard-focusable/operable), but wanted to flag my concern. Would we consider revisiting the designs and using a Button instead? I do like the chip for info-only users (users who don't have permission to restrict). Maybe the chip stays, non-interactable, and a new button is added "Restrict Client" or "Unrestrict Client" stacked under "Update Client Details"?

src/types/gqlTypes.ts
martha, line 1: It looks like the schema may need to be regenerated before merge

martha (approved):
Looks good to me! I have a non-blocking comment about the design.

src/modules/client/components/profileCard/ClientRestrictedChip.tsx
gigxz: Agreed, I shared this feedback with design

PR #1560, opened 2026-08-11
Add summary CDEs to Referral task card

martha:
## Description

* Add `ReferralStepSummaryDetails` to show custom data elements with the `TableSummary` display hook on completed referral step cards.
* Move `customDataElements` from `CeReferralStepFields` to `CeReferralStepSummaryFields` so summary CDEs are loaded with referral step cards.
* Render summary rows only when the step is completed and the CDE has a value; skip empty values.
* Update the `ReferralStepCard` Storybook `Done` story with sample summary CDEs, including repeating values and a CDE with no value.

Relates to https://github.com/greenriver/hmis-warehouse/pull/6847 (does not depend)

How to test: See issue

## Type of change
New feature

### DIFF
--- a/src/api/operations/ce.fragments.graphql
+++ b/src/api/operations/ce.fragments.graphql
@@ -305,15 +305,15 @@ fragment CeReferralStepSummaryFields on CeReferralStep {
   access {
     canPerformStep
   }
+  customDataElements {
+    ...CustomDataElementFields
+  }
 }
 
 fragment CeReferralStepFields on CeReferralStep {
   ...CeReferralStepSummaryFields
   formDefinition {
     ...FormDefinitionFields
-  }
-  customDataElements {
-    ...CustomDataElementFields
   }
 }
 
--- a/src/types/gqlTypes.ts
+++ b/src/types/gqlTypes.ts
@@ -18298,6 +18298,135 @@ export type CeReferralFieldsFragment = {
       name: string;
     } | null;
     access: { __typename?: 'CeReferralStepAccess'; canPerformStep: boolean };
+    customDataElements: Array<{
+      __typename?: 'CustomDataElement';
+      id: string;
+      key: string;
+      label: string;
+      fieldType: CustomDataElementType;
+      repeats: boolean;
+      displayHooks: Array<DisplayHook>;
+      value?: {
+        __typename?: 'CustomDataElementValue';
+        id: string;
+        valueBoolean?: boolean | null;
+        valueDate?: string | null;
+        valueFloat?: number | null;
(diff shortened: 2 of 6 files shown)


src/api/operations/ce.fragments.graphql
martha, line 1: For discussion (overall on this PR, not specific to this file):

As implemented right now, we're requesting _all_ CDEs on _all_ tasks, when we really only want summary CDEs on completed tasks.

I've been exploring whether it's worthwhile to further pursue optimization here, and I _think_ the answer is probably not - since we expect both the # of tasks and the # of CDEs on tasks to be small.

A couple ideas I explored:
- [`resolve_custom_data_elements`](https://github.com/greenriver/hmis-warehouse/blob/ccaa1f710a6268e2830b4f98df3c2e3e45c68bbe/drivers/hmis/app/graphql/types/hmis_schema/has_custom_data_elements.rb#L37-L51) already loads all CDE associations into an array in memory. The dataloader batches by association symbol, so it can't carry a per-request scope. We would need much more new dataloader machinery in order to support our normal AR-`BaseFilter` pattern on CDEs.
- Perhaps we could make the API of this query similar to other filterable queries, like `customDataElements(filters: { showInSummary: true })`, but skip the `BaseFilter` pattern and just do the filtering in-memory in the resolver. But this doesn't seem like a big performance savings because all CDEs are still fetched into memory; it only reduces the network request payload, not the records returned from the db.
- Another alternative would be to abandon the filtering idea and instead create a dedicated new backend field `summaryCustomDataElements` (maybe on the `FormSubmittable` concern, or else directly on the Step). This is probably my favorite idea but I'm still not sure it is worth the API churn. 

What do you think?

src/types/gqlTypes.ts
martha, line 1: AI's summary of the diff to this file:

- **Positional-only churn (no net new data on the wire):** `GetCeReferralStepQuery`, `StartCeReferralStepMutation`, `SubmitCeReferralStep.step`, `CeReferralStepFieldsFragment`. These already fetched CDEs; the generated block just moved location.
- **Genuinely new CDE payload:** `GetCeReferral` and `SubmitCeReferralStep.referral.steps` (needed — this is what the new summary card renders) and `AssignParticipants.referral.steps` (not needed — it only re-resolves steps to refresh assignees, and never renders CDEs).

src/api/operations/ce.fragments.graphql
gigxz: I agree it doesn't seem worth optimizing / API churn. CDE counts per task should stay pretty small

gigxz (approved):
Looks great!

PR #1566, opened 2026-08-13
fix: improvements to JWT/SSO logout failure handling

ttoomey:
## Description
- Reload the app when a failed sign-out is dismissed
- Simplify the "still signed in" dialog copy to plain language


gigxz (approved):
thanks!

PR #1569, opened 2026-08-18
PR Feedback Discovery: Sync CE top-level table text search to URL via searchQueryId

meborn:
## Description

Follow-up to [#1557](https://github.com/greenriver/hmis-frontend/pull/1557), targeting its feature branch. This PR addresses two URL-state bugs uncovered during review of [#9285](https://github.com/open-path/Green-River/issues/9285).

### Underlying bugs

**1. Redundant pagination writes were adding unnecessary history entries**

`GenericTableWithData` resets pagination to page 0 when search or filter variables change. `useTablePagination` previously updated the URL even when the table was already on page 0.

Because `useSearchParamsState` uses `navigate` without `replace`, those no-op updates could still add browser-history entries and interfere with the expected Back sequence. In the reported repro, Back skipped Referrals and returned to the page outside HMIS.

`setPage` now returns early when the requested page is already active.

**2. Independent URL writers could overwrite each other using stale params**

Pagination, filters, optional columns, and `searchQueryId` use separate `useSearchParamsState` instances. Each instance previously merged updates against the React Router `searchParams` captured during its last render.

If one writer updated the URL before another instance re-rendered, the second writer could rebuild the URL from a stale snapshot and accidentally drop the first writer’s params.

URL writes now merge against the live `window.location.search` value at call time. Identical query strings are also treated as no-ops so they do not trigger redundant navigation.

### Summary of changes

- Skip pagination updates when the requested page is already active
- Merge URL updates against the live address-bar params
- Preserve params owned by sibling URL-state hooks
- Skip navigation when the resulting query string is unchanged

This PR adds no backend changes. The base feature remains dependent on [hmis-warehouse#6822](https://github.com/greenriver/hmis-warehouse/pull/6822).

How to test:

1. Navigate: HMIS home → Referrals → Eligible Clients
2. Search for a client and wait for `searchQueryId` to appear
3. Open a client profile from the row menu
4. Use Back repeatedly:
   - return to the searched Eligible Clients list
   - return to Eligible Clients without search
   - return to Referrals
   - confirm history does not skip directly outside HMIS
5. Exercise search, pagination, and filters together; confirm unrelated URL params are preserved
6. On Referrals, combine search, pagination, filters, and optional columns; confirm one interaction does not remove params owned by another
7. Repeat URL pagination/filter interactions on a non-CE table to check for regressions

Manual QA confirmed that the original Back-history reproduction no longer occurs.

## Type of change

Bug fix (shared URL state and browser-history coordination)


src/hooks/useSearchParamState.ts
martha: Great work Megan, thank you for your efforts on understanding the underlying causes here! The search-params code has definitely been bug-prone and tricky. A few thoughts from me, but definitely curious to hear @gigxz’s wisdom on this one.

- This PR fixes the bug I found, and your explanation is clear. Thank you!
- The change you made to `useTablePagination` actually appears to be enough to prevent the bug, if my testing is correct.
- Even though your explanation of the change to `useSearchParamState` is clear, I understand the change much less well, so I’m more hesitant about it. A few thoughts:
  - If react-router-dom is returning something outdated in `useSearchParams`, is that expected, or a bug in react-router-dom, or our usage of it?
    - Maybe it arises from the way we're circumventing react-router-dom [here](https://github.com/greenriver/hmis-frontend/blob/de1cfa8c892133c9ec5a81038f750e71e14c535f/src/hooks/useSearchParamState.ts#L244-L249)?
    - or something related to the dual-router state (`BrowserRouter` in `AppProvider` + `createBrowserRouter` in `router.tsx`), see [here](https://github.com/greenriver/hmis-frontend/blob/1280115e606e8b26aff132d7dd7ba08b55be8506/src/routes/router.tsx#L2-L14). This was where I was going with my comment on the [upgrade ticket](https://github.com/open-path/Green-River/issues/7257#issuecomment-5283943379)
    - or possibly it's expected/intended behavior: https://github.com/remix-run/react-router/issues/9290 / https://reactrouter.com/api/hooks/useSearchParams#setsearchparams-function
  - We plan to update react-router-dom soon (it's a ticket in the current sprint!). In general, I feel like we should prefer to rely on react-router-dom's constructs rather than circumventing them. So switching to _sometimes_ use `window.location.search` instead makes me nervous.
    - From my AI review: "The hook now reads from two different sources: values (returned to consumers) still comes from react-router's searchParams (render snapshot, :195), while writes merge from window.location. This read/write asymmetry is the crux of the fix but also the main readability cost. It's the kind of thing the planned react-router upgrade + single-writer consolidation should retire."

Overall, I see 2 possible paths forward. I lean towards option 2 but curious for @gigxz's thoughts.
- Option 1: Merge this PR as is; add a pretty thorough QA checklist that checks for regressions; possibly add some basic search param system tests.
- Option 2: Merge only the changes to useTablePagination.ts for now; work on the react-router-dom upgrade next; move away from the dual-router state (`BrowserRouter` in `AppProvider` + `createBrowserRouter` in `router.tsx`); drop the [`navigate` workaround](https://github.com/greenriver/hmis-frontend/blob/de1cfa8c892133c9ec5a81038f750e71e14c535f/src/hooks/useSearchParamState.ts#L244-L249) in favor of `setSearchParams`; push this hook towards relying fully on react-router-dom's constructs, instead of taking a half-and-half approach.

Out-of-scope thoughts about a possible future refactor:
- Yesterday at our 1:1, we briefly discussed the idea of a refactor to consolidate these `useSearchParamState` calls. 
- My AI review also was pushing me in that direction. I'm not totally on board, but could be convinced. 
- The architecture we originally chose/wanted for this hook was that it'd be independently callable by any component rendered on the page, including multiple times within one page. We wanted it to be sort of like `useState` but for search params. Different `useState` calls shouldn't mess with each other, nor should they require consolidation into one big top-level `useState`, so our idea was the same here.
- However, from my AI review: "There are multiple independent hook instances, each memoizing its own render-time searchParams. Instance B rendering with a snapshot from before instance A's write is expected React behavior, not a router bug." So maybe, our originally desired architecture was actually a false dream all along...?

gigxz (commented):
@meborn I'd suggest reverting changes to `src/hooks/useSearchParamState.ts` and merging only the changes to `src/hooks/useTablePagination.ts`, as suggested by Martha https://github.com/greenriver/hmis-frontend/pull/1569#discussion_r3814579839 for the same reasons!

I tested with that file reverted to main; forward/back navigation looks acceptable, including tested with CE "Workspaces" selection

PR #1576, opened 2026-08-23
Split HUD and Custom assmts in dropdown menu

gigxz:
## Description

Issue: https://github.com/open-path/Green-River/issues/9544

Update UX of New Assessment Menu, so HUD and Custom assessments are differentiated.

- [x] Ensured system tests all pass on the backend

Screenshot 2026-08-23 at 11 07 22 AM

## Type of change
UX


meborn (approved):
I think this looks good. The only thing of concern would be accessibility. Right now the split wouldn't be clear to a screen reader. If this is important perhaps adding subheaders to the list or `aria-label` to the custom assessments so that the distinction is clear.

gigxz:
Merging as-is, I think thats acceptable since it's not a meaningful split and the updated "hud " prefixes are readable. thanks!

